PR #3, opened 2020-04-01
Sample QoS plugin

Huang-Wei:
This sample plugin serves as an example to showcase how to write a custom plugin as well its tests.

BTW: some initial CI settings are also included in this PR.


Huang-Wei:
/hold

I will submit prow job yaml to k/test-infra, and then verify if it works here.

Huang-Wei:
/cc @ahg-g @alculquicondor @ravisantoshgudimetla @damemi 

CI settings and test cases (unit and integration) are in place.

I will try to compose e2e test suite in another PR.

cmd/main.go
```diff
@@ -0,0 +1,36 @@
...
+	"time"
+
+	"k8s.io/kubernetes/cmd/kube-scheduler/app"
+	"sigs.k8s.io/scheduler-plugins/pkg/qos"
+)
+
+func main() {
+	rand.Seed(time.Now().UnixNano())
+	command := app.NewSchedulerCommand(
+		app.WithPlugin(qos.Name, qos.New),
```
alculquicondor, line 31: Add a comment indicating that here is where you add custom plugins

pkg/qos/REAMD.md
```diff
@@ -0,0 +1,24 @@
+# Overview
+
```
alculquicondor, line 2: the filename is misspelled as REAMD

pkg/qos/REAMD.md
```diff
@@ -0,0 +1,24 @@
...
+(Quality of Service) class](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/)
+of Pods.
+
+## Maturity Level
+
+<!-- Check one of the values: Sample, Alpha, Beta, GA -->
+
+- [x] 💡 Sample (for demonstrating and inspiring purpose)
+- [ ] 👶 Alpha (used in companies for pilot projects)
+- [ ] 👦 Beta (used in companies, developed actively and ready for production workloads)
```
alculquicondor, line 13: I would drop the "ready for production workloads" here. Or is this similar to k/k?

pkg/qos/REAMD.md
```diff
@@ -0,0 +1,24 @@
...
+- [ ] 👨 Stable (used in companies for production workloads)
+
+## QOS QueueSort Plugin
+
+Given a bunch of Pods having the same `.spec.priority`, this plugin
+enqueue the Pods with the following order:
```
alculquicondor, line 19: Sorts pods by `.spec.priority` and breaks ties by the [quality of service class](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#qos-classes).

test/util/scheduler.go
```diff
@@ -0,0 +1,57 @@
...
+
+	var podInformer coreinformers.PodInformer
+
+	podInformer = testCtx.InformerFactory.Core().V1().Pods()
+	var err error
+	eventBroadcaster := events.NewBroadcaster(&events.EventSinkImpl{
+		Interface: testCtx.ClientSet.EventsV1beta1().Events(""),
+	})
+
+	opts = append([]scheduler.Option{scheduler.WithBindTimeoutSeconds(600)}, opts...)
```
alculquicondor, line 33: why do we need to explicitly set the bind timeout?

test/util/taint.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"context"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+func RemoveAllTaintsOffNode(cs clientset.Interface, nodeName string) (*v1.Node, error) {
```
alculquicondor, line 27: I don't see this being used

test/integration/scheduler_test.go
```diff
@@ -0,0 +1,83 @@
...
+				nil,
+				0,
+				scheduler.WithProfiles(profile),
+				scheduler.WithFrameworkOutOfTreeRegistry(tt.registry),
+			)
+			defer testutils.CleanupTest(t, testCtx)
+
+			schedPlugins := testCtx.Scheduler.Profiles[v1.DefaultSchedulerName].ListPlugins()
+			// We only check the particular extension points defined in each test case.
+			for extension, expected := range tt.expectedPlugins {
```
alculquicondor, line 76: What is the point of this test?
Tests for scheduler creation are already in k/k.

test/integration/scheduler_test.go
```diff
@@ -0,0 +1,83 @@
...
+				QueueSort: &schedapi.PluginSet{
+					Enabled: []schedapi.Plugin{
+						{Name: qos.Name},
+					},
+					Disabled: []schedapi.Plugin{
+						{Name: "*"},
+					},
+				},
+			},
+			expectedPlugins: map[string][]schedapi.Plugin{
```
alculquicondor, line 52: is this test passing? I think we should be including the rest of default plugins here.

damemi:
Is this repo going to commit a `vendor` folder with its dependencies?

test/util/scheduler.go
Huang-Wei: copy/paste mistake...

test/integration/scheduler_test.go
Huang-Wei: It's aimed to ensure the out-of-tree plugin works well with default ones. Especially for some plugin which should be global unique, e.g., QueueSort plugin.

test/integration/scheduler_test.go
Huang-Wei: I added some specific logic below to only check the specified out-of-tree entries are present.

Huang-Wei:
> Is this repo going to commit a `vendor` folder with its dependencies?

My preference is not. I created a "update-vendor" entry in the Makefile for users to update their local vendor folder.

pkg/qos/REAMD.md
Huang-Wei: I'd say this doesn't have the exact and strict definition of k/k. We can refine the wordings later. What I want to highlight is the plugins should come from **real requirements** of companies, and the maturity level is in accordance of their usage. But I cannot give a detailed checklist for one to promote to another at this moment.

test/util/taint.go
Huang-Wei: Removed.

test/integration/scheduler_test.go
alculquicondor: sg

go.mod
```diff
@@ -0,0 +1,35 @@
+module sigs.k8s.io/scheduler-plugins
+
+go 1.13
+
+require (
+	github.com/google/go-cmp v0.3.0
+	k8s.io/api v0.18.0
+	k8s.io/apimachinery v0.18.0
+	k8s.io/client-go v0.18.0
+	k8s.io/kubernetes v1.18.0
```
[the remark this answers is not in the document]
alculquicondor: We might need to use a dash in some cases. For example, assume there are bugs in this repository and we need to release a new version that is tied to k/k 1.18.1.
Then we should tag it as 1.18.1-2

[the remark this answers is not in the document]
ahg-g: @Huang-Wei the plan sounds good to me.

alculquicondor:
No more comments from my side

go.mod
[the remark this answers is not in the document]
Huang-Wei: Good point.

BTW: I  checked [semver](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions). It seems `-` is used to present "ahead of" a release, e.g., k8s uses `1.18.0-alpha.0` to tag an early release ahead of 1.18.0. So we should following that as well, rather than using `1.18.1-2`, we should use `1.18.1+patch1` - `+` here represents the "post" semantics.

Huang-Wei:
Thanks all for reviewing!

/hold cancel

@alculquicondor wants to try /lgtm? (to see whether you have the inherited "implicit" k-sigs membership)

PR #2, opened 2020-03-26
KEP for co-scheduling plugin based on scheduler framework

denkensk:
Add co-scheduling plugin for the new scheduler framework.
This plugin will be put in kubernetes-sigs/scheduler-plugins firstly.
Kep for #1

/sig scheduling
/assign @Huang-Wei

Issue #1, opened 2020-02-03
Add co-scheduling plugin based on scheduler framework


denkensk:
### What would you like to be added:
Add co-scheduling plugin based on scheduler framework

### Why is this needed:
Kubernetes has become a popular solution for orchestrating containerized workloads. Due to limitation of Kubernetes scheduler, some offline workloads (ML/DL) are managed in a different system. To improve cluster utilization and operation efficiency, we'd like to treat Kubneretes as a unified management platform. But ML jobs are All-or-Nothing: they require all tasks of a job to be scheduled at the same time. If the job only start part of tasks, it will wait for other tasks to be ready to begin to work. In the worst case, all jobs are pending leading to a deadlock. To solve this problem, co-scheduling is needed for the scheduler. The new scheduler framework makes the goal possible.

KEP：https://github.com/kubernetes/enhancements/pull/1463

/assign
/cc @Huang-Wei

denkensk:
Close this issue because the pr is merged

PR #4, opened 2020-04-03
Co-scheduling plugin based on scheduler framework

denkensk:
Add co-scheduling plugin for the new scheduler framework.
This plugin will be put in kubernetes-sigs/scheduler-plugins firstly.

Kep : https://github.com/kubernetes/enhancements/pull/1463 https://github.com/kubernetes-sigs/scheduler-plugins/pull/2

Issue: https://github.com/kubernetes-sigs/scheduler-plugins/pull/1

### DIFF
--- a/pkg/coscheduling/README.md
+++ b/pkg/coscheduling/README.md
@@ -1,0 +1,116 @@
+# Overview
+
+This folder holds the coscheduling plugin implementations based on [Lightweight coscheduling based on back-to-back queue 
+sorting](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/kep/20200116-lightweight-coscheduling-based-on-back-to-back-queue-sorting.md).
+
+## Maturity Level
+
+<!-- Check one of the values: Sample, Alpha, Beta, GA -->
+
+- [ ] 💡 Sample (for demonstrating and inspiring purpose)
+- [x] 👶 Alpha (used in companies for pilot projects)
+- [ ] 👦 Beta (used in companies and developed actively)
+- [ ] 👨 Stable (used in companies for production workloads)
+
+## Tutorial
+### PodGroup
+We use a special label named pod-group.scheduling.sigs.k8s.io/name to define a PodGroup. Pods that set this label and use the same value belong to the same PodGroup. 
+```
+labels:
+     pod-group.scheduling.sigs.k8s.io/name: nginx
+     pod-group.scheduling.sigs.k8s.io/min-available: "2"
+```
+We will calculate the sum of the Running pods and the Waiting pods (assumed but not bind) in scheduler, if the sum is greater than or equal to the minAvailable, the Waiting pods
+will be created.
+
+Pods in the same PodGroup with different priorities might lead to unintended behavior, so need to ensure Pods in the same PodGroup with the same priority.
+
+### Expectation
+1. If 2 PodGroups with different priorities come in, the PodGroup with high priority has higher precedence.
+2. If 2 PodGroups with same priority come in when there are limited resources, the PodGroup created first one has higher precedence.
+
+### Config
+1. queueSort, permit and unreserve must be enabled in coscheduling.
+2. preFilter is enhanced feature to reduce the overall scheduling time for the whole group. It will check the total number of pods belonging to the same `PodGroup`. If the total number is less than minAvailable, the pod will reject in preFilter, then the scheduling cycle will interrupt. And the preFilter is user selectable according to the actual situation of users. If the minAvailable of PodGroup is relatively small, for example less than 5, you can disable this plugin. But if the minAvailable of PodGroup is relatively large, please enable this plugin to reduce the overall scheduling time.
+```
+apiVersion: kubescheduler.config.k8s.io/v1alpha2
+kind: KubeSchedulerConfiguration
+leaderElection:
+  leaderElect: false
+clientConnection:
+  kubeconfig: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
+profiles:
+- schedulerName: default-scheduler
+  plugins:
+    queueSort:
+      enabled:
+        - name: coscheduling
+      disabled:
+        - name: "*"
+    preFilter:
+      enabled:
+        - name: coscheduling
+    permit:
+      enabled:
+        - name: coscheduling
+    unreserve:
+      enabled:
+        - name: coscheduling
+```
+
+### Demo
+Suppose we have a cluster which can only afford 3 nginx pods. We create a ReplicaSet with replicas=6, and set the value of minAvailable to 3.
+```yaml
+apiVersion: apps/v1
+kind: ReplicaSet
+metadata:
+  name: nginx
+  labels:
+    app: nginx
+spec:
+  replicas: 6
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      name: nginx
+      labels:
+        app: nginx
+        pod-group.scheduling.sigs.k8s.io/name: nginx
+        pod-group.scheduling.sigs.k8s.io/min-available: "3"
+    spec:
+      containers:
+      - name: nginx
+        image: nginx
+        resources:
+          limits:
+            cpu: 3000m
+            memory: 500Mi
+          requests:
+            cpu: 3000m
+            memory: 500Mi
+```
+
+```script
+$ kubectl get pods
+NAME          READY   STATUS    RESTARTS   AGE
+nginx-4jw2m   0/1     Pending   0          55s
+nginx-4mn52   1/1     Running   0          55s
+nginx-c9gv8   1/1     Running   0          55s
+nginx-frm24   0/1     Pending   0          55s
+nginx-hsflk   0/1     Pending   0          55s
+nginx-qtj5f   1/1     Running   0          55s
+```
+
+If min-available is set to 4 at this time, all nginx pods are in pending state because the resource does not meet the requirements of minavailable
+```script
+$ kubectl get pods
+NAME          READY   STATUS    RESTARTS   AGE
+nginx-4vqrk   0/1     Pending   0          3s
+nginx-bw9nn   0/1     Pending   0          3s
+nginx-gnjsv   0/1     Pending   0          3s
+nginx-hqhhz   0/1     Pending   0          3s
+nginx-n47r7   0/1     Pending   0          3s
+nginx-n7vtq   0/1     Pending   0          3s
+```
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -1,0 +1,259 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package coscheduling
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"sync"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/api/v1/pod"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+)
+
+// Coscheduling is a plugin that implements the mechanism of gang scheduling.
+type Coscheduling struct {
+	frameworkHandle framework.FrameworkHandle
+	podLister       corelisters.PodLister
+	// Key is the name of Namespace/PodGroup.
+	podGroupInfos sync.Map
+}
+
+// PodGroupInfo is a wrapper to a PodGroup with additional information.
+// TODO implement a timeout based gc for the PodGroupInfos map
+type PodGroupInfo struct {
+	name string
+	// timestamp stores the timestamp of the initialization time of PodGroup.
+	timestamp time.Time
+}
+
+var _ framework.QueueSortPlugin = &Coscheduling{}
+var _ framework.PreFilterPlugin = &Coscheduling{}
+var _ framework.PermitPlugin = &Coscheduling{}
+var _ framework.UnreservePlugin = &Coscheduling{}
+
+const (
+	// Name is the name of the plugin used in Registry and configurations.
+	Name                 = "coscheduling"
+	PodGroupName         = "pod-group.scheduling.sigs.k8s.io/name"
+	PodGroupMinAvailable = "pod-group.scheduling.sigs.k8s.io/min-available"
+	// TODO make this configurable
+	PermitWaitingTime = 1 * time.Second
+)
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (cs *Coscheduling) Name() string {
+	return Name
+}
+
+// New initializes a new plugin and returns it.
+func New(_ *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
+	podLister := handle.SharedInformerFactory().Core().V1().Pods().Lister()
+	return &Coscheduling{frameworkHandle: handle,
+		podLister: podLister,
+	}, nil
+}
+
+// Less are used to sort pods in the scheduling queue.
+// 1. compare the priority of pod
+// 2. compare the timestamp of the initialization time of PodGroup
+// 3. compare the key of PodGroup
+func (cs *Coscheduling) Less(podInfo1 *framework.PodInfo, podInfo2 *framework.PodInfo) bool {
+	pod1 := podInfo1.Pod
+	pod2 := podInfo2.Pod
+	priority1 := pod.GetPodPriority(pod1)
+	priority2 := pod.GetPodPriority(pod2)
+
+	if priority1 != priority2 {
+		return priority1 > priority2
+	}
+
+	pgInfo1 := cs.getPodGroupInfo(podInfo1)
+	pgInfo2 := cs.getPodGroupInfo(podInfo2)
+	time1 := pgInfo1.timestamp
+	time2 := pgInfo2.timestamp
+
+	if !time1.Equal(time2) {
+		return time1.Before(time2)
+	}
+
+	key1 := fmt.Sprintf("%v/%v", podInfo1.Pod.Namespace, pgInfo1.name)
+	key2 := fmt.Sprintf("%v/%v", podInfo2.Pod.Namespace, pgInfo2.name)
+	return key1 < key2
+}
+
+func (cs *Coscheduling) getPodGroupInfo(p *framework.PodInfo) *PodGroupInfo {
+	podGroupName, min, err := GetPodGroupLabels(p.Pod)
+	if err == nil && podGroupName != "" && min > 1 {
+		key := fmt.Sprintf("%v/%v", p.Pod.Namespace, podGroupName)
+		pgInfo, ok := cs.podGroupInfos.Load(key)
+		if !ok {
+			pgInfo = &PodGroupInfo{
+				name:      podGroupName,
+				timestamp: p.InitialAttemptTimestamp,
+			}
+			cs.podGroupInfos.Store(key, pgInfo)
+		}
+		return pgInfo.(*PodGroupInfo)
+	}
+
+	// If the pod is regular pod, return object of PodGroupInfo but not store in PodGroupInfos.
+	// The purpose is to facilitate unified comparison.
+	return &PodGroupInfo{name: "", timestamp: p.InitialAttemptTimestamp}
+}
+
+// PreFilter validates that if the total number of pods belonging to the same `PodGroup` is less than `minAvailable`.
+// If so, the scheduling process will be interrupted directly to avoid the partial Pods holding system resources
+// until a timeout. It will reduce the overall scheduling time for the whole group
+func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleState, p *v1.Pod) *framework.Status {
+	podGroupName, minAvailable, err := GetPodGroupLabels(p)
+	if err != nil {
+		return framework.NewStatus(framework.Error, err.Error())
+	}
+	if podGroupName == "" || minAvailable <= 1 {
+		return framework.NewStatus(framework.Success, "")
+	}
+
+	total := cs.calculateTotalPods(podGroupName, p.Namespace)
+	if total < minAvailable {
+		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in PreFilter: %d",
+			p.Namespace, podGroupName, p.Name, minAvailable, total)
+		return framework.NewStatus(framework.Unschedulable, "less than minAvailable")
+	}
+
+	return framework.NewStatus(framework.Success, "")
+}
+
+func (cs *Coscheduling) PreFilterExtensions() framework.PreFilterExtensions {
+	return nil
+}
+
+// Permit is the functions invoked by the framework at "permit" extension point.
+func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState, p *v1.Pod, nodeName string) (*framework.Status, time.Duration) {
+	podGroupName, minAvailable, err := GetPodGroupLabels(p)
+	if err != nil {
+		return framework.NewStatus(framework.Error, err.Error()), 0
+	}
+	if podGroupName == "" || minAvailable <= 1 {
+		return framework.NewStatus(framework.Success, ""), 0
+	}
+
+	namespace := p.Namespace
+	// TODO get actually scheduled(bind successfully) account from the SharedLister
+	running := cs.calculateRunningPods(podGroupName, namespace)
+	waiting := cs.calculateWaitingPods(podGroupName, namespace)
+	current := running + waiting + 1
+
+	if current < minAvailable {
+		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: running(%d), waiting(%d)",
+			p.Namespace, podGroupName, p.Name, minAvailable, running, waiting)
+		// TODO Change the timeout to dynamic value depends on the size of the `PodGroup`
+		return framework.NewStatus(framework.Wait, ""), 10 * PermitWaitingTime
+	}
+
+	klog.V(3).Infof("The count of podGroup %v/%v/%v is up to minAvailable(%d) in Permit: running(%d), waiting(%d)",
+		p.Namespace, podGroupName, p.Name, minAvailable, running, waiting)
+	cs.frameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Namespace == namespace && waitingPod.GetPod().Labels[PodGroupName] == podGroupName {
+			klog.V(3).Infof("Permit allows the pod: %v/%v", podGroupName, waitingPod.GetPod().Name)
+			waitingPod.Allow(cs.Name())
+		}
+	})
+
+	return framework.NewStatus(framework.Success, ""), 0
+}
+
+// Unreserve rejects all other Pods in the PodGroup when one of the pods in the group times out.
+func (cs *Coscheduling) Unreserve(ctx context.Context, state *framework.CycleState, p *v1.Pod, nodeName string) {
+	podGroupName, exist := p.Labels[PodGroupName]
+	if !exist {
+		return
+	}
+
+	cs.frameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Namespace == p.Namespace && waitingPod.GetPod().Labels[PodGroupName] == podGroupName {
+			klog.V(3).Infof("Unreserve rejects the pod: %v/%v", podGroupName, waitingPod.GetPod().Name)
+			waitingPod.Reject(cs.Name())
+		}
+	})
+}
+
+// GetPodGroupLabels will check the pod if belongs to some podGroup. If so, it will return the
+// podGroupName、minAvailable of podGroup. If not, it will return "" as podGroupName.
+func GetPodGroupLabels(p *v1.Pod) (string, int, error) {
+	podGroupName, exist := p.Labels[PodGroupName]
+	if !exist || podGroupName == "" {
+		return "", 0, nil
+	}
+	minAvailable, exist := p.Labels[PodGroupMinAvailable]
+	if !exist || minAvailable == "" {
+		return "", 0, nil
+	}
+	minNum, err := strconv.Atoi(minAvailable)
+	if err != nil {
+		klog.Errorf("GetPodGroupLabels err in coschduling %v/%v : %v", p.Namespace, p.Name, err.Error())
+		return "", 0, err
+	}
+	return podGroupName, minNum, nil
+}
+
+func (cs *Coscheduling) calculateTotalPods(podGroupName, namespace string) int {
+	// TODO get the total pods from the scheduler cache and queue instead of the hack manner
+	selector := labels.Set{PodGroupName: podGroupName}.AsSelector()
+	pods, err := cs.podLister.Pods(namespace).List(selector)
+	if err != nil {
+		klog.Error(err)
+		return 0
+	}
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateRunningPods(podGroupName, namespace string) int {
+	pods, err := cs.frameworkHandle.SnapshotSharedLister().Pods().FilteredList(func(pod *v1.Pod) bool {
+		if pod.Labels[PodGroupName] == podGroupName && pod.Namespace == namespace && pod.Status.Phase == v1.PodRunning {
+			return true
+		}
+		return false
+	}, labels.NewSelector())
+
+	if err != nil {
+		klog.Error(err)
+		return 0
+	}
+
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateWaitingPods(podGroupName, namespace string) int {
+	waiting := 0
+	// Calculate the waiting pods.
+	// TODO keep a cache of podgroup size.
+	cs.frameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Labels[PodGroupName] == podGroupName && waitingPod.GetPod().Namespace == namespace {
+			waiting++
+		}
+	})
+
+	return waiting
+}
--- a/pkg/coscheduling/coscheduling_test.go
+++ b/pkg/coscheduling/coscheduling_test.go
@@ -1,0 +1,578 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package coscheduling
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	v1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+	"k8s.io/kubernetes/pkg/scheduler/listers"
+)
+
+const (
+	queueSortPlugin = "no-op-queue-sort-plugin"
+	bindPlugin      = "bind-plugin"
+)
+
+var emptyArgs = make([]config.PluginConfig, 0)
+
+var _ framework.QueueSortPlugin = &TestQueueSortPlugin{}
+
+// TestQueueSortPlugin is a no-op implementation for QueueSort extension point.
+type TestQueueSortPlugin struct{}
+
+func newQueueSortPlugin(_ *runtime.Unknown, _ framework.FrameworkHandle) (framework.Plugin, error) {
+	return &TestQueueSortPlugin{}, nil
+}
+
+func (pl *TestQueueSortPlugin) Name() string {
+	return queueSortPlugin
+}
+
+func (pl *TestQueueSortPlugin) Less(_, _ *framework.PodInfo) bool {
+	return false
+}
+
+var _ framework.BindPlugin = &TestBindPlugin{}
+
+// TestBindPlugin is a no-op implementation for Bind extension point.
+type TestBindPlugin struct{}
+
+func newBindPlugin(_ *runtime.Unknown, _ framework.FrameworkHandle) (framework.Plugin, error) {
+	return &TestBindPlugin{}, nil
+}
+
+func (t TestBindPlugin) Name() string {
+	return bindPlugin
+}
+
+func (t TestBindPlugin) Bind(ctx context.Context, state *framework.CycleState, p *corev1.Pod, nodeName string) *framework.Status {
+	return nil
+}
+
+type fakeLister struct {
+	called int
+}
+
+// List lists all Pods in the indexer.
+func (*fakeLister) List(selector labels.Selector) (ret []*corev1.Pod, err error) {
+	if selector.Matches(labels.Set{PodGroupName: "pg1"}) {
+		ret = append(ret, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "pg1-1", Namespace: "namespace1"}})
+		return ret, nil
+	}
+
+	if selector.Matches(labels.Set{PodGroupName: "pg2"}) {
+		ret = append(ret, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "pg2-1", Namespace: "namespace2"}, Status: corev1.PodStatus{Phase: corev1.PodRunning}})
+		ret = append(ret, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "pg2-2", Namespace: "namespace2"}, Status: corev1.PodStatus{Phase: corev1.PodRunning}})
+		ret = append(ret, &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "pg2-3", Namespace: "namespace2"}, Status: corev1.PodStatus{Phase: corev1.PodRunning}})
+		return ret, nil
+	}
+
+	return ret, nil
+}
+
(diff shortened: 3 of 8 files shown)


Huang-Wei:
@denkensk you can wait for #3 to be merged and rebase the codebase. Generally, we don't upload /vendor folder unless it deems necessary.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func (cs *Coscheduling) PreFilterExtensions() framework.PreFilterExtensions {
+	return nil
+}
+
+// Permit is the functions invoked by the framework at "permit" extension point.
+func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState, p *v1.Pod, nodeName string) (*framework.Status, time.Duration) {
+	podGroupName, minAvailable := cs.checkPod(p)
+	if podGroupName == "" || minAvailable == 0 {
```
everpeace, line 141:
```suggestion
	if podGroupName == "" || minAvailable <= 1 {
```

could we shortcut `minAvailable==1` here??

Huang-Wei:
@denkensk `pull-scheduler-plugins-verify-build` failure can be resolved after rebasing.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,260 @@
...
+const PodGroupMinAvailable = "pod-group.scheduling.sigs.k8s.io/min-available"
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (cs *Coscheduling) Name() string {
+	return Name
+}
+
+// New initializes a new plugin and returns it.
+func New(_ *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
+	rand.Seed(time.Now().Unix())
```
angao, line 53: `rand.Seed()` is a global function and has been called in `main` method, so I think it is redundant here

denkensk: yeah, I find it~ thanks.  I will remove it.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,257 @@
...
+		klog.Error(err)
+	}
+	klog.V(3).Infof("Pod is in a podGroup: %v/%v, minAvailable is %v", podGroupName, p.Name, minNum)
+	return podGroupName, minNum
+}
+
+func (cs *Coscheduling) calculateTotalPods(podGroupName string) int {
+	// Returns the list of pods by the podGroupName.
+	selector := labels.Set(labels.Set{PodGroupName: podGroupName}).AsSelector()
+	pods, err := cs.PodLister.List(selector)
```
angao, line 200: I think here should use `FilteredList()` method to filter the pod namespace if the podGroupName is the same in a different namespace.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		klog.Error(err)
+	}
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateAllocatedPods(podGroupName string) int {
+	// Returns the list of pods by the podGroupName.
+	selector := labels.Set(labels.Set{PodGroupName: podGroupName}).AsSelector()
+	pods, err := cs.FrameworkHandle.SnapshotSharedLister().Pods().List(selector)
```
angao, line 210: ditto

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,257 @@
...
+	klog.V(3).Infof("Pod is in a podGroup: %v/%v, minAvailable is %v", podGroupName, p.Name, minNum)
+	return podGroupName, minNum
+}
+
+func (cs *Coscheduling) calculateTotalPods(podGroupName string) int {
+	// Returns the list of pods by the podGroupName.
+	selector := labels.Set(labels.Set{PodGroupName: podGroupName}).AsSelector()
+	pods, err := cs.PodLister.List(selector)
+	if err != nil {
+		klog.Error(err)
```
angao, line 202: Add a line `return 0`?

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,257 @@
...
+		return 0
+	}
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateRunningPods(podGroupName string) int {
+	// Returns the list of pods by the podGroupName.
+	selector := labels.Set(labels.Set{PodGroupName: podGroupName}).AsSelector()
+	pods, err := cs.FrameworkHandle.SnapshotSharedLister().Pods().List(selector)
```
angao, line 221: ditto

angao: And the pod status can be filtered at `FilteredList`.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,257 @@
...
+		return framework.NewStatus(framework.Success, ""), 0
+	}
+
+	running := cs.calculateRunningPods(podGroupName)
+	waiting := cs.calculateWaitingPods(podGroupName)
+
+	if running+waiting+1 < minAvailable {
+		klog.V(3).Infof("The count of podGroup %v is not up to minAvailable(%d) in Permit: running(%d), waiting(%d)",
+			podGroupName, minAvailable, running, waiting)
+		return framework.NewStatus(framework.Wait, ""), 10 * time.Second
```
angao, line 148: I think of a generation algorithm about the dynamic timeout. We know that a group of pods is exactly the same. When the cluster node is unchanged, we can assume that the scheduling time of each pod is the same, then we can get a formula
(minAvailable - waitingPods - 1) * schedulingTime, where schedulingTime is the time that pod joins the scheduling queue to run to the permit plug-in. For example, if the minAvailable of a job is 5 and the time to schedule a pod is 10ms, then the waiting time for the first pod is (5-0-1) * 10ms, the waiting time for the second pod is (5-1-1) * 10ms, and so on for the rest of the pods. Considering network, load, etc., it is possible to add a random time to the timeout. WDYT?

pkg/coscheduling/coscheduling_test.go
```diff
@@ -0,0 +1,528 @@
...
+
+	"k8s.io/kubernetes/pkg/scheduler/apis/config"
+
+	"k8s.io/apimachinery/pkg/labels"
+	v1 "k8s.io/client-go/listers/core/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+)
```
angao, line 22: Can you format the code?

denkensk: Done

pkg/coscheduling/coscheduling.go
denkensk: Done

pkg/coscheduling/coscheduling.go
denkensk: Tks, I change to filter the status by `FilteredList `

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+	} else if podGroupName2 != "" {
+		return podInfo1.InitialAttemptTimestamp.Before(pgInfo2.InitialTimestamp)
+	}
+
+	return podInfo1.InitialAttemptTimestamp.Before(podInfo2.InitialAttemptTimestamp)
+}
+
+func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleState, p *v1.Pod) *framework.Status {
+	podGroupName, minAvailable, namespace := cs.checkPod(p)
+	if podGroupName == "" || minAvailable == 0 {
```
everpeace, line 120:
```suggestion
	if podGroupName == "" || minAvailable <= 1 {
```

could we shortcut when `minAvailable== 0 or 1` ??

denkensk: Cool~ I will shortcut when minAvailable== 1, thanks~

pkg/coscheduling/coscheduling.go
denkensk: thanks, I prefer the first version to use `minAvailable-running-waiting-1/minAvailable` * SomeTime (A constant value) as the timeout. I think using the schedulingTime(use the time.Now() - PodInfo.Timestamp) is a good option, Maybe we need more estimate to check it.

angao: SGTM

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+		klog.Error(err)
+		return "", 0, p.Namespace
+	}
+	klog.V(3).Infof("Pod is in a podGroup: %v/%v, minAvailable is %v", podGroupName, p.Name, minNum)
+	return podGroupName, minNum, p.Namespace
+}
+
+func (cs *Coscheduling) calculateTotalPods(podGroupName, namespace string) int {
+	// Returns the list of pods by the podGroupName.
+	selector := labels.Set(labels.Set{PodGroupName: podGroupName}).AsSelector()
```
angao, line 203:
```suggestion
	selector := labels.SelectorFromSet(labels.Set{PodGroupName: podGroupName})
```

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+	pods, err := cs.PodLister.Pods(namespace).List(selector)
+	if err != nil {
+		klog.Error(err)
+		return 0
+	}
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateAllocatedPods(podGroupName, namespace string) int {
+	allocated := 0
```
Huang-Wei, line 213: nit: you can just use the `Pods` returned by `FilteredList()`, and then return `len(pods)`. (applies elsewhere)

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState, p *v1.Pod, nodeName string) (*framework.Status, time.Duration) {
+	podGroupName, minAvailable, namespace := cs.checkPod(p)
+	if podGroupName == "" || minAvailable <= 1 {
+		return framework.NewStatus(framework.Success, ""), 0
+	}
+
+	running := cs.calculateRunningPods(podGroupName, namespace)
+	waiting := cs.calculateWaitingPods(podGroupName, namespace)
+
+	if running+waiting+1 < minAvailable {
```
Huang-Wei, line 148: nit: make `running+waiting+1` as a variable.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+		klog.V(3).Infof("The count of podGroup %v is not up to minAvailable(%d) in Permit: running(%d), waiting(%d)",
+			namespace+"/"+podGroupName, minAvailable, running, waiting)
+		return framework.NewStatus(framework.Wait, ""), time.Duration(minAvailable-running-waiting-1/minAvailable) * PermitWaitingTime
```
Huang-Wei, line 151: what's the point of `1/minAvailable`? I suppose it's evaluated to 0 when `minAvailable` >= 1 - which I think is the common case.

Huang-Wei:
> Could we introduce another annotation about priority of podgroup, then the Less() could sort pod based on it?

Any particular reason of introducing another priority instead of inheriting `pod.spec.priority`?

vincent-pli:
@Huang-Wei 
We can queue `task` base on `podgroup`

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+		return "", 0, p.Namespace
+	}
+	klog.V(3).Infof("Pod is in a podGroup: %v/%v, minAvailable is %v", podGroupName, p.Name, minNum)
+	return podGroupName, minNum, p.Namespace
+}
+
+func (cs *Coscheduling) calculateTotalPods(podGroupName, namespace string) int {
+	// Returns the list of pods by the podGroupName.
+	selector := labels.Set(labels.Set{PodGroupName: podGroupName}).AsSelector()
+	pods, err := cs.PodLister.Pods(namespace).List(selector)
```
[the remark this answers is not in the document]
denkensk: Yes, I add a TODO comment here

pkg/coscheduling/README.md
```diff
@@ -0,0 +1,80 @@
...
+  selector:
+    matchLabels:
+      app: nginx
+  template:
+    metadata:
+      name: nginx
+      labels:
+        app: nginx
+        pod-group.scheduling.sigs.k8s.io/name: nginx
+        pod-group.scheduling.sigs.k8s.io/min-available: "3"
```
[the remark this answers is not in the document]
denkensk: I find the most existing namings conventions is dash style for example `beta.kubernetes.io/instance-type` `node.kubernetes.io/instance-type` `node.kubernetes.io/windows-build`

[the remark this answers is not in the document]
Huang-Wei: SG.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+	} else if podGroupName1 != "" {
+		return pgInfo1.InitialTimestamp.Before(podInfo2.InitialAttemptTimestamp)
+	} else if podGroupName2 != "" {
+		return podInfo1.InitialAttemptTimestamp.Before(pgInfo2.InitialTimestamp)
+	}
+
+	return podInfo1.InitialAttemptTimestamp.Before(podInfo2.InitialAttemptTimestamp)
+}
+
+func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleState, p *v1.Pod) *framework.Status {
```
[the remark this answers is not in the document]
Huang-Wei: Please add a comment here.

test/integration/coscheduling_test.go
```diff
@@ -0,0 +1,286 @@
...
+	go testCtx.Scheduler.Run(testCtx.Ctx)
+	if err = wait.Poll(100*time.Millisecond, 10*time.Second, podScheduled(cs, ns, pgNames2[0])); err != nil {
+		t.Errorf("Failed to schedule Pod %q: %v", pgNames2[0], err)
+	}
+	if err = wait.Poll(100*time.Millisecond, 2*time.Second, podScheduled(cs, ns, pgNames2[1])); err != nil {
+		t.Errorf("Failed to schedule Pod %q: %v", pgNames2[1], err)
+	}
+	if err = wait.Poll(100*time.Millisecond, 2*time.Second, podScheduled(cs, ns, pgNames2[2])); err != nil {
+		t.Errorf("Failed to schedule Pod %q: %v", pgNames2[2], err)
+	}
```
Huang-Wei, line 270: It'd be good to use a single wait to check all Pods are running instead of three checks.

test/integration/coscheduling_test.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// Create 6 Pods belongs to two podGroup.
+	// pg1 has three pods, and the resources of the cluster can't meet the min of pg1.
+	// pg2 has three pods, but the resources of the cluster can meet the min of pg2.
+	var pods []*v1.Pod
+	pgNames1 := []string{"pod1-1", "pod1-2", "pod1-3"}
+	pgNames2 := []string{"pod2-1", "pod2-2", "pod2-3"}
+	pause := imageutils.GetPauseImageName()
+	for i := 0; i < len(pgNames1); i++ {
```
Huang-Wei, line 209: Current logic creates the Pods in a zigzag order: "pg1-1, pg2-1, pg1-2, pg2-2, ...". However, if I changed the order to "pg1-1, pg1-2, pg1-3, pg2-1, pg2-2, pg2-3", i.e., like this:

```diff
diff --git a/test/integration/coscheduling_test.go b/test/integration/coscheduling_test.go
index a9811e4..a26c992 100644
--- a/test/integration/coscheduling_test.go
+++ b/test/integration/coscheduling_test.go
@@ -221,8 +221,10 @@ func TestCoschedulingPlugin(t *testing.T) {
 			coscheduling.PodGroupMinAvailable: "3",
 		}
 		pods = append(pods, pod)
+	}
 
-		pod = st.MakePod().Namespace(ns).Name(pgNames2[i]).Container(pause).Obj()
+	for i := 0; i < len(pgNames2); i++ {
+		pod := st.MakePod().Namespace(ns).Name(pgNames2[i]).Container(pause).Obj()
 		pod.Spec.Containers[0].Resources = v1.ResourceRequirements{
 			Requests: v1.ResourceList{
 				v1.ResourceMemory: *resource.NewQuantity(100, resource.DecimalSI),
@@ -239,7 +241,7 @@ func TestCoschedulingPlugin(t *testing.T) {
 	}
 
 	// Create 6 Pods, We will expect them to be scheduled in a reversed order.
-	t.Logf("Start to create 6 Pods.")
+	t.Logf("Start to create %d Pods.", len(pods))
 	for i := range pods {
 		t.Logf("Creating Pod %q", pods[i].Name)
 		_, err := cs.CoreV1().Pods(pods[i].Namespace).Create(context.TODO(), pods[i], metav1.CreateOptions{})
```

The test would fail:

```script
--- FAIL: TestCoschedulingPlugin (17.94s)
    coscheduling_test.go:244: Start to create 6 Pods.
    coscheduling_test.go:246: Creating Pod "pod1-1"
    coscheduling_test.go:246: Creating Pod "pod1-2"
    coscheduling_test.go:246: Creating Pod "pod1-3"
    coscheduling_test.go:246: Creating Pod "pod2-1"
    coscheduling_test.go:246: Creating Pod "pod2-2"
    coscheduling_test.go:246: Creating Pod "pod2-3"
    coscheduling_test.go:265: Failed to schedule Pod "pod2-1": timed out waiting for the condition
    coscheduling_test.go:268: Failed to schedule Pod "pod2-2": timed out waiting for the condition
    coscheduling_test.go:271: Failed to schedule Pod "pod2-3": timed out waiting for the condition
FAIL
FAIL    sigs.k8s.io/scheduler-plugins/test/integration  23.061s
FAIL
```

However, I think the eventual result should be irrelevant, right?

test/integration/coscheduling_test.go
```diff
@@ -0,0 +1,286 @@
...
+		pod := st.MakePod().Namespace(ns).Name(pgNames1[i]).Container(pause).Obj()
+		pod.Spec.Containers[0].Resources = v1.ResourceRequirements{
+			Requests: v1.ResourceList{
+				v1.ResourceMemory: *resource.NewQuantity(200, resource.DecimalSI),
+			},
+			Limits: v1.ResourceList{
```
Huang-Wei, line 215: nit: is the Limits necessary? If not, we can remove it.

test/integration/coscheduling_test.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+		pod.Labels = map[string]string{
+			coscheduling.PodGroupName:         "pg2",
+			coscheduling.PodGroupMinAvailable: "3",
+		}
+		pods = append(pods, pod)
+	}
+
+	// Create 6 Pods, We will expect them to be scheduled in a reversed order.
+	t.Logf("Start to create 6 Pods.")
```
Huang-Wei, line 242:
```suggestion
	t.Logf("Start to create %d Pods.", len(pods))
```

test/integration/coscheduling_test.go
```diff
@@ -0,0 +1,286 @@
...
+	if err != nil {
+		t.Fatalf("Failed to create Node %q: %v", nodeName, err)
+	}
+
+	// Create 6 Pods belongs to two podGroup.
+	// pg1 has three pods, and the resources of the cluster can't meet the min of pg1.
+	// pg2 has three pods, but the resources of the cluster can meet the min of pg2.
+	var pods []*v1.Pod
+	pgNames1 := []string{"pod1-1", "pod1-2", "pod1-3"}
+	pgNames2 := []string{"pod2-1", "pod2-2", "pod2-3"}
```
Huang-Wei, line 207: I'd like to see more test cases:

1. 2 PodGroups with different pod number - like one has 2, the other has 5.
1. regular Pods are involved
1. 2 PodGroups with different priority (preemption may be triggered) 
1. (optional) any corner case needs to involve more than 2 PodGroups?

And the Pods can be composed in an **arbitrary** order.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,279 @@
...
+	"k8s.io/kubernetes/pkg/api/v1/pod"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+)
+
+// Coscheduling is a plugin that implements the mechanism of gang scheduling.
+type Coscheduling struct {
+	FrameworkHandle framework.FrameworkHandle
+	PodLister       corelisters.PodLister
+	PodGroupInfos   map[string]PodGroupInfo // Key is the name of Namespace/PodGroup.
+	AutoID          uint64
```
Fei-Guo, line 40: Nit: suggest to rename AutoID to SeqNum, an ID typically indicates a string type in my opinion.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+type PodGroupInfo struct {
+	// InitialTimestamp stores the timestamp of the initialization time of PodGroup.
+	InitialTimestamp time.Time
+	ID               uint64
+	MinAvailable     int
```
Fei-Guo, line 47: MinAvailable and Name fields can be removed. The former is not used anywhere, the latter can be deduced from key easily.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,279 @@
...
+	selector := labels.Set{PodGroupName: podGroupName}.AsSelector()
+	pods, err := cs.PodLister.Pods(namespace).List(selector)
+	if err != nil {
+		klog.Error(err)
+		return 0
+	}
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateAllocatedPods(podGroupName, namespace string) int {
```
Fei-Guo, line 226: Is this method is used anywhere?

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,279 @@
...
+	klog.V(3).Infof("The count of podGroup %v is up to minAvailable(%d) in Permit: running(%d), waiting(%d)",
+		podGroupName, minAvailable, running, waiting)
+	cs.FrameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Namespace == namespace && waitingPod.GetPod().Labels[PodGroupName] == podGroupName {
+			klog.V(3).Infof("Permit allows the pod: %v/%v", podGroupName, waitingPod.GetPod().Name)
+			waitingPod.Allow(cs.Name())
+		}
+	})
+
+	delete(cs.PodGroupInfos, fmt.Sprintf("%v/%v", namespace, podGroupName))
```
Fei-Guo, line 176: By deleting the entry here, the next Pod with the same podGroupName will form a new PodGroupInfo with a new ID.   is this intended? 

Maybe, we should implement a timeout based gc for the PodGroupInfos map.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,279 @@
...
+		return 0
+	}
+
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateRunningPods(podGroupName, namespace string) int {
+	running := 0
+	_, err := cs.FrameworkHandle.SnapshotSharedLister().Pods().FilteredList(func(pod *v1.Pod) bool {
+		if pod.Labels[PodGroupName] == podGroupName && pod.Namespace == namespace && pod.Status.Phase == v1.PodRunning {
```
Fei-Guo, line 245: It will take some time for Pod to reach Running phase after scheduling. Assuming the MinAvailable is 2, and two Pods pass the Permit and start binding, if the 3rd Pod of the same group arrives, will this Pod be put into wait list and eventually timeout if it is the last Pod in the PodGroup?

pkg/coscheduling/coscheduling.go
denkensk: SGTM, I add to TODO.

pkg/coscheduling/coscheduling.go
denkensk: Yes, it may lead to the 3rd Pod timeout. It's a tougher condition to Coscheduling. But the 3rd Pod timeout will not effect the job because the min pods is created.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,293 @@
...
+type PodGroupInfo struct {
+	// Timestamp is used to make sure that every PodGroup has a fair chance to try to schedule.
+	// It will avoid the starvation of the pod created later.
+	// Timestamp is initialized by the first podInfo's Timestamp and updated when the PodGroup
+	// is reject in Permit. But In some extreme case, like the size of PodGroup is very large,
+	// when the first pods is timeout in Permit but still left some pod in activeQ.
+	// The change of Timestamp may lead to some undefined behavior in the heap.
+	// But no effect was found in the actual test.
+	// TODO implement efficient requeuing of Unschedulable Pods
+	Timestamp    time.Time
```
denkensk, line 53: Please Pay attention to the change of this place.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,268 @@
...
+// If so, the scheduling process will be interrupted directly to avoid the partial Pods holding system resources
+// until a timeout. It will reduce the overall scheduling time for the whole group
+func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleState, p *v1.Pod) *framework.Status {
+	podGroupName, minAvailable := GetPodGroupLabels(p)
+	if podGroupName == "" || minAvailable <= 1 {
+		return framework.NewStatus(framework.Success, "")
+	}
+
+	total := cs.calculateTotalPods(podGroupName, p.Namespace)
+	if total < minAvailable {
```
lixiang233, line 136: The first several pods of a group may be Unschedulable in PreFilter because when they entering PreFilter, some other pods of the group may havn't been created , then all pods of the group will be Unschedulable in Permit.It's a long time until they get the next chance to be scheduled.The way of counting total pods of a group needs improvement, otherwise I prefer disable PreFilter.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,268 @@
...
+	"math"
+	"strconv"
+	"sync"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/klog/v2"
```
Huang-Wei, line 31: I feel we should still stick to klog v1 as v2 is not adopted even in the upstream.

pkg/coscheduling/coscheduling.go
Huang-Wei: I'm afraid this is one of the limitations of current scheduler framework - which doesn't support multi-queue, and also doesn't expose that as an extension point yet. So here I read it as a hack to work around that. Otherwise, we may have to start the 1st PG Pod's scheduling cycle right away and that may leads to inefficient waiting.

test/integration/coscheduling_test.go
```diff
@@ -0,0 +1,275 @@
...
+					{Name: coscheduling.Name},
+				},
+			},
+		},
+	}
+
+	testCtx := util.InitTestSchedulerWithOptions(
+		t,
+		testutils.InitTestMaster(t, "sched-coscheduling", nil),
+		false,
```
Huang-Wei, line 76: I'd suggest to set this to `true`, then you don't need L80~L88. Otherwise, this cannot simulate a real scheduler running scenario.

BTW: If I changed to true. The overall time spent for all integration tests are spent as below:

```
    --- PASS: TestCoschedulingPlugin/equal_priority,_sequentially_pg1_meet_min_and_pg2_not_meet_min (90.04s)
    --- PASS: TestCoschedulingPlugin/equal_priority,_not_sequentially_pg1_meet_min_and_pg2_not_meet_min (90.15s)
    --- PASS: TestCoschedulingPlugin/equal_priority,_sequentially_pg1_not_meet_min_and_regular_pods (3.12s)
    --- PASS: TestCoschedulingPlugin/equal_priority,_not_sequentially_pg1_not_meet_min_and_3_regular_pods (3.09s)
    --- PASS: TestCoschedulingPlugin/different_priority,_sequentially_pg1_meet_min_and_pg2_meet_min (84.11s)
    --- PASS: TestCoschedulingPlugin/different_priority,_not_sequentially_pg1_meet_min_and_pg2_meet_min (90.11s)
    --- PASS: TestCoschedulingPlugin/different_priority,_not_sequentially_pg1_meet_min_and_3_regular_pods (3.10s)
    --- PASS: TestCoschedulingPlugin/equal_priority,_not_sequentially_pg1_meet_min_and_p2_p3_not_meet_min (87.13s)
```

Some tests used ~90 seconds, which sounds a bit too much - I assume the first flushoff would take 30 seconds (by default), and each individual test should take less than 40 seconds? Could you help me understand the time cost here? Or, maybe it's wasting some unnecessary scheduling cycles?

alculquicondor:
I'd like to take a look. Could you wait for me until tomorrow?

test/integration/coscheduling_test.go
denkensk: I find that the flushoff will run every 30s. But it just moves the pods who stays in UnschedulableQ more than 60s. For example: if the scheduler starts in 0s, and the podA is created in 1s and is waiting in `permit` phase. And podA will be timeout in 11s after 10s, Because in the 2th cycle, PodA just stays 49s in UnschedulableQ. So the flush will move PodA in the 3th cycle in 90s. So I think it's the reason that why Some tests used ~90 seconds @Huang-Wei  I think we can improve it by the `Efficient Requeuing of Unschedulable Pods` in the future /cc @alculquicondor 
```go
// flushUnschedulableQLeftover moves pod which stays in unschedulableQ longer than the unschedulableQTimeInterval
// to activeQ.
func (p *PriorityQueue) flushUnschedulableQLeftover() {
	p.lock.Lock()
	defer p.lock.Unlock()

	var podsToMove []*framework.QueuedPodInfo
	currentTime := p.clock.Now()
	for _, pInfo := range p.unschedulableQ.podInfoMap {
		lastScheduleTime := pInfo.Timestamp
		if currentTime.Sub(lastScheduleTime) > unschedulableQTimeInterval {
			podsToMove = append(podsToMove, pInfo)
		}
	}

	if len(podsToMove) > 0 {
		p.movePodsToActiveOrBackoffQueue(podsToMove, UnschedulableTimeout)
	}
}
```

Huang-Wei: SGTM.

pkg/coscheduling/coscheduling.go
alculquicondor: It looks like people should be able to disable this PreFilter if they find it detrimental rather than useful. Am I correct?
If so, we should add this to the README.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,267 @@
...
+	if !exist {
+		return "", 0, p.Namespace
+	}
+	minAvailable, exist := p.Labels[PodGroupMinAvailable]
+	if !exist {
+		return "", 0, p.Namespace
+	}
+	minNum, err := strconv.Atoi(minAvailable)
+	if err != nil {
+		klog.Error(err)
```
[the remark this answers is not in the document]
alculquicondor: +1

pkg/coscheduling/coscheduling.go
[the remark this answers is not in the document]
alculquicondor: Can you clarify in the TODO that you should be using the scheduler cache + queue?

pkg/coscheduling/coscheduling.go
alculquicondor: What about using PodInfoGroup? Also, do you actually need to check they are running or just that they are scheduled? If the latter, you don't need to check the Phase. Only scheduled pods are visible in the snapshot.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+
+	return len(pods)
+}
+
+func (cs *Coscheduling) calculateWaitingPods(podGroupName, namespace string) int {
+	waiting := 0
+	// Calculate the waiting pods.
+	cs.FrameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Labels[PodGroupName] == podGroupName && waitingPod.GetPod().Namespace == namespace {
```
alculquicondor, line 246: This is very inefficient. Add a TODO to keep a cache of podgroup size.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+			waiting++
+		}
+	})
+
+	return waiting
+}
+
+func (cs *Coscheduling) GetSeqNum() uint64 {
```
alculquicondor, line 254: doesn't need to be exported

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (cs *Coscheduling) getPodGroupInfo(p *framework.PodInfo) *PodGroupInfo {
+	podGroupName, min := GetPodGroupLabels(p.Pod)
+	if podGroupName != "" {
+		key := fmt.Sprintf("%v/%v", p.Pod.Namespace, podGroupName)
+		pgInfo, ok := cs.PodGroupInfos.Load(key)
+		if !ok {
+			pgInfo = &PodGroupInfo{
+				Name:         podGroupName,
+				Timestamp:    p.InitialAttemptTimestamp,
```
alculquicondor, line 112: Given that the resolution of Timestamp is very good, I don't think we need SeqNum. We can just do any arbitrary comparison, like alphabetical order of the pod group name.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+		if waitingPod.GetPod().Namespace == namespace && waitingPod.GetPod().Labels[PodGroupName] == podGroupName {
+			klog.V(3).Infof("Permit allows the pod: %v/%v", podGroupName, waitingPod.GetPod().Name)
+			waitingPod.Allow(cs.Name())
+		}
+	})
+
+	return framework.NewStatus(framework.Success, ""), 0
+}
+
+// Unreserve is called by the scheduling framework when a reserved pod was
```
alculquicondor, line 180: The comment should say what you are doing here, rather than when this is called.

My understanding is: Reject all other Pods in the PodGroup when one of the pods in the group times out.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+// rejected in a later phase.
+func (cs *Coscheduling) Unreserve(ctx context.Context, state *framework.CycleState, p *v1.Pod, nodeName string) {
+	podGroupName, exist := p.Labels[PodGroupName]
+	if !exist {
+		return
+	}
+
+	cs.FrameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
```
alculquicondor, line 188: @Huang-Wei do you think it's possible that a rejected pod is called more than once? I feel like it is, due to the asynchronous implementation of Reject.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+	podGroupName, minAvailable := GetPodGroupLabels(p)
+	if podGroupName == "" || minAvailable <= 1 {
+		return framework.NewStatus(framework.Success, "")
+	}
+
+	total := cs.calculateTotalPods(podGroupName, p.Namespace)
+	if total < minAvailable {
+		klog.V(3).Infof("The count of podGroup %v/%v is not up to minAvailable(%d) in PreFilter: %d",
+			p.Namespace, podGroupName, minAvailable, total)
+		return framework.NewStatus(framework.Unschedulable, "less than minAvailable")
```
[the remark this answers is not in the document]
denkensk: In the current version, we rely on flush mechanism every one minute  to solve this problem if there isn't some condition changed in the cluster changes that makes the pod schedulable. But the performance is poor.
I think we can improve it by the `Efficient Requeuing of Unschedulable Pods` in the future. For eg: We can add logic in `func PodAdd` to bring them back when the number is equal or greater than min.

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,261 @@
...
+	PermitWaitingTime    = 1 * time.Second
+)
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (cs *Coscheduling) Name() string {
+	return Name
+}
+
+// New initializes a new plugin and returns it.
+func New(_ *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
```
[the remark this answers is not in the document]
denkensk: We are targeting 1.18 this version.

pkg/coscheduling/coscheduling.go
denkensk: At the beginning, I want to use some other comparison methods like alphabetic string comparison. But later, I think the auto increment key is easier to understand.

pkg/coscheduling/coscheduling.go
denkensk: In this place, I want to check if they are running instead of scheduled because 
1. The Unreserve and Permit is Asynchronous. So maybe some pods is timeout in Permit and before the Unreserve. They are scheduled but can be used to calculate the total count in Permit.
2. the pod belongs to the PodGroup may be created failed or pulling image failed..... This situation still causes the whole job to wait for a long time because of less than min.

pkg/coscheduling/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+<!-- Check one of the values: Sample, Alpha, Beta, GA -->
+
+- [ ] 💡 Sample (for demonstrating and inspiring purpose)
+- [x] 👶 Alpha (used in companies for pilot projects)
+- [ ] 👦 Beta (used in companies and developed actively)
+- [ ] 👨 Stable (used in companies for production workloads)
+
+## Tutorial
+### PodGroup
```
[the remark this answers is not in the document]
denkensk: Done

pkg/coscheduling/coscheduling.go
alculquicondor: but that's already conveyed in the Timestamp.

pkg/coscheduling/coscheduling.go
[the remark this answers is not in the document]
denkensk: Done

pkg/coscheduling/coscheduling.go
Huang-Wei: @alculquicondor Yes. It can be called multiple times, and the first call with `Unschedulable` status will be caught by the internal channel; other calls afterwards will be treated as no-ops.

pkg/coscheduling/coscheduling.go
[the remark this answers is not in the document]
Huang-Wei: @alculquicondor before 1.19.0 GA is out, `scheduler-plugins` have to stick to 1.18.X.

pkg/coscheduling/coscheduling.go
alculquicondor: sg

pkg/coscheduling/coscheduling.go
denkensk: > Or perhaps we can expose this information (if its assumed or actually scheduled) in the SharedLister.
SG. Maybe I can use the assumed here in this version 1.18 here and expose the status(assumed/actually) in framework in 1.19. WDYT? @alculquicondor

alculquicondor: Did you mean the binding?

Huang-Wei:
@denkensk the integration test timed out after 10 minutes.

denkensk:
> @denkensk the integration test timed out after 10 minutes.

thanks The timeout is fixed.  Any other comments? @Huang-Wei

Huang-Wei:
@denkensk Thanks! Please update the `go.sum` (by running `make update-vendor`) and squash the commits.

denkensk:
> @denkensk Thanks! Please update the `go.sum` (by running `make update-vendor`) and squash the commits.

@Huang-Wei  Done

Huang-Wei:
/approve

I will leave /lgtm to @alculquicondor .

pkg/coscheduling/coscheduling.go
```diff
@@ -0,0 +1,259 @@
...
+	if priority1 != priority2 {
+		return priority1 > priority2
+	}
+
+	pgInfo1 := cs.getPodGroupInfo(podInfo1)
+	pgInfo2 := cs.getPodGroupInfo(podInfo2)
+	time1 := pgInfo1.timestamp
+	time2 := pgInfo2.timestamp
+
+	if time1.Equal(time2) && pgInfo1.name != "" && pgInfo2.name != "" {
```
alculquicondor, line 97: nit: can you make this like line 88?

`if !time1.Equal(time2) { return ...}`

pkg/coscheduling/coscheduling.go
alculquicondor: this was not handled

pkg/coscheduling/coscheduling.go
alculquicondor: This should be part of the README, at least.

pkg/coscheduling/README.md
```diff
@@ -0,0 +1,116 @@
...
+We use a special label named pod-group.scheduling.sigs.k8s.io/name to define a PodGroup. Pods that set this label and use the same value belong to the same PodGroup. 
+```
+labels:
+     pod-group.scheduling.sigs.k8s.io/name: nginx
+     pod-group.scheduling.sigs.k8s.io/min-available: "2"
+```
+We will calculate the sum of the Running pods and the Waiting pods (assumed but not bind) in scheduler, if the sum is greater than or equal to the minAvailable, the Waiting pods
```
denkensk, line 23: and here for Running pods @alculquicondor

pkg/coscheduling/README.md
```diff
@@ -0,0 +1,116 @@
...
+Pods in the same PodGroup with different priorities might lead to unintended behavior, so need to ensure Pods in the same PodGroup with the same priority.
+
+### Expectation
+1. If 2 PodGroups with different priorities come in, the PodGroup with high priority has higher precedence.
+2. If 2 PodGroups with same priority come in when there are limited resources, the PodGroup created first one has higher precedence.
+
+### Config
+1. queueSort、permit and unreserve must be enabled in coscheduling.
+2. preFilter is user selectable according to the actual situation of users.
+```
```
denkensk, line 35: I add description of the configuration here to tell user can disable the preFilter. @alculquicondor

pkg/coscheduling/coscheduling.go
denkensk: Done

pkg/coscheduling/README.md
```diff
@@ -0,0 +1,116 @@
...
+will be created.
+
+Pods in the same PodGroup with different priorities might lead to unintended behavior, so need to ensure Pods in the same PodGroup with the same priority.
+
+### Expectation
+1. If 2 PodGroups with different priorities come in, the PodGroup with high priority has higher precedence.
+2. If 2 PodGroups with same priority come in when there are limited resources, the PodGroup created first one has higher precedence.
+
+### Config
+1. queueSort、permit and unreserve must be enabled in coscheduling.
```
alculquicondor, line 33: there is a extraneous character there.

pkg/coscheduling/README.md
alculquicondor: I think you should clarify why a user might want to use it or not.

denkensk: @alculquicondor  Done

alculquicondor: I kind of disagree with the statement. It depends more on the number of podgroups, rather on the size of a podgroup.

alculquicondor:
There are a lot of things that need to improve both in this plugin and the framework.

Feel free to open issues upstream for 1.19+

/lgtm

Huang-Wei:
@denkensk In case we lost the conversation here, could you please create an umbrella issue to track all TODOs explicitly?

denkensk:
@Huang-Wei @alculquicondor  Thanks for your help. I will list all TODOs both in this plugin and the framework later.

Issue #9, opened 2020-06-23
Add Capacity scheduling for ML/DL workloads based on scheduler framework


denkensk:
### What would you like to be added:
Add Capacity scheduling for ML/DL workloads based on scheduler framework

### Why is this needed:
There is increasing demand to use Kubernetes to manage batch workloads (ML/DL). In those cases, one challenge is to improve cluster utilization while ensuring that each user has a reasonable amount of resources. The problem can be partially addressed by the Kubernetes ResourceQuota. The native Kubernetes ResourceQuota API can be used to specify the maximum overall resource allocation per namespace. The quota enforcement is done through an admission check. A quota resource consumer (e.g., a Pod) cannot be created if the aggregated resource allocation exceeds the quota limit. The Kubernetes quota design has the following limitations:

1. The quota resource usage is aggregated based on the resource configurations (e.g., Pod cpu/mem requests specified in the Pod spec). Although this mechanism can guarantee that the actual resource consumption will never exceed the ResourceQuota limit, it might lead to low resource utilization had the actual resource consumption been much smaller than the limit. i.e., lead to internal resource fragmentation.

2. If we use ResourceQuota to strictly divide the resources of the cluster among all users to prevent clusters from running out of resources, it might also lead to low resource utilization because some users may have quota resources not being used.

Due to above limitations, the batch workloads (ML/DL) can't run in a Kubernetes cluster as efficiently as they do in other container orchestration platforms such as Yarn. In order to overcome above limitations, we introduce an "Queue" concept used in Yarn capacity scheduler in Kubernetes. Basically, the "Queue" has the notions of "max" and "min", where the "min” is the minimum resources that are needed to ensure the basic functionality/performance of the consumers and the "max" specifies the upper bound of the resource consumption of the consumers. By introducing "min" and "max",  Pod scheduling allows the following optimizations:

1. The slack between "min" and "max" can help to tolerate runtime failures. For example, if a Pod that consumes the entire "min" fails to run, new Pods can still be created if the "max" has not been reached. When using Kubernetes resource quota, once the Pod that consumes the entire quota is created, no other Pods can be created even if the Pod fails to run (e.g., the Pod stucks at the image pulling phase). 

[...]

PR #10, opened 2020-06-23
KEP for Capacity scheduling for ML/DL workloads based on scheduler framework

denkensk:
/sig scheduling
fix https://github.com/kubernetes-sigs/scheduler-plugins/issues/9

https://docs.google.com/document/d/1ViujTXLP1XX3WKYUTk6u5LTdJ1sX-tVIw9_t9_mLpIc/edit?usp=sharing


kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+### Non-Goals
+
+<!--
+What is out of scope for this KEP?  Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+## Proposal
+
+A  "CRD"is needed to interact with the scheduler. The CRD kind name is subject to change. We use "Queue" in the proposal. The CRD defines the min and max allocation for the managed resources such as cpu, memory or extended resources. "Queue" is namespace scoped. This version of proposal does not consider merging the queue into the existing ResourceQuota. A sample yaml is listed below:
```
Huang-Wei, line 101: Again, "Queue" doesn't sound appropriate.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+1. The slack between "min" and "max" can help to tolerate runtime failures. For example, if a Pod that consumes the entire "min" fails to run, new Pods can still be created if the "max" has not been reached. When using Kubernetes resource quota, once the Pod that consumes the entire quota is created, no other Pods can be created even if the Pod fails to run (e.g., the Pod stucks at the image pulling phase). 
+
+2. Improve overall resource utilization by allowing one queue user to "borrow" unused reserved "min" resources from other queues. A queue user's unused "min" resource can be used by other users, under the condition that there is a mechanism to guarantee the "victim" user can consume his "min" resource whenever he needs. Typically, this is done by implementing preemption.
+
+The success of “Queue” based capacity scheduling relies on a proper Pod preemption algorithm implementation. This KEP proposes the minimal scheduler extension to support the “Queue” based scheduling based on the scheduler framework.
+
+
+### Goals
+
+Use a scheduler plugin to support a quota based preemption algorithm.
```
Huang-Wei, line 90: Implement a scheduler plugin to honor the API of elastic resource quota. (and maybe also highlight some key steps, such as management of admission webhooks, preemption to ensure "borrowed" resource can be returned back etc.)

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+bogged down.
+-->
+
+#### Story 1
+We assume two elastic quotas are defined: QuotaA (guaranteed:4, hard:6) and QuotaB (guaranteed:6, hard:8), the quota unit is the number of GPUs. UserA in green color consumes QuotaA and UserB in red color consumes QuotaB. The entire cluster has 10 GPUs available hence the sum of quota guaranteed is equal to the cluster capacity.
+
+Initially, UserA consumes 2 GPUs and UserB consumes 3 GPUs. Both consumptions are within the reserved "guaranteed" quota range.
+<div style="text-align: center">
+<img src="./1.png"/>
+</div>
```
Huang-Wei, line 140: Use native markdown syntax can also work, right?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+  min:
+    cpu: 10
+    memory: 20Gi
+    nvidia.com/gpu: 1
+```
+
+The definitions of "min" and "max" have been explained above. Max is infinite by default, min is 0 by default. There is an admission webhook to ensure the aggregate resource configurations of all the Pods that consume a Queue does not exceed the Queue’s max resource. Again, the sum of min in all Queue must be less than or equal to the total resource capacity of the cluster. This is enforced by a separate admission webhook. The Queue controller will track the resource allocation.
+
+The Queue objects are created by a cluster administrator. One Queue object per namespace (this assumption can be relaxed if we later introduce a “label” to match Queue CR and Queue consumers).
+
```
Huang-Wei, line 123: I think we miss a section here to explain the following scenarios:

- when and how resources are borrowed
- when and how resources are returned back

Huang-Wei: And I would suggest we use a Pod's lifecycle to connect the dots into a whole story.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+## Design Details
+
+### Extention point
+#### Preempt
+(This work will be started after the preemption extension point of framework finished)
+
+In the current scheduling workflow, if a Pod fails in `Filter` phase because of `FitError`, it will enter the `Preempt` phase. Assuming the Preempt plugin is supported, we will make the following changes to support the “Queue” based preemption algorithm.
+
+The failed to schedule Pod is referred to as the “preemptor”. The original `selectVictimsOnNode` method selects all the pods with the lower priority than the preemptor’s priority as potential victims in a node. In this proposal, the victim selection logic needs to be changed in the following cases.
+If the preemptor belongs to a namespace without Queue defined, the preemption victims will be selected from other namespaces without Queue defined.
```
Huang-Wei, line 169: Is there any particular reason this is mandatory? Why the regular preemptor cannot preempt pods from a namespace with queue defined?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+If the preemptor belongs to a namespace with Queue defined:
+When preemptor.Request + Queue.allocated > Queue.min. The preemption victims will be selected from the pods belonging to the same namespace based on the Pod priority.
```
Huang-Wei, line 171: Is queue.allocated from the API spec (e.g., status) or cache?

Huang-Wei: > The preemption victims will be selected from the pods belonging to the same namespace based on the Pod priority.

The failure reason can be arbitrary, not particular with a scarce resource. So the algorithm described here (and below) seems to assume the failure reason is just because pods among the queues are competing, but what if it's not the case? In other words, (1) why preempting pods from the queues can resolve the unschedulable state, and (2) why preempting pods from the queues is optimal?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: This is mainly to consider the situation: some namespaces do not create the corresponding ElasticQuota. How to deal with the namespaces with ElasticQuota defined and without ElasticQuota defined. Maybe we have two options: 
1. Like this kep said, Strictly distinguish two different kinds of ElasticQuota in Preempt Cycle.
2. We add some default ElasticQuota for the namespace without ElasticQuota defined. But the min of this default ElasticQuota is hard to determine how much to set. If we set 0 in min, it means that it can be preempted by other namespaces. Do you have any better suggestions?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+There is increasing demand to use Kubernetes to manage batch workloads (ML/DL). In those cases, one challenge is to improve cluster utilization while ensuring that each user has a reasonable amount of resources. The problem can be partially addressed by the Kubernetes ResourceQuota. The native Kubernetes ResourceQuota API can be used to specify the maximum overall resource allocation per namespace. The quota enforcement is done through an admission check. A quota resource consumer (e.g., a Pod) cannot be created if the aggregated resource allocation exceeds the quota limit. The Kubernetes quota design has the following limitations:
```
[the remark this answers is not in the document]
denkensk: Done

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+1. The quota resource usage is aggregated based on the resource configurations (e.g., Pod cpu/mem requests specified in the Pod spec). Although this mechanism can guarantee that the actual resource consumption will never exceed the ResourceQuota limit, it might lead to low resource utilization had the actual resource consumption been much smaller than the limit. i.e., lead to internal resource fragmentation.
```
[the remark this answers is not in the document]
denkensk: Done

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+2. If we use ResourceQuota to strictly divide the resources of the cluster among all users to prevent clusters from running out of resources, it might also lead to low resource utilization because some users may have quota resources not being used.
+
+Due to above limitations, the batch workloads (ML/DL) can't run in a Kubernetes cluster as efficiently as they do in other container orchestration platforms such as Yarn. In order to overcome above limitations, we introduce an "Queue" concept used in Yarn capacity scheduler [1] in Kubernetes. Basically, the "Queue" has the notions of "max" and "min", where the "min” is the minimum resources that are needed to ensure the basic functionality/performance of the consumers and the "max" specifies the upper bound of the resource consumption of the consumers. By introducing "min" and "max",  Pod scheduling allows the following optimizations:
```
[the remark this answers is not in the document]
denkensk: Done  I change its name to `ElasticQuota`.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+1. The slack between "min" and "max" can help to tolerate runtime failures. For example, if a Pod that consumes the entire "min" fails to run, new Pods can still be created if the "max" has not been reached. When using Kubernetes resource quota, once the Pod that consumes the entire quota is created, no other Pods can be created even if the Pod fails to run (e.g., the Pod stucks at the image pulling phase). 
+
+2. Improve overall resource utilization by allowing one queue user to "borrow" unused reserved "min" resources from other queues. A queue user's unused "min" resource can be used by other users, under the condition that there is a mechanism to guarantee the "victim" user can consume his "min" resource whenever he needs. Typically, this is done by implementing preemption.
```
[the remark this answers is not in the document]
denkensk: I change the `queue user` to `the workloads from one quota`.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: I  add more detail in `PostFilter` to introduce `borrow` and `return`.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+1. The quota resource usage is aggregated based on the resource configurations (e.g., Pod cpu/mem requests specified in the Pod spec). Although this mechanism can guarantee that the actual resource consumption will never exceed the ResourceQuota limit, it might lead to low resource utilization had the actual resource consumption been much smaller than the limit. i.e., lead to internal resource fragmentation.
+
+2. If we use ResourceQuota to strictly divide the resources of the cluster among all users to prevent clusters from running out of resources, it might also lead to low resource utilization because some users may have quota resources not being used.
```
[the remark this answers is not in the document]
denkensk: There are some repetitions in the content of 1st and 2nd. I  merge them together.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: > Is queue.allocated from the API spec (e.g., status) or cache?
-----
from cache. Thanks I add this in this kep.

denkensk: >  why preempting pods from the queues can resolve the unschedulable state
----
We have two different types of workloads: offline and online. The application of online workload doesn‘t support preemption, so they also doesn‘t support the `borrow` and `return` in ElasticQuota. So the min and max is equal in online workload. And the rest is offline workloads. Most of the failures of pods from offline workloads come from a scarce resource. So the preemption from the queues can resolve the unschedulable state.

> why preempting pods from the queues is optimal?
----
The pods from the queues maybe not optimal. But the optional range needs to be limited by Quota. 
If quota.allocated + pod.request > quota.min, the preemption will happen Intra-Quota.
if quota.allocated + pod.request < quota.min, the preemption will happen Inter-Quota.

@Huang-Wei

denkensk:
/cc @Huang-Wei @yuanchen8911 
Please review it again, thanks so much

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+Due to above limitation, the batch workloads (ML/DL) can't run in a Kubernetes cluster as efficiently as they do in other container orchestration platforms such as Yarn. In order to overcome above limitations, we introduce an "ElasticQuota" concept used in [Yarn capacity scheduler](https://www.netjstech.com/2018/04/capacity-scheduler-in-yarn-hadoop.html) in Kubernetes. Basically, the "ElasticQuota" has the notions of "max" and "min", where the "min” is the minimum resources that are needed to ensure the basic functionality/performance of the consumers and the "max" specifies the upper bound of the resource consumption of the consumers. By introducing "min" and "max", capacity scheduling can be optimized as follows:
+
+1. The resource buffer between "min" and "max" can help to tolerate runtime failures. For example, if a Pod that consumes the entire "min" fails to run, new Pods can still be created if the "max" has not been reached. When using Kubernetes resource quota, once the Pod that consumes the entire quota is created,
+no other Pods can be created - even if the Pod failed to be scheduled, or failed on running (error
+when pulling image, etc.).
```
yuanchen8911, line 83: _"once the Pod that consumes the entire quota is created,
 no other Pods can be created - even if the Pod failed to be scheduled, or failed on running"_

This sentence is not easy to ready. Do you mean " Once a namespace reaches its quota limit, no Pods can be created, even if there are pods in the namespace that failed to be scheduled or run."?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+  min:
+    cpu: 10
+    memory: 20Gi
+    nvidia.com/gpu: 1
+```
+
+The definitions of "min" and "max" have been explained above. Max is infinite by default, min is 0 by default. The sum of min in all ElasticQuota must be less than or equal to the total resource capacity of the cluster.The ElasticQuota objects are created by a cluster administrator. One ElasticQuota object per namespace.
+
+### Relationship with ResourceQuota
+This version of proposal does not discuss merging the ElasticQuota into the existing ResourceQuota. But in the future, we will try to merge ElasticQuota into ResourceQuota when it's mature. We have three fields of resource in ElasticQuota(min max) and ResourceQuota(hard).
```
yuanchen8911, line 126: but min and max still need to respect the namespace's resourceQuota limit, i.e., `min <= max <= current resourceQuota`.  It is worth mentioning.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### PreFilter
+We will check if the (Pod.request + Quota.Allocated) is less than Quota.Max. If not, the scheduling cycle of PodA will fail. 
+
+#### PostFilter
+We will reimplement the method `FindCandidates` in `defaultPreempt`. The original `FindCandidates` method selects all the pods with the lower priority than the preemptor’s priority as potential victims in a node. 
+In this proposal, the victim selection logic needs to be changed in the following cases.
+- if Preemptor.Request + Quota.allocated > Quota.min: It means that its min or guaranteed resource is used or `borrowed` by other Quota. So we will ask the quotas who's allocated is more than max (it means thest quotas borrow the resource from other quota)
+
+- if Preemptor.Requst + Quota.allocated <= Quota.min: It means that its guaranteed isn't borrowed by other quotas. So that we will select the pods belongs to the same quota(namespace) with the lower priority than the preemptor’s priority as potential victims in a node.
```
yuanchen8911, line 172: Requst -> Request

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Extention point
+
+#### PreFilter
+We will check if the (Pod.request + Quota.Allocated) is less than Quota.Max. If not, the scheduling cycle of PodA will fail. 
+
+#### PostFilter
+We will reimplement the method `FindCandidates` in `defaultPreempt`. The original `FindCandidates` method selects all the pods with the lower priority than the preemptor’s priority as potential victims in a node. 
+In this proposal, the victim selection logic needs to be changed in the following cases.
+- if Preemptor.Request + Quota.allocated > Quota.min: It means that its min or guaranteed resource is used or `borrowed` by other Quota. So we will ask the quotas who's allocated is more than max (it means thest quotas borrow the resource from other quota)
```
yuanchen8911, line 170: "if _Preemptor.Request + Quota.allocated > Quota.min_: "

Should it be opposite?    " _Preemptor.Request + Quota.allocated <= Quota.min_"

"_So we will ask the quotas who's allocated is more than max (it means thest quotas borrow the resource from other quota)_"

"more than **min**"?

The sentence can be rephrased  as "a victim will be chosen from a namespace that allocates more resources than its min, i.e., borrowing resources from other namespaces" ?

How?  

1. Based on our offline discussion, it's randomly chosen.  
2. If evicting a single pod in namespace cannot release enough resources for the pod to be scheduled, will it continue to evict more pods in the same namespace or choose another one?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
yuanchen8911: Same question, should it be "_if Preemptor.Requst + Quota.allocated **>=** Quota.min_"?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+![4](./4.png)
+
+## Design Details
+
+### Extention point
+
+#### PreFilter
+We will check if the (Pod.request + Quota.Allocated) is less than Quota.Max. If not, the scheduling cycle of PodA will fail. 
+
+#### PostFilter
```
yuanchen8911, line 167: custom preemption is a critical piece of this proposal. I feel more details are required. Also, as elastic quota is associated with namespace. You may just "namespace to refer to quota for readability.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -0,0 +1,208 @@
...
+### Relationship with ResourceQuota
+This version of proposal does not discuss merging the ElasticQuota into the existing ResourceQuota. But in the future, we will try to merge ElasticQuota into ResourceQuota when it's mature. We have three fields of resource in ElasticQuota(min max) and ResourceQuota(hard). 
+- hard: it will check the resource in pod creation through an admission check. It will prevent submitting too many jobs at one time. The usage of hard is aggregated based on the resource configurations of all pods so it will be greater than or equal to max. 
+- max: it will check the resource in the Prefilter of Pod Scheduling Cycle. The usage of max is aggregated based on the resource configurations of successfully scheduled pods.
+- min: it means the guaranteed resource of Quota.
+
+They act on different phases of pod lifecycle. So it maybe possible to merge them together.
+
+
+### User Stories (optional)
```
yuanchen8911, line 134: @denkensk  (cc @Huang-Wei) In our use cases, we want to restrict where a namespace/user can "steal" resources from. A typical scenario is that excessive resources can only be dynamically shared within a team, i.e., among the namespaces belonging to the same team. To support this requirement, conceptually we'll need a two-level queue or a virtual namespace group and each queue/namespace belongs to a high level queue or group. Resource stealing/custom preemption will be only within a group or a top level queue. This would be an important feature in large clusters with several thousands nodes and multiple-tenants, in which t's not practical to randomly steal resources and perform preemption. The latest heretical namespace proposal can be a long term solution.  Thoughts?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
+There is increasing demand to use Kubernetes to manage batch workloads (ML/DL). In those cases, one challenge is to improve cluster utilization while ensuring that each user has a reasonable amount of resources. The problem can be partially addressed by the Kubernetes [ResourceQuota](https://kubernetes.io/docs/concepts/policy/resource-quotas/). The native Kubernetes ResourceQuota API can be used to specify the maximum overall resource allocation per namespace. The quota enforcement is done through an admission check. A quota resource consumer (e.g., a Pod) cannot be created if the aggregated resource allocation exceeds the quota limit. The Kubernetes quota design has the limitation: the quota resource usage is aggregated based on the resource configurations (e.g., Pod cpu/mem requests specified in the Pod spec). Although this mechanism can guarantee that the actual resource consumption will never exceed the ResourceQuota limit, it might lead to low resource utilization. For instance, actual resource
```
Huang-Wei, line 76:
```
...
A quota consumer (e.g., a Pod) cannot be created if the aggregated
resource allocation exceeds the quota limit. In other words, 
the overall resource usage is aggregated based on Pod's spec
(i.e., cpu/mem requests) when it's created.
...
```

BTW: putting all words in one line is difficult for reviewing. It'd be good to split them into different lines by soft line breaks.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
+consumption may be much smaller than the limit.
 
-1. The quota resource usage is aggregated based on the resource configurations (e.g., Pod cpu/mem requests specified in the Pod spec). Although this mechanism can guarantee that the actual resource consumption will never exceed the ResourceQuota limit, it might lead to low resource utilization had the actual resource consumption been much smaller than the limit. i.e., lead to internal resource fragmentation.
+Due to above limitation, the batch workloads (ML/DL) can't run in a Kubernetes cluster as efficiently as they do in other container orchestration platforms such as Yarn. In order to overcome above limitations, we introduce an "ElasticQuota" concept used in [Yarn capacity scheduler](https://www.netjstech.com/2018/04/capacity-scheduler-in-yarn-hadoop.html) in Kubernetes. Basically, the "ElasticQuota" has the notions of "max" and "min", where the "min” is the minimum resources that are needed to ensure the basic functionality/performance of the consumers and the "max" specifies the upper bound of the resource consumption of the consumers. By introducing "min" and "max", capacity scheduling can be optimized as follows:
```
Huang-Wei, line 79: "min” -> "min"

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
+There is increasing demand to use Kubernetes to manage batch workloads (ML/DL). In those cases, one challenge is to improve cluster utilization while ensuring that each user has a reasonable amount of resources. The problem can be partially addressed by the Kubernetes [ResourceQuota](https://kubernetes.io/docs/concepts/policy/resource-quotas/). The native Kubernetes ResourceQuota API can be used to specify the maximum overall resource allocation per namespace. The quota enforcement is done through an admission check. A quota resource consumer (e.g., a Pod) cannot be created if the aggregated resource allocation exceeds the quota limit. The Kubernetes quota design has the limitation: the quota resource usage is aggregated based on the resource configurations (e.g., Pod cpu/mem requests specified in the Pod spec). Although this mechanism can guarantee that the actual resource consumption will never exceed the ResourceQuota limit, it might lead to low resource utilization. For instance, actual resource 
+consumption may be much smaller than the limit.
```
Huang-Wei, line 77:
```
it might lead to low resource utilization as some pods may have claimed the
resources but failed to be scheduled.
```

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
Huang-Wei: I think the term "capacity scheduling" and "Resource Quota" should be officially defined, or you may want to declaim that these two terms are used interchangeably.

Huang-Wei: Or, you may want to say something like:

```
We proposed a new "ResourceQuota" mechanics to optimize current capacity scheduling as follows:
```

Huang-Wei: In order to overcome above limitations, we introduce an "ElasticQuota" concept used in
->
In order to overcome above limitations, an "ElasticQuota" concept used in
Yarn can be leveraged into Kubernetes.

Huang-Wei: Move the description of "min" and "max" to the first bullet. And here, maybe say something like:

```
Basically, the "ElasticQuota" aims to provide basic quota management
while also providing flexibility on a certain extend of overselling.
```

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
 
-2. If we use ResourceQuota to strictly divide the resources of the cluster among all users to prevent clusters from running out of resources, it might also lead to low resource utilization because some users may have quota resources not being used.
+1. The resource buffer between "min" and "max" can help to tolerate runtime failures. For example, if a Pod that consumes the entire "min" fails to run, new Pods can still be created if the "max" has not been reached. When using Kubernetes resource quota, once the Pod that consumes the entire quota is created,
```
Huang-Wei, line 81: Add the first bullet to explain min and max.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
+no other Pods can be created - even if the Pod failed to be scheduled, or failed on running (error
+when pulling image, etc.).
 
-Due to above limitations, the batch workloads (ML/DL) can't run in a Kubernetes cluster as efficiently as they do in other container orchestration platforms such as Yarn. In order to overcome above limitations, we introduce an "Queue" concept used in Yarn capacity scheduler [1] in Kubernetes. Basically, the "Queue" has the notions of "max" and "min", where the "min” is the minimum resources that are needed to ensure the basic functionality/performance of the consumers and the "max" specifies the upper bound of the resource consumption of the consumers. By introducing "min" and "max",  Pod scheduling allows the following optimizations:
+2. Improve overall resource utilization by allowing the workloads from one quota to "borrow" unused reserved "min" resources from other quotas. A quota's unused "min" resource can be used by other users, under the condition that there is a mechanism to guarantee the "victim" user can consume his "min" resource whenever he needs. Typically, this is done by implementing preemption.
```
Huang-Wei, line 85: Improve overall resource utilization by allowing the workloads... ->
Allow the workloads...

his -> its, he -> it
~~Typically, this is done by implementing preemption.~~

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
-
-2. Improve overall resource utilization by allowing one queue user to "borrow" unused reserved "min" resources from other queues. A queue user's unused "min" resource can be used by other users, under the condition that there is a mechanism to guarantee the "victim" user can consume his "min" resource whenever he needs. Typically, this is done by implementing preemption.
-
-The success of “Queue” based capacity scheduling relies on a proper Pod preemption algorithm implementation. This KEP proposes the minimal scheduler extension to support the “Queue” based scheduling based on the scheduler framework.
+The success of “ElasticQuota” based capacity scheduling relies on a proper Pod preemption algorithm implementation. This KEP proposes the minimal scheduler extension to support the “ElasticQuota” based scheduling based on the scheduler framework.
```
Huang-Wei, line 87: “ElasticQuota” -> "ElasticQuota"

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
Huang-Wei: Another question is: when to start the preemption? It seems that if the system resource is efficient, all elastic quotas can be fully used up to its "max". So doe that mean preemption can only happen when system resource has been used up globally?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -70,24 +73,24 @@ updates.
...
 
 
 ### Goals
 
-Use a scheduler plugin to support a quota based preemption algorithm. 
+- The API definetion of ElasticQuota
```
Huang-Wei, line 92: definition

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
Huang-Wei: I understand the concern, but at this moment, due to the lack of official definition of "tenant", it's hard to associate it the elastic RQ with a virtual concept. So IMO we can start with namespace, and try other multi-tenancy APIs such as hierarchical namespace, etc, at a later time.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -132,47 +141,38 @@ bogged down.
...
+![3](./3.png)
 
-Later, if UserB demands 8 GPUs, 5 more compared to his current consumption, his request cannot be immediately satisfied since only one unused GPU left in the cluster. Since UserB's "guaranteed" quota is 6 GPUs, the preemption algorithm ensures that three GPUs will be returned from UserA.  However, no more GPUs can be assigned to UserB since UserA's "guaranteed" quota has to be honored as well. In the end, UserA consumes 4 GPUs and UserB consumes 6 GPUs, satisfying each user's "guaranteed" quota.
-<div style="text-align: center">
-<img src="./4.png"/>
-</div>
+Later, if UserB demands 8 GPUs, 5 more compared to his current consumption, his request cannot be immediately satisfied since only one unused GPU left in the cluster. Since UserB's "min" quota is 6 GPUs, the preemption algorithm ensures that three GPUs will be returned from UserA.  However, no more GPUs can be assigned to UserB since UserA's "min" quota has to be honored as well. In the end, UserA consumes 4 GPUs and UserB consumes 6 GPUs, satisfying each user's "min" quota.
```
Huang-Wei, line 157: In this case, it doesn't help if UserB get another 3GPU (less than its requested 8 GPUs), isn't it?
Another question is: how to prioritize these two quotas upon preemption? By Pod's priority?

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
Huang-Wei: +1 to Yuan. I think here the "if clause" in the 2 bullets should be reversed.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: > In this case, it doesn't help if UserB get another 3GPU (less than its requested 8 GPUs), isn't it?
-----
I'm not sure I understand this question： the requested doesn't represent pod.spec.resources.requested. It means the sum of the resource requests for many jobs. 3GPU means the UserB get some resource to run the part of jobs.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: > If evicting a single pod in namespace cannot release enough resources for the pod to be scheduled, will it continue to evict more pods in the same namespace or choose another one?
----
The current design does not differentiate between different namespaces. Just Inter-Quota or Intra-Quota. If `if Preemptor.Request + Quota.allocated > Quota.min`, the preempt will happen Inter-Quota. All the pods belongs to other Quotas that allocates more resources than its min in a node will be selected as the potential victims. The subsequent selection strategy remains the current logic: priority, PDB, createtime.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: Thanks Done

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: Done. I change it to `once the Pod that consumes the entire quota is created and the namespace reaches its quota limit, no other Pods can be created - even if the Pod failed to be scheduled, or failed on running (error when pulling image, etc.).`

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: > Should it be opposite? " Preemptor.Request + Quota.allocated <= Quota.min"
---
Done thanks

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
```diff
@@ -8,11 +8,14 @@
     - [Goals](#goals)
```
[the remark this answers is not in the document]
denkensk: Done

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: Done

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: > how to prioritize these two quotas upon preemption? By Pod's priority?
---

The current design does not differentiate between different namespaces. Just Inter-Quota or Intra-Quota. If if Preemptor.Request + Quota.allocated > Quota.min, the preempt will happen Inter-Quota. All the pods belongs to other Quotas that allocates more resources than its min in a node will be selected as the potential victims. The subsequent selection strategy remains the current logic: priority, PDB, createtime.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: > we can start with namespace, and try other multi-tenancy APIs such as hierarchical namespace, etc, at a later time.
---
I agree with this. In the first version, we can divide the ElasticQuota based on the namespace. We can try to define a higher level of quota to associate with multi-tenant like heretical namespace or virtual cluster from Guo Fei at a later time.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: Yes. When 
1. there are free resources in the cluster
2.the allocated of quota has not reached max

The preemption will happen If one of the two items is not satisfied.

denkensk:
/cc @Huang-Wei @yuanchen8911
Please review it again, thanks

xq2005:
Do we need a section about using webhook to validate ElasticQuota based on the existing ResourceQuota?

denkensk:
> Do we need a section about using webhook to validate ElasticQuota based on the existing ResourceQuota?

---
@xq2005  Thanks.  Webhook and Controller will be discussed in the next release.

denkensk:
I squash the commits. Please take a look again.
/cc @Huang-Wei

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+There is increasing demand to use Kubernetes to manage batch workloads 
+(ML/DL). In those cases, one challenge is to improve cluster utilization 
+while ensuring that each user has a reasonable amount of resources. The 
+problem can be partially addressed by the Kubernetes [ResourceQuota
```
Huang-Wei, line 80:
```suggestion
problem can be partially addressed by the Kubernetes [ResourceQuota].
```

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+There is increasing demand to use Kubernetes to manage batch workloads 
+(ML/DL). In those cases, one challenge is to improve cluster utilization 
+while ensuring that each user has a reasonable amount of resources. The 
+problem can be partially addressed by the Kubernetes [ResourceQuota 
+(https://kubernetes.io/docs/concepts/policy/resource-quotas/). The native
```
Huang-Wei, line 81: You can define the link in the end of this section:

```
[ResourceQuota]: https://kubernetes.io/docs/concepts/policy/resource-quotas/
```

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+(e.g., Pod cpu/mem requests specified in the Pod spec). Although this 
+mechanism can guarantee that the actual resource consumption will never 
+exceed the ResourceQuota limit, it might lead to low resource utilization as 
+some pods may have claimed the resources but failed to be scheduled. For 
+instance, actual resource consumption may be much smaller than the limit.
+
+Due to above limitation, the batch workloads (ML/DL) can't run in a 
+Kubernetes cluster as efficiently as they do in other container orchestration
+platforms such as Yarn.In order to overcome above limitations, an 
+"ElasticQuota" concept used in [Yarn capacity scheduler](https://www
```
Huang-Wei, line 98: This cannot be truncated. Same as above, define this link at the end of this section.

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+.netjstec.com/2018/04/capacity-scheduler-in-yarn-hadoop.html) can be 
+leveraged into Kubernetes. Basically, the "ElasticQuota" has the notions of 
+"max" and "min".
+
+- max: the upper bound of the resource consumption of the consumers.
+
+- min: the minimum resources that are guranteed to ensure the basic
```
Huang-Wei, line 105: s/guranteed/guaranteed/

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+	Min  ResourceList
+	Max  ResourceList
+	Used ResourceList
+}
+
+// ResourceName is the name identifying various resources in a ResourceList.
+type ResourceName string
+
+// ResourceList is a set of (resource name, quantity) pairs.
+type ResourceList map[ResourceName]resource.Quantity
```
Huang-Wei, line 174: This has already been defined in upstream, right? If so, we can reuse the upstream version.

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+	metav1.TypeMeta
+	metav1.ObjectMeta
+	Spec   ElasticQuotaSpec
+	Status ElasticQuotaStatus
+}
+
+// ElasticQuotaSpec defines the Min and Max for Quota.
+type ElasticQuotaSpec struct {
+	Min ResourceList
+	Max ResourceList
```
Huang-Wei, line 160: Is `Guaranteed` / `BestEfforted` or (`Guaranteed` / `Limits`) more descrptive?

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,323 @@
...
+}
+
+// ElasticQuotaStatus defines the observed use.
+type ElasticQuotaStatus struct {
+	Min  ResourceList
+	Max  ResourceList
+	Used ResourceList
```
Huang-Wei, line 167: nit: `Occupied`?

kep/9-capacity-scheduling/kep.yaml
```diff
@@ -0,0 +1,26 @@
...
+creation-date: 2020-06-22
+reviewers:
+  - "@ahg-g"
+  - "@Huang-Wei"
+  - "@alculquicondor"
+  - "@yuanchen8911"
+approvers:
+  - "Huang-Wei"
+stage: alpha
+latest-milestone: "v1.20"
```
Huang-Wei, line 19: should be 1.19

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
Huang-Wei: This comment is still outstanding.

kep/9-capacity-scheduling/kep.yaml
denkensk: Done

kep/9-capacity-scheduling/README.md
denkensk: I think it's better to keep the same with `ResourceQuota` to future integration.
FYI 

https://github.com/kubernetes/kubernetes/blob/4db3a096ce8ac730b2280494422e1c4cf5fe875e/pkg/apis/core/types.go#L4838-L4845

kep/9-capacity-scheduling/README.md
denkensk: I think it's better to keep the same with Yarn for user to understand quickly.

kep/9-capacity-scheduling/README.md
denkensk: Done

kep/9-capacity-scheduling/README.md
denkensk: Done

kep/9-capacity-scheduling/README.md
denkensk: Done

kep/9-capacity-scheduling/README.md
denkensk: Done

kep/9-capacity-scheduling/README.md
denkensk: Done

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
denkensk: Done

denkensk:
/cc @Huang-Wei  
Update again. Please review it again. Thanks so much.

kep/9-capacity-scheduling-for-ml-dl-workloads-based-on-scheduler-framework/README.md
yuanchen8911: However, would it be possible to support a two-level structure, which will cover most use cases. For example, within each namespace, there can be multiple queues with min and max quotas.  I hope it won't introduce much complexity.

kep/9-capacity-scheduling/README.md
Huang-Wei: SG.

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,324 @@
...
+since UserB is not using them at that moment. 
+
+![3](./3.png)
+
+Later, if UserB demands 3 more GPUs compared to its current consumption, 
+its request cannot be immediately satisfied since only one unused GPU left in
+the cluster. Since UserB's "min" quota is 6 GPUs, the preemption algorithm 
+ensures that three GPUs will be returned back by preempting a pod which 
+requested 2 GPUs from UserA, i.e., from 6/3 to 4/6. Then if UserB requests 1 
+or 2 more, it won't be succeed as GPUs in UserA by that moment are all
```
Huang-Wei, line 257:
```suggestion
or 2 more GPUs, it won't succeed as GPUs in UserA by that moment are all 
```

denkensk:
I squash the commits. Please take a look again.
/cc @Huang-Wei

Huang-Wei:
/lgtm
/approve

Thanks @denkensk ! Looking forward to seeing it gets implemented.

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,324 @@
...
+We will reimplement the method `selectVictimsOnNode` in `defaultPreempt`. The
+original `selectVictimsOnNode` method selects all the pods with the lower
+priority than the preemptor’s priority as potential victims in a node. 
+
+In this proposal, the potential victims selection logic needs to be changed 
+in the following cases.
+
+- if Preemptor.Request + Quota.allocated <= Quota.min: It means that its min 
+or guaranteed resource is used or `borrowed` by other Quota. Potential 
+victims in a node will be chosen from Quotas that allocates more resources
```
xujyan, line 283: Even in this case, should the victim selection still go through the normal criteria (e.g., picking lower priority pods) and select pods from namespace that exceeded min quotas only as the last resort, e.g., if lower priority pods cannot be found?

This way if some pods are running on borrowed "high priority", they are nevertheless still "high priority" and maybe still deserve preferential treatment than lower priority pods when facing preemption?

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,324 @@
...
+In this proposal, the potential victims selection logic needs to be changed 
+in the following cases.
+
+- if Preemptor.Request + Quota.allocated <= Quota.min: It means that its min 
+or guaranteed resource is used or `borrowed` by other Quota. Potential 
+victims in a node will be chosen from Quotas that allocates more resources 
+than its min, i.e., borrowing resources from other Quotas.
+- if Preemptor.Request + Quota.allocated > Quota.min: It means that its 
+guaranteed isn't borrowed by other quotas. So that we will select the pods 
+belongs to the same quota(namespace) with the lower priority than the
```
xujyan, line 287: Today with regular quotas I don't think the scheduler necessarily restrict to the namespace when picking victims, should this rule also apply here?

kep/9-capacity-scheduling/README.md
```diff
@@ -0,0 +1,324 @@
...
+greater than or equal to max. 
+
+- max: it will check the resource in the Prefilter of Pod Scheduling Cycle. 
+The usage of max is aggregated based on the resource configurations of 
+successfully scheduled pods.
+
+- min: it means the guaranteed resource of Quota.
+
+min and max need to respect the namespace's ResourceQuota limit, i.e., `min 
+<= max <= hard` .
```
xujyan, line 215: The relationship need to be established [per PriorityClass if they are so specified](https://kubernetes.io/docs/concepts/policy/resource-quotas/#resource-quota-per-priorityclass), right?

This would imply that `ElasticQuota` also need to support priority classes?

kep/9-capacity-scheduling/README.md
denkensk: I've been thinking about this place for a long time.
> Even in this case, should the victim selection still go through the normal criteria (e.g., picking lower priority pods first) and select pods from a namespace that exceeded min quotas only as the last resort, e.g.

You means we select the lower priority pods firstly. But in this case, the pods belongs to elastic quota which is less than min will be selected as victims. And this pod will be deleted. Then the resource can't be guaranteed.

PR #15, opened 2020-06-28
update coscheduling stage to beta

denkensk:
Because Coscheduling plugin is used in companies now. So I update coscheduling stage to beta.
/cc @Huang-Wei

### DIFF
--- a/pkg/coscheduling/README.md
+++ b/pkg/coscheduling/README.md
@@ -8,8 +8,8 @@ sorting](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/kep/20
 <!-- Check one of the values: Sample, Alpha, Beta, GA -->
 
 - [ ] 💡 Sample (for demonstrating and inspiring purpose)
-- [x] 👶 Alpha (used in companies for pilot projects)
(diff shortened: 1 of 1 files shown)


Huang-Wei:
@denkensk is it possible to share some data, customer, scenarios, or which Alibaba cloud service is adopting the coscheduling plugin? If so, it would be way more convincing for more users to user, as well as developers to contribute.

wsxiaozhang:
> @denkensk is it possible to share some data, customer, scenarios, or which Alibaba cloud service is adopting the coscheduling plugin? If so, it would be way more convincing for more users to use, as well as developers to contribute.

@Huang-Wei , I think a table with 5 columns including "user/org name", "contact", "which plugins used", "phase", "scenarios" should be good for collecting usage info. We have some users who would like to share their info with public.

Huang-Wei:
@wsxiaozhang That looks nice! Could you add that into this PR?

That table can be put into https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/adopters (renamed to adopters.md looks better), and then linked back here.

Huang-Wei:
BTW: ideally I'd expect #19 to be fixed if updating to beta.

PR #17, opened 2020-06-29
QoS QueueSort: rename files&plugin and fix a comment

yuanchen8911:
This PR made the following changes to QoS QueueSort plugin by following the kubernetes convention.
  -  File name: queuesort.go -> queue_sort.go
  -  File name: queuesort_test.go -> queue_sort_test.go
  -  Plugin name: QoSSort -> QOSSort
  -  Fixed the comment of Less()


yuanchen8911:
The same failed integration test as coscheduling. Did I miss to change something?  Testing in our local clusters worked just fine.

pkg/qos/queue_sort.go
```diff
@@ -25,22 +25,22 @@ import (
 )
 
 // Name is the name of the plugin used in the plugin registry and configurations.
-const Name = "QoSSort"
+const Name = "QOSSort"
 
-// QoSSort is a plugin that implements QoS class based sorting.
-type Sort struct{}
+// QOSSort is a plugin that implements QoS class based sorting.
+type QOSSort struct{}
```
Huang-Wei, line 31: It's sometimes conventional to not include the package name (qos) in the type name. Ref: https://github.com/golang/go/wiki/CodeReviewComments#package-names.

I would suggest to keep it as, or later renamed to `GenericSort`.

Huang-Wei:
> The same failed integration test as coscheduling. Did I miss to change something? Testing in our local clusters worked just fine.

I created issue #19 to track this.

pkg/qos/queue_sort.go
yuanchen8911: > It's sometimes conventional to not include the package name (qos) in the type name. Ref: https://github.com/golang/go/wiki/CodeReviewComments#package-names.
> 
> I would suggest to keep it as, or later renamed to `GenericSort`.
Agree, changed it back to Sort.

yuanchen8911: If QueueSort is just one of plugins in package "qos" and other QoS related plugins can be added to the package in the future, the package name "qos" makes sense.  I'll submit a PR on the GenericSort plugin based on our discussions.  Thanks.

yuanchen8911: > > The same failed integration test as coscheduling. Did I miss to change something? Testing in our local clusters worked just fine.
> 
> I created issue #19 to track this.

After reverting the QueueSort plugin struct name to "Sort", it passed the integration test.

yuanchen8911:
@Huang-Wei The changes are straightforward and  minor. Any comments or concerns?  The same thing for the coscheduling PR. After they are merged, we will then look at more significant changes as discussed. Thanks.

Huang-Wei:
Thanks.

/lgtm
/approve

PR #18, opened 2020-06-30
TEST ONLY and will DELET

yuanchen8911:
Test if the current version will pass the integration test.


yuanchen8911:
Test is done.

Issue #20, opened 2020-06-30
coscheduling: check minAvailable of all pods in a PodGroup


yuanchen8911:
I think we should check minAvailable value of each pod in a PodGroup and make sure the values of the same PodGroup are the same. Also, minAvailable can be part of PodGroupInfo as the KEP does. WDYT? @Huang-Wei @denkensk

denkensk:
Thanks @yuanchen8911  It's make sense to me.  
This's the weakness to define `PodGroup` by the label of Pod. We can only assume that minAvailable value of each pod in a PodGroup are same. I agree to check minAvailable value in `PreFilter`.  But how to deal with these pods with different value of the same PodGroup? Maybe we can reject it in `PreFilter` but it will reschedule later.

yuanchen8911:
> Thanks @yuanchen8911  It's make sense to me.  
> 
> This's the weakness to define `PodGroup` by the label of Pod. We can only assume that minAvailable value of each pod in a PodGroup are same. I agree to check minAvailable value in `PreFilter`.  But how to deal with these pods with different value of the same PodGroup? Maybe we can reject it in `PreFilter` but it will reschedule later.

A simpler solution is to (1) set the number according to the first pod’s label value in getPodGroupInfo (2) ignore the other pods’ label values inPreFilter  (can check and report a warning if different in log). This should be consistent with the future CRD implementation. @denkensk

yuanchen8911:
> > Thanks @yuanchen8911  It's make sense to me.  
> > 
> > This's the weakness to define `PodGroup` by the label of Pod. We can only assume that minAvailable value of each pod in a PodGroup are same. I agree to check minAvailable value in `PreFilter`.  But how to deal with these pods with different value of the same PodGroup? Maybe we can reject it in `PreFilter` but it will reschedule later.
> 
> A simple solution is to (1) set the number according to the first pod’s label value in getPodGroupInfo (2) ignore the other pods’ label values inPreFilter  (can check and report a warning if different in log). This should be consistent with the future CRD implementation. @denkensk 

The changes required are more complex than what I thought earlier. 

How about considering a regular pod as a PodGroup with a single pod (stored in the PodGroupInfo map). We hence will have a uniform way to compare and process both regular pods and pod groups. For a  regular pod, we can just use its pod name as the group name. 

It will introduce additional overhead for scheduling regular pods (hope not much), but the design and implementation would be improved and more extendable. 

[...]

PR #21, opened 2020-07-01
[WIP] A refactoring of coscheduling code

yuanchen8911:
The PR aims to improve the current coscheduling implementation by using PodGroupInfo as the main data structure and unifying the processing of regular pods and PodGroup to simplify the plugin implementation. It also improves error checking. Please refer to issue #20. 

1. Extend PodGroupInfo to include all PodGroup related information and use it as the only information source for plugins. `PodGroupLabels` is only used to get `podGroupName` and `minAvailable` and generate the key for `PodGroupInfos` map. All plugins use the data in`PodGroupInfo`. 

```
type PodGroupInfo struct {
    // key is a unique PodGroup ID and currently implemented as <namespace>/<podgroup name>.
    key string
    // name is the PodGroup name and defined through a Pod label.
    // The PodGroup name of a regular pod is the pod name.
    name string
    // priority is the priority of pods in a PodGroup.
    // All pods in a PodGroup should have the same priority.
    priority int32
    // timestamp stores the timestamp of the initialization time of a PodGroup.
    timestamp time.Time
    // minAvailable is the minimum number of pods to be co-scheduled in a PodGroup.
    // All pods in a PodGroup should have the same minAvailable.
    minAvailable int32
}
```

2. A regular pod is considered a PodGroup with a single Pod. The PodGroup name is the pod name. As a result, plugins handle PodGroups and regular pods in the same way.  

3. Explicitly separate `getPodGroupInfo` and `setPodGroupInfo`. The former obtains PodGroup information from the map while the latter create a new PodGroup (if not existing) or update the existing PodGroupInfo (called by `Less` and `PreFilter`).   

```
// getPodGroupInfo returns the PodGroup that a pod belongs to.
func (cs *Coscheduling) getPodGroupInfo(p *v1.Pod) (*PodGroupInfo, bool) {
    podGroupName, _, _ := GetPodGroupLabels(p)
    key := getPodGroupKey(p.Namespace, podGroupName)
    pgInfo, exist := cs.podGroupInfos.Load(key)
    if !exist {
        return nil, false
    }
    return pgInfo.(*PodGroupInfo), true
}
```

```
// setPodGroupInfo creates or updates a PodGroup and returns it.
// (1) Create a new PodGroup if the PodGroup does not exist.
// (2) Update minAvailable and priority values of the existing PodGroup.
// (3) Return the new created or existing PodGroup.
func (cs *Coscheduling) setPodGroupInfo(p *v1.Pod, t time.Time) *PodGroupInfo {
    podGroupName, minAvailable, _ := GetPodGroupLabels(p)
    key := getPodGroupKey(p.Namespace, podGroupName)
    pgInfo, exist := cs.podGroupInfos.Load(key)
    if !exist {
        pgInfo = &PodGroupInfo{
            name:         podGroupName,
            key:          key,
            priority:     pod.GetPodPriority(p),
            timestamp:    t,
            minAvailable: minAvailable,
        }
        cs.podGroupInfos.Store(key, pgInfo)
    } else {
        pgInfo.(*PodGroupInfo).minAvailable = minAvailable
        pgInfo.(*PodGroupInfo).priority = pod.GetPodPriority(p)
    }
    return pgInfo.(*PodGroupInfo)
}
```

3. `setPodGroupInfo`creates a new PodGroup if the PodGroup does not exist. If a PodGroup exists, it updates the existing PodGroup's `minAvailable` and `priority`.  Without a mechanism to remove/update the stale PodGroup data in PodGroupInfos map, such an update is an improvement over the current implemenation as it can handle scenarios when an existing pod/podgroup is deleted, updated (e.g., minAvailable or resource spec.) and then relaunched. 

4.  Check if all the pods of a PodGroup have the same priority and minAvailable and report a warning if not.

5. Modified the test file accordingly. With the above changes, each individual plugin test will need to call setPodGroupInfo first.

6. Fixed some typos and made various wording changes to comments.

Tests in a local cluster worked as expected.

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -36,16 +36,26 @@ import (
 type Coscheduling struct {
 	frameworkHandle framework.FrameworkHandle
 	podLister       corelisters.PodLister
-	// Key is the name of Namespace/PodGroup.
+	// key is <namespace>/<PodGroup name> and value is *PodGroupInfo.
 	podGroupInfos sync.Map
 }
 
 // PodGroupInfo is a wrapper to a PodGroup with additional information.
-// TODO implement a timeout based gc for the PodGroupInfos map
+// TODO: implement a timeout based gc for the PodGroupInfos map.
 type PodGroupInfo struct {
+	// key is a unique PodGroup ID and currently implemented as <namespace>/<podgroup name>.
+	key string
+	// name is the PodGroup name and defined through a Pod label.
+	// The PodGroup name of a regular pod is the pod name.
 	name string
-	// timestamp stores the timestamp of the initialization time of PodGroup.
+	// priority is the priority of pods in a PodGroup.
+	// All pods in a PodGroup should have the same priority.
+	priority int32
+	// timestamp stores the timestamp of the initialization time of a PodGroup.
 	timestamp time.Time
+	// minAvailable is the minimum number of pods to be co-scheduled in a PodGroup.
+	// All pods in a PodGroup should have the same minAvailable.
+	minAvailable int32
 }
 
 var _ framework.QueueSortPlugin = &Coscheduling{}
@@ -60,7 +70,7 @@ const (
 	PodGroupName = "pod-group.scheduling.sigs.k8s.io/name"
 	// PodGroupMinAvailable specifies the minimum number of pods to be scheduled together in a pod group.
 	PodGroupMinAvailable = "pod-group.scheduling.sigs.k8s.io/min-available"
-	// PermitWaitingTime is the wait timeout returned by Permit plugin
+	// PermitWaitingTime is the wait timeout returned by Permit plugin.
 	// TODO make it configurable
 	PermitWaitingTime = 1 * time.Second
 )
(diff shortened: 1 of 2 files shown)


yuanchen8911:
@Huang-Wei  @denkensk   Can you review it when you get a chance? Thanks.

yuanchen8911:
@denkensk  @Huang-Wei  If we feel there are too many changes in this PR, I can break down them into multiple small PRs. I'm looking for your input on what changes are helpful and reasonable. I'd expect multiple rounds of iterations. Thank you.

denkensk:
Thanks @yuanchen8911 
> 2. A regular pod is considered a PodGroup with a single Pod. The PodGroup name is the pod name. As a result, plugins handle PodGroups and regular pods in the same way.
I'm worried that this will affect the performance of regular pod if we create a PodGroup for every single Pod.  Because the podGroupInfos is `sync.Map`. If the cluster is dominated by regular pod. This may cause performance degradation if there are too many reads and writes in podGroupInfos. WDYT? @yuanchen8911

yuanchen8911:
> Thanks @yuanchen8911 
> 
> > 2. A regular pod is considered a PodGroup with a single Pod. The PodGroup name is the pod name. As a result, plugins handle PodGroups and regular pods in the same way.
> 
> 
> 
> I'm worried that this will affect the performance of regular pod if we create a PodGroup for every single Pod.  Because the podGroupInfos is `sync.Map`. If the cluster is dominated by regular pod. This may cause performance degradation if there are too many reads and writes in podGroupInfos. WDYT? @yuanchen8911 
> 
> 

Yes, this will introduce additional overhead for regular pods. It’s a valid concern. I mentioned it in the issue discussion. Maybe we can run some performance experiments to compare two implementations?

yuanchen8911:
Will have a meeting with @denkensk to discus this PR and other features next Monday.

yuanchen8911:
> > Thanks @yuanchen8911
> > > 1. A regular pod is considered a PodGroup with a single Pod. The PodGroup name is the pod name. As a result, plugins handle PodGroups and regular pods in the same way.
> > 
> > 
> > I'm worried that this will affect the performance of regular pod if we create a PodGroup for every single Pod.  Because the podGroupInfos is `sync.Map`. If the cluster is dominated by regular pod. This may cause performance degradation if there are too many reads and writes in podGroupInfos. WDYT? @yuanchen8911
> 
> Yes, this will introduce additional overhead for regular pods. It’s a valid concern. I mentioned it in the issue discussion. Maybe we can run some performance experiments to compare two implementations?

@denkensk  Here is a discussion about sync.Map performance.  https://github.com/robaho/go-concurrency-test. TL;DR, using locks with built-in map can have better performance. We should probably change it to lock + regular map anyway.

yuanchen8911:
Resubmitted a revised version #24 and close this one.

PR #22, opened 2020-07-15
simplify coscheduling unit test

Huang-Wei:
This PR simplifies the coscheduling unit test:

- reuse scheduling testing utility to craft Pod object
- remove unnecessary fake artifacts
- simplify the UT's logic as well as making it more stateless


Huang-Wei:
@yuanchen8911 Would you like to review this?

denkensk:
@Huang-Wei  Thanks.
/lgtm

PR #23, opened 2020-07-15
[TEST] Try fix test error

denkensk:


Huang-Wei:
@denkensk The [past run](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_scheduler-plugins/23/pull-scheduler-plugins-integration-test/1283386143190028288) is actually very close to 10 min timeout:

```
ok  	sigs.k8s.io/scheduler-plugins/test/integration	460.321s
```

Issue #19, opened 2020-06-30
Integration test needs to be improved


Huang-Wei:
The integration test, esp. the coscheduling test, is slow and easily cause flakes. 

In my mac, it takes 550s to finish:

```
⇒  hack/integration-test.sh
+++ [0629 17:17:27] Checking etcd is on PATH
+++ [0629 17:17:27] Starting etcd instance
etcd --advertise-client-urls http://127.0.0.1:2379 --data-dir /var/folders/d8/p4t60rzs7rs6b6vmppqmth300000gn/T/tmp.s95okkq8Yk --listen-client-urls http://127.0.0.1:2379 --debug > "/dev/null" 2>/dev/null
Waiting for etcd to come up.
+++ [0629 17:17:28] On try 2, etcd: : {"health":"true"}
{"header":{"cluster_id":"14841639068965178418","member_id":"10276657743932975437","revision":"2","raft_term":"2"}}+++ [0629 17:17:28] Running integration test cases
ok  	sigs.k8s.io/scheduler-plugins/test/integration	550.233s
+++ [0629 17:28:39] Cleaning up etcd
+++ [0629 17:28:39] Integration test cleanup complete
```

However, the CI env would time out after 10 mins. I would expect the test to finish within 2 mins.

/assign @denkensk

yuanchen8911:
After reverting the QueueSort plugin struct name to "Sort", it passed the integration test.

yuanchen8911:
After reverting the QueueSort plugin struct name to "Sort", it passed the integration test.

Huang-Wei:
I don’t think that’s related lol
Yuan Chen <notifications@github.com> wrote:
“After reverting the QueueSort plugin struct name to "Sort", it passed the integration test.”

yuanchen8911:
Here's my experience with the recent PRs. It happened a few times that a same PR failed the integration first and then a  retest without any changes passed. Similarly, it passed the integration test first and another test with only typo fixes in  comments failed. There must be something wrong. @Huang-Wei  @denkensk

PR #26, opened 2020-07-17
Temporary disable coscheduling integration test

Huang-Wei:
Until #19 gets fixed, we should disable the coscheduling integration test to not impact other code changes.


Huang-Wei:
@denkensk sure thing. I will also take a look when I have time.

BTW: "/lgtm" needs to be at one line by itself.

PR #27, opened 2020-07-17
improve the current pod in accurary

DeliangFan:
Current Pod = Waited Pod + Binded Pod

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -164,20 +164,19 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	}
 
 	namespace := p.Namespace
-	// TODO get actually scheduled(bind successfully) account from the SharedLister
-	running := cs.calculateRunningPods(podGroupName, namespace)
+	bound := cs.calculateBoundPods(podGroupName, namespace)
 	waiting := cs.calculateWaitingPods(podGroupName, namespace)
-	current := running + waiting + 1
+	current := bound + waiting + 1
 
 	if current < minAvailable {
-		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: running(%d), waiting(%d)",
-			p.Namespace, podGroupName, p.Name, minAvailable, running, waiting)
+		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: bound(%d), waiting(%d)",
+			p.Namespace, podGroupName, p.Name, minAvailable, bound, waiting)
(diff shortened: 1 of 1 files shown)


pkg/coscheduling/coscheduling.go
```diff
@@ -164,20 +164,19 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	}
 
 	namespace := p.Namespace
-	// TODO get actually scheduled(bind successfully) account from the SharedLister
-	running := cs.calculateRunningPods(podGroupName, namespace)
+	binded := cs.calculateBindedPods(podGroupName, namespace)
```
Huang-Wei, line 167: Hmm... I don't quite think `bound` is better than `running`. Usually `running` means a Pod admitted by kubelet already, while `bound` just means the Pod's `.spec.nodeName` is already set.

pkg/coscheduling/coscheduling.go
```diff
@@ -233,10 +232,17 @@ func (cs *Coscheduling) calculateTotalPods(podGroupName, namespace string) int {
...
-			return true
+		if pod.Labels[PodGroupName] == podGroupName && pod.Namespace == namespace {
+			if pod.Status.Phase == v1.PodRunning {
+				return true
+			}
+			for _, c := range pod.Status.Conditions {
+				if c.Type == v1.PodScheduled || c.Type == v1.PodInitialized || c.Type == v1.ContainersReady || c.Type == v1.PodReady {
+					return true
+				}
+			}
```
Huang-Wei, line 245: Why this is needed?

DeliangFan: - running := cs.calculateRunningPods(podGroupName, namespace)
It only calculate the pods has bound to a node and all of the containers have been started, which may miss these bound pod whose PodPhase is still in Pending before containers start during pull images.

Is my opinion correct？

Huang-Wei: That's correct - running means at least one of the containers is already running.

Given that the core scheduler doesn't really look at the phase of a Pod, @denkensk do you think we should just check a Pod's spec.nodeName here?

DeliangFan: After a deep dive into the scheduler framework, I also believe that check spec.nodeName is much better than the Status。Conditions

denkensk: Thanks @Huang-Wei  I agree to just check a Pod's nodeName.

denkensk: Thanks @DeliangFan
If we just check the Pod's `spec.nodeName`,  we need to rename `calculateRunningPods `  to `calculateBindedPods`.

DeliangFan: Done, please review again

PR #32, opened 2020-07-24
Clean up the expired  PodGroupInfo regularly

denkensk:
Let me introduce our proposal in detail for your understanding in review. It is mainly divided into two parts.
-  How to manage the deleted status of PodGroup
1. When to set the status to be `deleted`?
The podGroup info is created or updated in `getOrUpdatedPodGroupInfo`.  The pods in podGroup can be divided into two types: scheduled or unscheduledPod. So we have different ways to clean up their PodGroup Info for these two kinds of pods. (1) mark the status as deleted when the pod scheduled success in `permit` (2) mark the status as deleted when the unschedulePod is deleted by event watch.

2. Why we marks the PodGroup as deletable when its first Pod gets deleted?
We have two choice to mark the PodGroup as deletable. 1. when its first pod gets deleted. 2 when the all the pods of PodGroup are deleted. The second will lead more frequent operations of sync.map because we need to store the pod in PodGroupInfo and rewrite it when one pods are added updated or deleted. it will degrade performance much. So we choose the set the PodGroup deleted when its first Pod gets deleted. 

2. When to remove the deleted status of PodGroup?
Sometimes after the pods of podGroup are deleted, there are still some pods left or created. We will remove the deleted status of PodGroup When it invokes `getOrUpdatedPodGroupInfo`
```go
value, exist := cs.podGroupInfos.Load(pgKey)
if exist {
    pgInfo := value.(*PodGroupInfo)
    if pgInfo.deletionTimestamp != nil {
        pgInfo.deletionTimestamp = nil
        cs.podGroupInfos.Store(pgKey, pgInfo)
    }
    return pgInfo, podMinAvailable
}
```
If the podGroup exits in podGroupInfos but its status is deleted, we will remove the deleted status of PodGroup and rewrite the podGroupInfos

- Regularly clean up the expired podGroup in go-routine.

The expiration time is 10min. The expired podGroup will be deleted by goroutine. The check operation runs every 30 seconds.

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -23,22 +23,26 @@ import (
 	"sync"
 	"time"
 
-	v1 "k8s.io/api/core/v1"
+	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+	"k8s.io/kubernetes/pkg/scheduler/util"
 )
 
 // Coscheduling is a plugin that implements the mechanism of gang scheduling.
-// TODO: implement a timeout based gc to clean up stale PodGroupInfo in the PodGroupInfos map.
 type Coscheduling struct {
 	frameworkHandle framework.FrameworkHandle
 	podLister       corelisters.PodLister
 	// key is <namespace>/<PodGroup name> and value is *PodGroupInfo.
 	podGroupInfos sync.Map
+	// clock is used to get the current time.
+	clock util.Clock
 }
 
 // PodGroupInfo is a wrapper to a PodGroup with additional information.
@@ -58,6 +62,8 @@ type PodGroupInfo struct {
 	// minAvailable is the minimum number of pods to be co-scheduled in a PodGroup.
 	// All pods in a PodGroup should have the same minAvailable.
 	minAvailable int
+	// deletionTimestamp stores the timestamp when the PodGroup marked as expired.
+	deletionTimestamp *time.Time
 }
 
 var _ framework.QueueSortPlugin = &Coscheduling{}
@@ -75,6 +81,11 @@ const (
 	// PermitWaitingTime is the wait timeout returned by Permit plugin.
 	// TODO make it configurable
 	PermitWaitingTime = 1 * time.Second
+	// PodGroupGCInterval is the period to run gc of PodGroup.
+	PodGroupGCInterval = 30 * time.Second
+	// If the deleted PodGroup stays longer than the PodGroupExpirationTime,
+	// The deleted PodGroup will be deleted from PodGroupInfos.
+	PodGroupExpirationTime = 10 * time.Minute
 )
 
 // Name returns name of the plugin. It is used in logs, etc.
@@ -85,9 +96,34 @@ func (cs *Coscheduling) Name() string {
 // New initializes a new plugin and returns it.
 func New(_ *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
 	podLister := handle.SharedInformerFactory().Core().V1().Pods().Lister()
-	return &Coscheduling{frameworkHandle: handle,
+	cs := &Coscheduling{frameworkHandle: handle,
 		podLister: podLister,
-	}, nil
+		clock:     util.RealClock{},
+	}
+	podInformer := handle.SharedInformerFactory().Core().V1().Pods().Informer()
+	podInformer.AddEventHandler(
+		cache.FilteringResourceEventHandler{
+			FilterFunc: func(obj interface{}) bool {
+				switch t := obj.(type) {
+				case *v1.Pod:
+					return responsibleForPod(t)
+				case cache.DeletedFinalStateUnknown:
+					if pod, ok := t.Obj.(*v1.Pod); ok {
+						return responsibleForPod(pod)
+					}
+					return false
+				default:
+					return false
+				}
+			},
+			Handler: cache.ResourceEventHandlerFuncs{
+				DeleteFunc: cs.markPodGroupAsExpired,
+			},
+		},
+	)
+	go wait.Until(cs.podGroupInfoGC, PodGroupGCInterval, nil)
+
+	return cs, nil
 }
 
 // Less is used to sort pods in the scheduling queue.
@@ -130,9 +166,16 @@ func (cs *Coscheduling) getOrCreatePodGroupInfo(pod *v1.Pod, ts time.Time) (*Pod
 
 	// If it is a PodGroup and present in PodGroupInfos, return it.
 	if len(pgKey) != 0 {
-		pgInfo, exist := cs.podGroupInfos.Load(pgKey)
+		value, exist := cs.podGroupInfos.Load(pgKey)
 		if exist {
-			return pgInfo.(*PodGroupInfo), podMinAvailable
+			pgInfo := value.(*PodGroupInfo)
+			// If the deleteTimestamp isn't nil, it means that the PodGroup is marked as expired before.
+			// So we need to set the deleteTimestamp as nil again to mark the PodGroup active.
+			if pgInfo.deletionTimestamp != nil {
+				pgInfo.deletionTimestamp = nil
+				cs.podGroupInfos.Store(pgKey, pgInfo)
+			}
+			return pgInfo, podMinAvailable
 		}
 	}
 
@@ -291,4 +334,46 @@ func (cs *Coscheduling) calculateBoundPods(podGroupName, namespace string) int {
 	}
 
 	return len(pods)
+}
+
+// markPodGroupAsExpired set the deletionTimestamp of PodGroup to mark PodGroup as expired.
+func (cs *Coscheduling) markPodGroupAsExpired(obj interface{}) {
+	pod := obj.(*v1.Pod)
+	podGroupName, podMinAvailable, _ := GetPodGroupLabels(pod)
+	if len(podGroupName) == 0 || podMinAvailable == 0 {
+		return
+	}
+
+	pgKey := fmt.Sprintf("%v/%v", pod.Namespace, podGroupName)
+	// If it's a PodGroup and present in PodGroupInfos, set its deletionTimestamp.
+	value, exist := cs.podGroupInfos.Load(pgKey)
+	if !exist {
+		return
+	}
+	pgInfo := value.(*PodGroupInfo)
+	if pgInfo.deletionTimestamp == nil {
+		now := cs.clock.Now()
+		pgInfo.deletionTimestamp = &now
+		cs.podGroupInfos.Store(pgKey, pgInfo)
+	}
+}
+
+// responsibleForPod selects pod that belongs to a PodGroup.
+func responsibleForPod(pod *v1.Pod) bool {
+	podGroupName, podMinAvailable, _ := GetPodGroupLabels(pod)
+	if len(podGroupName) == 0 || podMinAvailable == 0 {
+		return false
+	}
+	return true
+}
+
+func (cs *Coscheduling) podGroupInfoGC() {
+	cs.podGroupInfos.Range(func(key, value interface{}) bool {
+		pgInfo := value.(*PodGroupInfo)
+		if pgInfo.deletionTimestamp != nil && pgInfo.deletionTimestamp.Add(PodGroupExpirationTime).Before(cs.clock.Now()) {
+			klog.V(3).Infof("%v is out of date and has been deleted in PodGroup GC", key)
+			cs.podGroupInfos.Delete(key)
+		}
+		return true
+	})
 }
--- a/pkg/coscheduling/coscheduling_test.go
+++ b/pkg/coscheduling/coscheduling_test.go
@@ -18,19 +18,32 @@ package coscheduling
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


pkg/coscheduling/coscheduling.go
```diff
@@ -304,3 +342,59 @@ func (cs *Coscheduling) calculateWaitingPods(podGroupName, namespace string) int
...
+func (cs *Coscheduling) podGroupInfoGC() {
+	for {
+		cs.podGroupInfos.Range(func(key, value interface{}) bool {
+			pgInfo := value.(*PodGroupInfo)
+			if pgInfo.deleted {
+				pgInfo.deleteTime.Add(PodGroupGCInterval).Before(time.Now())
+				cs.podGroupInfos.Delete(key)
+				klog.V(3).Infof("%v is out of date and has been deleted in PodGroup GC", key)
+			}
+			return true
```
yuanchen8911, line 397: The function has no return value.

pkg/coscheduling/coscheduling.go
```diff
@@ -304,3 +342,59 @@ func (cs *Coscheduling) calculateWaitingPods(podGroupName, namespace string) int
...
+	if len(podGroupName) > 0 && podMinAvailable > 0 {
+		pgKey := fmt.Sprintf("%v/%v", pod.Namespace, podGroupName)
+		// If it is a PodGroup and present in PodGroupInfos, set it's deleted true.
+		if len(pgKey) != 0 {
+			value, exist := cs.podGroupInfos.Load(pgKey)
+			if exist {
+				pgInfo := value.(*PodGroupInfo)
+				if !pgInfo.deleted {
+					pgInfo.deleteTime = time.Now()
+				}
```
Huang-Wei, line 381: It marks the PodGroup as deletable when its first Pod gets deleted, is it a reasonable assumption?

pkg/coscheduling/coscheduling.go
```diff
@@ -304,3 +342,59 @@ func (cs *Coscheduling) calculateWaitingPods(podGroupName, namespace string) int
...
+		var ok bool
+		pod, ok = t.Obj.(*v1.Pod)
+		if !ok {
+			klog.Errorf("cannot convert to *v1.Pod: %v", t.Obj)
+			return
+		}
+	default:
+		klog.Errorf("cannot convert to *v1.Pod: %v", t)
+		return
+	}
```
yuanchen8911, line 361: I'm not sure you need to check the error cases since the function is called by PermitPlugin and EventHandler only. The former should pass a valid Pod and the latter already checks the error cases  through FilterFunc.

yuanchen8911:
> Just core code review, haven't looked into tests.

the test file and cases should be updated too. 

> /assign @Huang-Wei
> /cc @yuanchen8911

Should the test file be updated too?

denkensk:
> It's not clear when and how to set `pgInfo.deleted`to true.

------

There is two ways to set `pgInfo deleted`:
1. the pods are scheduled
2. the unscheduled pod  is deleted.
@yuanchen8911

pkg/coscheduling/coscheduling.go
```diff
@@ -85,9 +91,35 @@ func (cs *Coscheduling) Name() string {
...
+				default:
+					return false
+				}
+			},
+			Handler: cache.ResourceEventHandlerFuncs{
+				DeleteFunc: cs.cleanPodGroupInfoIfPresent,
+			},
+		},
+	)
+	go podInformer.Informer().Run(nil)
```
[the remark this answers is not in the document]
denkensk: Yes, we don't need to start it again.

pkg/coscheduling/coscheduling.go
denkensk: We have two choice to mark the PodGroup as deletable. 1. when its first pod gets deleted. 2 when the all the pods of PodGroup are deleted.
The second will lead more frequent operations of sync.map because we will store the pod in PodGroupInfo, it will degrade performance much. So we choose the set the PodGroup deleted. If some pods of PodGroup still left, we will remove the deleted status in 
```go
value, exist := cs.podGroupInfos.Load(pgKey)
		if exist {
			pgInfo := value.(*PodGroupInfo)
			if pgInfo.deletionTimestamp != nil {
				pgInfo.deletionTimestamp = nil
				cs.podGroupInfos.Store(pgKey, pgInfo)
			}
			return pgInfo, podMinAvailable
		}
```

pkg/coscheduling/coscheduling.go
denkensk: Yes Done

pkg/coscheduling/coscheduling.go
```diff
@@ -58,6 +60,8 @@ type PodGroupInfo struct {
 	// minAvailable is the minimum number of pods to be co-scheduled in a PodGroup.
 	// All pods in a PodGroup should have the same minAvailable.
 	minAvailable int
+	deleted      bool
+	deleteTime   time.Time
```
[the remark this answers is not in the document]
denkensk: Done

pkg/coscheduling/coscheduling.go
```diff
@@ -304,3 +354,40 @@ func (cs *Coscheduling) calculateWaitingPods(podGroupName, namespace string) int
 
 	return waiting
 }
+
+// markPodGroupAsExpired set the deletionTimestamp of PodGroup.
+func (cs *Coscheduling) markPodGroupAsExpired(obj interface{}) {
+	pod := obj.(*v1.Pod)
+	podGroupName, podMinAvailable, _ := GetPodGroupLabels(pod)
```
yuanchen8911, line 361: Again, does it make sense to replace `GetPodGroupLabels` with ` getOrUpdatedPodGroupInfo` for consistency and extensibility?

denkensk: I thought about it before. We can't replace `GetPodGroupLabels ` with `getOrUpdatedPodGroupInfo `  here. Let me give you an example: 
pod1, pod2 belong to pg1.
1.pod1 is deleted. it will mark pg1 deleted.
2.pod2 is deleted. if we invoke `getOrUpdatedPodGroupInfo`, it will remove the deleted status of pod2 firstly，then In the following process, we will mark pg1 deleted again. (this operation will write sync.map again)

Note: we remove the deleted status of the same pg in `getOrUpdatedPodGroupInfo`.

denkensk:
Let me introduce our proposal in detail for your understanding in review. It is mainly divided into two parts.
-  How to manage the deleted status of PodGroup
1. When to set the status to be `deleted`?
The podGroup info is created or updated in `getOrUpdatedPodGroupInfo`.  The pods in podGroup can be divided into two types: scheduled or unscheduledPod. So we have different ways to clean up their PodGroup Info for these two kinds of pods. (1) mark the status as deleted when the pod scheduled success in `permit` (2) mark the status as deleted when the unschedulePod is deleted by event watch.

2. Why we marks the PodGroup as deletable when its first Pod gets deleted?
We have two choice to mark the PodGroup as deletable. 1. when its first pod gets deleted. 2 when the all the pods of PodGroup are deleted. The second will lead more frequent operations of sync.map because we need to store the pod in PodGroupInfo and rewrite it when one pods are added updated or deleted. it will degrade performance much. So we choose the set the PodGroup deleted when its first Pod gets deleted. 

2. When to remove the deleted status of PodGroup?
Sometimes after the pods of podGroup are deleted, there are still some pods left or created. We will remove the deleted status of PodGroup When it invokes `getOrUpdatedPodGroupInfo`
```go
value, exist := cs.podGroupInfos.Load(pgKey)
if exist {
    pgInfo := value.(*PodGroupInfo)
    if pgInfo.deletionTimestamp != nil {
        pgInfo.deletionTimestamp = nil
        cs.podGroupInfos.Store(pgKey, pgInfo)
    }
    return pgInfo, podMinAvailable
}
```
If the podGroup exits in podGroupInfos but its status is deleted, we will remove the deleted status of PodGroup and rewrite the podGroupInfos

- Regularly clean up the expired podGroup in go-routine.

The expiration time is 10min. The expired podGroup will be deleted by goroutine. The check operation runs every 30 seconds.

denkensk:
> Let me introduce our proposal in detail for your understanding in review. It is mainly divided into two parts.
> 
> * How to manage the deleted status of PodGroup
> 
> 1. When to set the status to be `deleted`?
>    The podGroup info is created or updated in `getOrUpdatedPodGroupInfo`.  The pods in podGroup can be divided into two types: scheduled or unscheduledPod. So we have different ways to clean up their PodGroup Info for these two kinds of pods. (1) mark the status as deleted when the pod scheduled success in `permit` (2) mark the status as deleted when the unschedulePod is deleted by event watch.
> 2. Why we marks the PodGroup as deletable when its first Pod gets deleted?
>    We have two choice to mark the PodGroup as deletable. 1. when its first pod gets deleted. 2 when the all the pods of PodGroup are deleted. The second will lead more frequent operations of sync.map because we need to store the pod in PodGroupInfo and rewrite it when one pods are added updated or deleted. it will degrade performance much. So we choose the set the PodGroup deleted when its first Pod gets deleted.
> 3. When to remove the deleted status of PodGroup?
>    Sometimes after the pods of podGroup are deleted, there are still some pods left or created. We will remove the deleted status of PodGroup When it invokes `getOrUpdatedPodGroupInfo`
> 
> ```go
> value, exist := cs.podGroupInfos.Load(pgKey)
> if exist {
>     pgInfo := value.(*PodGroupInfo)
>     if pgInfo.deletionTimestamp != nil {
>         pgInfo.deletionTimestamp = nil
>         cs.podGroupInfos.Store(pgKey, pgInfo)
>     }
>     return pgInfo, podMinAvailable
> }
> ```
> 
> If the podGroup exits in podGroupInfos but its status is deleted, we will remove the deleted status of PodGroup and rewrite the podGroupInfos
> 
> * Regularly clean up the expired podGroup in go-routine.
> 
> The expiration time is 10min. The expired podGroup will be deleted by goroutine. The check operation runs every 30 seconds.

------
@Huang-Wei @yuanchen8911  please take a look again. I introduce our proposal in detail at the top of this page.

pkg/coscheduling/coscheduling.go
```diff
@@ -85,9 +95,43 @@ func (cs *Coscheduling) Name() string {
...
+			},
+		},
+	)
+	go wait.Until(cs.podGroupInfoGC, PodGroupGCInterval, nil)
+
+	return cs, nil
+}
+
+// FakeNew is used for test.
+func FakeNew(informerFactory informers.SharedInformerFactory, clock util.Clock) (*Coscheduling, error) {
```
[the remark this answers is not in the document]
Huang-Wei: remove `informerFactory` if it's not necessary.

pkg/coscheduling/coscheduling.go
```diff
@@ -85,9 +95,43 @@ func (cs *Coscheduling) Name() string {
...
+	cs := &Coscheduling{
+		clock: clock,
+	}
+	go wait.Until(cs.podGroupInfoGC, PodGroupGCInterval, nil)
```
Huang-Wei, line 133: Do you want to replace nil with a stop channel, so that the UT won't leave the goroutine running forever.

pkg/coscheduling/coscheduling.go
```diff
@@ -224,6 +275,7 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 			waitingPod.Allow(cs.Name())
 		}
 	})
+	cs.markPodGroupAsExpired(pod)
```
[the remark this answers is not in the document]
denkensk: > such as when some "allowed" pods happen to invoke getOrCreatePodGroupInfo, the it marks the PG as non-expired again?
---
The “allowed” pods won't invoke `getOrCreatePodGroupInfo` again. But the left pods of this PodGroup maybe invoke  `getOrCreatePodGroupInfo` again to mark the PG as non-expired. We may have two choice to handle it.

1. markPodGroupAsExpired in `permit` here and if someone invoke `getOrCreatePodGroupInfo` again. We will mark the PG as non-expired again. So we just need to watch the UnschedulePodDelete in informer.

2. watch the podDelete in informer to delete the PodGroup.

WDYT? @Huang-Wei

pkg/coscheduling/coscheduling.go
denkensk: Done

pkg/coscheduling/coscheduling.go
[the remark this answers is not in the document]
denkensk: Done

pkg/coscheduling/coscheduling.go
```diff
@@ -292,3 +344,40 @@ func (cs *Coscheduling) calculateBoundPods(podGroupName, namespace string) int {
 
 	return len(pods)
 }
+
+// markPodGroupAsExpired set the deletionTimestamp of PodGroup to mark PodGroup as expired.
+func (cs *Coscheduling) markPodGroupAsExpired(obj interface{}) {
+	pod := obj.(*v1.Pod)
+	podGroupName, podMinAvailable, _ := GetPodGroupLabels(pod)
+	if len(podGroupName) > 0 && podMinAvailable > 0 {
```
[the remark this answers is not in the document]
denkensk: Thanks,  It will reduce indent level. Done

pkg/coscheduling/coscheduling.go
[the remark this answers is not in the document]
Huang-Wei: I think we can just have (2) in place, WDYT?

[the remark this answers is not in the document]
denkensk: (2) is also ok to me. I have addressed it. Please take a look again. Thank you. @Huang-Wei

Huang-Wei:
Thanks!

/lgtm
/approve
/hold
in case @yuanchen8911 wants to take a final look.

pkg/coscheduling/coscheduling.go
yuanchen8911: OK, it can be optimized in the future versions.

pkg/coscheduling/coscheduling.go
yuanchen8911: > Thanks!
> 
> /lgtm
> /approve
> /hold
> in case @yuanchen8911 wants to take a final look.

lgtm to me too. @Huang-Wei

yuanchen8911:
@Huang-Wei  LGTM

PR #33, opened 2020-07-27
Coscheduling fix: bound contains both assigned and assumed waiting Pods

denkensk:
Fix the current pod in accuracy caused by https://github.com/kubernetes-sigs/scheduler-plugins/pull/27

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -204,19 +204,20 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	namespace := pod.Namespace
 	podGroupName := pgInfo.name
 	minAvailable := pgInfo.minAvailable
+	// bound includes both assigned and assumed Pods.
 	bound := cs.calculateBoundPods(podGroupName, namespace)
-	waiting := cs.calculateWaitingPods(podGroupName, namespace)
-	current := bound + waiting + 1
+	// The bound is calculated from the snapshot. The current pod does not exist in the snapshot during this scheduling cycle.
+	current := bound + 1
 
 	if current < minAvailable {
-		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: bound(%d), waiting(%d)",
-			pod.Namespace, podGroupName, pod.Name, minAvailable, bound, waiting)
+		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: current(%d)",
+			pod.Namespace, podGroupName, pod.Name, minAvailable, current)
 		// TODO Change the timeout to a dynamic value depending on the size of the `PodGroup`
 		return framework.NewStatus(framework.Wait, ""), 10 * PermitWaitingTime
 	}
 
-	klog.V(3).Infof("The count of PodGroup %v/%v/%v is up to minAvailable(%d) in Permit: bound(%d), waiting(%d)",
-		pod.Namespace, podGroupName, pod.Name, minAvailable, bound, waiting)
+	klog.V(3).Infof("The count of PodGroup %v/%v/%v is up to minAvailable(%d) in Permit: current(%d)",
+		pod.Namespace, podGroupName, pod.Name, minAvailable, current)
 	cs.frameworkHandle.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
 		if waitingPod.GetPod().Namespace == namespace && waitingPod.GetPod().Labels[PodGroupName] == podGroupName {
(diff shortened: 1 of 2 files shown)


pkg/coscheduling/coscheduling.go
```diff
@@ -206,7 +206,7 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	minAvailable := pgInfo.minAvailable
 	bound := cs.calculateBoundPods(podGroupName, namespace)
 	waiting := cs.calculateWaitingPods(podGroupName, namespace)
-	current := bound + waiting + 1
+	current := bound + waiting
```
Huang-Wei, line 209: Can you make it UTed? so that we can know which is the expected behavior

pkg/coscheduling/coscheduling.go
```diff
@@ -205,18 +205,18 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	podGroupName := pgInfo.name
 	minAvailable := pgInfo.minAvailable
 	bound := cs.calculateBoundPods(podGroupName, namespace)
-	waiting := cs.calculateWaitingPods(podGroupName, namespace)
-	current := bound + waiting + 1
+	// The bound is calculated from the snapshot. The current pod does not exist in the snapshot during this scheduling cycle.
+	current := bound + 1
```
denkensk, line 210: After we change `running` to `bound`, the pods in bound and waiting maybe duplication.

denkensk:
@Huang-Wei  Please take a look again.

pkg/coscheduling/coscheduling_test.go
```diff
@@ -479,10 +482,27 @@ func (f *fakeSharedLister) List(_ labels.Selector) ([]*v1.Pod, error) {
...
+	}
+	return pods, nil
 }
 
 func (f *fakeSharedLister) NodeInfos() listers.NodeInfoLister {
 	return nil
 }
+
+// Assumed simulates the operation of assumedPod in scheduling cycle.
+func (f *fakeSharedLister) Assumed(pod *v1.Pod, nodeName string) {
```
Huang-Wei, line 501: Assumed() -> assume()? (if this function is needed)

pkg/coscheduling/coscheduling_test.go
```diff
@@ -460,6 +461,8 @@ func TestPermit(t *testing.T) {
 				if got := f.RunPermitPlugins(context.TODO(), nil, tt.pods[i], ""); got.Code() != tt.expected[i] {
 					t.Errorf("[%v] want %v, but got %v", i, tt.expected[i], got.Code())
 				}
+				// The current pod does not exist in the snapshot during this scheduling cycle.
+				fakeSharedLister.Assumed(tt.pods[i], "node")
```
Huang-Wei, line 465: It can be simplfied to `tt.pods[i] = "node"`? If so, we can eliminate the Assume() function.

pkg/coscheduling/coscheduling_test.go
```diff
@@ -479,10 +482,27 @@ func (f *fakeSharedLister) List(_ labels.Selector) ([]*v1.Pod, error) {
 	return f.pods, nil
 }
 
-func (f *fakeSharedLister) FilteredList(_ listers.PodFilter, _ labels.Selector) ([]*v1.Pod, error) {
-	return nil, nil
+func (f *fakeSharedLister) FilteredList(podFilter listers.PodFilter, selector labels.Selector) ([]*v1.Pod, error) {
+	maxSize := len(f.pods)
+	pods := make([]*v1.Pod, 0, maxSize)
```
Huang-Wei, line 487:
```suggestion
	pods := make([]*v1.Pod, 0, len(f.pods))
```

pkg/coscheduling/coscheduling.go
```diff
@@ -205,18 +205,18 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	podGroupName := pgInfo.name
 	minAvailable := pgInfo.minAvailable
 	bound := cs.calculateBoundPods(podGroupName, namespace)
```
Huang-Wei, line 207: Add a comment above like:

```
// bound includes both assigned and assumed Pods.
```

pkg/coscheduling/coscheduling_test.go
denkensk: Done.

pkg/coscheduling/coscheduling.go
denkensk: Done.

pkg/coscheduling/coscheduling_test.go
denkensk: Done.

Huang-Wei:
/lgtm
/approve

Thanks!

PR #35, opened 2020-08-06
Add make sub-command to build docker image

Huang-Wei:
To setup an e2e test pipeline, we need to build the scheduler-plugin image.

This PR adds a `make local_image` command to achieve that.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,0 +1,18 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM alpine:3.12
+
+COPY kube-scheduler /bin/kube-scheduler
+CMD ["kube-scheduler"]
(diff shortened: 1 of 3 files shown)


Huang-Wei:
@denkensk could you take a look?

build/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM alpine:3.12
+
+COPY kube-scheduler /kube-scheduler
```
denkensk, line 17: I find the yaml of kube-scheduler in kubeadm like this 
```yaml
  containers:
  - command:
    - kube-scheduler
    - --authentication-kubeconfig=/etc/kubernetes/scheduler.conf
    - --authorization-kubeconfig=/etc/kubernetes/scheduler.conf
```
I think it's better to add kube-scheduler in `PATH`. Then the user won't to change `kube-scheduler` to `/kube-scheduler`.
Maybe like this
```yaml
FROM alpine:3.12
COPY kube-scheduler /kube-scheduler
ENV PATH="/:${PATH}"
ENTRYPOINT ["/kube-scheduler"]
```

Huang-Wei: Thanks for reviewing. I will give a try tomorrow to see if it's necessary to make the above change.

denkensk: FYI：https://github.com/kubernetes/kubernetes/blob/af29f81af6f5afc0ad48cd2d3d5472a9ca4d4cc8/cmd/kubeadm/app/phases/controlplane/manifests.go#L403

Huang-Wei: updated to put kube-scheduler to /bin, also fixed a bug to add `CGO_ENABLED=0`; otherwise it won't work in alpine.

PR #40, opened 2020-08-23
Coscheduling: make scheduling parameters configurable

yuanchen8911:
1. Made three cleanup parameters configurable as plugin args.

```
// Config defines the scheduling parameters for Coscheduling plugin
type Args struct {
    // PermitWaitingTime is the wait timeout in seconds.
    PermitWaitingTimeSeconds int64
    // PodGroupGCInterval is the period to run gc of PodGroup in seconds.
    PodGroupGCIntervalSeconds int64
    // If the deleted PodGroup stays longer than the PodGroupExpirationTime,
    // the PodGroup will be deleted from PodGroupInfos.
    PodGroupExpirationTimeSeconds int64
}
```
2. Added optional args to scheduler-config.yaml

```
# optional plugin configs
  pluginConfig:
  - name: Coscheduling
    args:
      permitWaitingTimeSeconds: 1
      podGroupGCIntervalSeconds: 10
      podGroupExpirationTimeSeconds: 30
```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -18,7 +18,6 @@ package coscheduling
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync"
@@ -39,12 +38,12 @@ import (
 // Args defines the scheduling parameters for Coscheduling plugin.
 type Args struct {
 	// PermitWaitingTime is the wait timeout in seconds.
-	PermitWaitingTime time.Duration
+	PermitWaitingTimeSeconds int64
 	// PodGroupGCInterval is the period to run gc of PodGroup in seconds.
-	PodGroupGCInterval time.Duration
+	PodGroupGCIntervalSeconds int64
 	// If the deleted PodGroup stays longer than the PodGroupExpirationTime,
 	// the PodGroup will be deleted from PodGroupInfos.
-	PodGroupExpirationTime time.Duration
+	PodGroupExpirationTimeSeconds int64
 }
 
 // Coscheduling is a plugin that implements the mechanism of gang scheduling.
@@ -100,15 +99,15 @@ func (cs *Coscheduling) Name() string {
 }
 
 // New initializes a new plugin and returns it.
-func New(arg *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
-	var args Args
-
-	var params []byte
-	if arg != nil {
-		params = arg.Raw
+func New(config *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
+	args := Args{
+		PermitWaitingTimeSeconds:      10,
+		PodGroupGCIntervalSeconds:     30,
+		PodGroupExpirationTimeSeconds: 600,
 	}
-	if err := parsePluginArgs(params, &args); err != nil {
-		return nil, fmt.Errorf("cannot parse plugin arguments: %v", args)
+
+	if err := framework.DecodeInto(config, &args); err != nil {
+		return nil, err
 	}
 
 	podLister := handle.SharedInformerFactory().Core().V1().Pods().Lister()
@@ -138,7 +137,7 @@ func New(arg *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plug
 			},
 		},
 	)
-	go wait.Until(cs.podGroupInfoGC, cs.args.PodGroupGCInterval, nil)
+	go wait.Until(cs.podGroupInfoGC, time.Duration(cs.args.PodGroupGCIntervalSeconds)*time.Second, nil)
 
 	return cs, nil
 }
@@ -273,7 +272,7 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: current(%d)",
 			pod.Namespace, podGroupName, pod.Name, minAvailable, current)
 		// TODO Change the timeout to a dynamic value depending on the size of the `PodGroup`
-		return framework.NewStatus(framework.Wait, ""), cs.args.PermitWaitingTime
+		return framework.NewStatus(framework.Wait, ""), time.Duration(cs.args.PermitWaitingTimeSeconds) * time.Second
 	}
 
 	klog.V(3).Infof("The count of PodGroup %v/%v/%v is up to minAvailable(%d) in Permit: current(%d)",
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/scheduler-plugins
 go 1.13
 
 require (
+	github.com/pkg/errors v0.8.1
```
Huang-Wei, line 6: Importing "errors" looks like an error :)

pkg/coscheduling/coscheduling.go
```diff
@@ -18,12 +18,14 @@ package coscheduling
...
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync"
 	"time"
 
-	"k8s.io/api/core/v1"
+	"github.com/pkg/errors"
```
Huang-Wei, line 27:
```suggestion
	"errors"
```

pkg/coscheduling/coscheduling.go
```diff
@@ -35,6 +37,17 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/util"
 )
 
+// Config defines the scheduling parameters for Coscheduling plugin
+type Config struct {
```
Huang-Wei, line 41: (optional): rename to "<PluginName>Args" - which is a [naming conversion] in the upstream, and once we migration deps to 1.19, we can have a **versioned** Args (for now just leave it as unversioned).

[naming conversion]: https://github.com/kubernetes/kubernetes/blob/1afc53514032a44d091ae4a9f6e092171db9fe10/pkg/scheduler/apis/config/types_pluginargs.go#L27

pkg/coscheduling/coscheduling.go
```diff
@@ -35,6 +37,17 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/util"
 )
 
+// Config defines the scheduling parameters for Coscheduling plugin
```
Huang-Wei, line 40: Add a period in the end.

pkg/coscheduling/coscheduling.go
```diff
@@ -370,10 +385,42 @@ func responsibleForPod(pod *v1.Pod) bool {
...
+		if pgInfo.deletionTimestamp != nil && pgInfo.deletionTimestamp.Add(cs.args.PodGroupExpirationTime).Before(cs.clock.Now()) {
 			klog.V(3).Infof("%v is out of date and has been deleted in PodGroup GC", key)
 			cs.podGroupInfos.Delete(key)
 		}
 		return true
 	})
 }
+
+// parsePluginConfig parses optional parameters for Coscheduling plugin
+func parsePluginConfig(args *runtime.Unknown, config *Config) error {
```
Huang-Wei, line 397: A heads-up: once we're in 1.19, we should be able to standardize the parsing/conversion/validation logic.

Huang-Wei: Please make this function UTed.

go.mod
yuanchen8911: remove errors package

pkg/coscheduling/coscheduling.go
yuanchen8911: remove error package

pkg/coscheduling/coscheduling.go
yuanchen8911: Added a unit test for parsePluginArgs

pkg/coscheduling/coscheduling.go
yuanchen8911: Changed as suggested.

pkg/coscheduling/coscheduling.go
```diff
@@ -256,7 +273,7 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 		klog.V(3).Infof("The count of podGroup %v/%v/%v is not up to minAvailable(%d) in Permit: current(%d)",
 			pod.Namespace, podGroupName, pod.Name, minAvailable, current)
 		// TODO Change the timeout to a dynamic value depending on the size of the `PodGroup`
-		return framework.NewStatus(framework.Wait, ""), 10 * PermitWaitingTime
+		return framework.NewStatus(framework.Wait, ""), 10 * cs.args.PermitWaitingTime
```
denkensk, line 276: I suggest deleting `10 * ` here.

xq2005: +1. 
And changing the default value into 10s on line 405. 
params.PermitWaitingTime = "10s"

pkg/coscheduling/coscheduling.go
```diff
@@ -370,10 +385,42 @@ func responsibleForPod(pod *v1.Pod) bool {
...
+	var params struct {
+		PermitWaitingTime      string `json: permitWaitingTime`
+		PodGroupGCInterval     string `json: podGroupGCInterval`
+		PodGroupExpirationTime string `json: podGroupExpirationTime`
+	}
+
+	if args == nil {
+		params.PermitWaitingTime = "1s"
+		params.PodGroupGCInterval = "30s"
+		params.PodGroupExpirationTime = "30m"
```
xq2005: Keeping the original default value "10m"?

pkg/coscheduling/coscheduling.go
Huang-Wei: +1

pkg/coscheduling/coscheduling.go
yuanchen8911: deleted it.

pkg/coscheduling/coscheduling.go
yuanchen8911: it's a mistake, corrected it.

yuanchen8911: > +1.
> And changing the default value into 10s on line 405.
> params.PermitWaitingTime = "10s"

The current default value is 30s.

denkensk: The current PermitWaitingTime is 1s. It's better to change it to 10s.

yuanchen8911: changed the default to 10s.

denkensk:
Thanks
/lgtm

yuanchen8911:
@Huang-Wei  Does it look good to me? If so, I'll squash and submit a final one. Thanks.

yuanchen8911:
Squashed

pkg/coscheduling/coscheduling.go
```diff
@@ -370,10 +387,42 @@ func responsibleForPod(pod *v1.Pod) bool {
...
+		if pgInfo.deletionTimestamp != nil && pgInfo.deletionTimestamp.Add(cs.args.PodGroupExpirationTime).Before(cs.clock.Now()) {
 			klog.V(3).Infof("%v is out of date and has been deleted in PodGroup GC", key)
 			cs.podGroupInfos.Delete(key)
 		}
 		return true
 	})
 }
+
+// parsePluginArgs parses optional parameters for Coscheduling plugin.
+func parsePluginArgs(params []byte, args *Args) error {
```
Huang-Wei, line 399: Can we leverage the upstream's utility:

https://github.com/kubernetes/kubernetes/blob/v1.18.0/pkg/scheduler/framework/v1alpha1/registry.go#L31

So only defaulting logic needs to be added.

pkg/coscheduling/coscheduling.go
```diff
@@ -94,11 +100,22 @@ func (cs *Coscheduling) Name() string {
...
 
 // New initializes a new plugin and returns it.
-func New(_ *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
+func New(arg *runtime.Unknown, handle framework.FrameworkHandle) (framework.Plugin, error) {
+	var args Args
+
+	var params []byte
+	if arg != nil {
+		params = arg.Raw
+	}
```
Huang-Wei, line 109: FYI: https://github.com/kubernetes/kubernetes/blob/v1.17.11/pkg/scheduler/framework/plugins/requestedtocapacityratio/requested_to_capacity_ratio.go#L40

pkg/coscheduling/coscheduling.go
```diff
@@ -35,6 +36,17 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/util"
 )
 
+// Args defines the scheduling parameters for Coscheduling plugin.
+type Args struct {
+	// PermitWaitingTime is the wait timeout in seconds.
+	PermitWaitingTime time.Duration
```
Huang-Wei, line 42: As it involves marshal/unmarshal, according to [api-conventions.md#units](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units),

let's make it `int64`, and rename it as "fooXYZSeconds" ([ref: api-conventions.md#naming-conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions))

(applies to the following parameters as well)

Huang-Wei:
@yuanchen8911 please check out the review I did earlier today https://github.com/kubernetes-sigs/scheduler-plugins/pull/40#pullrequestreview-475783228

pkg/coscheduling/coscheduling.go
yuanchen8911: > Can we leverage the upstream's utility:
> 
> https://github.com/kubernetes/kubernetes/blob/v1.18.0/pkg/scheduler/framework/v1alpha1/registry.go#L31
> 
> So only defaulting logic needs to be added.

I replaced "Unmarshal" with "framework.DecodeInto", but I don't think it can unmarshal json into time.Duration directly. So I still need to parse the string into time.Duration.

Huang-Wei: We'd better not use Duration as the type, see my comments below.

pkg/coscheduling/coscheduling.go
yuanchen8911: > FYI: https://github.com/kubernetes/kubernetes/blob/v1.17.11/pkg/scheduler/framework/plugins/requestedtocapacityratio/requested_to_capacity_ratio.go#L40

Again, I can simplify the implementation using framework.DecodeInto, but don't know how to unmarshal to time.Duration directly.

yuanchen8911: Thanks, made the changes using DecodeInto.

pkg/coscheduling/coscheduling.go
yuanchen8911: Changed as suggested.

yuanchen8911:
> @yuanchen8911 please check out the review I did earlier today https://github.com/kubernetes-sigs/scheduler-plugins/pull/40#pullrequestreview-475783228

Thanks, made the changes.

Huang-Wei:
Thanks @yuanchen8911 !

/approve
/lgtm

PR #42, opened 2020-08-24
Add kep for coscheduling base on CRD

cwdsuzhou:
This PR add a plugin named batch. Our original repo is [here](https://github.com/tenstack/batch-scheduler).

In Tencent, this plugin has been stably running for more than half a year.

/cc @Huang-Wei

### DIFF
--- a/kep/42-podgroup-coscheduling/README.md
+++ b/kep/42-podgroup-coscheduling/README.md
@@ -1,0 +1,130 @@
+# Coscheduling based on PodGroup CRD
+
+## Table of Contents
+
+<!-- toc -->
+- [Coscheduling based on PodGroup CRD](#Coscheduling based on PodGroup CRD)
+  - [Table of Contents](#table-of-contents)
+  - [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+  - [Use Cases](#use-cases)
+  - [Design Details](#design-details)
+    - [PodGroup](#podgroup)
+    - [Controller](#controller)
+    - [Extension points](#extension-points)
+      - [QueueSort](#queuesort)
+      - [PreFilter](#prefilter)
+      - [Permit](#permit)
+      - [PostBind](#postbind)
+  - [Known Limitations](#Known Limitations)
+<!-- /toc -->
+
+## Motivation
+Currently, through the default scheduler of Kubernetes, we cannot ensure a group of pods can be scheduled altogether. Under some scenes, it would waste resources since the whole application cannot work with only partial Pods' running, like Spark jobs, TensorFlow jobs, and so on. This proposal is aimed at solving the issue, by introducing a PodGroup CRD to do the heavy lifting on wiring a group of Pods.
+## Goals
+1. Base on the scheduling framework, implement the co-scheduling feature.
+2. Define a CRD name PodGroup to help pod scheduling.
+ 
+## Non-Goals
+Sort the job when we submit jobs to a cluster. Currently, we can only do this base on pods.
+ 
+## Use Cases
+Batch workloads such as Spark jobs, TensorFlow jobs that have to run altogether.
+
+## Design Details
+
+### PodGroup
+
+We define a CRD name PodGroup to help schedule, its definition is as follows:
+```go
+// PodGroupSpec represents the template of a pod group.
+type PodGroupSpec struct {
+	// MinMember defines the minimal number of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinMember uint32 `json:"minMember"`
+
+	// MinResources defines the minimal resource of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinResources *v1.ResourceList `json:"minResources,omitempty"`
+
+	// MaxScheduleTime defines the maximal time of members/tasks to wait before run the pod group;
+	MaxScheduleTime *metav1.Duration `json:"maxScheduleTime,omitempty"`
+}
+
+// PodGroupStatus represents the current state of a pod group.
+type PodGroupStatus struct {
+	// Current phase of PodGroup.
+	Phase PodGroupPhase `json:"phase"`
+
+    // OccupiedBy marks the workload (e.g., deployment, statefulset) UID that occupy the podgroup.
+    // It is empty if not initialized.
+	OccupiedBy string `json:"occupiedBy,omitempty"`
+
+	// The number of actively running pods.
+	// +optional
+	Scheduled uint32 `json:"scheduled"`
+
+	// The number of actively running pods.
+	// +optional
+	Running uint32 `json:"running"`
+
+	// The number of pods which reached phase Succeeded.
+	// +optional
+	Succeeded uint32 `json:"succeeded"`
+
+	// The number of pods which reached phase Failed.
+	// +optional
+	Failed uint32 `json:"failed"`
+
+	// ScheduleStartTime of the group
+	ScheduleStartTime metav1.Time `json:"scheduleStartTime"`
+}
+```
+
+### Controller
+
+We define a controller to reconcile PodGroup status, and we can query the job status through describing the PodGroup. Once a pod in a group failed, the Group Status is marked Failed. Controller would also help recover from abnormal cases, e.g. batch scheduling is interrupted due to
+cluster upgrade.
+
+### Extension points
+
+#### QueueSort
+
+To make sure a group of pods can be scheduled as soon as possible. We implemented this `extension point`. The main progress is as follows:
+
+1. Sort based on Pod Priority
+2. Pod does not belong to any group
+3. PodGroup creation time
+4. Pod creation time
+
+#### PreFilter
+
+This extension pre-filters pods to save scheduling cycles. This is especially helpful when there are not enough resources in a cluster. The overall flow works as below:
+
+1. If the pod doesn't belong to a pod group, allow it; otherwise, go to the following steps.
+2. If there are no other pending pod groups - say all other pod groups have already been scheduled, we allow the pod when its resource requirement is satisfied.
+3. If there are other pod groups not fully scheduled, we check if the pod group current pod belongs to is the closest to its completion - i.e., we check its completion progress by `len(waiting pods of a pod group) / pod group's minMember`.
+    - if it is, this pod is allowed.
+    - if it's not, we check if allowing it would still make the top-progressed pod group (or you want to say all progressed pod grouped?) scheduled:
+        - if so, this pod is allowed;
+        - if not, we check if the current pod has higher priority than the top-progressed pod group (I guess for convenience, you would still need the `PriorityClass` defined in PodGroup?). If it does, this pod is allowed and the top-progressed pod group gets rejected.
+
+For any pod that gets rejected, their pod group would be added to a backoff list and get retried until a TTL is met.
+
+#### Permit
+
+1. When the number of waiting pods in a PodGroup is less than `minMember` (defined in the PodGroup), the status `Wait` is returned. They will be added to cache with TLL (equal to MaxScheduleTime).
+2. When the number is equal or greater than `minMember`, send a signal to permit the waiting pods.
+
+We can define `MaxScheduleTime` for a PodGroup. If any pod times out, the whole group would be rejected.
+
+#### PostBind
+
+This extension is primarily used to record the PodGroup Status. When a pod is bound successfully, we would update the status of its affiliated PodGroup.
+
+### Known Limitations
+
+1. We cannot support group preemption very well now. Though we have tried to implement the extensions, it still cannot meet production requirements.
--- a/kep/42-podgroup-coscheduling/kep.yaml
+++ b/kep/42-podgroup-coscheduling/kep.yaml
@@ -1,0 +1,13 @@
+title: Coscheduling based on PodGroup CRD
+kep-number: 42
+authors:
+  - "@cwdsuzhou"
+owning-sig: sig-scheduling
+reviewers:
+  - "@Huang-Wei"
+  - "@denkensk"
+approvers:
+  - "@Huang-Wei"
+creation-date: 2020-08-24
+last-updated: 2020-08-24
+status: implementable


kep/batch/README.md
```diff
@@ -0,0 +1,140 @@
+# Batch scheduler based on scheduling framework
```
Huang-Wei, line 1: Can you move this out to a separate "KEP" PR so we can review the design separately?

An example is at https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/2-lightweight-coscheduling. In addition to a `README.md`, let's follow the community KEP conversion to put a `kep.yaml` to include the metadata there.

In terms of naming, IMO "batch" is a bit generic, can we make the folder as "podgroup-coscheduling", and the internet title here can be "Coscheduling based on PodGroup CRD", WDYT? (the wordings "based on scheduling framework" is unneeded as all proposals in this repo are based on scheduling framework)

cwdsuzhou: Good idea. I would separate the PR first. Thanks for kind suggestions

cwdsuzhou: PR updated!
@Huang-Wei

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+3. PodGroup creation time
+4. Pod creation time
+
+#### PreFilter
+
+This extension helps pre-filter pods. It is useful, especially when there are not enough resources in a cluster. The main proggress is as follows:
+
+1. Allow the pods do not belong to any group.
+2. If there are no groups scheduling, we check resource, if enough, we allow the pod.
+3. If there are groups running, we check if the current pod belong the max finished group, if it is, we allow it.
```
denkensk, line 119: What‘s the `Max finished group`？
@cwdsuzhou

denkensk:
@cwdsuzhou  Thanks for your contribution. The definition of PodGroup is very important for Coscheduling.

After I review the kep, I find the key points of this kep are similar with the previous one(https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/2-lightweight-coscheduling. )  like queueSort and Permit. 

Can we consider supporting podgroup based on the original coscheduling implemention Instead of reimplementing a plugin? Looking forward to your reply.

kep/podgroup-coscheduling/README.md
cwdsuzhou: The group who has the max progress.

cwdsuzhou:
> @cwdsuzhou Thanks for your contribution. The definition of PodGroup is very important for Coscheduling.
> 
> After I review the kep, I find the key points of this kep are similar with the previous one(https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/2-lightweight-coscheduling. ) like queueSort and Permit.
> 
> Can we consider supporting podgroup based on the original coscheduling implemention Instead of reimplementing a plugin? Looking forward to your reply.

Yes, that is similar. But we  consider more about resources and some add another controller to reconcile the podGroup status. Actually, the coscheduling has finished developing since Nov. 2019. Though the current coscheduling has a similar implementation, it may not cover some scenes or may have bad performance under some scene.  We have ran the same integration tests cases. The result is as follows:

image

image

denkensk:
> we consider more about resources and some add another controller to reconcile the podGroup status.
---
I think it's OK to reconcile the PodGroup status in controller and pre-check or reserve the resource like Volcano. These features are not in conflict with the current co-scheduling. @cwdsuzhou 

it may not cover some scenes or may have bad performance under some scene. 
---
Can your share the scenes not covered in the current Co-scheduling？

cwdsuzhou:
> > we consider more about resources and some add another controller to reconcile the podGroup status.
> 
> I think it's OK to reconcile the PodGroup status in controller and pre-check or reserve the resource like Volcano. These features are not in conflict with the current co-scheduling. @cwdsuzhou
> 
> > it may not cover some scenes or may have bad performance under some scene.
> 
> Can your share the scenes not covered in the current Co-scheduling？

300 resources in a cluster, a job requires 3 pods, 150 resources requested by per pod.  2 pods would schedule, but  actually  the job can not run. If another job consists of 3 pods and 100 resources required by per pod comes, this job would not run immediately. Because resources have been occupied by the first job.

denkensk:
> 300 resources in a cluster, a job requires 3 pods, 150 resources requested by per pod. 2 pods would schedule, but actually the job can not run. If another job consists of 3 pods and 100 resources required by per pod comes, this job would not run immediately. Because resources have been occupied by the first job.

---
This is a cluster where resources are highly used. The resources occupied by the first job will be released after the timeout. And also we can add some more policy like resource pre-check in Pre-Filter.

I don't think there is any problem that can't be solved so that we have to reimplement the coschduling plugins？
@cwdsuzhou

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+type PodGroupSpec struct {
+	// MinMember defines the minimal number of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinMember uint32 `json:"minMember"`
+
+	// If specified, indicates the PodGroup's priority. "system-node-critical" and
+	// "system-cluster-critical" are two special keywords which indicate the
+	// highest priorities with the former being the highest priority. Any other
+	// name must be defined by creating a PriorityClass object with that name.
```
xq2005, line 52: Does PriorityClass  mean another CRD or something else? Can you introduce more about this type?

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+	// If not specified, the PodGroup priority will be default or zero if there is no
+	// default.
+	// +optional
+	PriorityClassName string `json:"priorityClassName,omitempty"`
+
+	// MinResources defines the minimal resource of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinResources *v1.ResourceList `json:"minResources,omitempty"`
```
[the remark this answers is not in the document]
cwdsuzhou: pod

kep/podgroup-coscheduling/README.md
cwdsuzhou: name of Priority class, same as the default int K8s! But it is not used not, I wonder if I should remove it

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+Sort the job when we submit jobs to a cluster. Currently, we can only do this base on pods.
+ 
+## Use Cases
+Spark jobs, tensorflow jobs and other pods have to run together.
+
+## Design Details
+
+### PodGroup
+
+We define a CRD name PodGroup to help schedule, its definition is as follows:
```
Huang-Wei, line 40:
```suggestion
A CRD named `PodGroup` is proposed with the following API specs:
```

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+```build
```
Huang-Wei, line 41:
```suggestion
```go
```

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+	// +optional
+	PriorityClassName string `json:"priorityClassName,omitempty"`
+
+	// MinResources defines the minimal resource of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinResources *v1.ResourceList `json:"minResources,omitempty"`
+
+	// MaxScheduleTime defines the maximal time of members/tasks to wait before run the pod group;
+	MaxScheduleTime *metav1.Duration `json:"maxScheduleTime,omitempty"`
```
Huang-Wei, line 64: We should avoid defining API as Go type `Duration`, use `int64` instead, and rename it like `ScheduleTimeoutSeconds` or `WaitTimeoutSeconds`.

Ref:

- [api-conventions.md#units](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units)
- [api-conventions.md#naming-conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions)

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+### PodGroup
+
+We define a CRD name PodGroup to help schedule, its definition is as follows:
+```build
+// PodGroupSpec represents the template of a pod group.
+type PodGroupSpec struct {
+	// MinMember defines the minimal number of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinMember uint32 `json:"minMember"`
```
Huang-Wei, line 47: `int32` is more conventional. (applies elsewhere)

kep/podgroup-coscheduling/README.md
Huang-Wei, line 61: `MinResources v1.ResourceList` is more conventional.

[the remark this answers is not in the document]
Huang-Wei: To be honest, I don't quite get the point of `MinResources`.

The resource request is just one dimension of a PodGroup's constraints, how can you ensure all Pods of a PodGroup can be scheduled successfully by only looking at its resource requests?

> pod

but the comments said it's for all tasks.

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// PodGroupStatus represents the current state of a pod group.
+type PodGroupStatus struct {
+	// Current phase of PodGroup.
+	Phase PodGroupPhase `json:"phase"`
+
+	// OccupiedBy marks the podgroup occupied by which group.
+	// Owner reference would be used to filled it, it not initialize, it is empty
+	OccupiedBy string `json:"occupiedBy,omitempty"`
```
Huang-Wei, line 74: Are you trying to describe a hierarchical relation here? For example, a PodGroup is owned/referred by another PodGroup? If so, you can inject the relation into`ObjectMeta.OwnerReferences`, so don't need to define a new API field.

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+	Failed uint32 `json:"failed"`
+
+	// ScheduleStartTime of the group
+	ScheduleStartTime metav1.Time `json:"scheduleStartTime"`
+}
+```
+
+### Controller
+
+We define a controller to reconcile PodGroup status, and we can query the job status through describe the PodGroup. Any pod in a group failed, the Group Status is marked Failed. Controller would also help whem recovering from abnormal cases, e.g. batch scheduling is interpreted due to
```
Huang-Wei, line 99: - ... through describing ...
- ... Once a pod in a group failed ... is marked as Failed...
- ... Controller would also help to recover from ... 
- interpreted -> interrupted?

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+To make sure a group of pods can be scheduled as soon as possible. We implemented this `extension point`. The main progress is as follows:
+
+1. Sort based on Pod Priority
+2. Pod does not belong to any group
+3. PodGroup creation time
+4. Pod creation time
+
+#### PreFilter
+
+This extension helps pre-filter pods. It is useful, especially when there are not enough resources in a cluster. The main proggress is as follows:
```
Huang-Wei, line 115:
```suggestion
This extension helps pre-filter pods. It is useful, especially when there are not enough resources in a cluster. The overall flow works as below:
```

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+1. Allow the pods do not belong to any group.
```
Huang-Wei, line 117:
```suggestion
1. Allow the pods that do not belong to any group.
```

kep/podgroup-coscheduling/README.md
Huang-Wei: Please explain `max` in the document.

kep/podgroup-coscheduling/README.md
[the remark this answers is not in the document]
cwdsuzhou: Yes, this seems to be not necessary. Actually we did not set it in our current implementation, the value is achieved from Pod. I would remove it.

kep/podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Motivation
+Currently, through the default scheduler of Kubernetes, we cannot ensure a group of pods scheduled at the same time . Under some scene, it would waste resources since some pods need work together, like spark, tensorflow and so on . So, podgroup-coscheduling is aimed at solving the issue.
+
+## Goals
+1. Base on scheduling framework, implementing the gang scheduling.
+2. Define a CRD name PodGroup to help pod scheduling.
+ 
+## Non-Goals
+Sort the job when we submit jobs to a cluster. Currently, we can only do this base on pods.
```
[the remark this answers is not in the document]
cwdsuzhou: a general term of batch job

cwdsuzhou:
> First round of review.

Thank for review, I would update this kep according to your kind suggestions

kep/podgroup-coscheduling/README.md
cwdsuzhou: Not, in our current implement this is just UID of a deployment, statefulSet or Other CR, e.g tf job or mpi job.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+2. Pod does not belong to any group
+3. PodGroup creation time
+4. Pod creation time
+
+#### PreFilter
+
+This extension helps pre-filter pods. It is useful, especially when there are not enough resources in a cluster. The overall flow works as below:
+
+1. Allow the pods that do not belong to any group.
+2. If there are no groups scheduling, we check resource, if enough, we allow the pod.
```
Huang-Wei, line 103: Can you rephrase this? I suppose you want to say:

```suggestion
2. If there are no other pending pod groups - say all other pod groups have already been scheduled, we allow the pod when its resource requirement is satisfied.
```

I'm not quite sure what do you mean by `we check resource`, maybe you mean `MinResource`; if so, please reword it.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+3. If there are groups running, we check if the current pod belong the group having the max progress(num(Pods)/minMember), if it is, we allow it.
+4. Otherwise, we check if the max finished group can still run when allow this pod. If we can, allow it.
+5. Otherwise, we check if the pod has higher priority compared with the max finished one. If yes, we reject the pod belongs to the group and allow the current one.
```
Huang-Wei, line 106:
```suggestion
3. If there are other pod groups not fully scheduled, we check if the pod group current pod belongs to is the closest to its completion - i.e., we check its completion progress by `len(waiting pods of a pod group) / pod group's minMember`.
    - if it is, this pod is allowed.
    - if it's not, we check if allowing it would still make the top-progressed pod group (or you want to say all progressed pod grouped?) scheduled:
        - if so, this pod is allowed
        - if not, we check if the current pod has higher priority than the top-progressed pod group (I guess for convenience, you would still need the `PriorityClass` defined in PodGroup?). If it does, this pod is allowed and the top-progressed pod group gets rejected.
```

And remove bullet 4 & 5.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+This extension helps pre-filter pods. It is useful, especially when there are not enough resources in a cluster. The overall flow works as below:
+
+1. Allow the pods that do not belong to any group.
+2. If there are no groups scheduling, we check resource, if enough, we allow the pod.
+3. If there are groups running, we check if the current pod belong the group having the max progress(num(Pods)/minMember), if it is, we allow it.
+4. Otherwise, we check if the max finished group can still run when allow this pod. If we can, allow it.
+5. Otherwise, we check if the pod has higher priority compared with the max finished one. If yes, we reject the pod belongs to the group and allow the current one. 
+
+Any pod rejected to run, their group would be added to a denied list with a ttl.
```
Huang-Wei, line 108:
```suggestion
For any pod that gets rejected, their pod group would be added to a backoff list and get retried until a TTL is met.
```

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+#### Permit
+
+1. When number of pods cannot meet the `minMember` defines in the PodGroup, `Wait` is returned. They will be added to cache with TLL(equal to MaxScheduleTime).
+2. When number meet that, we would send a signal to permit the pods waiting. 
+
+We can define `MaxScheduleTime` for a PodGroup. If anyone of the pods times out, the whole group would be rejected.
+
+#### PostBind
+
+This extension is mainly used for helping record the PodGroup Status. When pod binds successfully, we would update the scheduling status of a PodGroup.
```
Huang-Wei, line 119:
```suggestion
This extension is primarily used to record the PodGroup Status. When a pod is bound successfully, we would update the status of its affiliated PodGroup.
```

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+2. If there are no groups scheduling, we check resource, if enough, we allow the pod.
+3. If there are groups running, we check if the current pod belong the group having the max progress(num(Pods)/minMember), if it is, we allow it.
+4. Otherwise, we check if the max finished group can still run when allow this pod. If we can, allow it.
+5. Otherwise, we check if the pod has higher priority compared with the max finished one. If yes, we reject the pod belongs to the group and allow the current one. 
+
+Any pod rejected to run, their group would be added to a denied list with a ttl.
+
+#### Permit
+
+1. When number of pods cannot meet the `minMember` defines in the PodGroup, `Wait` is returned. They will be added to cache with TLL(equal to MaxScheduleTime).
```
Huang-Wei, line 112:
```suggestion
1. When the number of waiting pods in a PodGroup is less than `minMember` (defined in the PodGroup), the status `Wait` is returned. They will be added to cache with TLL (equal to MaxScheduleTime).
```

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+2. When number meet that, we would send a signal to permit the pods waiting.
```
Huang-Wei, line 113:
```suggestion
2. When the number is equal or greater than `minMember`, send a signal to permit the waiting pods.
```

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+5. Otherwise, we check if the pod has higher priority compared with the max finished one. If yes, we reject the pod belongs to the group and allow the current one. 
+
+Any pod rejected to run, their group would be added to a denied list with a ttl.
+
+#### Permit
+
+1. When number of pods cannot meet the `minMember` defines in the PodGroup, `Wait` is returned. They will be added to cache with TLL(equal to MaxScheduleTime).
+2. When number meet that, we would send a signal to permit the pods waiting. 
+
+We can define `MaxScheduleTime` for a PodGroup. If anyone of the pods times out, the whole group would be rejected.
```
Huang-Wei, line 115: FYI: it's now defined in the latest codebase, and will be available in `master` branch until #45 gets reviewed.

Huang-Wei, line 115:
```suggestion
We can define `MaxScheduleTime` for a PodGroup. If any pod times out, the whole pod group would be rejected.
```

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### PostBind
+
+This extension is mainly used for helping record the PodGroup Status. When pod binds successfully, we would update the scheduling status of a PodGroup.
+
+We can define `MaxScheduleTime` for a PodGroup. If anyone of the pods times out, the whole group would be rejected.
```
Huang-Wei, line 121:
```suggestion
We can define `MaxScheduleTime` for a PodGroup. If any pod times out, the whole group would be rejected.
```

kep/42-podgroup-coscheduling/kep.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - "@cwdsuzhou"
+owning-sig: sig-scheduling
+reviewers:
+  - "@Huang-Wei"
+  - "@ahg-g"
+  - "@alculquicondor"
+  - "k82cn"
+  - "@resouer"
+  - "@hex108"
+  - "@everpeace"
```
Huang-Wei, line 13: Remove names who are not reviewing.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -46,20 +45,6 @@ type PodGroupSpec struct {
...
-	// name must be defined by creating a PriorityClass object with that name.
-	// If not specified, the PodGroup priority will be default or zero if there is no
-	// default.
-	// +optional
-	PriorityClassName string `json:"priorityClassName,omitempty"`
-
-	// MinResources defines the minimal resource of members/tasks to run the pod group;
-	// if there's not enough resources to start all tasks, the scheduler
-	// will not start anyone.
-	MinResources *v1.ResourceList `json:"minResources,omitempty"`
```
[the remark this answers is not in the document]
cwdsuzhou: In our original version, we fill this field from podSpec. Keeping this is also ok

kep/42-podgroup-coscheduling/README.md
cwdsuzhou: I think we should keep this.  Our original version supports from both args and the crd.  If the crd did not set the time, the args defines will work.

scene is as bellow:
A job need 10 pods may time out time after 5s, but a job need 10000 pods would like it to be 5min.

kep/42-podgroup-coscheduling/README.md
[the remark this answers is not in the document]
cwdsuzhou: If this defined, we would use this pro pre-fileter. If not, we compute the resource requirements from Pod Spec

cwdsuzhou:
@Huang-Wei PR addressed according to your latest suggestions, PTAL thanks

kep/42-podgroup-coscheduling/README.md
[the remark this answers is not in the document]
denkensk: I'm confused about how to set the `MinResource`. It's the total minimal resource of the PodGroup or the pod?

kep/42-podgroup-coscheduling/README.md
denkensk: I think it is better not to let the user to set the `MaxScheduleTime`. Users will prefer to set the `MaxScheduleTime` Greate than it needed to reserve the resource. This will lead the resource to be reserved for a long time.

> A job need 10 pods may time out time after 5s, but a job need 10000 pods would like it to be 5min.
---
We talked about it before. We can use a factor to compute the `WaitingTime`. The factor can be related to the number of pod.  I add Todo in code before.

kep/42-podgroup-coscheduling/README.md
denkensk: I don't agree to use `len(waiting pods of a pod group) / pod group's minMember` to check if the pg is the closest to its completion. The minMember and the resource requested in pod are different. Like this：
PG1: min 10 waiting 5 --> 50%
PG2 min 100 waiting 60 --> 60%

kep/42-podgroup-coscheduling/README.md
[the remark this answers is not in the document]
cwdsuzhou: Yes

kep/42-podgroup-coscheduling/README.md
cwdsuzhou: But for different cluster,  the factor would also be different. 
e.g. 100 nodes or 2000nodes

kep/42-podgroup-coscheduling/README.md
```diff
@@ -55,7 +60,7 @@ type PodGroupStatus struct {
 	Phase PodGroupPhase `json:"phase"`
 
 	// OccupiedBy marks the podgroup occupied by which group.
-	// Owner reference would be used to filled it, if not initialize, it is empty
+	// Owner reference would be used to filled it, if it is empty, if not initialize.
```
Huang-Wei: How about:

```
// OccupiedBy marks the workload (e.g., deployment, statefulset) UID that occupy the podgroup.
// It is empty if not initialized.
```

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,125 @@
...
+To make sure a group of pods can be scheduled as soon as possible. We implemented this `extension point`. The main progress is as follows:
+
+1. Sort based on Pod Priority
+2. Pod does not belong to any group
+3. PodGroup creation time
+4. Pod creation time
+
+#### PreFilter
+
+This extension helps pre-filter pods. It is useful, especially when there are not enough resources in a cluster. The overall flow works as below:
```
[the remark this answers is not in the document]
Huang-Wei: This comment is still outstanding.

`helps ... useful...` are a bit verbose and doesn't add extra info to make the words more understandable. Can you rephrase it?

kep/42-podgroup-coscheduling/README.md
Huang-Wei: This comment is still outstanding.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -96,7 +81,7 @@ type PodGroupStatus struct {
 
 ### Controller
 
-We define a controller to reconcile PodGroup status, and we can query the job status through describe the PodGroup. Any pod in a group failed, the Group Status is marked Failed. Controller would also help whem recovering from abnormal cases, e.g. batch scheduling is interpreted due to
+We define a controller to reconcile PodGroup status, and we can query the job status through describing the PodGroup. Onece a pod in a group failed, the Group Status is marked Failed. Controller would also help recover from abnormal cases, e.g. batch scheduling is interrupted due to
```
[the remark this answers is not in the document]
Huang-Wei: This comment is still outstanding.

kep/42-podgroup-coscheduling/README.md
Huang-Wei: @cwdsuzhou does the rewording comment https://github.com/kubernetes-sigs/scheduler-plugins/pull/42/files#r481413128 make sense to you?

Huang-Wei: Regarding @denkensk 's concern, I'd leave it for @cwdsuzhou to comment. Probably it's not a bad idea to also look at the absolute number to reach completion.

kep/42-podgroup-coscheduling/README.md
Huang-Wei: This comment is still outstanding.

kep/42-podgroup-coscheduling/README.md
Huang-Wei: This comment is still outstanding.

kep/42-podgroup-coscheduling/README.md
Huang-Wei: This comment is still outstanding.

kep/42-podgroup-coscheduling/README.md
cwdsuzhou: Before pods owned by other groups start scheduling, we would check whether the group with the max completion can continue scheduling when we schedule these pods. According `Sort` func. Current method may be more reasonable. Otherwise, the final scheduling results may conflict with the `Sort`.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### PreFilter
+
+This extension pre-filters pods to save scheduling cycles. This is especially helpful when there are not enough resources in a cluster. The overall flow works as below:
+
+1. If the pod doesn't belong to a pod group, allow it; otherwise, go to the following steps.
+2. If there are no other pending pod groups - say all other pod groups have already been scheduled, we allow the pod when its resource requirement is satisfied.
+3. If there are groups running, we check if the current pod belong the group having the max progress(num(Pods)/minMember), if it is, we allow it.
+4. Otherwise, we check if the max finished group can still run when allow this pod. If we can, allow it.
+5. Otherwise, we check if the pod has higher priority compared with the max finished one. If yes, we reject the pod belongs to the group and allow the current one.
```
Huang-Wei, line 111: Please see if the wordings in https://github.com/kubernetes-sigs/scheduler-plugins/pull/42/files#r481413128 makes more sense to you.

Huang-Wei (commented):
Some final comments. We're close to getting it merged.

kep/42-podgroup-coscheduling/README.md
cwdsuzhou: Sure,  I missed this last time.

kep/42-podgroup-coscheduling/README.md
denkensk: I think it's a good idea to find the podGroup most likely to be successfully scheduled . But I think the current method is not reasonable. Because the different pods may request different resource and with different scale. And also with different priority.
I think we need to think about the pod.resource.requested、absolute number and the priority to find the the podGroup most likely to be successfully scheduled.
@cwdsuzhou @Huang-Wei  WDYT?

kep/42-podgroup-coscheduling/README.md
denkensk: Yes, the size of cluster is also a factor to be considered with the the number of pod. This can be discussed.
And I am still worried about it. @Huang-Wei @cwdsuzhou 
> I think it is better not to let the user to set the MaxScheduleTime. Users will prefer to set the MaxScheduleTime greater than it needed to reserve the resource. This will lead the resource to be reserved for a long time.

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,128 @@
...
+type PodGroupSpec struct {
+	// MinMember defines the minimal number of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinMember uint32 `json:"minMember"`
+
+	// MinResources defines the minimal resource of members/tasks to run the pod group;
+	// if there's not enough resources to start all tasks, the scheduler
+	// will not start anyone.
+	MinResources *v1.ResourceList `json:"minResources,omitempty"`
```
denkensk, line 51: How about if the podgroup is made up of different types of pods? Like ps and trainer in tensorflow. How to set the MinResources ？
@cwdsuzhou

cwdsuzhou: This is the v1alpha1 version. I could add that if need and extend this CRD later. e.g

type  PodGroup struct {
     ...
     Groups [map]PodGroup 
     ...
}

Huang-Wei: I think one PodGroup should describe a particular type of workload. In terms of tensorflow application, a good model is to have PS and Worker referencing their corresponding (Sub-)PodGroup, and then there can be a (Parent-)PodGroup to wire those two (Sub-)PodGroup(s).

(not quite sure it would work, need a PoC on this)

Huang-Wei:
/approve

Thanks @cwdsuzhou ! Please squash the commits, then I will /lgtm.

cwdsuzhou:
> /approve
> 
> Thanks @cwdsuzhou ! Please squash the commits, then I will /lgtm.

Done, PTAL, thanks! @Huang-Wei

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,130 @@
...
+1. Sort based on Pod Priority
+2. Pod does not belong to any group
+3. PodGroup creation time
+4. Pod creation time
+
+#### PreFilter
+
+This extension pre-filters pods to save scheduling cycles. This is especially helpful when there are not enough resources in a cluster. The overall flow works as below:
+
+1. If the pod doesn't belong to a pod group, allow it; otherwise, go to the following steps.
```
xujyan, line 107: How is this "pod belonging to a pod group" relationship expressed?

kep/42-podgroup-coscheduling/README.md
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// PodGroupStatus represents the current state of a pod group.
+type PodGroupStatus struct {
+	// Current phase of PodGroup.
+	Phase PodGroupPhase `json:"phase"`
+
+    // OccupiedBy marks the workload (e.g., deployment, statefulset) UID that occupy the podgroup.
+    // It is empty if not initialized.
+	OccupiedBy string `json:"occupiedBy,omitempty"`
```
xujyan, line 64: As in, owned, like [`OwnerReference`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#ownerreference-v1-meta)?

kep/42-podgroup-coscheduling/README.md
xujyan: I thought per https://github.com/kubernetes-sigs/scheduler-plugins/pull/42#discussion_r478137253 `MinResources` was going to be removed?

kep/42-podgroup-coscheduling/README.md
cwdsuzhou: Yes

xujyan: I think it would be more intuitive to use the more common terminology?

kep/42-podgroup-coscheduling/README.md
Huang-Wei: - pod to PodGroup: pod gets applied with a PodGroup label or annotation
- PodGroup to pod: either implicitly by dynamic computing, or PodGroup's `OccupiedBy` - which refers to the the group of pods' controller (i.e., Deployment or StatefulSet).

@cwdsuzhou can confirm.

kep/42-podgroup-coscheduling/README.md
Huang-Wei: The canonical `meta.OwnerReference` _may_ be used to exercise the concept of hierarchical PodGroups, i.e., a child PodGroup "ownedBy" the other parent PodGroup, so using the term "OccupiedBy" in `status` can distinguish from that semantics IMO - to represent the semantics of PodGroup <-> Workload mapping.

@cwdsuzhou may comment further.

cwdsuzhou: > The canonical `meta.OwnerReference` _may_ be used to exercise the concept of hierarchical PodGroups, i.e., a child PodGroup "ownedBy" the other parent PodGroup, so using the term "OccupiedBy" in `status` can distinguish from that semantics IMO - to represent the semantics of PodGroup <-> Workload mapping.
> 
> @cwdsuzhou may comment further.

Yep ,this field is used for recording the related object e.g. workload or crd. But not absolutely same as the owner reference.

PR #43, opened 2020-08-27
Bump k8s deps to v1.18.8

Huang-Wei:
Kubernetes v1.19.0 is out today https://github.com/kubernetes/kubernetes/releases/tag/v1.19.0, to ensure we migrate the deps to v1.19.0 asap, we should branch off a `release-v1.18` branch based on the current `master`, and tag releases from there.

So let's bump k8s deps to the latest v1.18 releases first.


go.mod
```diff
@@ -3,33 +3,33 @@ module sigs.k8s.io/scheduler-plugins
...
 
 require (
-	k8s.io/api v0.18.0
-	k8s.io/apimachinery v0.18.0
-	k8s.io/client-go v0.18.0
+	k8s.io/api v0.18.8
+	k8s.io/apimachinery v0.18.8
+	k8s.io/client-go v0.18.8
 	k8s.io/klog v1.0.0
 	k8s.io/kubernetes v1.18.0
```
denkensk, line 10: v1.18.8 ?

Huang-Wei: Yes, I missed it. Thanks for the catch.

denkensk:
/lgtm
Thanks

PR #58, opened 2020-09-23
Rename MaxScheduleTime to ScheduleTimeoutSeconds

cwdsuzhou:
/cc @Huang-Wei

### DIFF
--- a/kep/42-podgroup-coscheduling/README.md
+++ b/kep/42-podgroup-coscheduling/README.md
@@ -50,8 +50,8 @@ type PodGroupSpec struct {
 	// will not start anyone.
 	MinResources *v1.ResourceList `json:"minResources,omitempty"`
 
-	// MaxScheduleTime defines the maximal time of members/tasks to wait before run the pod group;
-	MaxScheduleTime *metav1.Duration `json:"maxScheduleTime,omitempty"`
+	// ScheduleTimeoutSeconds defines the maximal time of members/tasks to wait before run the pod group;
+	ScheduleTimeoutSeconds *int32 `json:"scheduleTimeoutSeconds,omitempty"`
 }
 
 // PodGroupStatus represents the current state of a pod group.
@@ -116,7 +116,7 @@ For any pod that gets rejected, their pod group would be added to a backoff list
 
 #### Permit
 
-1. When the number of waiting pods in a PodGroup is less than `minMember` (defined in the PodGroup), the status `Wait` is returned. They will be added to cache with TLL (equal to MaxScheduleTime).
+1. When the number of waiting pods in a PodGroup is less than `minMember` (defined in the PodGroup), the status `Wait` is returned. They will be added to cache with TLL (equal to ScheduleTimeoutSeconds).
 2. When the number is equal or greater than `minMember`, send a signal to permit the waiting pods.
 
 We can define `MaxScheduleTime` for a PodGroup. If any pod times out, the whole group would be rejected.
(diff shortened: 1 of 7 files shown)


kep/42-podgroup-coscheduling/README.md
```diff
@@ -50,8 +50,8 @@ type PodGroupSpec struct {
 	// will not start anyone.
 	MinResources *v1.ResourceList `json:"minResources,omitempty"`
 
-	// MaxScheduleTime defines the maximal time of members/tasks to wait before run the pod group;
-	MaxScheduleTime *metav1.Duration `json:"maxScheduleTime,omitempty"`
+	// MaxScheduleTimeSeconds defines the maximal time of members/tasks to wait before run the pod group;
+	MaxScheduleTime *metav1.Duration `json:"maxScheduleTimeSeconds,omitempty"`
```
Huang-Wei, line 54: According to [api-conventions#units](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#units), the type should be an integer. Let's make it `*int32`.

And based on:

> We use the fooSeconds convention for durations, as discussed in the units subsection.
> - fooPeriodSeconds is preferred for periodic intervals and other waiting periods (e.g., over fooIntervalSeconds).
> - fooTimeoutSeconds is preferred for inactivity/unresponsiveness deadlines.
> - fooDeadlineSeconds is preferred for activity completion deadlines.

How about renaming it to `ScheduleTimeoutSeconds`?

cwdsuzhou: Seem better. Updated

PTAL

pkg/util/podgroup.go
```diff
@@ -27,7 +27,7 @@ import (
 	"sigs.k8s.io/scheduler-plugins/pkg/apis/podgroup/v1alpha1"
 )
 
-// DefaultWaitTime is 60s if MaxScheduleTime is not specified.
+// DefaultWaitTime is 60s if ScheduleTimeoutSeconds is not specified.
```
Huang-Wei, line 30: Forgot to mention that a good practice is to use `Defaults_XYZ` logic to define the expected defaulting behavior, and register the the function to `RegisterDefaults()` (in pkg/apis/podgroup/v1alpha1/zz_generated.defaults.go).

Feel free to do it in this PR, or add a TODO here and accomplish in a follow-up.

Huang-Wei (commented):
Apologizing for verbose comments here: would you mind updating `uint32` to `int32`? According to [api-conventions.md#primitive-types](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types):

> Do not use unsigned integers, due to inconsistent support across languages and libraries. Just validate that the integer is non-negative if that's the case.

And for `MinMember`, what if users leave it unset, can the PodGroup be created successfully then?

pkg/util/podgroup.go
cwdsuzhou: I would like  add a TODO here.

cwdsuzhou:
> Apologizing for verbose comments here: would you mind updating `uint32` to `int32`? According to [api-conventions.md#primitive-types](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types):
> 
> > Do not use unsigned integers, due to inconsistent support across languages and libraries. Just validate that the integer is non-negative if that's the case.
> 
> And for `MinMember`, what if users leave it unset, can the PodGroup be created successfully then?

I think we can add default value and validation in crd.yaml.
In this PR, I have changed uint32 to int32

pkg/util/podgroup.go
cwdsuzhou: PR updated

Huang-Wei:
Thanks @cwdsuzhou !

/lgtm
/approve

PR #60, opened 2020-09-23
Use concrete type in unversioned (internal) API fields

Huang-Wei:
Kubernetes supports smooth conversion between different versions of API objects, however, when the type in versioned pkg is defined as a pointer, the corresponding unversioned type doesn't need to defined as a pointer - as it's used internally and converted from the outbound versioned API object.

This PR corrects the nits across scheduler-plugin codebase.

Another change is on const string, I'm inclined to follow the pattern of upstream (e.g., `Protocol`) to make it as a concrete type, even in versioned pkg:

- https://github.com/kubernetes/kubernetes/blob/aad834b0733b17579dc07f7ced775278a5b9d978/staging/src/k8s.io/api/core/v1/types.go#L4378-L4382
- https://github.com/kubernetes/kubernetes/blob/aad834b0733b17579dc07f7ced775278a5b9d978/pkg/apis/core/v1/defaults.go#L74-L78

### DIFF
--- a/pkg/apis/config/v1beta1/defaults.go
+++ b/pkg/apis/config/v1beta1/defaults.go
@@ -25,7 +25,7 @@ var (
 	defaultPodGroupGCIntervalSeconds     int64 = 30
 	defaultPodGroupExpirationTimeSeconds int64 = 600
 
-	defaultNodeResourcesAllocatableMode ModeType = Least
+	defaultNodeResourcesAllocatableMode = Least
 
 	// defaultResourcesToWeightMap is used to set the default resourceToWeight map for CPU and memory
 	// used by the NodeResourcesAllocatable scoring plugin.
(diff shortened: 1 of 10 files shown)


Huang-Wei:
cc/ @patricklundquist @cwdsuzhou @yuanchen8911

cwdsuzhou:
LGTM! though I am not a reviewer.

pkg/apis/config/v1beta1/defaults.go
```diff
@@ -50,11 +50,11 @@ func SetDefaults_CoschedulingArgs(obj *CoschedulingArgs) {
 }
```
denkensk, line 50: If we change the args from point to int, `SetDefaults_CoschedulingArgs` also needs to be modified here.

Huang-Wei: We don't need to, do we? As in `SetDefaults_CoschedulingArgs` is in versioned pkg, the argument (versioned args) is still a pointer.

denkensk: Got it. Thanks.  please rebase the code.

Huang-Wei: Rebased.

PR #61, opened 2020-09-24
Trimaran: Real Load Aware Scheduling KEP 

zorro786:
* Add Real Load Aware Scheduling KEP

* Added images for safe plugin

* Try to change the image size.

* Try html for resizing images

* Update all safe plugin images

* Resizing images

* resize again

* Update real-load-aware-scheduling.md

* Update real-load-aware-scheduling.md

* Update real-load-aware-scheduling.md

* Update real-load-aware-scheduling.md

* Resize design image

Co-authored-by: Abdul Qadeer <aqadeer@paypal.com>


warmchang:
@wangchen615 @zorro786 , Thanks for your KEP! Pls sign the CLA, then run /check-cla

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,538 @@
...
+### Load Watcher/Load Analyser
+
+The load watcher and analyzer both run in a single process. The watcher is responsible for the cluster-wide aggregation of resource usage metrics like CPU, memory, network, and IO stats over windows of a specified duration. It stores these in its local cache and persists several aggregations in the host local DB for fault tolerance. The analyzer is responsible for the detection of bad metrics and any remediation. Bad metrics could be due to missing metrics or metrics with considerable errors, making them anomalies. It can also be extended in the future to use ML models for analyzing metrics.
+
+Load watcher would cache metrics in the last 15-minute, 10-minute, and 5-minute windows, which can be queried via REST API exposed.
+
+
+### DB
+
+This is a localhost database stored as a file. This DB aims to populate the load watcher's cache quickly if it is lost due to a crash and perform scheduling decisions without any impact on latency.
```
denkensk, line 81: If the scheduler is running as a pod or static pod，how to ensure persistence of file when the process crashes.
I think it's better add more detail about this.

zorro786:
@warmchang Yes currently working with legal teams for signing CLA.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
zorro786: The file will be stored in host file system, so it will be persisted across pod failures on the same node. For HA setups, each load watcher process will maintain its corresponding file locally. I will add this to doc :+1:

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+#### Score plugin
+
+This plugin would extend the Score extension point. K8s scheduler framework calls the Score function for each node separately when scheduling a pod.
+
+
+**Idea**
+
+
+Balancing load based on average would be risky sometimes, as it does not consider the bursty variations. A safe balancing plugin balances not only the average load but also the risk caused by load variations. Suppose we take the mean (M) and standard deviation (V) of all nodes’ utilization into a mu-sigma plot below. In that case, the safe balancing plugin will make placements such that all nodes’ utilization are aligned on the diagonal line, which is V + M = c. Here, c is a constant indicating the overall cluster utilization average plus the standard deviation.
```
Huang-Wei, line 205: Where do `M` and `V` come from? i.e., directly from LoadWatcher? or based on the raw data of LoadWather plus extra calculation? It'd good to use a concrete flow to explain how these values are generated, and hence make readers understand their semantics easily.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+**Algorithm**
+
+
+1. Get the requested resource for the pod to be scheduled as, <img src="https://render.githubusercontent.com/render/math?math=r">.
+2. Get the sliding window average (<img src="https://render.githubusercontent.com/render/math?math=M">) and standard deviation (<img src="https://render.githubusercontent.com/render/math?math=V">) of resource utilization fraction (range from 0 to 1) for all types of resources (CPU, Memory, GPU, etc.) of the current node to be scored. 
+3. Calculate the score of the current node for each type of resource: <img src="https://render.githubusercontent.com/render/math?math=S_i = M %2B r %2B V">
+4. Get a score for each type of resource and bound it to [0,1]: <img src="https://render.githubusercontent.com/render/math?math=S_i = \min(S_i, 1.0)">
+5. Calculate the node priority score per resource as: <img src="https://render.githubusercontent.com/render/math?math=U_i = (1 - S_i) \times MaxPriority ">
+6. Get the final node score as: <img src="https://render.githubusercontent.com/render/math?math=U = \min_i(U_i)">
```
Huang-Wei, line 221: Min(Ui) or Mini(Ui)?

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+
+According to the scores we have, node `N3` will be selected. The utilization fraction of nodes before and after the placement is as follows.
+
+<img src="images/image5.png" alt="before-placement" width="600" height="150"/>
+
+<img src="images/image7.png" alt="after-placement" width="600" height="150"/>
+
+
+If we plot these in mu-sigma plots, we can see the placement automatically pushes the utilization of nodes toward the diagonal line sigma = 1 - mu.
```
Huang-Wei, line 253: So the "1" here is the constant described above. Could you help me understand its real semantics? (as I assume this will be made configurable, so end-users would need to know how to configure it)

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+200 OK response example given above
+404 if no metrics are present
+```
+
+
+
+### **PostBind Plugin**
+
+A custom plugin will be added at post bind extension point, which maintains a time ordered state of node → pod mappings for pods scheduled successfully in last 5 minutes. This state will be maintained across scheduling cycles and used from BestFitBinPack Score plugin. This state will help to predict utilisation based on allocations when metrics are bad, especially in case 1 above (Unavailability of metrics).
```
Huang-Wei, line 356: It seems this act as a backup remediation method? If metrics are all good, we don't need it right?

If so, I'd suggest to get this implemented as a separate "controller" watching goroutine, as (1) it won't slow the whole scheduling cycle, and (2) it's able to watch Pod placements by external schedulers.

Huang-Wei (commented):
Thanks @zorro786 !

Sry for the late response.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+## Motivation
+
+Clusters are inefficient due to a lack of capacity utilization, which leads to increased costs due to more machines, maintenance, etc. The main reason for this is that the default scheduler in K8s doesn’t consider live node utilization values in scheduling decisions, and pod requests are not accurate in practice.
+
+
+### Goals
+
+1. Minimize the number of running nodes.
+2. CPU Utilisation per node should not go beyond X%.
+3. Instances of an app should spread across failure domains. The failure domain would be host at the lowest level and can extend to racks, zones, regions, etc.
```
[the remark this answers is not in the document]
zorro786: The standard bin packing we are doing to minimise average node utilisation in cluster has the above as a constraint for app replicas. Packing app replicas on the same node works against the goal of an application having replicas to provide fault tolerance.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+## Motivation
+
+Clusters are inefficient due to a lack of capacity utilization, which leads to increased costs due to more machines, maintenance, etc. The main reason for this is that the default scheduler in K8s doesn’t consider live node utilization values in scheduling decisions, and pod requests are not accurate in practice.
+
+
+### Goals
+
+1. Minimize the number of running nodes.
+2. CPU Utilisation per node should not go beyond X%.
```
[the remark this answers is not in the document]
zorro786: I guess "should not" is a strict term here as unexpected spikes in utilisation can cause it to go beyond X%. I will modify it as "Expected CPU Utilisation per node should not go beyond X%".

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+3. Instances of an app should spread across failure domains. The failure domain would be host at the lowest level and can extend to racks, zones, regions, etc.
+4. To not affect the behavior of default score plugins unless necessary.
+
+### Non-Goals
+
+1. The constraints above are the best effort of soft constraints. They are not hard constraints.
+2. Descheduling or pre-emption due to lousy scheduling decisions is not addressed in the initial design.
```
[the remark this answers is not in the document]
zorro786: As estimating correct utilisation of an application is a hard problem and we do our best effort from metrics, we may end up with "hot nodes" in the cluster for example, which would render our scoring (past) lousy.  I will rephrase to point it to scoring by the plugin and not overall scheduling decision.

Right preemption can confuse readers. We meant preemption after scheduling cycle is complete, not the pre-emption in extension points. I will rephrase.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+### Goals
+
+1. Minimize the number of running nodes.
+2. CPU Utilisation per node should not go beyond X%.
+3. Instances of an app should spread across failure domains. The failure domain would be host at the lowest level and can extend to racks, zones, regions, etc.
+4. To not affect the behavior of default score plugins unless necessary.
+
+### Non-Goals
+
+1. The constraints above are the best effort of soft constraints. They are not hard constraints.
```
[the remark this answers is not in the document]
zorro786: Filtering would mean making them hard constraints, that is not the goal. We still want to score nodes.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+|  N5  |  0  |  90  |        0       |
+
+
+
+### Safe Balancing Plugin
+
+
+#### PreScore plugin
+
+This extension point, which is invoked for all the nodes at once, will be used to cache metrics for all nodes using a single REST API  call. This optimization will reduce multiple calls for each node from the Score plugin below, reducing overall scheduling latency.
```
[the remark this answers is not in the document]
zorro786: We did some performance benchmarks and PreScore was a bottleneck. This will be replaced with a background go routine that keeps updating metrics from load watcher.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+<img src="images/design.png" alt="design" width="720" height="628"/>
+
+
+
+### Metrics Provider
+
+This service provides metrics backed by a time-series database — for example, Prometheus, InfluxDB, etc.
+
+
+### Load Watcher/Load Analyser
```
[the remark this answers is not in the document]
zorro786: This would be hosted publicly under [PayPal GH](https://github.com/paypal)

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+This uses the scheduler framework of K8s to incorporate our custom real load aware scheduler plugins without modifying the core scheduler code.
+
+
+### BestFitBinPack Plugins
+
+
+#### PreScore plugin
+
+This extension point which is invoked for all the nodes at once will be used to cache metrics for all nodes using a single REST API  call. This optimization will reduce multiple calls for each node from the Score plugin below, thereby reducing overall scheduling latency.
```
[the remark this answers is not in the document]
zorro786: As mentioned earlier PreScore would be replaced with a background go routine.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
zorro786: Directly from load watcher. I will explain this under watcher and add an example json payload in appendix.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
zorro786: This is not exactly a backup remediation. It is a critical component to predict correct utilisation. For example: Consider 3 pods scheduled together. For scheduling the 2nd and 3rd pods, we wouldn't have the correct utilisation values from the nodes because time to schedule (under normal circumstances) << metric-agent reporting interval (usually 1 minute). Without maintaining this state we will end up underestimating the actual utilisation values we get from load watcher.

zorro786: I did some benchmarking and this doesn't slow the scheduling (considerably), it adds one function call which doesn't do any significant work.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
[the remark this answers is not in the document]
zorro786: The interface will be generic, we will be contributing for Prometheus, SignalFX and K8s metrics provider.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
wangchen615: @Huang-Wei, good question. The `1` here indicates `100%` of utilization. What should be configured is a coefficient \ita, which indicates \mu + \ita x \sigma < 100 % and we choose \ita = 1 here. 

Here, we are assuming that the actual usage follows Gaussian distribution and we are using the 68-96-99.5 rule to model the risk of usage exceeding the node capacity. Please refer to [68-95-99.7 rule](https://en.wikipedia.org/wiki/68%E2%80%9395%E2%80%9399.7_rule)

So when \ita gets different values, we get different confidence of not exceeding the capacity.
- \ita = 1, we have a 68% chance that the actual usage is within the node capacity.
- \ita = 2, we have a 95% chance that the actual usage is within the node capacity.
- \ita = 3, we have a 99.7% chance that the actual usage is within the node capacity.

By default, we choose \ita = 1 as we would like to improve the overall utilization.
I will improve the description in the KEP.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+Clusters are inefficient due to a lack of capacity utilization, which leads to increased costs due to more machines, maintenance, etc. The main reason for this is that the default scheduler in K8s doesn’t consider live node utilization values in scheduling decisions, and pod requests are not accurate in practice.
+
+
+### Goals
+
+1. Minimize the number of running nodes.
+2. CPU Utilisation per node should not go beyond X%.
+3. Instances of an app should spread across failure domains. The failure domain would be host at the lowest level and can extend to racks, zones, regions, etc.
+4. To not affect the behavior of default score plugins unless necessary.
+
```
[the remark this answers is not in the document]
zorro786: Hmm we also use Post-Bind plugin, and may use more in the future. I think modifying the 1st goal as you suggested "Provide configurable scheduling plugins to increase cluster utilization." is enough.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
wangchen615: @Huang-Wei , good catch. It is Min(U_i)

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+## Summary
+
+Real Load Aware Scheduling minimizes cluster management costs and makes the cluster more efficient in terms of resource (CPU, Memory) utilization.
+
+
+## Motivation
+
+Clusters are inefficient due to a lack of capacity utilization, which leads to increased costs due to more machines, maintenance, etc. The main reason for this is that the default scheduler in K8s doesn’t consider live node utilization values in scheduling decisions, and pod requests are not accurate in practice.
```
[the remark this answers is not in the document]
zorro786: I will add some background.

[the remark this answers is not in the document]
zorro786: @Huang-Wei How does this sound?:

Kubernetes provides a declarative resource model that core components (scheduler and kubelet) honor to behave consistently and satisfy QoS guarantees.
However, using this model can lead to low-utilization of the cluster for the following reasons:

1. It is hard for users to estimate accurate resource usage for applications. Also, users may not understand the resource model and not set it at all.
2. The default in-tree scheduling plugins (Score) that Kubernetes provides don't consider live node utilization values.

This proposal utilizes real time resource usage to schedule pods with proposed plugins. The eventual goal is to increase cluster utilization and decrease costs of cluster management without breaking the contract of Kubernetes Resource Model.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
+# KEP - Real Load Aware Scheduling
+
+
+## Summary
+
+Real Load Aware Scheduling minimizes cluster management costs and makes the cluster more efficient in terms of resource (CPU, Memory) utilization.
```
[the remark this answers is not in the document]
zorro786: Updated, please check now.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+Example: 
+
+
+| Node | PTS | BFBP | Combined Score |
+|------|-----|------|----------------|
+|  N1  | 100 |  35  |       96       |
+|  N2  |  67 |  75  |       67       |
+|  N3  | 100 |  80  |       100      |
+|  N4  |  82 |  90  |       83       |
+|  N5  |  0  |  90  |        0       |
```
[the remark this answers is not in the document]
zorro786: Will do.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
```diff
@@ -0,0 +1,540 @@
...
+
+
+#### Story 2
+
+As a company-owned data center, we would like to minimize the hardware and maintenance associated with clusters by getting the maximum possible juice out of existing machines.
+
+
+#### Story 3
+
+As a K8s user, I would like to see more enriched single-level scheduler options in the community.
```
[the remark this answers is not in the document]
zorro786: I replaced this with a different story.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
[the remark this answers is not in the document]
zorro786: I have specified why not use them as separate plugins right under "Spreading with BestFitBinPack". Please let me know if it needs to be rephrased or it is confusing.

kep/real-load-aware-scheduling/real-load-aware-scheduling.md
wangchen615: Please see https://github.com/kubernetes-sigs/scheduler-plugins/pull/61/commits/56022e76e84139ac005a15f39c9eaf90b8264fdb

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
```diff
@@ -0,0 +1,608 @@
...
+
+**Algorithm**
+
+
+1. Get the requested resource for the pod to be scheduled as, <img src="https://render.githubusercontent.com/render/math?math=r">.
+2. Get the sliding window average (<img src="https://render.githubusercontent.com/render/math?math=M">) and standard deviation (<img src="https://render.githubusercontent.com/render/math?math=V">) of resource utilization fraction (range from 0 to 1) for all types of resources (CPU, Memory, GPU, etc.) of the current node to be scored. 
+3. Calculate the score of the current node for each type of resource: <img src="https://render.githubusercontent.com/render/math?math=S_i = M %2B r %2B V">
+4. Get a score for each type of resource and bound it to [0,1]: <img src="https://render.githubusercontent.com/render/math?math=S_i = \min(S_i, 1.0)">
+5. Calculate the node priority score per resource as: <img src="https://render.githubusercontent.com/render/math?math=U_i = (1 - S_i) \times MaxPriority ">
+6. Get the final node score as: <img src="https://render.githubusercontent.com/render/math?math=U =\min(U_i)">
```
Huang-Wei, line 223: Two questions:

1. In the initial design, we only care about CPU, so we only calculate the score for CPU, or we still calculate the others?
1. Once we consider multiple metrics, can we say we support [DRF](https://amplab.cs.berkeley.edu/publication/dominant-resource-fairness-fair-allocation-of-multiple-resources-types/)?

Huang-Wei:
Two more comments here (as Github doesn't allow me to comment on the "outdated" comments):

Re: https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r503633291

I got it. Another tip is to have an immutable "snapshot" for your metrics, so that during one scheduling cycle, this snapshot is the source of truth across different plugins - no matter it's a little bit behind the real time metrics or not. Otherwise, different plugins may obtain different versions of metrics, which would make the result unpredictable and hard to debug.

And in terms of implementation, for sure you should have a goroutine receiving metrics periodically or on demand. On the other hand, you can snapshot the metrics into `PreScoreState` (works like a session which is valid during one scheduling cycle) in `PreScore` phase, so you only have to do this once, and reuse the same copy of metric snapshot in both plugins, at the later `Score` phase.

Re: https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r503636836

It's not only about overhead/latency. `PostBind()` is triggered when a Pod gets scheduled, and more importantly, only when the Pod is supposed to by scheduled by the scheduling profile your plugins get registered to. Along with multi-profile scheduler, it's common to define multiple profiles, and the plugins describe here are only registered to one profile. In that case, you'd miss the chances to capture the events of Pods scheduled by other profiles. So PostBind() is not an ideal place to capture all Pods being scheduled. Use a regular controller to watch on pods whose `.spec.nodeName` gets set is the most generic way.

Huang-Wei:
cc/ @ahg-g and @alculquicondor in case you want to review this.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
```diff
@@ -0,0 +1,608 @@
...
+
+
+
+1. For each node, calculate S = (1000*PTS + PTS*BFBP)
+2. Scale each S to be between 0-100
+3. Return the scores
+
+**Algorithm Analysis**
+
+1. The algorithm favors nodes with high PTS scores with 1000*PTS part. The weight 1000 is chosen to give preference to PTS  and to avoid the congestion of scores upon scaling down, my magnifying the scores
```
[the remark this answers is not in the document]
zorro786: You are right, the reason for combining is explained below in bold:

"However, if we choose a static large weight for PTS, say 100, we can meet our goals. 
**The problem with this approach is that we will end up downplaying the scores of other important default in-tree plugins like Image Locality, Interpod Affinity, etc. with their default weights of 1**. 
**Also, with more plugins we plan to add in the future, managing weights will not be simple**. So for BFBP to work well with PTS without modifying weights, 
the following extended algorithm is proposed that is called when scheduling replicas of a service, replica-set, replication controller, etc."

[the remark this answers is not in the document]
zorro786: I think a large weight works well for "standalone" plugins which are kind of hard requirements. In our case spreading is not a hard requirement. If PTS scores of two nodes are nearly equal (say 70, 80), we want to favor nodes with better image locality (say 100, 50). If we use 1000 as PTS weight, we will end up selecting node with 80 score always, when we want to select node with 70 score (due to better image locality). This is what "downplaying" scores means above.

With multiple plugins and multiple weights, the interplay of scores gets complex and hard to understand/debug - this is what "managing weights will not be simple" means.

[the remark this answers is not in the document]
zorro786: The above goes with over goal of "To not affect the behavior of default score plugins unless necessary."

zorro786:
> Re: [#61 (comment)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r503636836)
> 
> It's not only about overhead/latency. `PostBind()` is triggered when a Pod gets scheduled, and more importantly, only when the Pod is supposed to by scheduled by the scheduling profile your plugins get registered to. Along with multi-profile scheduler, it's common to define multiple profiles, and the plugins describe here are only registered to one profile. In that case, you'd miss the chances to capture the events of Pods scheduled by other profiles. So PostBind() is not an ideal place to capture all Pods being scheduled. Use a regular controller to watch on pods whose `.spec.nodeName` gets set is the most generic way.

Ah I didn't realize `PostBind()` is per profile. Is it guaranteed that `.spec.nodeName` is set only after successful binding? If kubelet evicts the pod after successful binding, is the `.spec.nodeName` reverted? Since this state is time critical for plugin, what's the best way to watch for `.spec.nodeName` changes? Any APIs k8s provides for this event watching I can hook in to?

zorro786:
> Re: [#61 (comment)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r503633291)
> 
> I got it. Another tip is to have an immutable "snapshot" for your metrics, so that during one scheduling cycle, this snapshot is the source of truth across different plugins - no matter it's a little bit behind the real time metrics or not. Otherwise, different plugins may obtain different versions of metrics, which would make the result unpredictable and hard to debug.
> 
> And in terms of implementation, for sure you should have a goroutine receiving metrics periodically or on demand. On the other hand, you can snapshot the metrics into `PreScoreState` (works like a session which is valid during one scheduling cycle) in `PreScore` phase, so you only have to do this once, and reuse the same copy of metric snapshot in both plugins, at the later `Score` phase.
> 

Interesting idea - I think importance of latest metrics is higher here. Also, we wouldn't have two different score plugins working on utilisation metrics for core scoring, it would result in conflicts. In other words we shouldn't enable "SafeBalancing" and "BestFitBinPack" plugins together - it is up to the users to choose which one they want to use. I think this may be useful if we have two load aware plugins that work simultaneously, for example CPU & Memory plugins that are complementary to each other - but we would be using multi dimensional bin packing in "BestFitBinPack" to achieve this and not have another plugin for "Memory".

Also for the scoring aspect each node is scored independently based on its latest utilization and in terms of correctness this is better than having stale metrics (even with small difference). In fact the naive solution was to fetch latest metric for a given node each time when scoring -> the plugin cache from load watcher is after optimizations, so we very much want latest metrics always when possible. In terms of predictability and debugging - the formula is simple and very much independent of utilization values of other nodes, so I don't see much of a problem here.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
Huang-Wei: I'm not sure I like this.

It's inevitable that Score plugins may conflict with each other, or one plugin's impact gets diluted by others. Scoring is best-efforts anyway, expecting a 100% predictable scoring result is almost impossible. Even if you mixed a default weight of PTS into BFBP, it may not always work as expected. I'd still suggest BFBP be implemented on its own. You can guide the users to configure the weight of PTS when using BFBP.
On the other hand, mixing PTS score plugin would make the BFBP plugin hard to maintain. The PTS score plugin is not guaranteed to be exposed as a library, and can change its function API at any time, which makes the development of BFBP into risk.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
Huang-Wei: Bump.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
zorro786: I agree about the conflict - it is unavoidable to have 0 conflict. What is important to understand here is that we try to **minimise** the conflict where we can. This also goes along the idea of a fair scheduler with different plugins given the weight of 1 and none of them is favored more over others.

>>>The PTS score plugin is not guaranteed to be exposed as a library, and can change its function API at any time, which makes the development of BFBP into risk."

About this, yes it could be a problem - I thought the framework API is almost stable now :)

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
zorro786: @wangchen615 ^

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
Huang-Wei: @ahg-g @alculquicondor any thoughts about mixing a plugin into another?

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
wangchen615: @Huang-Wei , I think it is an orthogonal question. DRF can be supported in both cases where the real load on CPU/memory is considered or not. New DRF plugins need to be implemented. They can either take the available resources for new requests (in terms of allocatable) or the actual available resources (in terms of actual usage). Right now, we do not have a concrete plan to implement any DRF plugins yet. However, if there are some interests in DRF plugins in the community, we are happy to help. For the DRF plugin, it would be easy to start with the allocatable and then see whether we can adapt to the real load version when the load watcher is ready.

Huang-Wei:
/ok-to-test

@zorro786 can you sign the CLA?

zorro786:
@Huang-Wei Thank you, waiting on Legal team, process was started 2 weeks back. Expecting it to be done soon.

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+### Risks and Mitigations
+
+If utilization metrics are not available for a long time, we will fall back to the best fit bin pack based on allocations. There is no user action needed for this.
+
+
+## Design Details
+
+Our design consists of the following components as outlined in the diagram and described below. We propose two plugins namely "BestFitBinPack" and "SafeBalancing". Both of them use metrics from load watcher for scoring nodes with different algorithms.
+
+<img src="images/design.png" alt="design" width="720" height="628"/>
```
alculquicondor, line 67: nit:
`Design`

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+This plugin would extend the Score extension point. K8s scheduler framework calls the Score function for each node separately when scheduling a pod.
+
+Following is the algorithm:
+
+**Algorithm**
+
+1. Get the utilization of the current node to be scored. Call it A.
+2. Calculate the current pod's total CPU requests and overhead. Call it B.
+3. Calculate the expected utilization if the pod is scheduled under this node by adding i.e. U = A + B.
+4. If U &lt;= 50%, return U+50 as the score
```
alculquicondor, line 111: what if `x != 50`.
I would assume you want to use 50 as the default, but why is that?
In an scenario of multiple pods scheduled in a short amount of time, your live utilization data is likely to be outdated, so several Pods might be scheduled in the same node under the same "current utilization". Have you considered this?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+5. If 50% &lt; U &lt;= 100%, return 100 - U
```
alculquicondor, line 112: I still think this discontinuous function might be erratic when you have close to 50% utilization, specially given that the function uses live data. Have you experimented with other functions?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+### Notes/Constraints/Caveats
+
+Enabling our plugin(s) will cause conflict with 2 default scoring plugins: "NodeResourcesLeastAllocated" and "NodeResourcesBalancedAllocation" plugins. So it is strongly advised to disable them when enabling plugin(s) mentioned in this proposal.
+For the 3rd Goal, since we utilise PodTopologySpread plugin scoring and to prevent double scoring, it is recommended to disable "PodTopologySpread" Score plugin only. The motivation behind using this plugin is explained under "Spreading with BestFitBinPack".
+
+
+### Risks and Mitigations
+
+If utilization metrics are not available for a long time, we will fall back to the best fit bin pack based on allocations. There is no user action needed for this.
```
alculquicondor, line 60: nit: Add reference to where this is analyzed. `#bad-metrics`?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+This is a localhost database stored as a file. This DB aims to populate the load watcher's cache quickly if it is lost due to a crash and perform scheduling decisions without any impact on latency.
+The file will be stored in host file system, so it will be persisted across pod failures on the same node. For HA setups, each load watcher process will maintain its corresponding file locally.
+
+
+### Scheduler Plugins
+
+This uses the scheduler framework of K8s to incorporate our customized real load aware scheduler plugins without modifying the core scheduler code. The plugins we proposed mainly include the following two.
+
+- BestFitBinPack Plugin: It is a node sorting plugin that sorts nodes by their actual resource utilization in a way that all utilized nodes have around x% of utilization. We use x% = 50% below to describe the plugin.
```
alculquicondor, line 93: I would suggest a name that reflect the "live" nature of these plugins (as opposed to static requests)

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+3. The positive slope begins from 50 and not from 0 because the range of score is from 0-100, and we want to maximize our score output for nodes we wish to favor so that the score is more substantial and other plugins do not affect it much.
+4. There is a break in the graph with a high drop due to the "penalty" we have.
+
+#### Spreading with BestFitBinPack
+
+
+To meet the 2nd constraint in the problem statement, we utilize the existing PodTopologySpread (PTS) plugin provided by K8s. 
+This is to avoid baking in another spreading algorithm within BFBP when there is an existing in-tree plugin and conflict our 4th goal.  
+It was proved with experiments that using a dynamic weight-based approach (changing weights of PTS and BFBP in every scheduling cycle) is not a good idea, neither it is supported by Kubernetes. 
+However, if we choose a static large weight for PTS, say 100, we can meet our goals.
```
alculquicondor, line 163: why 100 instead of the default (2)? Have you experimented with 1.19?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+1. It is hard for users to estimate accurate resource usage for applications. Also, users may not understand the resource model and not set it at all.
+2. The default in-tree scheduling plugins (Score) that Kubernetes provides don't consider live node utilization values.
+
+This proposal utilizes real time resource usage to schedule pods with proposed plugins. The eventual goal is to increase cluster utilization and decrease costs of cluster management without breaking the contract of Kubernetes Resource Model.
+
+### Goals
+
+1. Provide configurable scheduling plugins to increase cluster utilization.
+2. Expected CPU Utilisation per node should not go beyond X%.
+3. Instances of an app should spread across failure domains. The failure domain would be host at the lowest level and can extend to racks, zones, regions, etc.
```
alculquicondor, line 22: why is this a goal, given that no plugin will be introduced to implement it

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+The problem with this approach is that we will end up downplaying the scores of other important default in-tree plugins like Image Locality, Interpod Affinity, etc. with their default weights of 1. 
+Also, with more plugins we plan to add in the future, managing weights will not be simple. So for BFBP to work well with PTS without modifying weights, 
+the following extended algorithm is proposed that is called when scheduling replicas of a service, replica-set, replication controller, etc.
+
+
+**Algorithm**
+
+
+
+1. For each node, calculate S = (1000\*PTS + PTS\*BFBP)
```
alculquicondor, line 173: are you proposing to do this inside your plugin?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+Minimizing machine costs by utilizing all nodes is the main objective for efficient cluster management. To achieve this goal, we can make the Kubernetes scheduler aware of the gap between resource allocation and actual resource utilization. Taking advantage of the gap may help pack pods more efficiently, while the default scheduler that only considers pod requests and allocable resources on nodes cannot.
+
+## Motivation
+
+Kubernetes provides a declarative resource model that core components (scheduler and kubelet) honor to behave consistently and satisfy QoS guarantees.
+However, using this model can lead to low-utilization of the cluster for the following reasons:
+
+1. It is hard for users to estimate accurate resource usage for applications. Also, users may not understand the resource model and not set it at all.
+2. The default in-tree scheduling plugins (Score) that Kubernetes provides don't consider live node utilization values.
```
alculquicondor, line 14: Playing devils' advocate here: Have you considered using the Vertical Pod Autoscaler and the existing plugins?

kep/61-real-load-aware-scheduling/README.md
zorro786: That doesn't resize the image :), this one is of high quality and default size is quite big.

kep/61-real-load-aware-scheduling/README.md
zorro786: This would be made configurable from plugin args. We considered 50 as default because it is our requirement, it can be anything otherwise :)

zorro786: > In an scenario of multiple pods scheduled in a short amount of time, your live utilization data is likely to be outdated, so several Pods might be scheduled in the same node under the same "current utilization". Have you considered this?

Yes sir this is explained under "Bad Metrics" and taken care under "Scheduled Pods State". Also refer to [this](https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r503636836) discussion.

kep/61-real-load-aware-scheduling/README.md
zorro786: 2 and 3 are more like constraints for Goal 1, they are not independent goals...

kep/61-real-load-aware-scheduling/README.md
zorro786: Yes.

kep/61-real-load-aware-scheduling/README.md
alculquicondor: Even if your default is 50, you should show the generic formula, as it is not clear how to generalize from here

kep/61-real-load-aware-scheduling/README.md
alculquicondor: Why? This should be up to users

kep/61-real-load-aware-scheduling/README.md
zorro786: We _did_ experiment the final scoring function with 1.19, not the above part of assigning a static weight. It may be 2 for all we know, it doesn't matter as it is just an example.

alculquicondor: It does matter, to justify merging the scores in a single plugin

kep/61-real-load-aware-scheduling/README.md
zorro786: Good point - given the nature of live data and error rate involved, the value of U should actually be 10 less than the desired utilisation value -> as per our experiments. I think I should specify this in Notes/Caveats..

kep/61-real-load-aware-scheduling/README.md
Huang-Wei: My option: VPA may become more practical when in-place pod req update is supported in the upstream. For now, every adjustment VPA makes needs to restart the Pod (correct me if I'm wrong), which is somewhat unacceptable is most scenarios.

alculquicondor: Yeah, in place updates was what was in my mind.
It doesn't look like there has been any progress on that :/ https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20181106-in-place-update-of-pod-resources.md

wangchen615: Yes, I was trying to help with the in-place pod update KEP. As it required some API changes, it may take a longer time to merge upstream. We need to wait at least until 1.20.

Besides, VPA with in-place update cannot solve all the goals we would like to achieve. 
- For example, the risk balancing plugin is taking the usage variations into consideration. Without understanding how the actual usage varies over time, it is impossible to estimate the risk. 
- The BFBP plugin aims to maintain all nodes' average utilization of around x%. VPA applied to a group of pods may have more accurate requests, but they may not help in achieving the overall node utilization goal in a cluster. Even if VPA is applied for all pods in a cluster, taking the average load on nodes in scoring nodes is much easier than taking the requests as they might be fluctuating all the time.

kep/61-real-load-aware-scheduling/README.md
zorro786: Sure, will update. However will keep 50 for example that follows.

kep/61-real-load-aware-scheduling/README.md
zorro786: Not sure if I follow you, for justification part, please follow this discussion https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r505098463 
Since we are not proposing to use a static weight in this, Even if I do the experiment again for static weight with 1.19, I will just update "100" value above to what works - say "2". It doesn't change the proposal idea.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
alculquicondor: > The problem with this approach is that we will end up downplaying the scores of other important default in-tree plugins like Image Locality, Interpod Affinity, etc. with their default weights of 1.

If we just keep the plugins separate, with a weight of 2 (default), we wouldn't affect other plugins.

>  You can guide the users to configure the weight of PTS when using BFBP.

+1

kep/61-real-load-aware-scheduling/README.md
zorro786: @alculquicondor Do you believe the value of "x" could be much lower in 1.19?

kep/61-real-load-aware-scheduling/README.md
zorro786: Something like "LoadAware" prefix added? -> LoadAwareBestFitBinPack, LoadAwareSafeBalancing ?

kep/61-real-load-aware-scheduling/README.md
alculquicondor: yes

kep/61-real-load-aware-scheduling/README.md
alculquicondor: but users can already configure that using the framework

zorro786:
I signed it

zorro786:
@Huang-Wei CLA check passed.

kep/61-real-load-aware-scheduling/README.md
zorro786: I got your point, let me add an option to turn off spreading in docs and use static weights instead as an alternative.

kep/61-real-load-aware-scheduling/README.md
zorro786: Okay, I have removed the "100" value, and specified the extended algorithm would be configurable to turn off and use static weights. Since experiments will take a while, I would add what value we recommend if using static configuration in a later PR and want this PR merged. @Huang-Wei @alculquicondor

alculquicondor: The problem is not so much about the value that you suggest, but the fact that you are merging 2 plugins into one. I don't think there is justification for that, as we have already pointed out in https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r505098463

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
zorro786: @alculquicondor I doubt if weight of 2 will solve our goals in 1.19. If it does then I agree the justification above is not valid. Let me try this out.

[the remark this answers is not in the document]
alculquicondor: topology spreading in 1.19 is significantly different from 1.18 kubernetes/kubernetes#90475

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,612 @@
...
+
+
+#### Story 2
+
+Increasing resource utilization as much as possible may not be the right solution for all clusters. As it always takes some time to scale up a cluster to handle the sudden spikes of load, cluster admins would like to leave adequate room for the bursty load, so there is enough time to add more nodes to the cluster.
+
+
+### Notes/Constraints/Caveats
+
+Enabling our plugin(s) will cause conflict with 2 default scoring plugins: "NodeResourcesLeastAllocated" and "NodeResourcesBalancedAllocation" plugins. So it is strongly advised to disable them when enabling plugin(s) mentioned in this proposal.
```
damemi, line 54: Will this also conflict with `NodeResourcesMostAllocated`? (it's disabled by default but good to mention if so). The bin packing plugin seems similar to it

kep/61-real-load-aware-scheduling/README.md
damemi: in this case BestFitBinPack based just on allocations seems similar to the existing `NodeResourcesMostAllocated` (where NRMA is just BFBP with a target utilization of 100%). If that's the case, maybe we could look at just extending NRMA to have a configurable target utilization + live metrics, rather than introducing a new plugin?

Though I may be oversimplifying it, but just a thought

damemi (commented):
I'm a little late to this, but to add a couple comments I think this could also be helpful when combined with [LowNodeUtilization](https://github.com/kubernetes-sigs/descheduler/#lownodeutilization) in the descheduler (or could even be imported there)

And not sure if this was mentioned but would the metrics these plugins use be provided by the Pod resource metrics KEP? https://github.com/kubernetes/enhancements/pull/1916

alculquicondor:
> And not sure if this was mentioned but would the metrics these plugins use be provided by the Pod resource metrics KEP?

Actually no, as those deal with requests and limits.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
zorro786: @alculquicondor good job for your work on PTS. I found no problems when I tested it for spreading with new default weight (as compared to 1.18). Since this works with weight of 2 for simple use cases, I believe for complex topologies or with more number of nodes, we wouldn't have to use a "large weight" like earlier to achieve our goals. I have removed the extended algorithm from KEP.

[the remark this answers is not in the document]
zorro786: @Huang-Wei @alculquicondor please review/approve.

kep/61-real-load-aware-scheduling/README.md
zorro786: Yes it will. I added that.

kep/61-real-load-aware-scheduling/README.md
zorro786: Since this is more like a fall-back in case of failures, and it would be utilized by multiple types of "load aware" plugins (Both SafeBalancing and BestFitBinPack in this PR) extending `NodeResourcesMostAllocated` may not be a good idea. Perhaps falling back to `default-scheduler` profile is one way to achieve this, if real load plugins can not work.

kep/61-real-load-aware-scheduling/61-real-load-aware-scheduling.md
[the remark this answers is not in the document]
Huang-Wei: SG, you don't need to specify the default weight for now. Once you got a reasonable default number, return and update the KEP.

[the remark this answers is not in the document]
Huang-Wei: SG. If you found that a larger weight is needed (for most cases), just update the wordings in this KEP.

kep/61-real-load-aware-scheduling/README.md
Huang-Wei: Can we say "RealtimeLoadAware"? A bit verbose though...

I guess "BestFix" doesn't add additional context, can we name it "<prefix to indicate it's based on live data>" + "BinPack"?

Huang-Wei:
@zorro786 @wangchen615 Thanks for addressing the comments. Given that we've got a consensus to not mix PodTopologySpread into BestFitBinPack, I think we're close to getting it merged.

The last comment seems to be the naming: https://github.com/kubernetes-sigs/scheduler-plugins/pull/61#discussion_r508600200.

@alculquicondor @damemi Thanks a lot for reviewing this! Please bump the left exceptional comments if there is any.

kep/61-real-load-aware-scheduling/README.md
wangchen615: @Huang-Wei , for safe balancing plugin, we can just call it SafeBalancing as risks can only be balanced with usage variations under consideration.

For BFBP,  it really does more than bin packing. It will do spreading when all nodes' usage exceeding x%. I will leave it to @zorro786.

Besides, we have an agreed name for the whole framework, called "Trimaran". Given that Kubernetes relates to boating, `Trimaran` has three hulls spaced out apart from one another, so it can reduce wave impact, increase stability, and maintain its beam moving forward.

> Can we say "RealtimeLoadAware"? A bit verbose though...
> 
> I guess "BestFix" doesn't add additional context, can we name it "<prefix to indicate it's based on live data>" + "BinPack"?

zorro786: @Huang-Wei "Load Aware" by default refers to real load. We say load balancer and not real load balancer for example. How about "LoadAwareBinPack"? -> As @wangchen615 mentioned it does more than "Best fit" bin pack, it moves to "Least fit" variant of bin pack if usage exceeds x%. Problem with this name is if we want to add "Least Fit" variant plugin later the name will be ambiguous and needs to be changed. So right now can't think of a better name than "LoadAwareBestFitBinPack".....

zorro786: "MetricAwareBestFitBinPack" is an alternative but I prefer "Load"

kep/61-real-load-aware-scheduling/README.md
alculquicondor: Could you still link to `#bad-metrics` so readers can know that there is more to this topic?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,572 @@
...
+
+### Scheduler Plugins
+
+This uses the scheduler framework of K8s to incorporate our customized real load aware scheduler plugins without modifying the core scheduler code. The plugins we proposed mainly include the following two.
+
+- BestFitBinPack Plugin: It is best fit variant of bin pack algorithm that scores nodes by their actual resource utilization in a way that all utilized nodes have around x% of utilization.
+- Safe Balancing Plugin: It is a node sorting plugin that sorts nodes base on both the mean and the standard deviation of node resource utilization. It aims to balance not only the average load but also the risk caused by load variations.
+
+
+### BestFitBinPack Plugin
```
alculquicondor, line 98: For whatever reason, I can't comment in the old thread.

I would name it `TargetNodeLoad`

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,572 @@
...
+This plugin would extend the Score extension point. K8s scheduler framework calls the Score function for each node separately when scheduling a pod.
+
+Following is the algorithm:
+
+**Algorithm**
+
+1. Get the utilization of the current node to be scored. Call it A.
+2. Calculate the current pod's total CPU requests and overhead. Call it B.
+3. Calculate the expected utilization if the pod is scheduled under this node by adding i.e. U = A + B.
+4. If U &lt;= X%, return U+X as the score
```
alculquicondor, line 112: say X = 70 and U = 60, then limits don't work with these formulas.

And again, I don't think I heard back about the discontinuous function. Did you get to try a continuous function?
It would be interesting to see which formula gives you better utilization over a period of time.

By continuous I mean something like: 50 to 100 score from 0 to 50% utilization. Then 100 to 0 score from 50 to 100% utilization

kep/61-real-load-aware-scheduling/README.md
alculquicondor: Can you split the conflicts for each of your plugins? Do they both conflict with all the plugins mentioned here?

kep/61-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,572 @@
...
+3. The positive slope begins from 50 and not from 0 because the range of score is from 0-100, and we want to maximize our score output for nodes we wish to favor so that the score is more substantial and other plugins do not affect it much.
+4. There is a break in the graph with a high drop due to the "penalty" we have.
+
+
+### Safe Balancing Plugin
+
+
+#### Score plugin
+
+This plugin would extend the Score extension point. K8s scheduler framework calls the Score function for each node separately when scheduling a pod.
```
alculquicondor, line 165: Is this plugin the same as NodeResourcesBalancedAllocation but with live data?

alculquicondor (commented):
There are a lot of mentions of configurable options all over the KEP. Could you include the PluginArgs struct so that we can visualize and revise their semantics in a single location?

kep/61-real-load-aware-scheduling/README.md
zorro786: Yes you are right, X=50% is a special use case per our requirements, since I recently changed the algorithm to generalize I missed this part. I will add the correct generalized equations. Good catch.

The discontinuity here can be thought of as a penalty for nodes going above target utilisation as we want to be aggressive in keeping the node under target when possible, and then move from best fit to least fit variant of bin packing if no nodes are available with < target utilisation. A break in continuity is needed to incorporate penalty in scoring function. The continuous function you mentioned is closer to the target, which we want to avoid. Based on this reasoning we didn't try a continuous function, and this function works good per experiments and simulations.

kep/61-real-load-aware-scheduling/README.md
zorro786: @wangchen615 ^^

kep/61-real-load-aware-scheduling/README.md
zorro786: They both conflict with plugins mentioned, hence the `plugin(s)` part.

kep/61-real-load-aware-scheduling/README.md
zorro786: Target is a bit misleading because we miss the target if we have no free nodes. Also achieving "Target load" is just a constraint it is not main goal, main goal is to increase cluster utilisation.
How about `LoadAwareNodePacking`? We always pack the nodes, whether below target or above, and it is load aware.

kep/61-real-load-aware-scheduling/README.md
zorro786: Done

zorro786: > The architecture required to obtain and cache the metrics might be a non-neglibile risk, even if the feature is disabled. I prefer this evolves independently first. But we could upstream the "target utilization" bits and somehow reuse that code in here.

For some reason this newer comment is shown after my older comment and I missed this. Yeah we don't have a clear implementation yet or design for how the fall back works, it adds quite a bit of complexity -> "for example when does fall back kick in, is it for all cluster nodes or just a few nodes?"
We would like to analyse this further with time and see the best solution. If you want I can remove this part from KEP.

zorro786: @Huang-Wei your thoughts on this?

kep/61-real-load-aware-scheduling/README.md
wangchen615: > Is this plugin the same as NodeResourcesBalancedAllocation but with live data?

@alculquicondor, No. This scoring plugin is taking the usage variations + the usage mean to sort nodes. It's more than balancing the allocation as it also balances the risk of under-provisioning.

kep/61-real-load-aware-scheduling/README.md
Huang-Wei: I agree that you can build this standalonely. However the ideal case is to decouple the implementations can reuse the commons as much as possible, specifically:

1. Abstract the metric source - like in the upstream, it's from the pod's spec and cluster's initial capacity; while in this plugin, it's from realtime metrics
1. With the metric source serving as an contract/interface, we can build a more common "utilization" based algorithm, which can be used by both upstream and the plugins discussed here.

Maybe you can mention this in the KEP, but up to you:)

kep/61-real-load-aware-scheduling/README.md
Huang-Wei: > We say load balancer and not real load balancer for example.

That's fair :) LoadAwareBestFitBinPack sounds good to me. @alculquicondor any thoughts?

kep/61-real-load-aware-scheduling/README.md
Huang-Wei: I understand you want to penalize the node that has utilization larger than X%, but @alculquicondor 's point is that a node with utilization of 51% should be penalized that much, comparing to a node with 49% utilization - the score drops from nearly 100 to 50.

So, does it make sense to update the formula on the range of [50%, 100%] to be `2•(100-U)` when X% is 50%? And a general formula is like: `100•(100-U) / (100-X)`.

zorro786: "utilization of 51% should be penalized that much, comparing to a node with 49% utilization"
Better way to put this is we want to "favor" nodes <= 50% utilization and "penalise" nodes with > 50% utilisation.

With 2*(100-U) the problem is that we will score node with utilization 40 and 65 to be same i.e. 90. This would go against the constraint we have as we have equal chance of selecting underutilised and overutilised nodes. Hope this example makes it clear for the need of a break.

kep/61-real-load-aware-scheduling/README.md
damemi: `LoadAwareBinPack`? The "BestFit" seems kind of implied imo

kep/61-real-load-aware-scheduling/README.md
Huang-Wei: > With 2*(100-U) the problem is that we will score node with utilization 40 and 65 to be same i.e. 90.

How about using a logarithmic formula to ensure node with utilization 65% gets scored less than 40%?

Huang-Wei: Hold on... Isn't 40% gets a Score of 90, and 65% gets a Score of 70?

zorro786: If the issue is about continuity, we _could_ explore a continuous non linear function for utilisation > 50%, and analyse how it compares to a break. We can leave it for a future research project.

zorro786: Yup sorry not 65 -> 55*

kep/61-real-load-aware-scheduling/README.md
zorro786: I am fine with `LoadAwareBestFitBinPack` or `LoadAwareBinPack` both. You guys decide and let me know  :sweat_smile:

kep/61-real-load-aware-scheduling/README.md
alculquicondor: > Also achieving "Target load" is just a constraint it is not main goal, main goal is to increase cluster utilisation.

I disagree. If that was your goal, you shouldn't have the penalty once the utilization is over 50. This plugin is trying to have nodes in the same target utilization, and if there are not enough resources left, it won't be able to. But that happens for any other Plugin.

@damemi @Huang-Wei moving the conversation of naming here, because github UI is great :)

kep/61-real-load-aware-scheduling/README.md
alculquicondor: > We can leave it for a future research project or enhancement.

Fair, it's a score function after all.

kep/61-real-load-aware-scheduling/README.md
zorro786: Okay your point is fair. I see three candidates:

1) `LoadAwareBinPack`
2) `LoadAwareBestFitBinPack`
3) `TargetNodeLoad`

My vote is for 1 or 2. One thing missing from 3 for me is bin pack behavior. It being a standard algorithm helps in user adoption for me just by looking at name.

wangchen615: Voting for 2. `LoadAwareBestFitBinPack`

kep/61-real-load-aware-scheduling/README.md
lenhattan86: I think It is too hard to have a perfect score function that works for everyone. We may have 2 separate score functions if we need but lets keep the version for now.

kep/61-Trimaran-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,588 @@
...
+
+#### Score plugin
+
+This plugin would extend the Score extension point. K8s scheduler framework calls the Score function for each node separately when scheduling a pod.
+
+
+**Idea**
+
+
+Balancing load based on average would be risky sometimes, as it does not consider the bursty variations. A safe balancing plugin balances not only the average load but also the risk caused by load variations. Suppose we take the mean (M) and standard deviation (V) of all nodes’ utilization into a mu-sigma plot below. In that case, the safe balancing plugin will make placements such that all nodes’ utilization are aligned on the diagonal line, which is V + M = c. Here, c is a constant indicating the overall cluster utilization average plus the standard deviation.
```
alculquicondor, line 179: Could you add a one line description of what this Plugin tries to favor?

My thinking is that it prefers nodes with low variance but high utilization or high variance but low utilization. Correct?

The name of the plugin doesn't give any clue on what the plugin does.

wangchen615: @alculquicondor , yes, I will add the following into the KEP.

`Considering load on all nodes fluctuates dynamically over time, the plugin favors nodes that have lower risks of the load exceeding capacity.`

kep/61-Trimaran-real-load-aware-scheduling/README.md
```diff
@@ -0,0 +1,588 @@
...
+**Plugin Config Args**
+```go
+type PluginArgs struct {
+    TargetCPUUtilisation    int
+    DefaultCPURequests      v1.ResourceList
+}
+```
+
+
+### Safe Balancing Plugin
```
alculquicondor, line 168: What about "Safe Utilization"?

wangchen615: @alculquicondor, sounds good too. We refer to balancing as it balances the risks and the risk denotes the fluctuating utilization. How about `Risk Balancing plugin`?

alculquicondor: But what is it balancing? Honestly, it's going to be hard to come up with a good short name. But at least it should have a similar suffix to the other plugin, given that it uses the same source of information and it's trying to do something about the Load or Utilization.

atantawi: I like “risk balancing”. The suggestion “safe utilization” may imply that we’re determining a value of utilization that is safe, and that’s not true.

atantawi: "Safe Loading"?

wangchen615: Will rename the `Safe Balancing` plugin to `LoadVariationRiskBalancing` plugin.

kep/61-real-load-aware-scheduling/README.md
zorro786: Updated to `TargetLoadPacking` per discussion.

zorro786:
@Huang-Wei @alculquicondor Thank you for your reviews. All reviews have been addressed. If anything missed please bump.

Huang-Wei:
Thanks @zorro786. Please squash the commits. It's ready to be merged

alculquicondor:
LGTM after squash

zorro786:
@Huang-Wei rebased to master and squashed.

Huang-Wei:
/approve

Thanks @zorro786 , @wangchen615 and @alculquicondor !

PR #62, opened 2020-09-26
Work Around Git Command Failure In Image Build

seanmalloy:
The automated container image builds are failing with the below error in
the logs.

fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
docker build -t gcr.io/k8s-staging-scheduler-plugins/kube-scheduler: .
invalid argument "gcr.io/k8s-staging-scheduler-plugins/kube-scheduler:" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.

This work around will fall back to using the date command if the
"git describe" command fails. This will ensure that the RELEASE_VERSION
variable is always set to a non-empty string.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ LOCAL_CONTROLLER_IMAGE=controller:latest
 # into. The default is to push to the staging
 # registry, not production(k8s.gcr.io).
 RELEASE_REGISTRY?=gcr.io/k8s-staging-scheduler-plugins
-RELEASE_VERSION?=$(shell git describe --tags --match "v*")
+RELEASE_VERSION?=$(shell git describe --tags --match "v*" || date +%Y-%m-%d-%H-%M-%S)
 RELEASE_IMAGE:=kube-scheduler:$(RELEASE_VERSION)
 RELEASE_CONTROLLER_IMAGE:=controller:$(RELEASE_VERSION)
 


seanmalloy:
I honestly do not understand why the "git describe" command is failing in the automated build environment, but I am hopeful that this change will work around the issue. We are using the exactly same "git describe" command in the deschduler Makefile and the automated image builds are working just fine: Ref: https://github.com/kubernetes-sigs/descheduler/blob/5d65a9ad683f6bfad965ef92000d5e330a230b30/Makefile#L18.

Keep in mind the container image tags that are pushed to the staging registry are not the tags that are used in the production `k8s.gcr.io` registry, so the time stamp tags will not show up in the production registry.

/assign @Huang-Wei 
/cc @denkensk

Huang-Wei:
@seanmalloy Re:

> fatal: not a git repository (or any parent up to mount point /)

Can it because the `git describe` command is not running under the project repo? I guess there _may be_ a tiny config parameter controlling the working direction?

seanmalloy:
@Huang-Wei I submitted #66 as an alternative. I'm not sure if that will fix the build issue, but maybe it is worth a shot.

I double checked again all the configurations for descheduler vs scheduler-plugins repo for the automated build process, but I cannot find anything that I missed. I'm not aware of a setting to set the work directory.

Beyond these two possible fixes I think I would need to ask in `#wg-k8s-infra` on Slack for help. Let me know if you are hesitant to merge either of these fixes and I can try to ask for further troubleshooting assistance on Slack.

Huang-Wei:
@seanmalloy We're not in a hurry. Let's give #66 a try, while also see if any hints can get obtained from infra guys.

Really appreciate all the work!

seanmalloy:
/hold

I will try to get a better understanding of exactly why the `git describe` command is failing, so let's put this on hold for now.

seanmalloy:
A more correct fix is proposed in #67.

/close

PR #64, opened 2020-09-27
Improve coscheduling's PodGroup cleanup

yuanchen8911:
When a new pod with a PodGroup is submitted, if the PodGroup exists but has to be cleaned up (GC) yet, the current implementation resets the PodGroup active but still uses the existing PodGroup's minAvailable. Thus, deleting a PodGroup, updating its minAvailable and resubmitting it won't work.

This PR makes a simple improvement by updating the existing PodGroup's minAvailable value by the new Pod's. It works for general cases and allows us to use a larger GC interval.


denkensk:
Thanks.
/lgtm

yuanchen8911:
@Huang-Wei Can you review it when you have a moment? Just one line😀 Thanks!

Huang-Wei:
/approve

Thanks @yuanchen8911 .

PR #65, opened 2020-09-27
Speed Up Docker Image Build Times

seanmalloy:
Prior to this change the docker image builds would always compile both
binaries(scheduler and controller). Now only build the a controller
binary for the controller docker image, and only build the scheduler
binary for the scheduler docker image.


denkensk:
Thanks so much. 
/lgtm

PR #66, opened 2020-09-29
Update "make clean" To Only Delete Specific Files

seanmalloy:
This is an attempt to fix the broken automated container image builds
for k8s.gcr.io. The below error is currently being reported in the build
logs.

fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
docker build -t gcr.io/k8s-staging-scheduler-plugins/kube-scheduler: .
invalid argument "gcr.io/k8s-staging-scheduler-plugins/kube-scheduler:"
for "-t, --tag" flag: invalid reference format
See 'docker build --help'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,4 +83,5 @@ verify-gofmt:
 
 .PHONY: clean
 clean:
-	rm -rf ./bin
+	rm -f ./bin/controller
+	rm -f ./bin/kube-scheduler


seanmalloy:
This is an alternative to #62. I'm not sure if this will fix the broken automated image builds or not. The theory is that by somehow deleting the entire `./bin` directory in the automated build process is causing problems.

This was the only other difference I could find between the descheduler and schedler-plugins build automation. The descheduler does not run an `rm` commands as part of the image build automation.

Huang-Wei:
Let's give it a try.

/lgtm
/approve

PR #73, opened 2020-10-12
Capacity Scheduling Plugin

denkensk:


pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+	prefilterStatue := computePodResourceRequest(pod)
+	snapshotState := c.createSnapshot()
+	elasticQuotaInfo := snapshotState.elasticQuotas[pod.Namespace]
+
+	if !elasticQuotaInfo.lessThanMax(prefilterStatue.Resource) {
+		return framework.NewStatus(framework.Error, "MoreThanMax")
+	}
+	state.Write(preFilterStateKey, prefilterStatue)
+	state.Write(SnapshotStateKey, snapshotState)
+	return framework.NewStatus(framework.Success, "")
```
Huang-Wei, line 203: Where is the logic of verifying the pods **needs** a preemption? I assume it should be in Filter():

- including the incoming pod's req, current EQ's `used` <= `min`, then for sure return `Success`.
- including the incoming pod's req, `min` < current EQ's `used` <= `max` and sum(eq's `used`) > sum(eq's `min`), so we should return `Unschedulable`

Huang-Wei (commented):
Thanks @denkensk !

Haven't reviewed preemption logic and tests. Have some questions, would be good to get them clear and continue reviewing.

**Question:**

Should we disallow the Pod in Filter() when sum(elastic quota's used) > sum(elastic quota's min)? Or, we just let other Filter plugins to decide whether the pod is schedulable? In the former case, sum(elastic quota's min) is the limit for all namespaces pods that have associated elastic quota. While in the latter case, the limit is the cluster's capacity.

**Suggestions:**

1. A standalone controller aims to update ElasticQuota's `status.used` properly. Otherwise, operator cannot easily check the usage of a ElasticQuota.
1. The logic of maintaining the ElasticQuotaInfo cache should be robust:
    - if in-cache ElasticQuotaInfo is not present, fall back to obtain it from ElasticQuotaLister (if exists)
    - act on events of ElasticQuota properly (don't need to update `status.used`). Just update the ElasticQuotaInfo's min and max, and delete the entry properly.
    - act on events of assignedPods, to maintain an in-cache version of "used" capacity. (Do not use ElasticQuota's `status.used` as controller binary may not be up)
    - don't let a assignedPod event slip easily - e.g., upon a scheduler's restart, it's possible that the pod is associated with a ElasticQuota, it's just the ElasticQuota event hasn't arrived yet.
    - keep the case that the scheduler may lose network connection for a period (not crash), and then resume. So think about if the cache is synced properly.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+)
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (c *CapacityScheduling) Name() string {
+	return Name
+}
+
+type PluginConfig struct {
+	kubeconfig string
+}
```
yuanchen8911, line 103: It's already defined in apis/config/types.go. This one should be deleted.

pkg/capacityscheduling/README.md
```diff
@@ -0,0 +1,209 @@
...
+        - name: "*"
+    reserve:
+      enabled:
+        - name: CapacityScheduling
+  pluginConfig:
+    - name: CapacityScheduling
+      # optional plugin configs
+      # default path is /etc/kubernetes/scheduler.conf
+      args:
+        kubeConfigPath: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
```
yuanchen8911, line 44: Is it redundant with 
```
 clientConnection:
   kubeconfig: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
```
Can we get kubeconig path from `clientConnection`?

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+
+	// Check if there is elastic quota in the preemptor's namespace.
+	if preemptorWithElasticQuota {
+		moreThanMinWithPreemptor = preemptorElasticQuotaInfo.moreThanMinWithPod(preFilterState.Resource)
+	}
+
+	for _, p := range nodeInfo.Pods {
+		pElasticQuotaInfo, pWithElasticQuota := snapshotState.elasticQuotas[p.Pod.Namespace]
+		// Preemption only happens between in the same type of namespace: with or without.
+		if preemptorWithElasticQuota && pWithElasticQuota {
```
gongx, line 441: We are using "&&" ("and" operator here). 

- So, line 443 "if preemptorWithElasticQuota" is always true?

- line 471 will never be accessed?

Given the condition: "Preemption only happens between in the same type of namespace: with or without.", we only execute preemption when both of preemptorWithElasticQuota and pWithElasticQuota are true, or both of them are false, so line 441 condition should be XOR instead of AND?

pkg/capacityscheduling/capacity_scheduling.go
gongx: bq. including the incoming pod's req, min < current EQ's used <= max and sum(eq's used) > sum(eq's min), so we should return Unschedulable.

I do not understand why we should return Unschedulable here. As long as the current EQ's total used is smaller than the max, 

- If the cluster still has the available resource, it could be assigned

- Otherwise, based on the priority, the preemption could happen in the same namespace(Line 444)

pkg/capacityscheduling/README.md
```diff
@@ -0,0 +1,209 @@
...
+leaderElection:
+  leaderElect: false
+clientConnection:
+  kubeconfig: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
+profiles:
+- schedulerName: default-scheduler
+  plugins:
+    preFilter:
+      enabled:
+        - name: CapacityScheduling
```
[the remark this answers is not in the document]
denkensk: I agree with yuan's suggestion. We really need to have guidance on how to name the plugins.
But CapacityScheduling is proper nouns in yarn. It's easier for users to understand what we're going to do.

pkg/capacityscheduling/capacity_scheduling.go
yuanchen8911: @denkensk?

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+	Namespace   string
+	Min         framework.Resource
+	Max         framework.Resource
+	Used        framework.Resource
+	UsedOverMin bool
+}
+
+func (e *ElasticQuotaInfo) reserveResource(request framework.Resource) {
+	e.Used.Memory += request.Memory
+	e.Used.MilliCPU += request.MilliCPU
```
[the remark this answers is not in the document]
denkensk: In this version, I think it's better not add storage or huge-file?  These may not be easy to describe in the Cluster Quota.
WDYT?

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+		klog.Errorf("failed to check whether pod %s/%s is assumed: %v", pod.Namespace, pod.Name, err)
+		return
+	}
+	if isAssumed {
+		klog.V(4).Infof("pod %s/%s is assumed", pod.Namespace, pod.Name)
+		c.deleteAssumedPod(pod)
+		return
+	}
+	elasticQuotaInfo, exit := c.elasticQuotas[pod.Namespace]
+	if exit {
```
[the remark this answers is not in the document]
denkensk: https://github.com/kubernetes-sigs/scheduler-plugins/pull/73/files#diff-d2629e9bc4a82622042892618ce271fef08fca9723f7d2993f16e9099db78134R162-R164

The plugins will sync the elastic event firstly. I think this will solve the problem of sequence of events like the scheduler restart?

pkg/capacityscheduling/capacity_scheduling.go
denkensk: Thanks for your catch. Yes, it should be XOR. I will update it.

pkg/capacityscheduling/README.md
```diff
@@ -0,0 +1,209 @@
...
+We assume two elastic quotas are defined: QuotaA (min:4, max:6) and QuotaB 
+(min:4, max:6), the quota unit is the number of CPUs. UserA in green color 
+consumes QuotaA and UserB in red color consumes QuotaB. The entire cluster 
+has 8 CPUs available hence the sum of quota min is equal to the cluster capacity.
+
+- Create namespace and elasticQuota
+
+```script
+cat <<EOF > eq1.yaml
+apiVersion: capacityscheduling.sigs.k8s.io/v1alpha1
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+			return nil, fmt.Errorf("want args to be of type CapacitySchedulingArgs, got %T", obj)
+		}
+		kubeConfigPath = args.KubeConfigPath
+	}
+
+	c := &CapacityScheduling{
+		frameworkHandle: handle,
+		elasticQuotas:   make(map[string]*ElasticQuotaInfo),
+		pdbLister:       getPDBLister(handle.SharedInformerFactory()),
+		lock:            &sync.RWMutex{},
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+		assumedPods:     make(map[string]bool),
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,869 @@
...
+	if err != nil {
+		klog.Errorf("failed to check whether pod %s/%s is assumed: %v", pod.Namespace, pod.Name, err)
+		return
+	}
+	if isAssumed {
+		klog.V(4).Infof("pod %s/%s is assumed", pod.Namespace, pod.Name)
+		c.deleteAssumedPod(pod)
+		return
+	}
+	elasticQuotaInfo, exit := c.elasticQuotas[pod.Namespace]
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
denkensk: Done

pkg/capacityscheduling/README.md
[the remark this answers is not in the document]
denkensk: Yes, it will redundant with clientConnection and other plugins which need kubeconfig.
I think we can expose this parameter in upstream for plugins to use. Our internal implementation does this.

pkg/capacityscheduling/candidate.go
```diff
@@ -0,0 +1,34 @@
...
+
+import extenderv1 "k8s.io/kube-scheduler/extender/v1"
+
+type candidate struct {
+	victims *extenderv1.Victims
+	name    string
+}
+
+// Victims returns s.victims.
+func (s *candidate) Victims() *extenderv1.Victims {
```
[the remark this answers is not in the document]
denkensk: Done

denkensk:
> Should we disallow the Pod in PreFilter (or Filter()) when sum(elastic quota's used) > sum(elastic quota's min)? Or, we just let other Filter plugins to decide whether the pod is schedulable? In the former case, sum(elastic quota's min) is the limit for all namespaces pods that have associated elastic quota. While in the latter case, the limit is the cluster's capacity.

> Where is the logic of verifying the pods needs a preemption? I assume it should be in Filter():
> - including the incoming pod's req, current EQ's used <= min, then for sure return Success.
> - including the incoming pod's req, min < current EQ's used <= max and sum(eq's used) > sum(eq's min), so we should return Unschedulable

----
I think we maybe not need to add an extra prefilter/filter plugins to check if `sum(elastic quota's used) > sum(elastic quota's min)`. Because even if the `sum(elastic quota's used) > sum(elastic quota's min)`, the pods still can be scheduled if the used < max and the cluster still has the available resource as the comments of @gongx https://github.com/kubernetes-sigs/scheduler-plugins/pull/73#discussion_r507012902. 

> In the former case, sum(elastic quota's min) is the limit for all namespaces pods that have associated elastic quota. 

I understand that you worried about the other namespaces which haven't associated elastic quota. Ideally, users are expected to be able to create elastic quota for every namespaces. I'm open to more discussion here.

denkensk:
> Should we disallow the Pod in PreFilter (or Filter()) when sum(elastic quota's used) > sum(elastic quota's min)? Or, we just let other Filter plugins to decide whether the pod is schedulable? In the former case, sum(elastic quota's min) is the limit for all namespaces pods that have associated elastic quota. While in the latter case, the limit is the cluster's capacity.

> Where is the logic of verifying the pods needs a preemption? I assume it should be in Filter():
> - including the incoming pod's req, current EQ's used <= min, then for sure return Success.
> - including the incoming pod's req, min < current EQ's used <= max and sum(eq's used) > sum(eq's min), so we should return Unschedulable

----
I think we maybe not need to add an extra prefilter/filter plugins to check if `sum(elastic quota's used) > sum(elastic quota's min)`. Because even if the `sum(elastic quota's used) > sum(elastic quota's min)`, the pods still can be scheduled if the used < max and the cluster still has the available resource as the comments of @gongx https://github.com/kubernetes-sigs/scheduler-plugins/pull/73#discussion_r507012902. 

> In the former case, sum(elastic quota's min) is the limit for all namespaces pods that have associated elastic quota. 

I understand that you worried about the other namespaces which haven't associated elastic quota. Ideally, users are expected to be able to create elastic quota for every namespaces. I'm open to more discussion here.
@Huang-Wei

denkensk:
> A standalone controller aims to update ElasticQuota's status.used properly. Otherwise, operator cannot easily check the usage of a ElasticQuota.

Thanks I will submit a separate PR soon to sync the ElasticQuota's status.

pkg/capacityscheduling/README.md
denkensk: Yes, it will redundant with clientConnection and other plugins which need kubeconfig.
I think we can expose this parameter in upstream for plugins to use.

pkg/capacityscheduling/README.md
```diff
@@ -0,0 +1,209 @@
...
+nginx-2mxdd   1/1     Running   0          6m49s
+nginx-6gbgx   1/1     Running   0          6m49s
+nginx-6kj6v   0/1     Pending   0          7s
+nginx-bxvcg   0/1     Pending   0          6m49s
+nginx-kcrtb   0/1     Pending   0          6m49s
+nginx-ptxp2   1/1     Running   0          6m49s
+nginx-ajc69   0/1     Running   0          7s
+nginx-xmzkn   1/1     Running   0          6m49s
+
+$ kubectl get pods -n quota2
```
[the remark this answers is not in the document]
denkensk: Done I change it to provide a simpler example without GPUs.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,882 @@
...
+					}
+				} else {
+					// If Preemptor.Request + Quota.allocated <= Quota.min: It
+					// means that its min or guaranteed resource is used or
+					// `borrowed` by other Quota. Potential victims in a node
+					// will be chosen from Quotas that allocates more resources
+					// than its min, i.e., borrowing resources from other
+					// Quotas.
+					if pElasticQuotaInfo.overUsed() &&
+						!pElasticQuotaInfo.lessThanMinWithoutPod(computePodResourceRequest(pod).Resource) {
```
gongx, line 458: I have a little concern about this condition. Assume, in namespace A, we have two pods. The total used resources are larger than the Quota.min. But if we remove any of the pods, the total used resources would be smaller than the Quota.min. Then two pods would not be a preemption candidate. So, by using such a condition, it is possible that we can not find any candidate. In that case, we will break the resource guarantee.
I think that we can probably still add this kind of pod into the candidate list, but it would have less probability to be preempted. Thoughts ?

denkensk: Introduce the background, I add the check for `lessThanMinWithoutPod` for this condition:
1. we have two quotas: quota1(cpu min 3) quota2(cpu min 3) capacity is 6
2. we submit two pods(pod1A/pod1B -- cpu request 2 for every pod) for quota1, the two pods will be scheduled successfully. At this time, (quota1 used is 4, min is 3)
3. we submit two pods(pod2A/pod2B -- cpu request 2  for every pod) for quota2, pod2A is scheduled successfully. At this time,  (quota1 used is 4, min is 3) and (quota1 used is 2, min is 3). In this case, if we choose to add pod1A or pod1B as a preemption candidate, the pod1A is preempted. At this time,  (quota1 used is 2, min is 3) and (quota1 used is 4, min is 3).
4. Then pod1B is recreated by controller, then pod1B will preempt pod2A/pod2B. In this way, preempt will continue to appear between quota1 and quota2.

For this case, I won't choose the pod as  a preemption candidate if the total used resources would be smaller than the Quota.min after the pod is removed.

WDYT？Looking forward to your reply

pkg/capacityscheduling/README.md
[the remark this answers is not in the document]
Huang-Wei: Agree. We may refactor the naming at some time, and put plugins with similar semantics under the package.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,882 @@
...
+						potentialVictims = append(potentialVictims, p.Pod)
+						if err := removePod(p.Pod); err != nil {
+							return nil, 0, false
+						}
+					}
+				}
+			} else {
+				if podutil.GetPodPriority(p.Pod) < podPriority {
```
Huang-Wei, line 466: Does this mean you'd mixin the defaultpreemption behavior? So that you'd like to disable defaultpreemption upon configuration?

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,882 @@
...
+
+	podPriority := podutil.GetPodPriority(pod)
+	preemptorElasticQuotaInfo, preemptorWithElasticQuota := snapshotState.elasticQuotas[pod.Namespace]
+
+	// Check if there is elastic quota in the preemptor's namespace.
+	if preemptorWithElasticQuota {
+		moreThanMinWithPreemptor = preemptorElasticQuotaInfo.moreThanMinWithPod(preFilterState.Resource)
+	}
+
+	for _, p := range nodeInfo.Pods {
```
Huang-Wei, line 432: I actually expected a more efficient algorithm to loop over pods by elastic quota's namespace.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,882 @@
...
+	pod *v1.Pod,
+	nodeInfo *framework.NodeInfo,
+	pdbs []*policy.PodDisruptionBudget,
+	snapshotState *SnapshotState,
+	preFilterState *PreFilterState,
+) ([]*v1.Pod, int, bool) {
+	var potentialVictims []*v1.Pod
+	var moreThanMinWithPreemptor bool
+
+	removePod := func(rp *v1.Pod) error {
```
Huang-Wei, line 405: As `PreFilterExtensions` is not implemented, removePod and addPod won't change the PreFilterState in terms of ElasticQuota. So does that mean upon a PreFilter error raised by this plugin, the dry run would always fail?

Huang-Wei (commented):
Haven't reviewed UT yet.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,882 @@
...
+			elasticQuota := elasticQuotas[0]
+			elasticQuotaInfo = &ElasticQuotaInfo{
+				Namespace: pod.Namespace,
+				Min:       framework.Resource{},
+				Max:       framework.Resource{},
+				Used:      framework.Resource{},
+			}
+			elasticQuotaInfo.Max.Add(elasticQuota.Spec.Max)
+			elasticQuotaInfo.Min.Add(elasticQuota.Spec.Min)
+			c.elasticQuotas[elasticQuota.Namespace] = elasticQuotaInfo
```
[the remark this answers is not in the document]
denkensk: Here is
1. We will check the cache(elasticQuotas) if it's eq exist. 
```
elasticQuotaInfo, ok := c.elasticQuotas[pod.Namespace]
```
2. If not, we will take the initiative to query the eq by namespace.
```go
elasticQuotas, err := c.elasticQuotaLister.ElasticQuotas(pod.Namespace).List(labels.NewSelector())
```
3. We may get the list of elasticQuotas. In this version, we assume that users can only create one eq in a namespace. So we will check the length of elasticQuotas and just take elasticQuotas[0]
```
if len(elasticQuotas) > 0 {
			elasticQuota := elasticQuotas[0]
```

4. If the length is more than 0. we will create structure elasticQuotaInfo and init the data by elasticQuotas[0].
```go
elasticQuota := elasticQuotas[0]
			elasticQuotaInfo = &ElasticQuotaInfo{
				Namespace: pod.Namespace,
				Min:       framework.Resource{},
				Max:       framework.Resource{},
				Used:      framework.Resource{},
			}
			elasticQuotaInfo.Max.Add(elasticQuota.Spec.Max)
			elasticQuotaInfo.Min.Add(elasticQuota.Spec.Min)
			c.elasticQuotas[elasticQuota.Namespace] = elasticQuotaInfo
```

pkg/capacityscheduling/capacity_scheduling.go
denkensk: Yes. I disable defaultpreemption upon configruration.
```go
    postFilter:
      enabled:
      - name: CapacityScheduling
      disabled:
      - name: "*"
```

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,882 @@
...
+						}
+					}
+				} else {
+					// If Preemptor.Request + Quota.allocated <= Quota.min: It
+					// means that its min or guaranteed resource is used or
+					// `borrowed` by other Quota. Potential victims in a node
+					// will be chosen from Quotas that allocates more resources
+					// than its min, i.e., borrowing resources from other
+					// Quotas.
+					if pElasticQuotaInfo.overUsed() &&
```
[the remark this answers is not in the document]
denkensk: moreThanMinWithPreemptor is for the preemptor.
pElasticQuotaInfo.overUsed() is for the other pods on nodes.

This means that if Preemptor.Request + PreemptorQuota.allocated <= PreemptorQuota.min and OtherPodQuota.used > OtherPodQuota.min.

so other pod will be the Potential victims.

pkg/capacityscheduling/capacity_scheduling.go
denkensk: Ditto  
pElasticQuotaInfo is for other pods.
preemptorElasticQuotaInfo is for preemptor.

gongx: @denkensk 
Thank you for the explanation. And yes, in the case you mentioned, the scheduler does not need to do preemption. But let us discuss the following cases. Assume, we have two quotas: quota1(CPU min 3) quota2(CPU min 3) capacity is 6.

Case1: quota1 used is 4, quota2 used is 2. When quota2 requests 2 CPU. The scheduler does not do preemption even quota1 used more resources. Because quota2 asked for more resources than the minimum resource. Also, we need to output the status/logs clearly so that the users can understand why this 2 CPU request can not be satisfied.

Case2: quota1 used is 4 (two pods, 2cpu for every pod), quota2 used is 2. When quota2 requests 1 CPU, the scheduler needs to do the preemption and delete one pod from quota1. Because we must guarantee the min resource for quota2. After that, the controller would re-request 2 CPUs for quota1, but the scheduler can not schedule it and does not do the preemption.

I understand that the preemption strategy does not sound very user-friendly. But I think that the CS should provide:
minimum resource == guaranteed resource which should/has to be satisfied at any time/ anywhere.  
max resource == best-effort resource which can not be guaranteed.

Also, we need to document the min/max concepts more clearly.
What do you think?

denkensk: Thanks for your reply.

> Case1: quota1 used is 4, quota2 used is 2. When quota2 requests 2 CPU. The scheduler does not do preemption even quota1 used more resources. Because quota2 asked for more resources than the minimum resource. Also, we need to output the status/logs clearly so that the users can understand why this 2 CPU request can not be satisfied.

Case2: quota1 used is 4 (two pods, 2cpu for every pod), quota2 used is 2. When quota2 requests 1 CPU, the scheduler needs to do the preemption and delete one pod from quota1. Because we must guarantee the min resource for quota2. After that, the controller would re-request 2 CPUs for quota1, but the scheduler can not schedule it and does not do the preemption.

Thank you for your detailed explanation. You are right. I delete the check for if we remove any of the pods, the total used resources would be smaller than the Quota.min. And also add them into the candidate list.

> Also, we need to document the min/max concepts more clearly.

Done, I add the  description about min/max concepts in README.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,816 @@
...
+	// Check if there is elastic quota in the preemptor's namespace.
+	if preemptorWithElasticQuota {
+		moreThanMinWithPreemptor = preemptorElasticQuotaInfo.overUsed(preFilterState.Resource, preemptorElasticQuotaInfo.Min, resourceAdd)
+	}
+
+	// sort the pods in node by the priority class
+	sort.Slice(nodeInfo.Pods, func(i, j int) bool { return !util.MoreImportantPod(nodeInfo.Pods[i].Pod, nodeInfo.Pods[j].Pod) })
+	for _, p := range nodeInfo.Pods {
+		pElasticQuotaInfo, pWithElasticQuota := elasticQuotaSnapshotState.elasticQuotaInfos[p.Pod.Namespace]
+		// Preemption only happens between in the same type of namespace: with or without.
```
Huang-Wei, line 478: What if the pod gets unschedulabe is just due to regular constraints, and preemptorWithElasticQuota happens to not equal to pWithElasticQuota? I recalled that you expect this plugin to replace (instead of co-exist with) defaultpreemption plugin.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,816 @@
...
+		if preemptorWithElasticQuota == pWithElasticQuota {
+			// If there is elastic quota in the preemptor's namespace.
+			if preemptorWithElasticQuota {
+				if moreThanMinWithPreemptor {
```
Huang-Wei, line 482: The logic is able to come down here is due to one of the following reasons:

1. the pod requests too much RQ, which is larger than the PQ's `max`. In this case, the pod should try to preempt pods in the same namespace
1. the pod's request results in global RQ usage greater than the global RQ's `min`. In this case, any pod subjects to RQ may be preemptable
1. (1) and (2) may happen at the same time

So we should firstly identify the exact reason the logic reaches here. The current if...else block seems not to have covered all 3 cases above.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,816 @@
...
+
+	c.Lock()
+	defer c.Unlock()
+	c.elasticQuotaInfos[eq.Namespace] = elasticQuotaInfo
+}
+
+func (c *CapacityScheduling) updateElasticQuota(oldObj, newObj interface{}) {
+	oldEQ := oldObj.(*v1alpha1.ElasticQuota)
+	newEQ := newObj.(*v1alpha1.ElasticQuota)
+	newElasticQuotaInfo := newElasticQuotaInfo(newEQ.Namespace, newEQ.Spec.Min, newEQ.Spec.Max, nil)
```
Huang-Wei, line 594: nit: `newElasticQuotaInfo` is a function name, prefer to use `newEQInfo`

pkg/capacityscheduling/capacity_scheduling.go
denkensk: If the pods get unschedulabe because of regular constraints，it will preempt other pods which belongs to ElasticQuota which eq.used > eq.min.

pkg/capacityscheduling/capacity_scheduling.go
denkensk: case 1 is right. 
case2: Not any pod subjects to RQ. I think it is just the pods which belong to ElasticQuota which eq.used > eq.min.  The pods belong to ElasticQuota which eq.used < eq.min is guaranteed.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,816 @@
...
+	preFilterState := computePodResourceRequest(pod)
+
+	state.Write(preFilterStateKey, preFilterState)
+	state.Write(ElasticQuotaSnapshotKey, snapshotElasticQuota)
+
+	elasticQuotaInfos := snapshotElasticQuota.elasticQuotaInfos
+	eq := snapshotElasticQuota.elasticQuotaInfos[pod.Namespace]
+	if eq == nil {
+		return framework.NewStatus(framework.Success, "skipCapacityScheduling")
+	}
```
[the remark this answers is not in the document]
denkensk: > state.Write(preFilterStateKey, preFilterState)
	state.Write(ElasticQuotaSnapshotKey, snapshotElasticQuota)
---
These values will be useful later.

pkg/capacityscheduling/capacity_scheduling.go
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,816 @@
...
+
+	if eq.overUsed(preFilterState.Resource, eq.Max, resourceAdd) {
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, fmt.Sprintf("Pod %v/%v is rejected in Prefilter because ElasticQuota %v is more than Max", pod.Namespace, pod.Name, eq.Namespace))
```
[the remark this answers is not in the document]
denkensk: Yes you are right.

pkg/capacityscheduling/elasticquota.go
```diff
@@ -0,0 +1,201 @@
...
+
+func (e ElasticQuotaInfos) totalUsedMoreThanTotalMin() bool {
+	used := framework.NewResource(nil)
+	min := framework.NewResource(nil)
+
+	for _, elasticQuotaInfo := range e {
+		used.Add(elasticQuotaInfo.Used.ResourceList())
+		min.Add(elasticQuotaInfo.Min.ResourceList())
+	}
+	if moreThanMin(*used, *min) {
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,816 @@
...
+	}
+
+	if elasticQuotaInfos.totalUsedMoreThanTotalMin() {
+		return framework.NewStatus(framework.UnschedulableAndUnresolvable, fmt.Sprintf("Pod %v/%v is rejected in Prefilter because total ElasticQuota used is more than min", pod.Namespace, pod.Name))
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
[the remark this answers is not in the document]
Huang-Wei: `eq == nil` indicates the pod isn't subject to EQ, so isn't it unnecessary to calculate the pre-states?

pkg/capacityscheduling/elasticquota.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, elasticQuotaInfo := range e {
+		used.Add(elasticQuotaInfo.Used.ResourceList())
+		min.Add(elasticQuotaInfo.Min.ResourceList())
+	}
+
+	used.Add(podRequest.ResourceList())
+	if moreThanMin(*used, *min) {
+		return true
+	}
+	return false
```
Huang-Wei, line 52: L49~L52 can be simplified to `return moreThanMin()`.

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: It's still used in elasticquota.go#clone().

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: > it will preempt other pods which belongs to ElasticQuota which eq.used > eq.min.

What if preempting RQ's managed pods isn't helpful - for example, a pod subject to RQ, but failed due to some regular constraints like PodAffinity. We may need to increase the searching scope of regular pods as we don't know the exact reason why it's judged as "unschedulable".

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: > case2: Not any pod subjects to RQ. I think it is just the pods which belong to ElasticQuota which eq.used > eq.min. The pods belong to ElasticQuota which eq.used < eq.min is guaranteed.

Sounds reasonable.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,841 @@
...
+		}
+		return nil
+	}
+
+	podPriority := podutil.GetPodPriority(pod)
+	preemptorElasticQuotaInfo, preemptorWithElasticQuota := elasticQuotaSnapshotState.elasticQuotaInfos[pod.Namespace]
+
+	// Check if there is elastic quota in the preemptor's namespace.
+	if preemptorWithElasticQuota {
+		moreThanMinWithPreemptor = preemptorElasticQuotaInfo.overUsed(preFilterState.Resource, preemptorElasticQuotaInfo.Min, resourceAdd)
```
Huang-Wei, line 471: I think it's `moreThanMaxWithPreemptor`, not more than min.

- If it's more than max, for sure we should preempt pods belonging to the same EQ.
    - If preempting pods from the same EQ works, we're done.
    - If not, it indicates we may need to **additionally** preempt pods belongs to other EQs that have over-used their min.
- If it's not, then pods that belong to an EQ whose usage > min are potential victims.

WDYT?

gongx: I think that "moreThanMinWithPreemptor" makes more sense. 

- If it is more than min, we have to preempt pods belonging to the same EQ. We can not preempt pods belongs to other EQs. Because those extra resources belong to best-effort resources which can not be guaranteed. So, we can not preempt pods from other EQs

-  If it is less than min, it means some other EQs use more resources than its guaranteed resources which they need to return back.

Using "moreThanMaxWithPreemptor" might cause more issues, for example, if one user asks for some resources which cause totalResource > min, but < max. At this time, I think that we can only preempt the pods from the same EQ. We should not preempt pods from other EQ even if they use more resources than min.

Huang-Wei: > * If it is more than min, we have to preempt pods belonging to the same EQ. We can not preempt pods belongs to other EQs. Because those extra resources belong to best-effort resources which can not be guaranteed. So, we can not preempt pods from other EQs

What if preempting pods within the EQ doesn't help? Is it legal to preempt over-used resources from other EQs if the current pod has a higher priority? If not, it seems low-priority pods in other EQs that comes first can reserve extra resource without worrying about being preempted. Which doesn't look correct to me.

So my point is that among EQs that over-used resources, Pod's priority is the key to determine whether it can preempt others.

gongx: Exactly. I totally agree with you. Here is what I mean: 

- request resource + allocated resource <= min, then preempt from other EQs

- request resource + allocated resource > max, then preempt from the same EQ

- min < request resource + allocated resource < max, preempt based on the priority. It can preempt pods from the same EQ or from the other EQs whose total resource > min.

denkensk: This is actually divided into two layers:  EQ and Priority Class.
I think it's multi tenant. The Priority Class in the same EQ or tenant is comparable. The priority Class in the different EQ or tenant isn't comparable.
The extra resource is first come, first served. If min < request resource + allocated resource < max, the preempt only allowed to preempt the pods from the same EQ.

denkensk: > I think it's multi tenant. The Priority Class in the same EQ or tenant is comparable. The priority Class in the different EQ or tenant isn't comparable.
A little more: 
Different users(tenants) have different ways to use priority Class. tenant A: high(100) low(10) tenant B: high(1000) low(100). So The priority Class in the different EQ or tenant isn't comparable.

pkg/capacityscheduling/capacity_scheduling.go
denkensk: If preemptorWithElasticQuota happens to not equal to pWithElasticQuota, it means one of preemptor or p doesn't belong to some EQ.  So the p can't be selected as victim. 
This is mainly to prevent mutual influence between the two sides. As I see, The Priority Class in the same EQ or tenant is comparable. The priority Class in the different EQ or tenant isn't comparable. This is the same here.

pkg/capacityscheduling/capacity_scheduling.go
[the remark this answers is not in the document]
denkensk:
```
    postFilter:
      enabled:
      - name: CapacityScheduling
      disabled:
      - name: "*"
```
Because we use CapacityScheduling replace the defaultpreemption. So although the pod isn't subject to EQ, it also can preempt other pods which also isn't subject to EQ. So pre-states also need to be used in PostFilter.

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: Note that k8s doesn't have an official API called tenant yet. So if the implementation is assuming tenant == namespace, it should be documented well, otherwise the behavior would really confuse users.

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: > The extra resource is first come, first served. If min < request resource + allocated resource < max, the preempt only allowed to preempt the pods from the same EQ.

I'm afraid this behavior is not mentioned in the KEP. I'm not sure it's the right way as it violates the API semantics of Priority Class. Could you submit a PR describing this behavior in the KEP's [story section](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/9-capacity-scheduling#story-1)? and I may bring it up to the community meeting (next Thursday) for discussion.

Huang-Wei: > Different users(tenants) have different ways to use priority Class. tenant A: high(100) low(10) tenant B: high(1000) low(100). So The priority Class in the different EQ or tenant isn't comparable.

This would surprise users IMHO. Is it possible to make it a plugin option? I'm fine with making it default.

denkensk: After offline discussion, we decided that the current version process is as follows to make sure that they are used in the same way as yarn users.
1. When quota.used + preemptor.request < min, victims will be selected from the EQ which used > min
2. When quota.used + preemptor.request > min, victims will be selected from the same EQ.
@Huang-Wei @gongx

Huang-Wei: @denkensk can you add these 2 cases into the KEP's story sections. Thanks!

BTW: a nit on (1), `<` should be `<=` :)

test/integration/capacity_scheduling_test.go
```diff
@@ -0,0 +1,518 @@
...
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "t1-p4", Namespace: "ns2"},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "t1-p5", Namespace: "ns2"},
+				},
+			},
+		},
+		{
+			name:       "intra namespace preempt with quota",
```
Huang-Wei, line 246:
```suggestion
			name:       "in-namespace preemption",
```

test/integration/capacity_scheduling_test.go
```diff
@@ -0,0 +1,518 @@
...
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "t2-p5", Namespace: "ns1"},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "t2-p6", Namespace: "ns2"},
+				},
+			},
+		},
+		{
+			name:       "intra namespace preempt without quota",
```
Huang-Wei, line 304:
```suggestion
			name:       "regular preemption without Elastic Quota",
```

test/integration/capacity_scheduling_test.go
```diff
@@ -0,0 +1,518 @@
...
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "t3-p5", Namespace: "ns1"},
+				},
+				{
+					ObjectMeta: metav1.ObjectMeta{Name: "t3-p6", Namespace: "ns2"},
+				},
+			},
+		},
+		{
+			name:       "inter namespace preempt without quota",
```
Huang-Wei, line 333: nit: this is quite similar to the previous one, so we can remove it.

test/integration/capacity_scheduling_test.go
```diff
@@ -0,0 +1,518 @@
...
+		autoCreate := false
+		t.Logf("namespaces %+v", ns)
+		_, err = cs.CoreV1().ServiceAccounts(ns).Create(ctx, &v1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{Name: "default", Namespace: ns}, AutomountServiceAccountToken: &autoCreate}, metav1.CreateOptions{})
+		if err != nil && !errors.IsAlreadyExists(err) {
+			t.Fatalf("Failed to create ns default: %v", err)
+		}
+	}
+
+	for _, tt := range []struct {
```
Huang-Wei, line 179: Can you add a couple of cases?

1. EQ1 overused, but preempting pods in EQ1 doesn't help. (this is possible is preemptor requests a huge amount of resources)
1. EQ1 not overused, but pods in other over-used EQ have a higher priority, and hence preemptor is not able to preempt.
1. Pods spread in multiple nodes (it's ok to showcase that the current implementation doesn't support cross-node preemption)

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -0,0 +1,841 @@
...
+			},
+		})
+
+	schedSharedInformerFactory.Start(nil)
+	if !cache.WaitForCacheSync(nil, elasticQuotaInformer.HasSynced) {
+		return nil, fmt.Errorf("timed out waiting for caches to sync %v", Name)
+	}
+
+	podInformer := handle.SharedInformerFactory().Core().V1().Pods().Informer()
+	podInformer.AddEventHandler(
```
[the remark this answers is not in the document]
denkensk: Yes. We will add the corresponding controller logic for the `status` field of ElasticQuota. This will be used by other operator or users. But in order to prevent data synchronization delay, I think it's better to maintain the status in the scheduler itself also.

test/util/utils.go
```diff
@@ -131,3 +136,43 @@ func BuildKubeConfigFile(config *restclient.Config) string {
 	}
 	return ""
 }
+
+func MakePods(podName string, namespace string, memReq int64, cpuReq int64, priority int32, uid string, nodeName string) *corev1.Pod {
```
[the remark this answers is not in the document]
denkensk: This is mainly to make the code concise. 
for example:
```go
st.MakePod().Namespace("ns1").Name("pod1").Container(pause).Priority(priority).Node("node1").UID("pod1").ZeroTerminationGracePeriod().Req({corev1.ResourceRequirements{
		Requests: corev1.ResourceList{
			corev1.ResourceMemory: *resource.NewQuantity(100, resource.DecimalSI),
			corev1.ResourceCPU:    *resource.NewMilliQuantity(100, resource.DecimalSI),
		},
	}).Obj()
```
We can use as 
```
util.MakePods("t1-p1", "ns1", 50, 10, midPriority, "t1-p1", "")
```
If there are a lot of pods, it will be more obvious

test/integration/capacity_scheduling_test.go
denkensk: Done

pkg/capacityscheduling/capacity_scheduling_test.go
```diff
@@ -0,0 +1,326 @@
...
+)
+
+func TestPreFilter(t *testing.T) {
+	type podInfo struct {
+		podName      string
+		podNamespace string
+		memReq       int64
+	}
+
+	tests := []struct {
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
denkensk: Sure I will update it into KEP in another pr,

pkg/capacityscheduling/capacity_scheduling_test.go
```diff
@@ -112,6 +85,41 @@ func TestPreFilter(t *testing.T) {
...
 			},
 		},
+		{
+			name: "the sum of used is bigger than the sum of min",
+			podInfos: []podInfo{
+				{podName: "ns2-p1", podNamespace: "ns2", memReq: 500},
+			},
+			elasticQuotas: map[string]*ElasticQuotaInfo{
+				"ns1": {
+					Namespace: "ns2",
```
Huang-Wei, line 95: s/ns2/ns1

pkg/capacityscheduling/capacity_scheduling_test.go
```diff
@@ -112,6 +85,41 @@ func TestPreFilter(t *testing.T) {
...
+				"ns2": {
+					Namespace: "ns2",
+					Min: &framework.Resource{
+						Memory: 1000,
+					},
+					Max: &framework.Resource{
+						Memory: 2000,
+					},
+					Used: &framework.Resource{
+						Memory: 300,
```
Huang-Wei, line 115: Although it's a UT, making it `200` is more reasonable (currently summed Used is 300 + 1800 > summed Min)

test/integration/capacity_scheduling_test.go
```diff
@@ -330,31 +329,175 @@ func TestCapacityScheduling(t *testing.T) {
 			},
 		},
 		{
-			name:       "inter namespace preempt without quota",
+			name:       "in-namespace preemption failed because ",
```
Huang-Wei, line 332: because ...

Huang-Wei (commented):
Some final nits. LGTM otherwise.

The integration tests are pretty neat. Thanks @denkensk !

manifests/capacityscheduling/crd.yaml
```diff
@@ -27,10 +27,28 @@ spec:
               properties:
                 min:
                   type: object
+                  additionalProperties:
+                    anyOf:
+                      - type: integer
+                      - type: string
+                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
```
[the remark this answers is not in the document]
denkensk: Just reuse internal pattern in company ^_^

Huang-Wei:
/lgtm
/approve

Thanks @denkensk! It's another milestone achieved using scheduler framework to resolve a real user pain point.

FYI: @ahg-g @alculquicondor @damemi .

PR #57, opened 2020-09-22
Implementation of coscheduling base on crd[Part 2]: coscheduling refactor-must have

cwdsuzhou:
/cc @Huang-Wei

PR #75, opened 2020-10-12
Optimize Permit and Unreserve plugins in coscheduling

yuanchen8911:
Instead of iterating all the waiting pods in` Permit` and `Unreserve` plugins, 

-  Add a new field `waitingPods` for a PodGroup to `PodGroupInfo`.
- `Permit`:  **Allow** all the waiting pods for a PodGroup if the minimum Pod number is met. Otherwise, add the Pod to the PodGroup's `waitingPods`. 
- `Unreserve`: **Reject** all pods in `waitingPod` for the PodGroup.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -4,4 +4,7 @@ approvers:
   - ahg-g
   - Huang-Wei
 reviewers:
+  - cwdsuzhou
   - denkensk
+  - seanmalloy
+  - yuanchen8911


Huang-Wei:
@yuanchen8911 we may revisit this PR after #57 gets merged first.

/hold

yuanchen8911:
> @yuanchen8911 we may revisit this PR after #57 gets merged first.
> 
> /hold

OK.  waitingPod is implemented via channel in K8s (`scheduler/framework/runtime`). More importantly, we should probably simplify the the CRD-based coscheduling implementation? Please refer to the offline discussions today. @Huang-Wei  (/cc @cwdsuzhou)

Huang-Wei:
> do you think it's still useful to continue to improve the existing PodGroup label coscheduling as long as the features/enhancement are relevant?

I'd prefer to focus on merging the implementation for now. Then add enhancements based on the new implementation. This would save us efforts.

PR #79, opened 2020-10-18
Fix kubeconfig Options in cmd/controller 

lianghao208:
KubeConfig cant be used since the parse of the KubeConfig options are set to MasterUrl, Its supposed to be KubeConfig file path.


denkensk:
Thanks 
/lgtm

Huang-Wei:
Thanks for the catch!

/approve

PR #88, opened 2020-10-30
Prescheduling logic to controller

cwdsuzhou:
1. Add prescheduling status
2. Add filling occupied by filed


cwdsuzhou:
ping @Huang-Wei  for review

pkg/controller/podgroup.go
```diff
@@ -266,3 +276,17 @@ func (ctrl *PodGroupController) patchPodGroup(old, new *schedv1alpha1.PodGroup)
 	}
 	return nil
 }
+
+func fillOccupiedObj(pg *schedv1alpha1.PodGroup, pod *v1.Pod) {
+	refs := make([]string, 0)
```
Huang-Wei, line 281: nit: `var refs []string`

pkg/controller/podgroup.go
```diff
@@ -266,3 +276,17 @@ func (ctrl *PodGroupController) patchPodGroup(old, new *schedv1alpha1.PodGroup)
...
+	for _, ownerRef := range pod.OwnerReferences {
+		refs = append(refs, fmt.Sprintf(pod.Namespace, ownerRef.Name))
```
Huang-Wei, line 283: "fmt.Sprintf(pod.Namespace, ownerRef.Name)" looks weird, I guess you want to use a `/` separator?

pkg/controller/podgroup.go
```diff
@@ -266,3 +276,17 @@ func (ctrl *PodGroupController) patchPodGroup(old, new *schedv1alpha1.PodGroup)
...
+	}
+	if len(pg.Status.OccupiedBy) == 0 {
+		return
+	}
+	if len(refs) != 0 {
+		sortkeys.Strings(refs)
```
Huang-Wei, line 289: Is it equivalent to `sort.Strings(refs)`? If yes, let's eliminate the dependency on `sortkeys`.

pkg/controller/podgroup.go
cwdsuzhou: Yes, sorry for mistake

pkg/controller/podgroup.go
cwdsuzhou: removed

cwdsuzhou:
@Huang-Wei  PR addressed

pkg/controller/podgroup.go
```diff
@@ -266,3 +275,16 @@ func (ctrl *PodGroupController) patchPodGroup(old, new *schedv1alpha1.PodGroup)
...
+func fillOccupiedObj(pg *schedv1alpha1.PodGroup, pod *v1.Pod) {
+	var refs []string
+	for _, ownerRef := range pod.OwnerReferences {
+		refs = append(refs, fmt.Sprintf("%s/%s", pod.Namespace, ownerRef.Name))
+	}
+	if len(pg.Status.OccupiedBy) == 0 {
+		return
+	}
+	if len(refs) != 0 {
+		pg.Status.OccupiedBy = strings.Join(refs, ",")
```
Huang-Wei, line 288: Sort `refs`?

pkg/controller/podgroup.go
Huang-Wei: I meant you can still sort them, but we can just use the default `sort` library.

cwdsuzhou: done

pkg/controller/podgroup.go
cwdsuzhou: add sort.Strings()

Huang-Wei:
@cwdsuzhou could you take a look at the flake? I created #99 for tracking.

This PR should be irrelevant.

/lgtm
/approve

PR #96, opened 2020-11-07
Cleanup and update Coscheduling parameters

yuanchen8911:
1. Removed the configuration parameters for the old PodGroup label based implementation.
2. Made `lastDeniedPodGroupExpirationTime ` configurable.
3. Fixed some format, naming and comments issues.   

@Huang-Wei  Here are some suggestions about apiVersion and plugin arguments.
1.  I'm not sure whether we need to keep the old config version in `pkg/apis/config`.  How about just switching to `v1beta1`, but renaming it `v1alpha1`? No idea why we use `v1beta1`.   
1. Can we get `KubeconfigPath` and KubeMaster from the command line args or kubeconfig define in the scheduler policy file?
1. Use SetDefaults to set the default parameter values for all plugins.
1. More consistent naming and conventions (e.g., using pointers instead of primitive types)

### DIFF
--- a/pkg/apis/config/types.go
+++ b/pkg/apis/config/types.go
@@ -23,28 +23,28 @@ import (
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// CoschedulingArgs defines the parameters for Coscheduling plugin.
 type CoschedulingArgs struct {
 	metav1.TypeMeta
 
 	// PermitWaitingTime is the wait timeout in seconds.
 	PermitWaitingTimeSeconds int64
-	// PodGroupGCInterval is the period to run gc of PodGroup in seconds.
-	PodGroupGCIntervalSeconds int64
-	// If the deleted PodGroup stays longer than the PodGroupExpirationTime,
-	// the PodGroup will be deleted from PodGroupInfos.
-	PodGroupExpirationTimeSeconds int64
+	// DeniedPGExpirationTimeSeconds is the expiration time of the denied podgroup store.
+	DeniedPGExpirationTimeSeconds int64
 	// KubeMaster is the url of api-server
 	KubeMaster string
 	// KubeConfigPath for scheduler
 	KubeConfigPath string
 }
 
-// modes type.
+// ModeType is a "string" type.
 type ModeType string
 
 const (
+	// Least is the string "Least".
 	Least ModeType = "Least"
-	Most  ModeType = "Most"
+	// Most is the string "Most".
+	Most ModeType = "Most"
 )
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
@@ -66,6 +66,7 @@ type NodeResourcesAllocatableArgs struct {
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// CapacitySchedulingArgs defines the scheduling parameters for CapacityScheduling plugin.
 type CapacitySchedulingArgs struct {
 	metav1.TypeMeta
 
--- a/pkg/apis/config/v1beta1/defaults.go
+++ b/pkg/apis/config/v1beta1/defaults.go
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// +k8s:defaulter-gen=true
+
 package v1beta1
 
 import (
@@ -21,9 +23,8 @@ import (
 )
 
 var (
-	defaultPermitWaitingTimeSeconds      int64 = 10
-	defaultPodGroupGCIntervalSeconds     int64 = 30
-	defaultPodGroupExpirationTimeSeconds int64 = 600
+	defaultPermitWaitingTimeSeconds      int64 = 60
+	defaultDeniedPGExpirationTimeSeconds int64 = 20
 
 	defaultNodeResourcesAllocatableMode = Least
 
@@ -39,19 +40,20 @@ var (
 	defaultKubeConfigPath string = "/etc/kubernetes/scheduler.conf"
 )
 
-func SetDefaults_CoschedulingArgs(obj *CoschedulingArgs) {
+// SetDefaultsCoschedulingArgs sets the default parameters for Coscheduling plugin.
+func SetDefaultsCoschedulingArgs(obj *CoschedulingArgs) {
 	if obj.PermitWaitingTimeSeconds == nil {
 		obj.PermitWaitingTimeSeconds = &defaultPermitWaitingTimeSeconds
 	}
-	if obj.PodGroupGCIntervalSeconds == nil {
-		obj.PodGroupGCIntervalSeconds = &defaultPodGroupGCIntervalSeconds
+	if obj.DeniedPGExpirationTimeSeconds == nil {
+		obj.DeniedPGExpirationTimeSeconds = &defaultDeniedPGExpirationTimeSeconds
 	}
-	if obj.PodGroupExpirationTimeSeconds == nil {
-		obj.PodGroupExpirationTimeSeconds = &defaultPodGroupExpirationTimeSeconds
-	}
+
+	// TODO(k/k#96427): get KubeConfigPath and KubeMaster from configuration or command args.
 }
 
-func SetDefaults_NodeResourcesAllocatableArgs(obj *NodeResourcesAllocatableArgs) {
+// SetDefaultsNodeResourcesAllocatableArgs sets the defaults parameters for NodeResourceAllocatable.
+func SetDefaultsNodeResourcesAllocatableArgs(obj *NodeResourcesAllocatableArgs) {
 	if len(obj.Resources) == 0 {
 		obj.Resources = defaultNodeResourcesAllocatableResourcesToWeightMap
 	}
@@ -61,7 +63,8 @@ func SetDefaults_NodeResourcesAllocatableArgs(obj *NodeResourcesAllocatableArgs)
 	}
 }
 
-func SetDefaults_CapacitySchedulingArgs(obj *CapacitySchedulingArgs) {
+// SetDefaultsCapacitySchedulingArgs sets the default parameters for CapacityScheduling plugin.
+func SetDefaultsCapacitySchedulingArgs(obj *CapacitySchedulingArgs) {
 	if obj.KubeConfigPath == nil {
 		obj.KubeConfigPath = &defaultKubeConfigPath
 	}
--- a/pkg/apis/config/v1beta1/zz_generated.defaults.go
+++ b/pkg/apis/config/v1beta1/zz_generated.defaults.go
@@ -28,22 +28,22 @@ import (
 // Public to allow building arbitrary schemes.
 // All generated defaulters are covering - they call all nested defaulters.
 func RegisterDefaults(scheme *runtime.Scheme) error {
-	scheme.AddTypeDefaultingFunc(&CapacitySchedulingArgs{}, func(obj interface{}) { SetObjectDefaults_CapacitySchedulingArgs(obj.(*CapacitySchedulingArgs)) })
-	scheme.AddTypeDefaultingFunc(&CoschedulingArgs{}, func(obj interface{}) { SetObjectDefaults_CoschedulingArgs(obj.(*CoschedulingArgs)) })
+	scheme.AddTypeDefaultingFunc(&CapacitySchedulingArgs{}, func(obj interface{}) { SetObjectDefaultsCapacitySchedulingArgs(obj.(*CapacitySchedulingArgs)) })
+	scheme.AddTypeDefaultingFunc(&CoschedulingArgs{}, func(obj interface{}) { SetObjectDefaultsCoschedulingArgs(obj.(*CoschedulingArgs)) })
 	scheme.AddTypeDefaultingFunc(&NodeResourcesAllocatableArgs{}, func(obj interface{}) {
-		SetObjectDefaults_NodeResourcesAllocatableArgs(obj.(*NodeResourcesAllocatableArgs))
+		SetObjectDefaultsNodeResourcesAllocatableArgs(obj.(*NodeResourcesAllocatableArgs))
 	})
 	return nil
 }
 
-func SetObjectDefaults_CapacitySchedulingArgs(in *CapacitySchedulingArgs) {
-	SetDefaults_CapacitySchedulingArgs(in)
+func SetObjectDefaultsCapacitySchedulingArgs(in *CapacitySchedulingArgs) {
+	SetDefaultsCapacitySchedulingArgs(in)
 }
 
-func SetObjectDefaults_CoschedulingArgs(in *CoschedulingArgs) {
-	SetDefaults_CoschedulingArgs(in)
+func SetObjectDefaultsCoschedulingArgs(in *CoschedulingArgs) {
+	SetDefaultsCoschedulingArgs(in)
 }
 
-func SetObjectDefaults_NodeResourcesAllocatableArgs(in *NodeResourcesAllocatableArgs) {
-	SetDefaults_NodeResourcesAllocatableArgs(in)
+func SetObjectDefaultsNodeResourcesAllocatableArgs(in *NodeResourcesAllocatableArgs) {
+	SetDefaultsNodeResourcesAllocatableArgs(in)
 }
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -58,12 +58,13 @@ const (
 	Name = "Coscheduling"
 )
 
-// New initializes a new plugin and returns it.
+// New initializes and returns a new Coscheduling plugin.
 func New(obj runtime.Object, handle framework.FrameworkHandle) (framework.Plugin, error) {
 	args, ok := obj.(*config.CoschedulingArgs)
 	if !ok {
 		return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
 	}
+
 	conf, err := clientcmd.BuildConfigFromFlags(args.KubeMaster, args.KubeConfigPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to init rest.Config: %v", err)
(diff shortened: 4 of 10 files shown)


pkg/apis/config/v1beta1/defaults.go
```diff
@@ -37,21 +36,27 @@ var (
...
+	if obj.DeniedPGExpirationTimeSeconds == nil {
+		obj.DeniedPGExpirationTimeSeconds = &defaultDeniedPGExpirationTimeSeconds
 	}
-	if obj.PodGroupExpirationTimeSeconds == nil {
-		obj.PodGroupExpirationTimeSeconds = &defaultPodGroupExpirationTimeSeconds
+	if obj.KubeConfigPath == nil {
+		obj.KubeConfigPath = &defaultKubeConfigPath
+	}
+	if obj.KubeMaster == nil {
+		obj.KubeMaster = &defaultKubeMaster
```
cwdsuzhou, line 54: I not  sure if we need default values here

yuanchen8911:
> /retest

No idea why the integration test failed.

pkg/apis/config/v1beta1/defaults.go
yuanchen8911: Maybe not. Actually, I don't think we need kubeCongfigPath and kubeMaster. They are defined in kubeconfig.

cwdsuzhou:
> > /retest
> 
> No idea why the integration test failed.

Have opened https://github.com/kubernetes-sigs/scheduler-plugins/pull/101  to fix it

pkg/apis/config/v1beta1/defaults.go
cwdsuzhou: I  also think we can remove this when framework supporting directly getting kubeClient form framework

yuanchen8911: @cwdsuzhou  I removed default `KubeConfigPath` and `KubeMaster`.  Hopefully we can get the two parameters from the existing configurations or command args.  Submitted an issue https://github.com/kubernetes/kubernetes/issues/96427.

pkg/apis/config/types.go
```diff
@@ -23,28 +23,28 @@ import (
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
+// CoschedulingArgs defines the paramters for Coscheduling plugin.
```
Huang-Wei, line 26:
```suggestion
// CoschedulingArgs defines the parameters for Coscheduling plugin.
```

pkg/apis/config/v1beta1/defaults.go
```diff
@@ -39,19 +38,20 @@ var (
...
-	if obj.PodGroupGCIntervalSeconds == nil {
-		obj.PodGroupGCIntervalSeconds = &defaultPodGroupGCIntervalSeconds
-	}
-	if obj.PodGroupExpirationTimeSeconds == nil {
-		obj.PodGroupExpirationTimeSeconds = &defaultPodGroupExpirationTimeSeconds
+	if obj.DeniedPGExpirationTimeSeconds == nil {
+		obj.DeniedPGExpirationTimeSeconds = &defaultDeniedPGExpirationTimeSeconds
 	}
+
+	//TODO: get KubeConfigPath and KubeMaster from configuration or command args.
```
Huang-Wei, line 50: Please help to add the k/k issue number here: `TODO(k/k#12345)`

pkg/coscheduling/coscheduling.go
```diff
@@ -32,7 +32,6 @@ import (
 	"k8s.io/klog/v2"
 	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
 	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
-
```
Huang-Wei, line 35: Let's keep this empty line as a sepepartor.

pkg/apis/config/v1beta1/defaults.go
Huang-Wei, line 50:
```suggestion
	// TODO: get KubeConfigPath and KubeMaster from configuration or command args.
```

pkg/apis/config/v1beta1/zz_generated.defaults.go
```diff
@@ -28,22 +28,5 @@ import (
 // Public to allow building arbitrary schemes.
 // All generated defaulters are covering - they call all nested defaulters.
 func RegisterDefaults(scheme *runtime.Scheme) error {
-	scheme.AddTypeDefaultingFunc(&CapacitySchedulingArgs{}, func(obj interface{}) { SetObjectDefaults_CapacitySchedulingArgs(obj.(*CapacitySchedulingArgs)) })
```
Huang-Wei, line 31: Why remove this? For a follow-up exercising v1alpha1 args?

Huang-Wei:
> 1\. I'm not sure whether we need to keep the old config version in `pkg/apis/config`.  How about just switching to `v1beta1`, but renaming it `v1alpha1`? No idea why we use `v1beta1`.

That is to ensure it's consistent with the upstream args. I haven't exercised to use different versions of args in one scheduler config yaml.

> 2\. Can we get `KubeconfigPath` and KubeMaster from the command line args or kubeconfig define in the scheduler policy file?

Let's chase it in the upstream.

> 3\. Use SetDefaults to set the default parameter values for all plugins.

For now, it's already the case. Can you elaborate?

> 4\. More consistent naming and conventions (e.g., using pointers instead of primitive types)

For versioned objs, it's a good practice to set it as pointers (although not always the case). And set them to values in internal versions. So that we can easily tell nil from zero value.

pkg/apis/config/v1beta1/zz_generated.defaults.go
yuanchen8911: No idea. It's generated code. Restored it to the original version.

pkg/apis/config/v1beta1/defaults.go
yuanchen8911: done

pkg/apis/config/v1beta1/zz_generated.defaults.go
yuanchen8911: No idea, it's code generated by code-generated. restored it.

pkg/coscheduling/coscheduling.go
```diff
@@ -55,15 +55,31 @@ var _ framework.PostBindPlugin = &Coscheduling{}
 
 const (
 	// Name is the name of the plugin used in Registry and configurations.
-	Name = "Coscheduling"
+	Name                                       = "Coscheduling"
+	defaultPermitWaitingTimeSeconds      int64 = 60
+	defaultDeniedPGExpirationTimeSeconds int64 = 20
```
Huang-Wei, line 60: We can remove this, as `pkg/apis/config/v1beta1/defaults.go` has already done the defalting logic.

pkg/coscheduling/coscheduling.go
```diff
@@ -55,15 +55,31 @@ var _ framework.PostBindPlugin = &Coscheduling{}
...
+		if !ok {
+			return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
+		}
+	}
+	// TODO: Replace it with v1beta1.SetDefaultsCoschedulingArgs
+	if args.PermitWaitingTimeSeconds == 0 {
+		args.PermitWaitingTimeSeconds = defaultPermitWaitingTimeSeconds
+	}
+	if args.DeniedPGExpirationTimeSeconds == 0 {
+		args.DeniedPGExpirationTimeSeconds = defaultDeniedPGExpirationTimeSeconds
```
Huang-Wei, line 80: Same as above, no need to add duplicated defaulting logic.

pkg/coscheduling/coscheduling.go
```diff
@@ -55,15 +55,31 @@ var _ framework.PostBindPlugin = &Coscheduling{}
...
+// New initializes and returns a new Coscheduling plugin.
 func New(obj runtime.Object, handle framework.FrameworkHandle) (framework.Plugin, error) {
-	args, ok := obj.(*config.CoschedulingArgs)
-	if !ok {
-		return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
+	var (
+		args *config.CoschedulingArgs
+		ok   bool
+	)
+	if obj != nil {
```
Huang-Wei, line 69: Any particular reason checking  if it's nil?

pkg/coscheduling/coscheduling.go
yuanchen8911: We'll then need to import the package`config/v1beta1`.  In that case, we should probably just use `SetDefaultsCoschedulingArgs` instead.

yuanchen8911: > We can remove this, as `pkg/apis/config/v1beta1/defaults.go` has already done the defalting logic.

We'll then need to import the package`config/v1beta1`.  In that case, we should probably just use `SetDefaultsCoschedulingArgs` instead.

pkg/coscheduling/coscheduling.go
yuanchen8911: > Same as above, no need to add duplicated defaulting logic.

Similarly, if we use the version`v1beta1,` we won't need them.

pkg/coscheduling/coscheduling.go
yuanchen8911: It skips setting the defaults  if the configuration is empty.  If the configuration cannot be empty, we can just return with an error. Again, if we use the api version `v1beta1,` we won't need it.

pkg/coscheduling/coscheduling.go
```diff
@@ -55,15 +55,31 @@ var _ framework.PostBindPlugin = &Coscheduling{}
...
+		args, ok = obj.(*config.CoschedulingArgs)
+		if !ok {
+			return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
+		}
+	}
+	// TODO: Replace it with v1beta1.SetDefaultsCoschedulingArgs
```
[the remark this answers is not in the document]
yuanchen8911: Can you elaborate it a bit more? Thanks.

yuanchen8911:
@Huang-Wei  Thanks for reviewing the PR. Most comments on  `CoschedulingArgs` and its default values can be addressed if we switch to the version` v1beta1` and use SetDefaults function. We'll have a neater implementation.  What's your recommendation?

pkg/coscheduling/coscheduling.go
Huang-Wei: I'm not sure I'm following. In terms of args, aren't we already using `v1beta1`, and `SetDefaultsCoschedulingArgs` is called implicitly?

https://github.com/kubernetes-sigs/scheduler-plugins/blob/011599fb15f2d57dbec7afa086814602d7f24664/pkg/apis/config/v1beta1/register.go#L29

pkg/coscheduling/coscheduling.go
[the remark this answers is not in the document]
Huang-Wei: I mean:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/011599fb15f2d57dbec7afa086814602d7f24664/pkg/apis/config/v1beta1/register.go#L47-L54

And schedschemev1beta1 is the one derived from in-tree.

pkg/coscheduling/coscheduling.go
Huang-Wei: Again, I'm not following what do you mean by "if we switch...".

Huang-Wei (commented):
Basically I think making the following changes would be good enough:

```diff
diff --git a/pkg/coscheduling/coscheduling.go b/pkg/coscheduling/coscheduling.go
index b9a3170..76a2c0f 100644
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -55,29 +55,14 @@ var _ framework.PostBindPlugin = &Coscheduling{}
 
 const (
 	// Name is the name of the plugin used in Registry and configurations.
-	Name                                       = "Coscheduling"
-	defaultPermitWaitingTimeSeconds      int64 = 60
-	defaultDeniedPGExpirationTimeSeconds int64 = 20
+	Name = "Coscheduling"
 )
 
 // New initializes and returns a new Coscheduling plugin.
 func New(obj runtime.Object, handle framework.FrameworkHandle) (framework.Plugin, error) {
-	var (
-		args *config.CoschedulingArgs
-		ok   bool
-	)
-	if obj != nil {
-		args, ok = obj.(*config.CoschedulingArgs)
-		if !ok {
-			return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
-		}
-	}
-	// TODO: Replace it with v1beta1.SetDefaultsCoschedulingArgs
-	if args.PermitWaitingTimeSeconds == 0 {
-		args.PermitWaitingTimeSeconds = defaultPermitWaitingTimeSeconds
-	}
-	if args.DeniedPGExpirationTimeSeconds == 0 {
-		args.DeniedPGExpirationTimeSeconds = defaultDeniedPGExpirationTimeSeconds
+	args, ok := obj.(*config.CoschedulingArgs)
+	if !ok {
+		return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
 	}
 
 	conf, err := clientcmd.BuildConfigFromFlags(args.KubeMaster, args.KubeConfigPath)
```

yuanchen8911:
> Basically I think making the following changes would be good enough:
> 
> ```diff
> diff --git a/pkg/coscheduling/coscheduling.go b/pkg/coscheduling/coscheduling.go
> index b9a3170..76a2c0f 100644
> --- a/pkg/coscheduling/coscheduling.go
> +++ b/pkg/coscheduling/coscheduling.go
> @@ -55,29 +55,14 @@ var _ framework.PostBindPlugin = &Coscheduling{}
>  
>  const (
>  	// Name is the name of the plugin used in Registry and configurations.
> -	Name                                       = "Coscheduling"
> -	defaultPermitWaitingTimeSeconds      int64 = 60
> -	defaultDeniedPGExpirationTimeSeconds int64 = 20
> +	Name = "Coscheduling"
>  )
>  
>  // New initializes and returns a new Coscheduling plugin.
>  func New(obj runtime.Object, handle framework.FrameworkHandle) (framework.Plugin, error) {
> -	var (
> -		args *config.CoschedulingArgs
> -		ok   bool
> -	)
> -	if obj != nil {
> -		args, ok = obj.(*config.CoschedulingArgs)
> -		if !ok {
> -			return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
> -		}
> -	}
> -	// TODO: Replace it with v1beta1.SetDefaultsCoschedulingArgs
> -	if args.PermitWaitingTimeSeconds == 0 {
> -		args.PermitWaitingTimeSeconds = defaultPermitWaitingTimeSeconds
> -	}
> -	if args.DeniedPGExpirationTimeSeconds == 0 {
> -		args.DeniedPGExpirationTimeSeconds = defaultDeniedPGExpirationTimeSeconds
> +	args, ok := obj.(*config.CoschedulingArgs)
> +	if !ok {
> +		return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
>  	}
>  
>  	conf, err := clientcmd.BuildConfigFromFlags(args.KubeMaster, args.KubeConfigPath)
> ```

Yes, that's the changes I'd like to make. Instead of setting the default values explicitly, we rely on `SetDefaultsCoschedulingArgs`.  

By switching to `v1beta1`, I mean we need to replace  `import sigs.k8s.io/scheduler-plugins/pkg/apis/config `with  `import sigs.k8s.io/scheduler-plugins/pkg/apis/config/v1beta`. Also, we need to change the default parameter types from int64/string to pointers to in64/string (vebeta1)

yuanchen8911:
> Basically I think making the following changes would be good enough:
> 
> ```diff
> diff --git a/pkg/coscheduling/coscheduling.go b/pkg/coscheduling/coscheduling.go
> index b9a3170..76a2c0f 100644
> --- a/pkg/coscheduling/coscheduling.go
> +++ b/pkg/coscheduling/coscheduling.go
> @@ -55,29 +55,14 @@ var _ framework.PostBindPlugin = &Coscheduling{}
>  
>  const (
>  	// Name is the name of the plugin used in Registry and configurations.
> -	Name                                       = "Coscheduling"
> -	defaultPermitWaitingTimeSeconds      int64 = 60
> -	defaultDeniedPGExpirationTimeSeconds int64 = 20
> +	Name = "Coscheduling"
>  )
>  
>  // New initializes and returns a new Coscheduling plugin.
>  func New(obj runtime.Object, handle framework.FrameworkHandle) (framework.Plugin, error) {
> -	var (
> -		args *config.CoschedulingArgs
> -		ok   bool
> -	)
> -	if obj != nil {
> -		args, ok = obj.(*config.CoschedulingArgs)
> -		if !ok {
> -			return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
> -		}
> -	}
> -	// TODO: Replace it with v1beta1.SetDefaultsCoschedulingArgs
> -	if args.PermitWaitingTimeSeconds == 0 {
> -		args.PermitWaitingTimeSeconds = defaultPermitWaitingTimeSeconds
> -	}
> -	if args.DeniedPGExpirationTimeSeconds == 0 {
> -		args.DeniedPGExpirationTimeSeconds = defaultDeniedPGExpirationTimeSeconds
> +	args, ok := obj.(*config.CoschedulingArgs)
> +	if !ok {
> +		return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
>  	}
>  
>  	conf, err := clientcmd.BuildConfigFromFlags(args.KubeMaster, args.KubeConfigPath)
> ```

Yes, that's the changes I'd like to make. Instead of setting the default values explicitly, we rely on `SetDefaultsCoschedulingArgs`.   I'm submitting an updated version now.

Huang-Wei:
> I may have misunderstood apiVersion management.

It's easy to get confused indeed. In terms of PluginArgs, we followed the KubeSchedulerConfiguration version, which is v1beta1; on the other hand, each plugin has its (optional) CRD API.

yuanchen8911:
squashed commits.

pkg/coscheduling/coscheduling.go
```diff
@@ -55,15 +55,31 @@ var _ framework.PostBindPlugin = &Coscheduling{}
...
+	var (
+		args *config.CoschedulingArgs
+		ok   bool
+	)
+	if obj != nil {
+		args, ok = obj.(*config.CoschedulingArgs)
+		if !ok {
+			return nil, fmt.Errorf("want args to be of type CoschedulingArgs, got %T", obj)
+		}
+	}
```
Huang-Wei, line 74: Please revert this.

pkg/coscheduling/coscheduling.go
Huang-Wei, line 80: This can be removed.

pkg/coscheduling/coscheduling.go
yuanchen8911: sorry, my bad. messed up different versions.

pkg/coscheduling/coscheduling.go
yuanchen8911: sorry, committed the wrong version.

Huang-Wei:
Thanks @yuanchen8911 .

/lgtm
/approve

Issue #25, opened 2020-07-16
Sample plugins to exercise each extension point


Huang-Wei:
It's nice to come up with a series of sample plugins to exercise each extension point, and accompanied with a README guiding newcomers on how that's built.

Generally, the use-cases that the sample plugin tries to solve can be _hypothetical_, however, we do hope it's a or part of a problem in real-case world.

- [x] Sort - A QOS Sort plugin to sort the pods by their QOS level. #3
- [ ] PreFilter
- [ ] Filter
- [x] PostFilter - A cross-node preemption PostFitler plugin to try preempting pods on multiple nodes #56 
- [ ] PreScore
- [x] Score & NormalizeScore - A Score plugin to take (1) terminating Pods and (2) nominated Pods into accounts #103
- [ ] Permit
- [ ] Reserve
- [ ] PreBind
- [ ] Bind

yuchen-sun:
Hi @Huang-Wei  can I do this task?

denkensk:
> Hi @Huang-Wei can I do this task?

@yuchen-sun Of course, you can comment here which plugin you are interested in and take it to do.

Huang-Wei:
@yuchen-sun Sure, which bullet you want to work on?

yuchen-sun:
@Huang-Wei @denkensk THX,may be PreFilter plugin

Huang-Wei:
Feel free to take it, and it would be nice if you can think of a use case for that.

lianghao208:
Is that possible these series of sample plugins  become a new opensource project that newcomers can easily build?

lianghao208:
@Huang-Wei, @denkensk Can I add a sample in pkg/qos that use prefilterPlugin to ignore the Pods which is in BestEffort (requests and limits not set) QOS level? it`s not a regular function but an example exercise tho.  Any suggestions?

Huang-Wei:
@lianghao208 The example would depend on the schdule-delay annotation, right? If so, I don't quite think the sample would make a lot of sense. I believe there should be a better example to demonstrate PreFilter without relying on a "fake" annotation, WDYT?

lianghao208:
[...]

PR #103, opened 2020-11-11
Add Pod State Scheduling Plugin

lianghao208:
a sample score plugin implementation mentioned and discussed in #25 
It takes terminating pods  and nominated pods into accounts.

/kind feature
/cc @Huang-Wei
/assign @Huang-Wei

### DIFF
--- a/cmd/scheduler/main.go
+++ b/cmd/scheduler/main.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/scheduler-plugins/pkg/coscheduling"
 	"sigs.k8s.io/scheduler-plugins/pkg/crossnodepreemption"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesources"
+	"sigs.k8s.io/scheduler-plugins/pkg/podstate"
 	"sigs.k8s.io/scheduler-plugins/pkg/qos"
 
 	// Ensure scheme package is initialized.
@@ -44,6 +45,7 @@ func main() {
 		app.WithPlugin(noderesources.AllocatableName, noderesources.NewAllocatable),
 		// Sample plugins below.
 		app.WithPlugin(crossnodepreemption.Name, crossnodepreemption.New),
+		app.WithPlugin(podstate.Name, podstate.New),
 		app.WithPlugin(qos.Name, qos.New),
 	)
 
--- a/pkg/podstate/README.md
+++ b/pkg/podstate/README.md
@@ -1,0 +1,20 @@
+# Overview
+
+This folder holds a sample plugin implementation based 
+on the number of terminating and nominated Pods on a Node.
+
+## Maturity Level
+
+<!-- Check one of the values: Sample, Alpha, Beta, GA -->
+
+- [x] 💡 Sample (for demonstrating and inspiring purpose)
+- [ ] 👶 Alpha (used in companies for pilot projects)
+- [ ] 👦 Beta (used in companies and developed actively)
+- [ ] 👨 Stable (used in companies for production workloads)
+
+## Pod State Plugin
+
+This is a score plugin that takes terminating and nominated Pods into accounts in the following manner:
+- the nodes that have more terminating Pods will get a higher score as those terminating Pods would be physically removed eventually from nodes
+- the nodes that have more nominated Pods (which carry .status.nominatedNodeName) will get a lower score as the nominated nodes are supposed to accommodate some preemptor pod in a 
+future scheduling cycle.
--- a/pkg/podstate/pod_state.go
+++ b/pkg/podstate/pod_state.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package podstate
+
+import (
+	"context"
+	"fmt"
+	"math"
+
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+)
+
+type PodState struct {
+	handle framework.FrameworkHandle
+}
+
+var _ = framework.ScorePlugin(&PodState{})
+
+// Name is the name of the plugin used in the Registry and configurations.
+const Name = "PodState"
+
+func (ps *PodState) Name() string {
+	return Name
+}
+
+// Score invoked at the score extension point.
+func (ps *PodState) Score(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string) (int64, *framework.Status) {
+	nodeInfo, err := ps.handle.SnapshotSharedLister().NodeInfos().Get(nodeName)
+	if err != nil {
+		return 0, framework.NewStatus(framework.Error, fmt.Sprintf("getting node %q from Snapshot: %v", nodeName, err))
+	}
+
+	// pe.score favors nodes with terminating pods instead of nominated pods
+	// It calculates the sum of the node's terminating pods and nominated pods
+	return ps.score(nodeInfo)
+}
+
+// ScoreExtensions of the Score plugin.
+func (ps *PodState) ScoreExtensions() framework.ScoreExtensions {
+	return ps
+}
+
+func (ps *PodState) score(nodeInfo *framework.NodeInfo) (int64, *framework.Status) {
+	var terminatingPodNum, nominatedPodNum int64
+	// get nominated Pods for node from nominatedPodMap
+	nominatedPodNum = int64(len(ps.handle.PreemptHandle().NominatedPodsForNode(nodeInfo.Node().Name)))
+	for _, p := range nodeInfo.Pods {
+		// Pod is terminating if DeletionTimestamp has been set
+		if p.Pod.DeletionTimestamp != nil {
+			terminatingPodNum++
+		}
+	}
+	return terminatingPodNum - nominatedPodNum, nil
+}
+
+func (ps *PodState) NormalizeScore(ctx context.Context, state *framework.CycleState, pod *v1.Pod, scores framework.NodeScoreList) *framework.Status {
+	// Find highest and lowest scores.
+	var highest int64 = -math.MaxInt64
+	var lowest int64 = math.MaxInt64
+	for _, nodeScore := range scores {
+		if nodeScore.Score > highest {
+			highest = nodeScore.Score
+		}
+		if nodeScore.Score < lowest {
+			lowest = nodeScore.Score
+		}
+	}
+
+	// Transform the highest to lowest score range to fit the framework's min to max node score range.
+	oldRange := highest - lowest
+	newRange := framework.MaxNodeScore - framework.MinNodeScore
+	for i, nodeScore := range scores {
+		if oldRange == 0 {
+			scores[i].Score = framework.MinNodeScore
+		} else {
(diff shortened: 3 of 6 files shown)


lianghao208:
Hi, @Huang-Wei 
I`ve finished. Would you like to review this PR?

pkg/podeffect/pod_effect_test.go
```diff
@@ -0,0 +1,163 @@
...
+			for i := range gotList {
+				if !reflect.DeepEqual(test.expectedList[i].Score, gotList[i].Score) {
+					t.Errorf("expected %#v, got %#v", test.expectedList[i].Score, gotList[i].Score)
+				}
+			}
+		})
+	}
+}
+
+func makeNodeInfo(node string, terminatingPodNumber int, nominatedPodNumber int) *framework.NodeInfo {
```
Huang-Wei, line 108:
```suggestion
func makeNodeInfo(node string, terminatingPodNumber, nominatedPodNumber int) *framework.NodeInfo {
```

pkg/podeffect/pod_effect_test.go
```diff
@@ -0,0 +1,163 @@
...
+	ni := framework.NewNodeInfo()
+	for i := 0; i < terminatingPodNumber; i++ {
+		podInfo := &framework.PodInfo{
+			Pod: makeTerminatingPod("tpod_" + node + strconv.Itoa(i+1)),
```
Huang-Wei, line 112: Please use `fmt.Sprintf()` to format it. (applies elsewhere)

pkg/podeffect/pod_effect_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+var _ framework.SharedLister = &fakeSharedLister{}
+
+type fakeSharedLister struct {
+	nodes []*framework.NodeInfo
+}
+
+func (f *fakeSharedLister) NodeInfos() framework.NodeInfoLister {
+	return fakeframework.NodeInfoLister(f.nodes)
+}
```
Huang-Wei, line 163: You can use `test/util/fake.go`

pkg/podeffect/pod_effect_test.go
```diff
@@ -0,0 +1,163 @@
...
+func TestPodStatusEffect(t *testing.T) {
+
+	tests := []struct {
+		nodeInfos    []*framework.NodeInfo
+		wantErr      string
+		expectedList framework.NodeScoreList
+		name         string
+	}{
+		{
+			nodeInfos:    []*framework.NodeInfo{makeNodeInfo("machine1", 5, 0), makeNodeInfo("machine2", 3, 0)},
```
Huang-Wei, line 33: Let's replace "machine" with "node".

pkg/podeffect/pod_effect_test.go
```diff
@@ -0,0 +1,163 @@
...
+	frameworkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
+	framework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1"
+	fakeframework "k8s.io/kubernetes/pkg/scheduler/framework/v1alpha1/fake"
+	st "k8s.io/kubernetes/pkg/scheduler/testing"
+	testutil "sigs.k8s.io/scheduler-plugins/test/util"
+)
+
+func TestPodStatusEffect(t *testing.T) {
+
+	tests := []struct {
```
Huang-Wei, line 26: Can you expand the nodes number to 3 or 4, so that the score doesn't fall into a binary result: MinNodeScore and MaxNodeScore?

pkg/podeffect/pod_effect_test.go
```diff
@@ -0,0 +1,163 @@
...
+				gotList = append(gotList, framework.NodeScore{Name: test.nodeInfos[i].Node().Name, Score: score})
+			}
+
+			status := plugin.ScoreExtensions().NormalizeScore(context.Background(), nil, nil, gotList)
+			if !status.IsSuccess() {
+				t.Errorf("unexpected error: %v", status)
+			}
+
+			for i := range gotList {
+				if !reflect.DeepEqual(test.expectedList[i].Score, gotList[i].Score) {
```
Huang-Wei, line 100: As you're just comparing `string`, it's good enough to check if the 2 scores are equal.

Huang-Wei (commented):
In addition to the comments below, could you:

1. register it to `cmd/scheduler` and make it UTed in `cmd/scheduler/main_test.go`
1. add an integration test

pkg/podeffect/pod_effect_test.go
lianghao208: test/util/fake.go provides NewFakeSharedLister() which accept *v1.Node as params, but what I need is *framework.NodeInfo because it concludes podInfo. I refer to allocatable_test.go.

lianghao208:
Hi, @Huang-Wei 
Thanks for the review, I just finished modifying the codes.

pkg/podeffect/pod_effect_test.go
lianghao208: Maybe I can improve it by opening another PR. WDYT?

pkg/podstate/README.md
```diff
@@ -0,0 +1,23 @@
...
+- [ ] 👨 Stable (used in companies for production workloads)
+
+## Pod State Plugin
+
+TThis is a score plugin to take terminating and nominated Pods into accounts.
+As those terminating Pods would be physically removed eventually from nodes, 
+the nodes which has more terminating Pods, will get higher score.
+For the nominated nodes (e.g. set by Pods in preemption), they would be given 
+a lower score, as the nodes are supposed to accormordate some preemptor pod in a 
+future scheduling cycle.
```
Huang-Wei, line 22:
```suggestion
This is a score plugin that takes terminating and nominated Pods into accounts in the following manner:
- the nodes that have more terminating Pods will get a higher score as those terminating Pods would be physically removed eventually from nodes
- the nodes that have more nominated Pods (which carry .status.nominatedNodeName) will get a lower score as the nominated nodes are supposed to accommodate some preemptor pod in a 
future scheduling cycle.
```

pkg/podstate/pod_state.go
```diff
@@ -0,0 +1,86 @@
+package podstate
```
Huang-Wei, line 1: Oh, don't forget to add copyright header. (applies elsewhere)

pkg/podstate/pod_state_test.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+		})
+	}
+}
+
+func makeNodeInfo(node string, terminatingPodNumber, nominatedPodNumber int) *framework.NodeInfo {
+	ni := framework.NewNodeInfo()
+	for i := 0; i < terminatingPodNumber; i++ {
+		podInfo := &framework.PodInfo{
+			Pod: makeTerminatingPod(fmt.Sprintf("tpod_%s_%s", node, strconv.Itoa(i+1))),
```
Huang-Wei, line 116: We don't need `strconv.Itoa`, do we?

```suggestion
			Pod: makeTerminatingPod(fmt.Sprintf("tpod_%s_%v", node, i+1)),
```

pkg/podeffect/pod_effect_test.go
Huang-Wei: Given this private fakeSharedLister is simply enough, you can keep it as is.

cmd/scheduler/main.go
```diff
@@ -45,6 +46,7 @@ func main() {
 		// Sample plugins below.
 		app.WithPlugin(crossnodepreemption.Name, crossnodepreemption.New),
 		app.WithPlugin(qos.Name, qos.New),
+		app.WithPlugin(podstate.Name, podstate.New),
```
Huang-Wei, line 49: move it ahead of qos, so as to make they're sorted alphabetically.

Huang-Wei (commented):
Can you add an integration test?

pkg/podstate/pod_state_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+func TestPodState(t *testing.T) {
+	tests := []struct {
+		nodeInfos    []*framework.NodeInfo
+		wantErr      string
+		expectedList framework.NodeScoreList
+		name         string
+	}{
+		{
+			nodeInfos:    []*framework.NodeInfo{makeNodeInfo("node1", 5, 0), makeNodeInfo("node2", 3, 0)},
```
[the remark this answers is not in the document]
lianghao208: Can I  just pass 0 terminating pods and 0 nominated pods as params to specify regular Pods only?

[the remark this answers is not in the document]
Huang-Wei: Not a fan of that. It's always better to make the semantics of a function explicit than implicit.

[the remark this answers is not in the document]
lianghao208: Very good point, I just resolved this.

lianghao208:
@Huang-Wei
I`ve resolved the problems. Thanks!

Huang-Wei:
@lianghao208 it's close to merge. The last missing piece is integration test. Can you add one under `test/integration`? (You can refer to the existing ones for reference)

lianghao208:
> @lianghao208 it's close to merge. The last missing piece is integration test. Can you add one under `test/integration`? (You can refer to the existing ones for reference)

@Huang-Wei 
I have added the integration test, but it only works for the terminating pods now. Making nominated Pods in integration test is a bit complicated. ( For making terminating pods, just simply delete the pods and set GracePeriodSeconds to a large number) Is there an easy way to make nominated Pods in integration test?

Huang-Wei:
@lianghao208 That's a good question. I think we have to simulate the real preemption. Here is my thought:

1. Prepare 2 nodes: node-a and node-b that each has 2 cpus
1. Prepare 3 pods: `p-a1`, `p-a2` on node-a, with low-priority and each requests 1 cpu; `p-b` on node-b, with high-priority and reqeust 2 cpus
1. Schedule a `preemptor pod` with medium-priority, asking for 2 cpus, which would trigger preemption to delete 2 pods from node-a
1. For now, the 2 pods on node-a would stay in terminating phase forever, because that integration test doesn't have a real kubelet, and hence the preemptor pod stays pending
1. Schedule another `test pod` with high-priority, reqeusting 0 cpu. So it's expected this `test pod` would land on node-a as there are 2 terminating pods and 1 nominated pod.

lianghao208:
> 1. Prepare 2 nodes: node-a and node-b that each has 2 cpus
> 2. Prepare 3 pods: `p-a1`, `p-a2` on node-a, with low-priority and each requests 1 cpu; `p-b` on node-b, with high-priority and reqeust 2 cpus
> 3. Schedule a `preemptor pod` with medium-priority, asking for 2 cpus, which would trigger preemption to delete 2 pods from node-a
> 4. For now, the 2 pods on node-a would stay in terminating phase forever, because that integration test doesn't have a real kubelet, and hence the preemptor pod stays pending
> 5. Schedule another `test pod` with high-priority, reqeusting 0 cpu. So it's expected this `test pod` would land on node-a as there are 2 terminating pods and 1 nominated pod.

@Huang-Wei 
Thanks, I just figured it out. I make two pods request for the same host port to create ports conflict. The one has priority will be in preemption. Is this an easier way to simulate the real preeption?

Huang-Wei:
> The one has priority will be in preemption. Is this an easier way to simulate the real preeption?

That _may_ work, but IMO use resource requests may be easier.

lianghao208:
@Huang-Wei 
I've finished the integration test.

Huang-Wei (commented):
Some final nits! We're close to merge.

lianghao208:
@Huang-Wei 
Done. I think it is able to merge now.

test/integration/pod_state_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+			// Ensure the Pod is scheduled successfully.
+			if err := wait.Poll(1*time.Second, 60*time.Second, func() (bool, error) {
+				return podScheduled(cs, ns, tt.pod.Name), nil
+			}); err != nil {
+				t.Errorf("pod %q failed to be scheduled: %v", tt.pod.Name, err)
+			}
+
+			// Lastly, verify pod gets scheduled to the expected node.
+			pod, _ := cs.CoreV1().Pods(ns).Get(context.TODO(), tt.pod.Name, metav1.GetOptions{})
```
Huang-Wei, line 140: Ah I forgot to comment on this: don't swallow err :)

Huang-Wei:
Thanks!

/lgtm
/approve

PR #107, opened 2020-11-12
Update README With Official Container Image Details

seanmalloy:
Document container image location for end usrs now
that release v0.18.9 was created and the first official
container image has been pushed to k8s.gcr.io.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,35 @@
 
 Repository for out-of-tree scheduler plugins based on scheduler framework.
 
-Official container images are coming soon. They will eventually be available at the below location.
+## Install
+
+Starting with release v0.18.9 container images are available in the official k8s container registry
+`k8s.gcr.io/scheduler-plugins/kube-scheduler`. See the [Compatibility Matrix section](#compatibility-matrix)
+for the list of images.
+
 ```shell
 docker pull k8s.gcr.io/scheduler-plugins/kube-scheduler:$TAG
-docker pull k8s.gcr.io/scheduler-plugins/controller:$TAG
 ```
+
+In a future release an official container image will be provided for the scheduler-plugins controller. For
+example `docker pull k8s.gcr.io/scheduler-plugins/controller:$TAG`.
+
+## Compatibility Matrix
+The below compatibility matrix shows the k8s client package(client-go, apimachinery, etc) versions that the
+scheduler-plugins are compiled with.
+
+The minor version of the scheduler-plugins matches the minor version of the k8s client
+packages that it is compiled with. For example scheduler-plugins v0.18.x releases are built with k8s v1.18.x
+dependencies.
+
+The scheduler-plugins patch versions come in two different varieties(single digit or three digits). The single digit
+patch versions(i.e. v0.18.9) exactly align with the the k8s client package versions that the scheduler plugins are built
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,12 +2,18 @@
...
-Official container images are coming soon. They will eventually be available at the below location.
+## Install
+
+Starting with release v0.18.9 container images are available in the official k8s container registry
+`k8s.gcr.io/scheduler-plugins/kube-scheduler`.
+
 ```shell
-docker pull k8s.gcr.io/scheduler-plugins/kube-scheduler:$TAG
-docker pull k8s.gcr.io/scheduler-plugins/controller:$TAG
+docker pull k8s.gcr.io/scheduler-plugins/kube-scheduler:v0.18.9
```
denkensk, line 11: If there is a list showing the existing tags, it may be more convenient for users to choose.

denkensk (commented):
Thanks a lot

README.md
seanmalloy: I could build a mark down table right here in the README similar to what we the descheduler does. This would need to be maintained by hand.

https://github.com/kubernetes-sigs/descheduler#compatibility-matrix

It doesn't look like the `docker` command supports listing images or tags from a remove registry. I think I could come up with a curl command to list the tag in the registry. I think using `gcloud` would be an option too, but most people probably won't have that installed.

Do you prefer a documented curl command or a table in the mark down file?

denkensk: I think a mark down table in README is better. it may be more convenient for users to choose the version.
We can modify it manually after release.
WDYT?

seanmalloy: Yep ... that is fine. I'll update the PR to add this change. Thanks for the feedback.

seanmalloy: I amended the commit and added a compatibility matrix to the README.md file. Let me know if you want any additional changes.

README.md
```diff
@@ -2,12 +2,37 @@
...
+dependencies.
+
+The scheduler-plugins patch versions come in two different varieties(single digit or three digits). The single digit
+patch versions(i.e. v0.18.9) exactly align with the the k8s client package versions that the scheduler plugins are built
+with. The three digit patch versions(i.e. v0.18.800) are used to indicated that the k8s client package versions have not
+changed since the previous release, and that only scheduler plugins code(features or bug fixes) was changed.
+
+Scheduler Plugins  | Compiled With k8s Version | Container Image                                     |
+-------------------|---------------------------|-----------------------------------------------------|
+v0.18.800          | v1.18.8                   | N/A                                                 |
```
Huang-Wei, line 33: Let's remove this line.

Huang-Wei (commented):
Thanks @seanmalloy .

One nit, LGTM otherwise.

README.md
seanmalloy: @Huang-Wei fixed in commit 5fb10fbce068a4f26d58bf1a867742f464bc3c85. Let me know if you want me to squash commits.

Huang-Wei: Yeap, please squash the commits.

seanmalloy: @Huang-Wei commits have been squashed.

Huang-Wei:
/lgtm
/approve

Thanks!

PR #109, opened 2020-11-19
Implement resource pre-check

cwdsuzhou:
This PR implements resource check at filter. We add the parameter `EnableResourceCheck` to enable resource check.

Currently, this resource check can not absolutely to avoid unschedulable group. We should add Post-Filter to reject all if anyone does not fit in the following PRs.

/assign @Huang-Wei 
/cc @denkensk


mateuszlitwin:
This change assumes that all pods are identical, do not have anti affinities, etc. In other words this check only works for specific pod groups, so maybe it should be enabled/disable in the pod group resource rather than plugin config?

Huang-Wei:
@mateuszlitwin I think the resource (pre)check is just coarse-grained. It's not perfect to pre-check every case, however, in practice, it can be beneficial to fit quite a certain amount of scenarios. Is that right, @cwdsuzhou ?

cwdsuzhou:
> @mateuszlitwin I think the resource (pre)check is just coarse-grained. It's not perfect to pre-check every case, however, in practice, it can be beneficial to fit quite a certain amount of scenarios. Is that right, @cwdsuzhou ?

Yes，we do not prefer to covering every scenario. If we would like it be precise, more checks should do here, e.g tolerations and so on, but that would cost too much.

IMO, for unschedulable pods, reject them in post-filter may be better.

mateuszlitwin:
0. What problem is this check addressing? Is this performance? Or maybe how long resources are being held reserved?
1. This check is neither sufficient nor necessary condition for pod group scheduling. In some scenarios it will prevent pod group from scheduling even when if it could. I think it is a bad idea to add these checks to the plugin, because it will make scheduling unpredictable and hard to reason about. I feel this feature is targeting a specific scenario, but the `coscheduling` plugin is meant for the general usage.

I think we could change the check so it is always a necessary condition (regardless of what pods are in a pod group) and could make that check always enabled. For instance, necessary condition could be something like: 

```
currently_available = sum resources available across all nodes for each resource type
for each resource type:
  # requires_at_least is how many resources pod group require to be available in the most optimistic case
  requires_at_least = sum resource limits for first `minAvailable` smallest resource requests from pod group pods
  if requires_at_least > currently_available[resource_type]:
    reject pod group  # we definitely cannot schedule the pod group
    return
```

cwdsuzhou:
> What problem is this check addressing? Is this performance? Or maybe how long resources are being held reserved?

cluster resource can not satisfy the group pods requests.

>This check is neither sufficient nor necessary condition for pod group scheduling. In some scenarios it will prevent pod >group from scheduling even when if it could. I think it is a bad idea to add these checks to the plugin, because it will make >scheduling unpredictable and hard to reason about. I feel this feature is targeting a specific scenario, but the coscheduling >plugin is meant for the general usage.

1. Can you give out your cases
2. IMO this check fits quite a certain amount of scenarios. This can reject group when prefilter.


> I think we could change the check so it is always a necessary condition (regardless of what pods are in a pod group) and 
> could make that check always enabled. For instance, necessary condition, that is similar to your check

Since you think `This check is neither sufficient nor necessary condition for pod group scheduling`, why you recommend the similar way? 

We have discussed with @Huang-Wei in a previous PR, node resource fit has do the same thing, so we can directly apply it.

Reason why not we enable it by default is that this check may cause more and have not implemented Post-Filter. For a large cluster with enough resources, this check may be unnecessary.

cwdsuzhou:
> > What problem is this check addressing? Is this performance? Or maybe how long resources are being held reserved?
> 
> cluster resource can not satisfy the group pods requests.
> 
> > This check is neither sufficient nor necessary condition for pod group scheduling. In some scenarios it will prevent pod >group from scheduling even when if it could. I think it is a bad idea to add these checks to the plugin, because it will make >scheduling unpredictable and hard to reason about. I feel this feature is targeting a specific scenario, but the coscheduling >plugin is meant for the general usage.
> 
> 1. Can you give out your cases
> 2. IMO this check fits quite a certain amount of scenarios. This can reject group when prefilter.
> 
> > I think we could change the check so it is always a necessary condition (regardless of what pods are in a pod group) and
> > could make that check always enabled. For instance, necessary condition, that is similar to your check
> 
> Since you think `This check is neither sufficient nor necessary condition for pod group scheduling`, why you recommend the similar way?
> 
> We have discussed with @Huang-Wei in a previous PR, node resource fit has do the same thing, so we can directly apply it.
> 
> Reason why not we enable it by default is that this check may cause more and have not implemented Post-Filter. For a large cluster with enough resources, this check may be unnecessary.

https://github.com/kubernetes-sigs/scheduler-plugins/pull/57#discussion_r497059846

cwdsuzhou:
ping @Huang-Wei for review

pkg/coscheduling/core/core.go
```diff
@@ -112,7 +119,37 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
...
+	if !pgMgr.enableResourceCheck {
+		return nil
+	}
+	if _, ok := pgMgr.permittedPG.Get(pgFullName); ok {
+		return nil
+	}
+	matchCount := 0
+	for _, node := range nodes {
+		nodeCopy := node.Clone()
+		for {
```
Huang-Wei, line 137: Shouldn't we check if matchCount is not less than minAvailable here? Otherwise, the for loop above may have been run way more than it ought to, right?

```go
for matchCount < int(pg.Spec.MinMember) {
    ...
}
```

test/util/utils.go
```diff
@@ -57,6 +58,11 @@ func MakeNodesAndPods(labels map[string]string, existingPodsNum, allNodesNum int
 	// build nodes
 	for i := 0; i < allNodesNum; i++ {
 		node := st.MakeNode().Name(fmt.Sprintf("node%d", i))
+		res := map[corev1.ResourceName]resource.Quantity{
+			corev1.ResourceCPU:  resource.MustParse("1"),
+			corev1.ResourcePods: resource.MustParse("20")}
+		node.Status.Allocatable = res
+		node.Status.Capacity = res
```
Huang-Wei, line 65: Can be standarized to:

```go
		res := map[corev1.ResourceName]string{
			corev1.ResourceCPU:  "1",
			corev1.ResourcePods: "20",
		}
		node := st.MakeNode().Name(fmt.Sprintf("node%d", i)).Capacity(res)
```

pkg/apis/config/types.go
```diff
@@ -35,6 +35,8 @@ type CoschedulingArgs struct {
 	KubeMaster string
 	// KubeConfigPath for scheduler
 	KubeConfigPath string
+	// EnableResourceCheck enables check cluster resource at pre-filer
```
Huang-Wei, line 38: // XYZ applies additional logic in PreFilter phase to check whether
// minimum resource request of a PodGroup can be satisfied.
// It defaults to false if not set.

(XYZ to be determined)

Huang-Wei:
@mateuszlitwin Re your suggestion in https://github.com/kubernetes-sigs/scheduler-plugins/pull/109#issuecomment-731374849:

IIUC, the current implementation does what you suggested. It's just done in another way - it doesn't pre-summarize the usage for each resource type on all nodes; instead, it checks node one by one and returns early when `matchCount` is satisfies, this in turn results in an optimal time complexity in the average case. WDYT?

pkg/coscheduling/core/core.go
cwdsuzhou: Good catch, thanks.

pkg/apis/config/types.go
```diff
@@ -35,6 +35,8 @@ type CoschedulingArgs struct {
...
+	EnableResourceCheck bool
```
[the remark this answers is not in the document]
cwdsuzhou: Renamed

cwdsuzhou:
@Huang-Wei  PR updated, PTAL

pkg/coscheduling/core/core.go
```diff
@@ -112,7 +119,37 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
 		return fmt.Errorf("cannot found engough pods, "+
 			"current pods number: %v, minMember of group: %v", len(pods), pg.Spec.MinMember)
 	}
-	return nil
+
+	// TODO(cwdsuzhou): This resource check can not absolutely to avoid unschedulable group. We should add Post-Filter to reject all if anyone does not fit.
```
Huang-Wei, line 123: This comment sounds a bit ambiguous to me, esp. the PostFilter part.

I think you can just focus on explaining how it works in PreFilter as well as emphasizing its limitation (only PreFilter resource constraints so even if a PodGroup passed here, it may not necessarily pass Filter due to other constraints such as affinity/taints, etc.)

mateuszlitwin:
@Huang-Wei 

Consider the scenario where your pod group has two types of pods:

- type A: 100 pods requesting 8 GPUs and 64 CPUs
- type B: 100 pods requesting 0 GPUs and 80 CPUs

Pod group size / min available is 200.

Let say cluster have 100 nodes with 8 GPUs/64CPUs available and 100 CPU-only nodes with 80 CPU nodes.
In this scenario it should be possible to schedule all pods:

- type A pods would be scheduled on GPU nodes
- type B pods would be scheduled on CPU-only nodes

This check however will try to

- (when type A pod is scheduled) find matching nodes for 200 x (8 GPUs, 64 CPUs) pods , and fail, pod group will not be permitted
- (when type B pod is scheduled) find matching nodes for 200 x (0 GPUs, 80 CPUs) pods and fail, pod group will not be permitted

Check I proposed does not have this issue: it checks if there are 100x8 GPUs and 100x64+100x80 CPUs available in the cluster.

Huang-Wei:
@mateuszlitwin I see your point now. You're referring to a scenario that employs different workloads in a single PodGroup. The current implementation, however, assumes a unique workalod spec to be associated with a PodGroup.

I have a question here, in the multi-workload case, how do you calculate the summed PodGroup requests, only by the incoming Pod? IIUC, it's difficult to be aware of exactly how many workload specs are present in advance. Probably we have to define the summed requests explicitly in PodGroup API spec.

BTW: are you using this pattern frequently in your business scenario? (also cc/ @denkensk @yuanchen8911)

cwdsuzhou:
PR updated!

Since pre-check may cost more. We still add the option to PluginArgs. Reason why not we add it in PodGroup is that enable this may affect the cluster scheduling and other users. So IMO, it is may better add it in the pluginArg.

When `enableResourcePreCheck` is true:
     If MinResource  is not nil, we would compare left Resource and MinResource.
     If MinResource is nil, we go to NodeFit

`enableResourcePreCheck` is false：
    We do not check resource.



@Huang-Wei @deads2k  PTAL

pkg/coscheduling/core/core.go
```diff
@@ -112,7 +119,46 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
...
+	// TODO(cwdsuzhou): This resource check can not absolutely to avoid unschedulable group.
+	// Only PreFilter resource constraints so even if a PodGroup passed here,
+	// it may not necessarily pass Filter due to other constraints such as affinity/taints.
+	nodes, err := pgMgr.snapshotSharedLister.NodeInfos().List()
+	if err != nil {
+		return err
+	}
+	if !pgMgr.enableResourcePreCheck {
+		return nil
+	}
```
Huang-Wei, line 132: This should be applied only when MinResource is nil (L136).

I'd suggest to refine the logic flow like this:

```go
if _, ok := pgMgr.permittedPG.Get(pgFullName); ok {
    return nil
}

if pg.Spec.MinResources != nil {
    // logic here, and add it to permitted cache if needed
    return Success or Failure
}

if !pgMgr.enableResourcePreCheck {
    return nil
}

// Logic to handle PreCheck on Pod's req * minNum
```

pkg/coscheduling/core/core.go
```diff
@@ -112,7 +119,46 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
...
+				nodeCopy.AddPod(pod)
+			}
+			if matchCount >= int(pg.Spec.MinMember) {
+				pgMgr.permittedPG.Add(pgFullName, pgFullName, *pgMgr.scheduleTimeout)
+				return nil
+			}
+		}
+	} else {
+		clusterLeftResource := pgMgr.getLeftClusterResource()
+		if clusterResourceSatisfied(clusterLeftResource, framework.NewResource(*pg.Spec.MinResources)) {
```
Huang-Wei, line 157: Two comments:

1. Combine the two methods, and return early when we walked to NodeX and have found that minResources are already satisfied.
1. Return a concrete error indicating which type of resource is not satisfied; instead of a binary true or false.

Huang-Wei: BTW: I guess there is already some existing code snippet in k/k and hence can be reused here?

mateuszlitwin:
@Huang-Wei I do not want to go off-topic in this PR and I am happy to use different channels, like slack.

Long story short: yes, we have large jobs that are composed of 100s of pods of a few different sizes and responsibilities; they are deployed as part of the single ML experiment, so we want to schedule them together. Example workloads in my company: https://openai.com/blog/openai-five/#rapid

Regarding "it's difficult to be aware of exactly how many workload specs are present in advance" - in our case we care about gang scheduling - all pods or nothing, that is minAvailable == numberOfAllPods, so it is possible to wait for all pod resources to be created and added to the scheduler to understand what needs to be scheduled.

yuanchen8911:
> PR updated!
> 
> Since pre-check may cost more. We still add the option to PluginArgs. Reason why not we add it in PodGroup is that enable this may affect the cluster scheduling and other users. So IMO, it is may better add it in the pluginArg.
> 
> When `enableResourcePreCheck` is true:
> If MinResource is not nil, we would compare left Resource and MinResource.
> If MinResource is nil, we go to NodeFit
> 
> `enableResourcePreCheck` is false：
> We do not check resource.
> 
> @Huang-Wei @deads2k PTAL

I'm not sure we really need `enableResourcePreCheck` option. If `minResource` is specified, do PreCheck. Otherwise skip it.  Also,` minResource` should be the total minimum resources across all pod specs,  e.g., spark driver + n * spark executors.

Huang-Wei:
> I'm not sure we really need `enableResourcePreCheck` option.

By default it's disabled, so generally, it should be fine.

> Also,` minResource` should be the total minimum resources across all pod specs

Yes, it's the current semantics.

Huang-Wei:
Inspired by the design of NodeAffinityArgs (https://github.com/kubernetes/kubernetes/pull/96202), maybe it's better to make the semantics more explicit. Here is another though:

- Introduce an option in plugin args named `ResourcePreCheck`, which is to type enum and can be one of the following values:
    - `Disabled`: in this case, we don't do pre check at all
    - `PodGroupMinResource`: we check the `MinResource` specified in individual PodGroup
    - `MultiplySinglePodResource`: we check if a single pod's resource * its `MinAvaialable` is satistifed.

WDYT?

cwdsuzhou:
> Inspired by the design of PodTopologySpreadConstraintsDefaulting ([kubernetes/kubernetes#95048](https://github.com/kubernetes/kubernetes/pull/95048)), maybe it's better to make the semantics more explicit. Here is another thought:
> 
> * Introduce an option in plugin args named `ResourcePreCheckType`, which is of type enum and can be one of the following values:
>   
>   * `Disabled`: in this case, we don't do pre-check at all
>   * `PodGroupMinResource`: we check the `MinResource` specified in individual PodGroup
>   * `MultiplySinglePodResource`: we check if a single pod's resource * its `MinAvaialable` is satisfied.
> 
> WDYT?

After we discussed offline, we would only check resource based on `MinResource`

pkg/coscheduling/core/core.go
cwdsuzhou: > BTW: I guess there is already some existing code snippet in k/k and hence can be reused here?

I try to reuse the logic in K8s, but it seems hard to do this. Maybe we can expose some func in `Kubernetes` to do this. At this time, we can not re-use.

pkg/coscheduling/core/core.go
```diff
@@ -238,7 +271,102 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
...
+		}
+	}
+	klog.V(4).Infof("Node %v left resource %+v", info.Node().Name, leftResource)
+	return &leftResource
+}
+
+func formatErrors(notFits []noderesources.InsufficientResource) error {
+	err := fmt.Errorf("Cluster resourc not enough")
+	for _, notFit := range notFits {
+		err = errors.Wrap(err, notFit.Reason)
```
Huang-Wei, line 369: Is it mandatory to introduce this `github.com/pkg/errors` dependency? If not, I'd like to keep it simple and only use built-in Go library.

pkg/coscheduling/core/core.go
```diff
@@ -112,6 +117,34 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
...
 			"current pods number: %v, minMember of group: %v", len(pods), pg.Spec.MinMember)
 	}
+
+	nodes, err := pgMgr.snapshotSharedLister.NodeInfos().List()
+	if err != nil {
+		return err
+	}
+
+	minResources := pg.Spec.MinResources
+	if minResources == nil {
```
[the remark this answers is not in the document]
cwdsuzhou: this is a *ResourceList type

pkg/coscheduling/core/core.go
```diff
@@ -238,7 +271,102 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
 	return count
 }
 
+func (pgMgr *PodGroupManager) CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest framework.Resource) (*framework.Resource, []noderesources.InsufficientResource) {
```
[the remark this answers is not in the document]
cwdsuzhou: Seems this can avoid computing cluster left resources.  I have  taken a similar method

pkg/coscheduling/core/core.go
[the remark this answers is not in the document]
cwdsuzhou: I  think we should change this to ResourceList in another PR

pkg/coscheduling/core/core.go
```diff
@@ -238,7 +269,60 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
...
+				delete(resourceRequest, name)
+				continue
+			}
+			resourceRequest[name] = quant
+		}
+		if len(resourceRequest) == 0 {
+			return nil
+		}
+	}
+	return fmt.Errorf("Resource gap: %v", resourceRequest)
```
Huang-Wei, line 298:
```suggestion
	return fmt.Errorf("resource gap: %v", resourceRequest)
```

pkg/coscheduling/core/core.go
```diff
@@ -112,6 +116,33 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
...
+	if err != nil {
+		return err
+	}
+
+	minResources := *pg.Spec.MinResources
+	podQuantity := resource.NewQuantity(int64(pg.Spec.MinMember), resource.DecimalSI)
+	minResources[corev1.ResourcePods] = *podQuantity
+	err = pgMgr.CheckClusterResource(nodes, minResources)
+	if err != nil {
+		klog.Error(err)
```
Huang-Wei, line 141: Please prepend the PodGroup name to the error.

Huang-Wei (commented):
Some nits. LGTM otherwise.

pkg/coscheduling/core/core.go
```diff
@@ -58,9 +59,11 @@ type PodGroupManager struct {
 	// scheduleTimeout is the default time when group scheduling.
 	// If podgroup's ScheduleTimeoutSeconds set, that would be used.
 	scheduleTimeout *time.Duration
-	// lastDeniedPG store the pg name if a pod can not pass pre-filer,
+	// lastDeniedPG stores the pg name if a pod can not pass pre-filer,
 	// or anyone of the pod timeout
 	lastDeniedPG *gochache.Cache
+	// permittedPG stores the pg name which has passed the pre resource check.
+	permittedPG *gochache.Cache
```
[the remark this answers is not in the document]
cwdsuzhou: Can not in both at same time. I would not like combine them for that they may have different expire time or we can deal with the cache in next PR (PostFilter), postFilter would  add some logics, e.g. remove from permitted or add to denied. If you would like combine them, we can do that.

pkg/coscheduling/core/core.go
```diff
@@ -238,7 +269,60 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
 	return count
 }
 
+func (pgMgr *PodGroupManager) CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest corev1.ResourceList) error {
+	for k, info := range nodeList {
+		if info == nil {
+			continue
+		}
+		if info.Node() == nil {
+			klog.Warningf("Node: %v can not found", k)
```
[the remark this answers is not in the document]
cwdsuzhou: thanks

pkg/coscheduling/core/core.go
```diff
@@ -238,7 +269,60 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
...
+			continue
+		}
+
+		if info.Node().Spec.Unschedulable {
+			continue
+		}
```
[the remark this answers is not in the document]
cwdsuzhou: done, thanks

pkg/coscheduling/core/core.go
cwdsuzhou: Done

cwdsuzhou:
PR addressed thanks @Huang-Wei

Huang-Wei:
/lgtm
/approve

Thanks!

PR #112, opened 2020-11-23
Add PreFilterExtensions for Coscheduling

cwdsuzhou:
Add PreFilterExtensions implementation for Coscheduling.  We should not allow pod to preempt the pod in a same pod group


Huang-Wei (commented):
In most cases, pods belonging to the same PodGroup would have the same priority. So I guess optimization in this PR may not be necessary as preemptor can only preempt low-priority Pods. WDYT?

cwdsuzhou:
> In most cases, pods belonging to the same PodGroup would have the same priority. So I guess optimization in this PR may not be necessary as preemptor can only preempt low-priority Pods. WDYT?

Thanks


/close

PR #119, opened 2020-11-30
Topology aware scheduler plugin

AlexeyPerevalov:
Initially it was proposed as built-in plugin https://github.com/kubernetes/enhancements/pull/1858

Co-authored-by: Swati Sehgal <swsehgal@redhat.com>
Signed-off-by: Alexey Perevalov <alexey.perevalov@huawei.com>

### DIFF
--- a/kep/119-node-resource-topology-aware-scheduling/README.md
+++ b/kep/119-node-resource-topology-aware-scheduling/README.md
@@ -1,0 +1,261 @@
+# Topology Aware Scheduling
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Topology format](#topology-format)
+  - [CRD API](#crd-api)
+  - [Plugin implementation details](#plugin-implementation-details)
+    - [Topology information in the NodeResourceTopologyMatch plugin](#topology-information-in-the-noderesourcetopologymatch-plugin)
+    - [Description of the Algorithm](#description-of-the-algorithm)
+  - [Accessing NodeResourceTopology CRD](#accessing-noderesourcetopology-crd)
+- [Use cases](#use-cases)
+- [Known limitations](#known-limitations)
+- [Test plans](#test-plans)
+- [Graduation criteria](#graduation-criteria)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature enablement and rollback](#feature-enablement-and-rollback)
+- [Implementation history](#implementation-history)
+<!-- /toc -->
+
+# Summary
+
+This document describes behaviour of the Kubernetes Scheduler which takes worker node topology into account.
+
+# Motivation
+
+After Topology Manager was introduced the problem of launching pod in the
+cluster where worker nodes have different NUMA topology and different amount
+of resources in that topology became actual. Pod could be scheduled on the node
+where total amount of resources are enough, but resource distribution could not
+satisfy the appropriate Topology policy. In this case the pod failed to start. Much
+better behaviour for scheduler would be to select appropriate node where kubelet admit
+handlers may pass.
+
+
+## Goals
+
+- Make scheduling process more precise when we have NUMA topology on the
+worker node by using scheduler plugin.
+
+## Non-Goals
+
+- Change the PodSpec to allow requesting a specific node topology manager policy
+- This Proposal requires exposing NUMA topology information. This KEP doesn't
+describe how to expose all necessary information it just declare what kind of
+information is necessary. Please see [topology aware resource provisioning daemon KEP][4] for more information.
+
+# Proposal
+
+This plugin implements a simplified version of Topology Manager and hence is different from original topology manager algorithm.
+Plugin would check for the ability to run pod only in case of single-numa-node policy on the
+node, since it is the most strict policy, it implies that the launch on the node with
+other existing policies will be successful if the condition for single-numa-node policy passed for the worker node.
+Proposed plugin will use [CRD][1] to identify which topology policy is enabled on the node.
+To work, this plugin requires topology information of the available resource on the worker nodes.
+
+## Topology format
+
+Available resources with topology of the node should be stored in CRD. Format of the topology described
+[in this document][1].
+
+[The daemon][5] which runs outside of the kubelet will collect all necessary information on running pods, based on allocatable resources of the node and consumed resources by pods it will provide available resources in CRD, where one CRD instance represents one worker node. The name of the CRD instance is the name of the worker node.
+
+## CRD API
+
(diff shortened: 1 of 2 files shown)


kep/119-node-resource-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,261 @@
...
+
+* Beta
+- [ ] Add node E2E tests.
+- [ ] Provide beta-level documentation.
+
+# Production Readiness Review Questionnaire
+
+## Feature enablement and rollback
+* **How can this feature be enabled / disabled in a live cluster?**
+    - This plugin doesn't require special feature gate, but it expects: TopologyManager and CPUManager feature gate enabled on the worker node\
```
AlexeyPerevalov, line 251: If I'm not wrong if special  KubeSchedulerConfiguration is not defined, than plugin is not loading.

Huang-Wei:
@AlexeyPerevalov We haven't come to a conclusion where to hold the API, have we?

If we get a consensus on https://github.com/kubernetes/kubernetes/pull/96275, then we can implement it in in-tree scheduler.

AlexeyPerevalov:
> @AlexeyPerevalov We haven't come to a conclusion where to hold the API, have we?
> 
> If we get a consensus on [kubernetes/kubernetes#96275](https://github.com/kubernetes/kubernetes/pull/96275), then we can implement it in in-tree scheduler.

We discussed the ability to include it as built-in API in the sig-architecture meeting, where opinions divided. The staging is also raised concerns. So we will collect usage feedbacks for NodeResourceTopology plugin and for API,
API could be placed in scheduler-plugins or in somewhere in https://github.com/k8stopologyawareschedwg,

swatisehgal:
Discussion upstream is still going on and we are pursuing acceptance of in-tree scheduler plugin (and will continue to do so). While we work on showcasing the advantages of Topology aware scheduling in a more measurable way in order to make a strong case for in-tree acceptance of this feature,  it was suggested that we consider an out-of-tree approach for better feature velocity. Hence the enhancement proposal here.

Huang-Wei:
@AlexeyPerevalov @swatisehgal Thanks for the clarification. This proposal looks good to me (actually we have already reviewed this KEP in the upstream). It's just that I'd like to hold the merge in this repo until we got a conclusion on where to place the APIs. Let me know your thoughts.

swatisehgal:
@Huang-Wei Thanks for reviewing the proposal. In this KEP, in the CRD API section [here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/dbacf977773361a64c9de44398f16cc8f8ef6512/kep/119-node-resource-topology-aware-scheduling/README.md#crd-api)  we are proposing that the API be placed in kubernetes-sigs/scheduler-plugins in pkg/apis/noderesourcetopology-api directory. What do you think about that? Getting this scheduler plugin merged as an out-of tree scheduler plugin would be a great starting point for us as it would allow relevant stakeholders to play around with it allowing us to gather feedback and help mature the plugin and the API. This would also help drive conversations upstream to figure out how we as a community want to shape and introduce this feature natively in kubernetes. Let us know your thoughts.

Huang-Wei:
@swatisehgal does that mean the work pushing the API to staging and scheduler plugin to in-tree being postponed? Or you're still pursuing that?

The other aspect is the API review. Similar to the Topology Manager feature, sig-node should be responsible to own and review the topology API. Sig-scheduling is out of expertise in this area actually. So if we store the CRD API in scheduler-plugins, can we involved sig-node to review it to ensure the API design is appropriate?

swatisehgal:
> @swatisehgal does that mean the work pushing the API to staging and scheduler plugin to in-tree being postponed? Or you're still pursuing that?

We need to do some work to showcase the value-proposition of topology-aware scheduling to make a strong case for the API and scheduler-plugin in-tree. We think that this will take some time but when we are ready with that we will resume the discussion  with SIG-arch and SIG-node. So, to answer your question we are pursuing pushing in-tree in the background but postponing the discussion for now.
> The other aspect is the API review. Similar to the Topology Manager feature, sig-node should be responsible to own and review the topology API. Sig-scheduling is out of expertise in this area actually. So if we store the CRD API in scheduler-plugins, can we involved sig-node to review it to ensure the API design is appropriate?

We will take care of this and bring this to SIG-Node. Thanks for letting us know the expectations.

kep/119-node-resource-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,335 @@
...
+	Costs      CostList         `json:"costs,omitempty"`
+	Attributes AttributeList    `json:"attributes,omitempty"`
+	Resources  ResourceInfoList `json:"resources,omitempty"`
+}
+
+type ZoneList []Zone
+
+type ResourceInfo struct {
+	Name        string             `json:"name"`
+	Allocatable intstr.IntOrString `json:"allocatable"`
```
draveness, line 116: How can we maintain the consistency for `NodeResourceTopology.Allocatable` and reserved CPUs. Let's suppose if schedule a pod on the node. 

+ In Filter and Score extension, we use the `NodeResourceTopology.Allocatable` to filter and select the host.
+ In Reserve extension, we reserve several CPUs for the current pod (kubelet don't know the reserved CPUs)
+ In the PostBind extension, we should release the CPUs as Kubelet would update the CRD to reflect the CPU usage, e.g. reduce the allocatable field.

Since kubelet and scheduler do not talk to each other, is it possible that there could be a small timeslot between releasing the CPUs in the PostBind extension and kubelet update the CRD? And is a wrong scheduling decision could be made in that gap?

AlexeyPerevalov: Yes, this KEP didn't cover issue you mentioned, but it should. We can't follow approach of noderesources/fit.go built-in plugin, where plugin relies on NodeInfo.Allocatable, which updates once pod nominated, and it reduces the gap. We can't follow it, because we can't predict which NUMA node will be chosen on the worker node.
I think in future we can try to add some "hint", e.g. annotation at PostBind stage to inform kubelet which NUMA node was selected and use it to reduce amount of available resource of the worker node, as it implemented for Fit plugin.

draveness: Is it possible if we select a socket or sockets during scheduling and the kubelet allocates the pod on the socket. I think after we introduce NUMA into scheduling, the socket may become the minimum scheduling unit.

AlexeyPerevalov: Socket it's another term, several NUMA nodes could be on one socket. Here we're talking about NUMA nodes, yes as I said, if we would manage to add some annotation it's possible for kubelet to make its decision based on scheduler's evaluation.

draveness: Can we introduce some common approach to control the behaviour on how kubelet set cgroups, including cpuset, cfs and etc.

AlexeyPerevalov: I think external resource management like https://github.com/nokia/CPU-pooler or https://github.com/intel/cri-resource-manager/ could address cgroup management issue.

PR #122, opened 2020-12-04
Fix case sensitive args

AlexeyPerevalov:
While playing with different plugins I found some strange places in example manifests.


manifests/capacityscheduling/scheduler-config.yaml
```diff
@@ -21,4 +21,4 @@ profiles:
   pluginConfig:
   - name: CapacityScheduling
     args:
-      kubeConfigPath: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
\ No newline at end of file
+      kubeconfigpath: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
```
AlexeyPerevalov, line 24: the general notation isCamelStyle, so maybe fix it in the pkg/apis/config/v1beta1/types.go?
I'm asking because it could affect something.

Huang-Wei: Thanks for the catch. Yes, please fix the json tag in `pkg/apis/config/v1beta1/types.go`:

```go
KubeConfigPath *string `json:"kubeConfigPath,omitempty"`
```

damemi (commented):
/ok-to-test
/lgtm

PR #124, opened 2020-12-07
Test cases for resource pre-check

cwdsuzhou:
Add some cases for resource pre-check

### DIFF
--- a/test/integration/coscheduling_test.go
+++ b/test/integration/coscheduling_test.go
@@ -351,6 +351,63 @@ func TestCoschedulingPlugin(t *testing.T) {
 			},
 			expectedPods: []string{"t7-p1-1", "t7-p1-2", "t7-p1-3"},
 		},
+		{
+			name: "equal priority, not sequentially pg1 meet min and p2 p3 not meet min, pgs have min resources",
+			pods: []*v1.Pod{
+				WithContainer(st.MakePod().Namespace(ns.Name).Name("t8-p1-1").Req(map[v1.ResourceName]string{v1.ResourceMemory: "50"}).Priority(
+					midPriority).Label(coschedulingutil.PodGroupLabel, "pg8-1").ZeroTerminationGracePeriod().Obj(), pause),
+				WithContainer(st.MakePod().Namespace(ns.Name).Name("t8-p2-1").Req(map[v1.ResourceName]string{v1.ResourceMemory: "100"}).Priority(
+					midPriority).Label(coschedulingutil.PodGroupLabel, "pg8-2").ZeroTerminationGracePeriod().Obj(), pause),
+				WithContainer(st.MakePod().Namespace(ns.Name).Name("t8-p3-1").Req(map[v1.ResourceName]string{v1.ResourceMemory: "100"}).Priority(
+					midPriority).Label(coschedulingutil.PodGroupLabel, "pg8-3").ZeroTerminationGracePeriod().Obj(), pause),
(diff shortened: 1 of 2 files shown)


test/integration/coscheduling_test.go
```diff
@@ -229,8 +229,8 @@ func TestCoschedulingPlugin(t *testing.T) {
 					midPriority).Label(coschedulingutil.PodGroupLabel, "pg2-2").ZeroTerminationGracePeriod().Obj(), pause),
 			},
 			podGroups: []*v1alpha1.PodGroup{
-				util.MakePG("pg2-1", ns.Name, 3, nil, nil),
-				util.MakePG("pg2-2", ns.Name, 4, nil, nil),
+				util.MakePG("pg2-1", ns.Name, 3, nil, &v1.ResourceList{v1.ResourceMemory: resource.MustParse("150")}),
+				util.MakePG("pg2-2", ns.Name, 4, nil, &v1.ResourceList{v1.ResourceMemory: resource.MustParse("400")}),
```
Huang-Wei, line 233: I'd suggest duplicating this test with minResource rather than replacing the existing one.

test/integration/coscheduling_test.go
```diff
@@ -408,6 +442,10 @@ func makeCRD() *apiextensionsv1.CustomResourceDefinition {
 										Type:    "integer",
 										Minimum: &min,
 									},
+									"minResources": {
+										Type:                   "object",
+										XPreserveUnknownFields: &preserveUnknownFields,
```
Huang-Wei, line 447: What's the intention here?

cwdsuzhou: I found in apiextension/v1, minResources would be ignored if we do not set XPreserveUnknownFields, but in apiextension/v1beta1, it works

test/integration/coscheduling_test.go
cwdsuzhou: Got it

cwdsuzhou:
@Huang-Wei PTAL updated

test/integration/coscheduling_test.go
cwdsuzhou: I found in apiextension/v1beta1, default value of this field is true

cwdsuzhou: https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/v1beta1/defaults.go#L73

Huang-Wei: If I understand  https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/types.go#L65-L71 cleary, it's good enough to just add `minResources` to OpenAPI schema, rather than additionally defining "XPreserveUnknownFields", right?

Huang-Wei:
> @Huang-Wei PTAL updated

Just one nit: https://github.com/kubernetes-sigs/scheduler-plugins/pull/124#discussion_r538853936. LGTM otherwise.

test/integration/coscheduling_test.go
cwdsuzhou: PR changed

Have changed `minResources`  to `AdditionalPropertities`

See https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/validation/validation.go#L756

cwdsuzhou:
> > @Huang-Wei PTAL updated
> 
> Just one nit: [#124 (comment)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/124#discussion_r538853936). LGTM otherwise.

PTAL, explanation added!

Huang-Wei:
Thanks!

/lgtm
/approve

PR #126, opened 2020-12-07
Add PostFilter implementation

cwdsuzhou:
Add postFilter implementation

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -49,6 +49,7 @@ type Coscheduling struct {
 
 var _ framework.QueueSortPlugin = &Coscheduling{}
 var _ framework.PreFilterPlugin = &Coscheduling{}
+var _ framework.PostFilterPlugin = &Coscheduling{}
 var _ framework.PermitPlugin = &Coscheduling{}
 var _ framework.ReservePlugin = &Coscheduling{}
 var _ framework.PostBindPlugin = &Coscheduling{}
@@ -130,11 +131,60 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
 // 1. Whether the PodGroup that the Pod belongs to is on the deny list.
 // 2. Whether the total number of pods in a PodGroup is less than its `minMember`.
 func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod) *framework.Status {
+	// If any validation failed, a no-op state data is injected to "state" so that in later
+	// phases we can tell whether the failure comes from PreFilter or not.
 	if err := cs.pgMgr.PreFilter(ctx, pod); err != nil {
 		klog.Error(err)
+		state.Write(cs.getStateKey(), NewNoopStateData())
 		return framework.NewStatus(framework.Unschedulable, err.Error())
 	}
 	return framework.NewStatus(framework.Success, "")
+}
+
+// PostFilter is used to rejecting a group of pods if a pod does not pass PreFilter or Filter.
+func (cs *Coscheduling) PostFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod,
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	// Check if the failure comes from PreFilter or not.
+	_, err := state.Read(cs.getStateKey())
+	if err == nil {
+		state.Delete(cs.getStateKey())
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable)
+	}
+
+	pgName, pg := cs.pgMgr.GetPodGroup(pod)
+	if pg == nil {
+		klog.V(4).Info("Pod does not belong to any group")
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable, "can not find pod group")
+	}
+
+	// This indicates there are already enough Pods satisfying the PodGroup,
+	// so don't bother to reject the whole PodGroup.
+	assigned := cs.pgMgr.CalculateAssignedPods(pg.Name, pod.Namespace)
+	if assigned >= int(pg.Spec.MinMember) {
+		klog.V(4).Infof("%v pods of %v assigned", assigned, pgName)
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable)
+	}
+
+	// If the gap is less than/equal 10%, we may want to try subsequent Pods
+	// to see they can satisfy the PodGroup
+	notAssiginedPercentage := float32(int(pg.Spec.MinMember)-assigned) / float32(pg.Spec.MinMember)
+	if notAssiginedPercentage <= 0.1 {
+		klog.V(4).Infof("%.2f/100 pods of group %v have not assigned", float32(assigned*100), pgName)
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable)
+	}
+
+	// It's based on an implicit assumption: if the nth Pod failed,
+	// it's inferrable other Pods belonging to the same PodGroup would be very likely to fail.
+	cs.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Namespace == pod.Namespace && waitingPod.GetPod().Labels[util.PodGroupLabel] == pg.Name {
+			klog.V(3).Infof("PostFilter rejects the pod: %v/%v", pgName, waitingPod.GetPod().Name)
+			waitingPod.Reject(cs.Name())
+		}
+	})
+	cs.pgMgr.AddDeniedPodGroup(pgName)
+	cs.pgMgr.DeletePermittedPodGroup(pgName)
+	return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable,
+		fmt.Sprintf("PodGroup %v gets rejected due to Pod %v is unschedulable even after PostFilter", pgName, pod.Name))
 }
 
 // PreFilterExtensions returns a PreFilterExtensions interface if the plugin implements one.
@@ -199,6 +249,7 @@ func (cs *Coscheduling) Unreserve(ctx context.Context, state *framework.CycleSta
 		}
 	})
 	cs.pgMgr.AddDeniedPodGroup(pgName)
+	cs.pgMgr.DeletePermittedPodGroup(pgName)
 }
 
 // PostBind is called after a pod is successfully bound. These plugins are used update PodGroup when pod is bound.
@@ -214,4 +265,19 @@ func (cs *Coscheduling) rejectPod(uid types.UID) {
 		return
 	}
 	waitingPod.Reject(Name)
+}
+
+func (cs *Coscheduling) getStateKey() framework.StateKey {
+	return framework.StateKey(fmt.Sprintf("Prefilter-%v", cs.Name()))
+}
+
+type noopStateData struct {
+}
+
+func NewNoopStateData() framework.StateData {
+	return &noopStateData{}
+}
+
+func (d *noopStateData) Clone() framework.StateData {
+	return d
 }
--- a/cmd/scheduler/main_test.go
+++ b/cmd/scheduler/main_test.go
@@ -204,6 +204,9 @@ profiles:
     filter:
       disabled:
       - name: "*"
+    postFilter:
+      enabled:
+      - name: Coscheduling
     preScore:
       disabled:
       - name: "*"
@@ -440,7 +443,7 @@ profiles:
 					"BindPlugin":       {{Name: "DefaultBinder"}},
 					"PreFilterPlugin":  {{Name: "Coscheduling"}},
 					"PostBindPlugin":   {{Name: "Coscheduling"}},
-					"PostFilterPlugin": {{Name: "DefaultPreemption"}},
+					"PostFilterPlugin": {{Name: "DefaultPreemption"}, {Name: "Coscheduling"}},
 					"QueueSortPlugin":  {{Name: "Coscheduling"}},
 					"ReservePlugin":    {{Name: "VolumeBinding"}, {Name: "Coscheduling"}},
 					"PermitPlugin":     {{Name: "Coscheduling"}},
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -48,6 +48,8 @@ type Manager interface {
 	GetPodGroup(*corev1.Pod) (string, *v1alpha1.PodGroup)
 	GetCreationTimestamp(*corev1.Pod, time.Time) time.Time
 	AddDeniedPodGroup(string)
+	DeletePermittedPodGroup(string)
+	CalculateAssignedPods(string, string) int
 }
 
 // PodGroupManager defines the scheduling operation called
@@ -157,7 +159,7 @@ func (pgMgr *PodGroupManager) Permit(ctx context.Context, pod *corev1.Pod, nodeN
 		return false, fmt.Errorf("PodGroup not found")
 	}
 
-	assigned := pgMgr.calculateAssignedPods(pg.Name, pg.Namespace)
+	assigned := pgMgr.CalculateAssignedPods(pg.Name, pg.Namespace)
 	// The number of pods that have been assigned nodes is calculated from the snapshot.
 	// The current pod in not included in the snapshot during the current scheduling cycle.
 	ready := int32(assigned)+1 >= pg.Spec.MinMember
@@ -226,6 +228,11 @@ func (pgMgr *PodGroupManager) AddDeniedPodGroup(pgFullName string) {
 	pgMgr.lastDeniedPG.Add(pgFullName, "", *pgMgr.lastDeniedPGExpirationTime)
 }
 
+// DeletePodGroup delete a podGroup that pass Pre-Filter but reach PostFilter.
+func (pgMgr *PodGroupManager) DeletePermittedPodGroup(pgFullName string) {
+	pgMgr.permittedPG.Delete(pgFullName)
+}
+
 // PatchPodGroup patches a podGroup.
 func (pgMgr *PodGroupManager) PatchPodGroup(pgName string, namespace string, patch []byte) error {
 	if len(patch) == 0 {
@@ -249,8 +256,8 @@ func (pgMgr *PodGroupManager) GetPodGroup(pod *corev1.Pod) (string, *v1alpha1.Po
 	return fmt.Sprintf("%v/%v", pod.Namespace, pgName), pg
 }
 
-// calculateAssignedPods returns the number of pods that has been assigned a node: assumed or bound.
-func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace string) int {
+// CalculateAssignedPods returns the number of pods that has been assigned a node: assumed or bound.
+func (pgMgr *PodGroupManager) CalculateAssignedPods(podGroupName, namespace string) int {
 	nodeInfos, err := pgMgr.snapshotSharedLister.NodeInfos().List()
 	if err != nil {
 		klog.Errorf("Cannot get nodeInfos from frameworkHandle: %v", err)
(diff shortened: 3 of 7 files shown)


cwdsuzhou:
/hold
should wait https://github.com/kubernetes-sigs/scheduler-plugins/pull/124 merged first

Huang-Wei:
@cwdsuzhou #124 has been merged, you can proceed to rebase here and let me know when this PR is reviewable.

mateuszlitwin:
Question: will this work with preemption enabled? AFAIK if the filter did not find nodes, then scheduler will try to preempt some lower priority pods to make a space for the pod group pod. In this implementation you reject all pod groups pods every time this happens. Ideally we would reject if preemption did not find victims?

I am worried that if there say 1000 pods and preemption is required for all of them, then scheduling will be super slow because we will have to attempt other 1000 pods every time there is a preemption.

cwdsuzhou:
> @cwdsuzhou #124 has been merged, you can proceed to rebase here and let me know when this PR is reviewable.

IMO PR is reviewable

cwdsuzhou:
> EDIT: After further inspection I think it will work, I forgot that default preemption is now a plugin which can run before this plugin, and if preemption succeeded then this plugin will not be called.
> 
> Question: will this work with preemption enabled? AFAIK if the filter did not find nodes, then scheduler will try to preempt some lower priority pods to make a space for the pod group pod. In this implementation you reject all pod groups pods every time this happens.
> 
> I am worried that if there say 1000 pods and preemption is required for all of them, then scheduling will be super slow because we will have to attempt other 1000 pods every time there is a preemption.
> 
> Ideally we would reject all pod group pods if preemption did not find any victims?

Yep!

cwdsuzhou:
ping

cwdsuzhou:
@Huang-Wei PTAL thanks

pkg/coscheduling/coscheduling.go
```diff
@@ -137,6 +138,30 @@ func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleSta
 	return framework.NewStatus(framework.Success, "")
 }
 
+// PostFilter is use for rejecting the group pods if a pod can not pass resource check when filter.
+func (cs *Coscheduling) PostFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod,
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	// check if failed due to preFilter failed, if not, filteredNodeStatusMap should not be nil
```
Huang-Wei, line 144: This comment doesn't look correct - PreFilter would return `Unschedulable` in most cases, then filteredNodeStatusMap is populated on each node:

https://github.com/kubernetes/kubernetes/blob/v1.19.5/pkg/scheduler/core/generic_scheduler.go#L273-L275

pkg/coscheduling/coscheduling.go
```diff
@@ -137,6 +138,30 @@ func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleSta
...
+	if filteredNodeStatusMap != nil {
```
Huang-Wei, line 145: Nil filteredNodeStatusMap occurs only when PreFilter returns `Error`:

https://github.com/kubernetes/kubernetes/blob/v1.19.5/pkg/scheduler/core/generic_scheduler.go#L264-L266

In other words, non-nil filteredNodeStatusMap is the common case, right?

pkg/coscheduling/coscheduling.go
```diff
@@ -137,6 +138,30 @@ func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleSta
 	return framework.NewStatus(framework.Success, "")
 }
 
+// PostFilter is use for rejecting the group pods if a pod can not pass resource check when filter.
```
Huang-Wei, line 141:
```suggestion
// PostFilter rejects a PodGroup's pods if <pod> is unschedulable
// after attempts of preemption.
// Note: this plugin is expected to work when at least one preemption plugin is enabled and called earlier.
```

pkg/coscheduling/coscheduling.go
```diff
@@ -137,6 +138,30 @@ func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleSta
...
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	// check if failed due to preFilter failed, if not, filteredNodeStatusMap should not be nil
+	if filteredNodeStatusMap != nil {
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Success, "")
+	}
+
+	pgName, pg := cs.pgMgr.GetPodGroup(pod)
+	if pg == nil {
+		klog.V(4).Info("Pod does not belong to any group")
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Success, "")
```
Huang-Wei, line 152: return `Unschedulable` b/c the logic is able to come down here is b/c the previous PostFitler plugins returned non-successful status, we should be consistent with those statuses.

pkg/coscheduling/coscheduling.go
cwdsuzhou: We would like skip Prefilter Error, only react to Filter Error.
If Prefilter returns Unschedulable, filteredNodeStatusMap is not nil, so we skip.

cwdsuzhou: Thanks, will change, filteredNodeStatusMap must not be nil except some special cases

pkg/coscheduling/coscheduling.go
cwdsuzhou: Sorry, have not understood very well.

Huang-Wei: I mean:
- suppose it's the 2nd PostFilter plugin: why the logic comes down here is b/c the 1st PostFilter failed (in terms of defaultpreemption, it cannot find any candidate), so returning `Success` actually indicate it's a successful running, which isn't the case, is it?
- more generally, suppose it's configured as the Nth PostFilter plugins, returning `Success` will prevent further PostFilter plugins from their running, which doesn't quite make sense.

pkg/coscheduling/coscheduling.go
```diff
@@ -137,6 +138,25 @@ func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleSta
...
 
+// PostFilter is use for rejecting the group pods if a pod does not pass PreFilter or Filter.
+func (cs *Coscheduling) PostFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod,
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	pgName, pg := cs.pgMgr.GetPodGroup(pod)
+	if pg == nil {
+		klog.V(4).Info("Pod does not belong to any group")
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Success, "")
+	}
+	cs.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
```
Huang-Wei, line 149: Now the logic is running unconditionally. Let's look at the possibles cases:

- current pod failed due to not meeting PG's `minAvailable`. In this case, there are no waiting pods, so the loop here is a no-op. Although may have some perf overhead, generally it's fine.
- current pod meets PG's `minAvailable` but failed due to other constraints. So in this case, IMO it's too aggressive to reject the other waiting Pods? Because per my understanding, this Pod may NOT be the bottleneck to block the whole PG? i.e., maybe another Pod later can be scheduled successfully and hence "approve()" the waiting pods. I'm open to discussion on this. @mateuszlitwin @denkensk @yuanchen8911 it'd be appreciable if you can share your thoughts.

pkg/coscheduling/coscheduling.go
cwdsuzhou: thanks  will change this pr

pkg/coscheduling/coscheduling.go
xq2005: @Huang-Wei From my understanding, this pod does not block the whole PG. But this PG may block following PGs.

pkg/coscheduling/coscheduling.go
```diff
@@ -215,3 +244,18 @@ func (cs *Coscheduling) rejectPod(uid types.UID) {
 	}
 	waitingPod.Reject(Name)
 }
+
+func (cs *Coscheduling) getStateKey(pod *v1.Pod) framework.StateKey {
+	return framework.StateKey(fmt.Sprintf("%v-%v", cs.Name(), core.GetNamespacedName(pod)))
```
Huang-Wei, line 249: CycleState is per pod, so you don't need to use Pod's name as the key. Usually we just use "PreFilter" + plugin name as the key.

(Ref: VolumeBinding's example: https://github.com/kubernetes/kubernetes/blob/42c00bc523c54ea72177af33a806e611850e44c9/pkg/scheduler/framework/plugins/volumebinding/volume_binding.go#L42)

pkg/coscheduling/coscheduling.go
```diff
@@ -132,11 +133,38 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
...
 }
 
+// PostFilter is use for rejecting the group pods if a pod does not pass PreFilter or Filter.
+func (cs *Coscheduling) PostFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod,
+	filteredNodeStatusMap framework.NodeToStatusMap) (*framework.PostFilterResult, *framework.Status) {
+	// check if failed when pre filter
+	_, err := state.Read(cs.getStateKey(pod))
+	if err == nil {
+		state.Delete(cs.getStateKey(pod))
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable, "")
```
Huang-Wei, line 149: `framework.NewStatus(framework.Unschedulable)`, or provide the specific error reason `framework.NewStatus(framework.Unschedulable, ...)`

pkg/coscheduling/coscheduling.go
```diff
@@ -132,11 +133,38 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
 func (cs *Coscheduling) PreFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod) *framework.Status {
```
Huang-Wei, line 133: Can you add a bullet comment saying that unsatisfied minResources would also return an PreFilter error?

pkg/coscheduling/coscheduling.go
Huang-Wei: @cwdsuzhou The current logic elimiated the first case I mentioned above. But the second case can still happen IMO. For example, the `minAvailable` is 3, however, there comes 5 Pods, so it doesn't sound like a good idea to reject all when the 3rd Pod failed the scheduling?

Huang-Wei: @xq2005 can you elaborate? are you implying we still should reject all Pods?

xq2005: @Huang-Wei Case 2 example:  pg1 and pg2 are waiting for scheduling, gp1 minAvailable is 4 and pg2 minAvailable is 2.  At this time there are 3 free pods resource in the cluster. pg1's 3 pods are waiting and pg2's pods are "Unschedulable". 

I think sometime this feature is useful, but it is **too aggressive** and may import resource starvation. We need more discussion about it.

cwdsuzhou: @xq2005 We mean we should not reject all when we have 5 pods, I need minMember 3 pods, cluster resource can meet 3 but can not meet 5. At this time, we should not reject the 5 pods.

The case you meant would not happen
1. we will check resource 
2. If any one failed of 4 pods failed when filter. The whole group would be rejected and resource will be released.

xq2005: @cwdsuzhou I used resource "pods" as an example, other resources may trigger the same behavior. My case wants to show new logic is useful.

Refer Wei's comment
> For example, the minAvailable is 3, however, there comes 5 Pods, so it doesn't sound like a good idea to reject all when the 3rd Pod failed the scheduling?

The 4th/5th pod may successful. It is too aggressive to reject pg when meeting the first failed pod in this pg.

cwdsuzhou:
@Huang-Wei @denkensk 

As we discussed offline,  we would keep similar with `Unreserve` now.

We can take actions to optimize both `PostFilter` and `Unreserve` in the future if necessary.

PTAL , thanks!

pkg/coscheduling/coscheduling.go
```diff
@@ -132,11 +133,45 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
...
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable, "can not find pod group")
+	}
+
+	assigned := cs.pgMgr.CalculateAssignedPods(pg.Name, pod.Namespace)
+	if assigned >= int(pg.Spec.MinMember) {
+		klog.Errorf("assigned %v", assigned)
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable)
+	}
+
+	cs.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
```
Huang-Wei, line 164: The updated logic is a bit conservative now.

However, I think it's based on an implicit assumption: if the n<sub>th</sub> Pod failed, it's inferrable other Pods belonging to the same PodGroup would be very likely to fail. Am I understanding it right?

If so, we should comment it so as not to confuse readers.

pkg/coscheduling/coscheduling.go
```diff
@@ -132,11 +133,45 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
...
+	}
+
+	pgName, pg := cs.pgMgr.GetPodGroup(pod)
+	if pg == nil {
+		klog.V(4).Info("Pod does not belong to any group")
+		return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable, "can not find pod group")
+	}
+
+	assigned := cs.pgMgr.CalculateAssignedPods(pg.Name, pod.Namespace)
+	if assigned >= int(pg.Spec.MinMember) {
```
Huang-Wei, line 159: nit: it'd be good to add a comment explaining:

```
This indicates there are already enough Pods satisfying the PodGroup,
so don't bother to reject the whole PodGroup.
```

pkg/coscheduling/coscheduling.go
```diff
@@ -132,11 +133,45 @@ func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
...
+
+	cs.frameworkHandler.IterateOverWaitingPods(func(waitingPod framework.WaitingPod) {
+		if waitingPod.GetPod().Namespace == pod.Namespace && waitingPod.GetPod().Labels[util.PodGroupLabel] == pg.Name {
+			klog.V(3).Infof("PostFilter rejects the pod: %v/%v", pgName, waitingPod.GetPod().Name)
+			waitingPod.Reject(cs.Name())
+		}
+	})
+	cs.pgMgr.AddDeniedPodGroup(pgName)
+	cs.pgMgr.DeletePermittedPodGroup(pgName)
+	return &framework.PostFilterResult{}, framework.NewStatus(framework.Unschedulable, "reject all of the whole group because pod failed when filter")
```
Huang-Wei, line 172: nit: reword the msg to "PodGroup %v gets rejected due to Pod %v is unschedulable even after PostFilter"

Huang-Wei (commented):
The current logic looks more conservative. LGTM except for some nits.

pkg/coscheduling/coscheduling.go
cwdsuzhou: > However, I think it's based on an implicit assumption: if the nth Pod failed, it's inferrable other Pods belonging to the same PodGroup would be very likely to fail. Am I understanding it right?

Yes, will add comments

pkg/coscheduling/coscheduling.go
```diff
@@ -215,3 +251,18 @@ func (cs *Coscheduling) rejectPod(uid types.UID) {
 	}
 	waitingPod.Reject(Name)
 }
+
+func (cs *Coscheduling) getStateKey() framework.StateKey {
+	return framework.StateKey(fmt.Sprintf("Prefilter-%v", cs.Name()))
+}
+
+type nilStateData struct {
```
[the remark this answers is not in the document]
cwdsuzhou: will change

cwdsuzhou:
@Huang-Wei  PR addressed, PTAL, thanks

Huang-Wei (commented):
/approve

Thanks @cwdsuzhou .

Will leave /lgtm to @denkensk .

cwdsuzhou:
ping @denkensk  for review

denkensk:
Sorry for late reply.
If we have about 1000 pods in a PodGroup and MinAvailable is 500，400 pods are waiting in Permit. The 401st pod failed in Filter. Then the 501st pod is scheduled successfully. Because scheduling big jobs takes long time. During this period, the total amount of resources will change. The 401st pod failed doesn't mean the latter pods also failed.
@Huang-Wei @cwdsuzhou

cwdsuzhou:
> Sorry for late reply.
> If we have about 1000 pods in a PodGroup and MinAvailable is 500，400 pods are waiting in Permit. The 401st pod failed in Filter. Then the 501st pod is scheduled successfully. Because scheduling big jobs takes long time. During this period, the total amount of resources will change. The 401st pod failed doesn't mean the latter pods also failed.
> @Huang-Wei @cwdsuzhou

@denkensk  Yes, you are right, but we can not ensure the 502nd pod successfully scheduled, right?  As we discussed offline, a better solution is waiting for a period of time. However, we can not find a good solution for this now.  If we wait for too long, we may cost some resources and lead to some pods starving.

To keep up with `Unreserve`, we would like reject the group of pods. If we find a better solution, we can change the both.

denkensk:
I'm not sure if I understand the problem scenarios.

> As we discussed offline, a better solution is waiting for a period of time. However, we can not find a good solution for this now.
 
The Waiting time in permit can be configured by users themselves now. Is it a solution for this? 

> If we wait for too long, we may cost some resources and lead to some pods starving.

It can be configured by user. If the user think the default time is too long. They can set the time to a smaller value.

cwdsuzhou:
> I'm not sure if I understand the problem scenarios.
> 
> > As we discussed offline, a better solution is waiting for a period of time. However, we can not find a good solution for this now.
> 
> The Waiting time in permit can be configured by users themselves now. Is it a solution for this?
> 
> > If we wait for too long, we may cost some resources and lead to some pods starving.
> 
> It can be configured by user. If the user think the default time is too long. They can set the time to a smaller value.

IMO, the Waiting time defined in PodGroup CRD should only be used when permitting pods, which means the pods have passed the `Filter`. If the resource is not enough which means a pod failed in `Filter`, I do not think we should wait for so long. e.g. 1000 pods, then 1st pod failed or the 1000th pod failed should not wait for a same period of time. But I can a good method now , so I would like reject all of the pod group.

Huang-Wei:
I see @denkensk 's point - in a high-churn cluster where run-to-completion pods paly a large part, the benefits of rejecting the whole PodGroup **may not** outweigh the cost of waiting for the resting pods to see whether they can satisfy the whole PodGroup.

So probably we can come up with a soft threshold here, based on which we decide to reject the PG or not:

```python
# If assigned >= minAvailable, keep the current logic

# If the gap is less than 10%, we may want to try subsequent Pods to see they can satisfy the PodGroup
if (minAvailable - assigned) / minAvailable < 10%:
    pass
else:
    reject the whole PodGroup
``` 

@cwdsuzhou @denkensk WDYT?

cwdsuzhou:
> I see @denkensk 's point - in a high-churn cluster where run-to-completion pods paly a large part, the benefits of rejecting the whole PodGroup **may not** outweigh the cost of waiting for the resting pods to see whether they can satisfy the whole PodGroup.
> 
> So probably we can come up with a soft threshold here, based on which we decide to reject the PG or not:
> 
> ```python
> # If assigned >= minAvailable, keep the current logic
> 
> # If the gap is less than 10%, we may want to try subsequent Pods to see they can satisfy the PodGroup
> if (minAvailable - assigned) / minAvailable < 10%:
>     pass
> else:
>     reject the whole PodGroup
> ```
> 
> @cwdsuzhou @denkensk WDYT?

Seems good. @denkensk WDYT?

denkensk:
Sounds good to me.
@cwdsuzhou

cwdsuzhou:
I have updated this PR based on  the suggestions!


PTAL @denkensk @Huang-Wei , thanks!

Issue #116, opened 2020-11-24
Autogen files are not updated on running `make autogen`


zorro786:
Using Go version (1.14 and 1.15) on OSX, I am unable to generate code gen files for types defined under `pkg/apis/config/types.go` and `pkg/apis/config/v1beta/types.go`. 

```
$: make autogen
hack/update-vendor.sh
hack/update-generated-openapi.sh
Generating Kubernetes OpenAPI
```

The command runs successfully but zz_generated* files are not updated.

Huang-Wei:
@zorro786 please run `hack/update-codegen.sh` and ensure scheduler-plugins code are put under $GOPATH.

zorro786:
Tried that too
```
➜  scheduler-plugins git:(real-load-plugin) ✗ hack/update-codegen.sh
Generating deepcopy funcs
Generating defaulters
Generating conversions
Generating deepcopy funcs
Generating clientset for scheduling:v1alpha1 at sigs.k8s.io/scheduler-plugins/pkg/generated/clientset
Generating listers for scheduling:v1alpha1 at sigs.k8s.io/scheduler-plugins/pkg/generated/listers
Generating informers for scheduling:v1alpha1 at sigs.k8s.io/scheduler-plugins/pkg/generated/informers
```
Both from $GOPATH directory and outside that. Doesn't generate any files.

AlexeyPerevalov:
hack/update-codegen.sh count on scheduler-plugins situated at path $HOME/go/src/sigs.k8s.io/scheduler-plugins,
where you GOPATH is $HOME/go
so if your scheduler-plugins repo outside of $GOPATH/src/sigs.k8s.io the place where generated files will be is $HOME/go/src/sigs.k8s.io/scheduler-plugins.
Are you expect it at you current directory e.g. $HOME/src/github/kubernetes-sigs/scheduler-plugins?

zorro786:
@AlexeyPerevalov Thank you for clarification. My directory was `$GOPATH/src/github.com/k8s.io/scheduler-plugins`. It will be great if the script takes care of copying over files from `$GOPATH/src/sigs.k8s.io/scheduler-plugins`, if generated outside GOPATH.

Huang-Wei:
[...]

Issue #127, opened 2020-12-07
how to install coscheduling?


jackielyc:
I don‘t find any related information in this git.

damemi:
Hi @jackielyc, to run the coscheduling plugin you'll need to compile it into a custom scheduler and deploy that scheduler in your cluster, then it will be available to enable just like any other plugin.

One option to do that is to just build this repo, which will include all the default (in-tree) scheduler plugins as well as any that are hosted here. Or, you can use this repo as an example to build your own custom scheduler and only import the plugins you need.

Hope that helps!

Huang-Wei:
Fixed by [doc/install.md](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/install.md).

PR #128, opened 2020-12-07
Additional documentation for development purpose

AlexeyPerevalov:
Issue: https://github.com/kubernetes-sigs/scheduler-plugins/issues/116

Signed-off-by: Alexey Perevalov <alexey.perevalov@huawei.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,8 @@ You can reach the maintainers of this project at:
 - [Slack](https://kubernetes.slack.com/messages/sig-scheduling)
 - [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-sig-scheduling)
 
+You can find an [instruction how to build and run out-of-tree plugin here](doc/develop.md) .
+
 ### Code of conduct
 
 Participation in the Kubernetes community is governed by the [Kubernetes Code of Conduct](code-of-conduct.md).
--- a/doc/develop.md
+++ b/doc/develop.md
@@ -1,0 +1,98 @@
+# Developer guide
+
+## How to build
+
+This section is about building on the local host, e.g. for debug purpose.
+Your `GOPATH` environment variable should be set.
+After cloning it into the `$GOPATH/src/sigs.k8s.io` directory
+you can build an image by the following command:
+```shell
+make local-image
+```
+After that you'll take localhost:5000/scheduler-plugins/kube-scheduler:latest and
+localhost:5000/scheduler-plugins/controller:latest images.
+
+Your source code has to be located at `$GOPATH/src/sigs.k8s.io`, if you would like to change some type definition of the interfaces
+and regenerate their implementation by running
+```shell
+hack/update-codegen.sh
+```
+command.
+If you would like to build just the binaries you can do it by
+```shell
+make
+```
+command.
+If you changed the branch or added new dependency you might need to regenerate vendor directory, for this you can use
+```shell
+make autogen
+```
+command.
+
+## How to debug
+By default debug information is stripped, to keep it in the binary you need to remove -w command line option from
+ldflags in the Makefile.
+You can run all unit tests by **make unit-test** command or if you would like to run exact unit test
+```shell
+GOFLAGS="-v" go test ./pkg/util/podgroup.go ./pkg/util/constants.go ./pkg/util/podgroup_test.go
+```
+-v in GOFLAGS was used for verbosity
+You can also run integration test locally, but etcd port 2379 on your system should be free.
+```shell
+make integration-test
+```
+
+## How to start
+If you would like to start produced kube-scheduler image you can use it in your static kube-scheduler manifests or any kind of
+deployment spec as following:
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: schedulingplugin
+  namespace: kube-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      component: scheduler
+      tier: control-plane
+  template:
+    metadata:
+      labels:
+        component: scheduler
+        tier: control-plane
+    spec:
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
+      containers:
+        - image: localhost:5000/scheduler-plugins/kube-scheduler:latest
+          imagePullPolicy: Never
+          command:
+          - /bin/kube-scheduler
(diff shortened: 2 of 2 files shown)


AlexeyPerevalov:
While working on documentation for contributors, I found some aspects:
1. Debug symbols. I just described how to "enable" it, but it worth adding make target for debug binaries, and then describe how to attach into running scheduler.
2. I didn't include deployment specification neither for scheduler, nor for controller, but probably it could be interesting as well, especially for issues such that: https://github.com/kubernetes-sigs/scheduler-plugins/issues/127
3. README.md already has section Install, I didn't touch it, from my point of view it is incomplete, it just describes how to obtain images, but it doesn't tell what to do with it after.
4. Makefile doesn't **help** target as in k/k/Makefile, with built-in command description.

Point #1 and #4 - it's some kind of tasks and requires additional issue and PR. And maybe discussion, whether it's needed or not.
The 2 and 3 point could be addressed in this PR.

README.md
```diff
@@ -41,6 +41,8 @@ You can reach the maintainers of this project at:
 - [Slack](https://kubernetes.slack.com/messages/sig-scheduling)
 - [Mailing List](https://groups.google.com/forum/#!forum/kubernetes-sig-scheduling)
 
+[Here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/develop.md) you can find an instruction how to build and run out-of-tree plugin.
```
damemi, line 44: more link text would be good, maybe change this to something like `You can find an [instruction how to build and run out-of-tree plugins here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/develop.md).`

doc/develop.md
```diff
@@ -0,0 +1,34 @@
...
+command
+If you would like to build just the binaries you can do it by
+```shell
+make
+```
+
+## How to debug
+By default debug information is stripped, to keep it in the binary you need to remove -w command line option from
+ldflags in the Makefile. 
+
```
damemi, line 29: nit: extra newline

doc/develop.md
```diff
@@ -0,0 +1,34 @@
+# Developer guid
```
damemi, line 1: typo: `Developer guide`

damemi (commented):
/ok-to-test

I think this is a great start for some of the frequent questions we see about how to use this repo. Maybe even add some more info about how this is an example custom scheduler built with the framework too?

damemi:
@AlexeyPerevalov I think your points on documenting how to actually run the scheduler are a good idea too. There seems to be a lot of confusion on how to "use" these plugins that would be helpful to point people to

AlexeyPerevalov:
> /ok-to-test
> 
> I think this is a great start for some of the frequent questions we see about how to use this repo. Maybe even add some more info about how this is an example custom scheduler built with the framework too?

Could you please clarify, do you mean to add some links to https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/  or maybe a  link to keps/sig-scheduling/624-scheduling-framework/README.md?

damemi:
@AlexeyPerevalov I was more referring to what you said in https://github.com/kubernetes-sigs/scheduler-plugins/pull/128#issuecomment-739748988 (I made my review comment before I saw your other comment, sorry). Those are also great ideas of things to link to though.

doc/develop.md
```diff
@@ -0,0 +1,33 @@
...
+```shell
+make
+```
+
+## How to debug
+By default debug information is stripped, to keep it in the binary you need to remove -w command line option from
+ldflags in the Makefile.
+
+## How to start
+If you would like to start produced kube-scheduler image you can use it in your static kube-scheduler manifests or any kind of
```
Huang-Wei, line 31: Missing some words here?

doc/develop.md
```diff
@@ -0,0 +1,33 @@
...
+```
+After that you'll take localhost:5000/scheduler-plugins/kube-scheduler:latest and
+localhost:5000/scheduler-plugins/controller:latest images.
+
+Your source code has to be located at `$GOPATH/src/sigs.k8s.io`, if you would like to change some type definition of the interfaces
+and regenerate their implementation by running
+```shell
+hack/update-codegen.sh
+```
+command
```
Huang-Wei, line 20:
```suggestion
command.
```

Huang-Wei:
> The 2 and 3 point could be addressed in this PR.

Given that it's a developer guide instead of a user guide, this doc should put more focus on (1) how to run/debug it locally, and (2) probably add a section to introduce how to test (unit/integration test) locally, which should be helpful for developers.

We can compose other docs for users later, to guide users on how to use/configure the binary.

seanmalloy (approved):

Huang-Wei:
> /assign @Huang-Wei

https://github.com/kubernetes-sigs/scheduler-plugins/pull/128#discussion_r539604107 is good to be resolved :)

doc/develop.md
```diff
@@ -0,0 +1,33 @@
+# Developer guide
+
+## How to build
+
+This section is about building on the local host, e.g. for debug purpose.
+Your `GOPATH` environment variable should be set.
```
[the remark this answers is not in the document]
seanmalloy: I way I read it this comment is resolved by https://github.com/kubernetes-sigs/scheduler-plugins/pull/128/files#diff-66e9efc4e747562f39d585f75eb79b480404a58111ccae618b8c98d09f36c3faR26-R30. But I could be wrong.

seanmalloy:
> > /assign @Huang-Wei
> 
> [#128 (comment)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/128#discussion_r539604107) is good to be resolved :)

The way I read it this comment is resolved by https://github.com/kubernetes-sigs/scheduler-plugins/pull/128/files#diff-66e9efc4e747562f39d585f75eb79b480404a58111ccae618b8c98d09f36c3faR26-R30. But I could be wrong.

AlexeyPerevalov:
> 
> 
> > /assign @Huang-Wei
> 
> [#128 (comment)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/128#discussion_r539604107) is good to be resolved :)

It was resolved on line 26 of develop.md

Huang-Wei:
> It was resolved on line 26 of develop.md

Oh, I misread.

/approve

Thanks @AlexeyPerevalov !

PR #134, opened 2020-12-16
Bugfix: cluster resource computation

cwdsuzhou:
Fix cluster resource computation. 

We should sub the pods that have been assumed (in waitingWap), which would have existed in node snapshot. When we Pre-filter a pod that belongs to the same PodGroup, we should simulate removing the waiting pods; otherwise res of those pods will be doulbe-calculated..

/assign @Huang-Wei

### DIFF
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -136,7 +136,7 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
 	minResources := pg.Spec.MinResources.DeepCopy()
 	podQuantity := resource.NewQuantity(int64(pg.Spec.MinMember), resource.DecimalSI)
 	minResources[corev1.ResourcePods] = *podQuantity
-	err = pgMgr.CheckClusterResource(nodes, minResources)
+	err = CheckClusterResource(nodes, minResources, pgFullName)
 	if err != nil {
 		klog.Errorf("PreFilter pod group %v error: %v", pgFullName, err)
 		pgMgr.AddDeniedPodGroup(pgFullName)
@@ -269,13 +269,15 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
 	return count
 }
 
-func (pgMgr *PodGroupManager) CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest corev1.ResourceList) error {
+// CheckClusterResource checks if resource capacity of the cluster can satisfy <resourceRequest>.
+// It returns an error detailing the resource gap if not satisfied; otherwise returns nil.
+func CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest corev1.ResourceList, desiredPodGroupName string) error {
 	for _, info := range nodeList {
 		if info == nil || info.Node() == nil {
 			continue
 		}
 
-		nodeResource := getNodeResource(info).ResourceList()
+		nodeResource := getNodeResource(info, desiredPodGroupName).ResourceList()
 		for name, quant := range resourceRequest {
 			quant.Sub(nodeResource[name])
 			if quant.Sign() <= 0 {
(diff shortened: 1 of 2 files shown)


Huang-Wei:
> We should sub the pod that has passed all Filters, which would exit in node snapshot.

I'm not sure I'm following this. Do you mean the pods in waitingMap? If so, those Pods have already been assumed, so in other words, they have been counted into the "allocated" portion of the snapshot.

cwdsuzhou:
> > We should sub the pod that has passed all Filters, which would exit in node snapshot.
> 
> I'm not sure I'm following this. Do you mean the pods in waitingMap? If so, those Pods have already been assumed, so in other words, they have been counted into the "allocated" portion of the snapshot.

Yes，so  when we compute the cluster resource we should sub the resource for the pods belong to the same group.

Huang-Wei:
ok, you mean that when checking MinResource in `PreFilter`, some Pods may have been assumed, but we're still checking the whole amount of MinResource - the waiting Pod may have been double-counted, right? If so, I'm fine with this PR, and it'd be appreciable if you can add a test to reproduce this bug.

cwdsuzhou:
> ok, you mean that when checking MinResource in `PreFilter`, some Pods may have been assumed, but we're still checking the whole amount of MinResource - the waiting Pod may have been double-counted, right? If so, I'm fine with this PR, and it'd be appreciable if you can add a test to reproduce this bug.

Sure. But I would like this can be merged as soon as possible since this is a bug. I prefer open another PR to add more integration tests when this PR and the post filter feature merged. WDYT？

pkg/coscheduling/core/core.go
```diff
@@ -269,13 +269,13 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
 	return count
 }
 
-func (pgMgr *PodGroupManager) CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest corev1.ResourceList) error {
+func (pgMgr *PodGroupManager) CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest corev1.ResourceList, desiredPodGroupName string) error {
```
Huang-Wei, line 272: Please add a comment for this function.

Off the topic: should we add the PG into permitted cache in L121?

Huang-Wei:
> But I would like this can be merged as soon as possible since this is a bug. I prefer open another PR to add more integration tests when this PR and the post filter feature merged.

It's exactly because it's a bug so that we need coverage :) I would expect a simple UT test to cover this.

cwdsuzhou:
> > But I would like this can be merged as soon as possible since this is a bug. I prefer open another PR to add more integration tests when this PR and the post filter feature merged.
> 
> It's exactly because it's a bug so that we need coverage :) I would expect a simple UT test to cover this.

Thanks, will add some UT

cwdsuzhou:
@Huang-Wei UT add, PTAL

pkg/coscheduling/core/core.go
cwdsuzhou: > Please add a comment for this function.
> 
> Off the topic: should we add the PG into permitted cache in L121?

IMO, add or not is ok , permitted cache is used for avoiding computing  resources repeatedly

pkg/coscheduling/core/core_test.go
```diff
@@ -231,6 +231,65 @@ func TestPermit(t *testing.T) {
...
+		corev1.ResourceMemory: *resource.NewQuantity(300, resource.DecimalSI),
+	}
+	node.Status.Allocatable = rl
+	node.Status.Capacity = rl
+	snapshot := testutil.NewFakeSharedLister(nil, []*corev1.Node{node})
+	nodeInfo, _ := snapshot.List()
+
+	pod := st.MakePod().Name("t1-p1-3").Req(map[corev1.ResourceName]string{corev1.ResourceMemory: "100"}).Label(util.PodGroupLabel,
+		"pg1-1").ZeroTerminationGracePeriod().Obj()
+	scheudledSnapshot := testutil.NewFakeSharedLister([]*corev1.Pod{pod}, []*corev1.Node{node})
```
Huang-Wei, line 247: scheudledSnapshot -> snapshotWithAssumedPod

pkg/coscheduling/core/core_test.go
```diff
@@ -231,6 +231,65 @@ func TestPermit(t *testing.T) {
...
 }
 
+func TestCheckClusterResource(t *testing.T) {
+	node := st.MakeNode().Name("fake-node").Obj()
+	rl := corev1.ResourceList{
+		corev1.ResourcePods:   *resource.NewQuantity(32, resource.DecimalSI),
+		corev1.ResourceMemory: *resource.NewQuantity(300, resource.DecimalSI),
+	}
+	node.Status.Allocatable = rl
+	node.Status.Capacity = rl
```
Huang-Wei, line 241: Can be simplfied to:

```go
nodeRes := map[corev1.ResourceName]string{corev1.ResourceMemory: "300"}
node := st.MakeNode().Name("fake-node").Capacity(nodeRes).Obj()
```

pkg/coscheduling/core/core.go
```diff
@@ -269,13 +269,14 @@ func (pgMgr *PodGroupManager) calculateAssignedPods(podGroupName, namespace stri
 	return count
 }
 
-func (pgMgr *PodGroupManager) CheckClusterResource(nodeList []*framework.NodeInfo, resourceRequest corev1.ResourceList) error {
+// CheckClusterResource check the cluster left resources, will return resource gap if left resources can not satisfy the request
```
Huang-Wei, line 272:
```suggestion
// CheckClusterResource checks if resource capacity of the cluster can satisfy <resourceRequest>.
// It returns an error detailing the resource gap if not satisfied; otherwise returns nil.
```

Huang-Wei (commented):
Some nits. LGTM otherwise.

cwdsuzhou:
> Some nits. LGTM otherwise.

PR addressed @Huang-Wei

Huang-Wei:
/lgtm
/approve

Thanks!

PR #143, opened 2021-01-29
Topology aware scheduler per numa

AlexeyPerevalov:
This pull request implements KEP https://github.com/kubernetes-sigs/scheduler-plugins/pull/119

It implements idea collected in this document: https://docs.google.com/presentation/d/1SR6XSIFsHkiWTws66LABpTiZaRwYk8IoRts4HQWE8Bc/edit#slide=id.g752682a7d2_68_129

Mostly focused on approach where some simplified version of TopologyManager is filtering pods at scheduler time.
Initially was proposed as built-in plugin (implementation https://github.com/kubernetes/kubernetes/pull/90708 design described in KEP(kubernetes/enhancements#1858)

/kind feature


swatisehgal:
Associated Enhancement Proposal: [Here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/119).

swatisehgal:
@Huang-Wei Could you please take a look at this PR and the associated implementation [here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/143). We are now focusing on out-of-tree enablement of Topology-aware Scheduling and the CRD API is [http://github.com/k8stopologyawareschedwg/noderesourcetopology-api](http://github.com/k8stopologyawareschedwg/noderesourcetopology-api). Thanks in advance!

pkg/noderesourcetopology/match.go
```diff
@@ -0,0 +1,358 @@
+/*
```
Huang-Wei, line 1: Can be renamed to filter.go

pkg/noderesourcetopology/match.go
```diff
@@ -0,0 +1,358 @@
...
+	if nodeInfo.Node() == nil {
+		return framework.NewStatus(framework.Error, fmt.Sprintf("Node is nil %s", nodeInfo.Node().Name))
+	}
+	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort {
+		return nil
+	}
+
+	nodeName := nodeInfo.Node().Name
+
+	topologyPolicies := getTopologyPolicies(tm.nodeTopologies, nodeName)
```
Huang-Wei, line 222: If there is no particular reason (e.g., indexing), it's unnecessary to  maintain a local copy of `nodeTopologies`. Instead, inject a xyzInformerFactory().xyzInfomer.Lister() to the plugin instance, and then use xyzLister's methods to fetch the xyz objects when needed. Then, all the onXYZAdd/Update/Delete methods can also be removed.

pkg/noderesourcetopology/match.go
```diff
@@ -0,0 +1,358 @@
...
+	for _, policyName := range topologyPolicies {
+		if handler, ok := tm.topologyPolicyHandlers[policyName]; ok {
```
Huang-Wei, line 224: Given that the two objects are actually identical, and there is no struct fields associated with each, I'd suggest just create 2 stateless PolicyFilter() methods  (e.g., `PodLevelFilter()` and `SingleNUMANodeFilter()`)to handle them individually.

Huang-Wei (commented):
First round of review. (will review core Filter logic and tests after most of the comments get resolved)

pkg/apis/config/register.go
```diff
@@ -41,6 +41,8 @@ func addKnownTypes(scheme *runtime.Scheme) error {
 		&NodeResourcesAllocatableArgs{},
 		&CapacitySchedulingArgs{},
 		&TargetLoadPackingArgs{},
+		&NodeResourcesAllocatableArgs{},
```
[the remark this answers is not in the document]
AlexeyPerevalov: done

pkg/apis/config/types.go
```diff
@@ -41,11 +41,18 @@ type CoschedulingArgs struct {
...
+type TopologyManagerPolicy string
+
 const (
 	// Least is the string "Least".
 	Least ModeType = "Least"
 	// Most is the string "Most".
 	Most ModeType = "Most"
+
+	// to preserve consistency keep it in pkg/apis/core/types.go"
+	SingleNUMANodeTopologyManagerPolicy TopologyManagerPolicy = "SingleNUMANode"
```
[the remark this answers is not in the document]
AlexeyPerevalov: these constants (all of type TopologyManagerPolicy) represent behavior of the resource management component on the node (now it's TopologyManager, but it could be some external resource manager). This patch series adds support only for SingleNUMANode (it's single-numa-node policy of the TopologyManager) and PodTopologyScope (it's pod level resource counting in TopologyManager). In CRD it's array of string.  So here could be a string too or we can keep TopologyManagerPolicy type in noderesourcetopology-api.

Regarding name of the type, since it represent behavior/policy of the node's resource management based on hardware topology. I think names like TopologyPolicy/TopologyManagerPolicy/ResourceTopologyPolicy show the purpose.

pkg/noderesourcetopology/match.go
AlexeyPerevalov: nodeTopologies are populating in Add/Update callback, it's not a poll model, it's push model, when plugin getting informed about node topology state modification (e.g. after new pod get launched)

pkg/noderesourcetopology/match.go
```diff
@@ -0,0 +1,358 @@
...
+const (
+	// NodeResourceTopologyMatchName is the name of the plugin used in the plugin registry and configurations.
+	NodeResourceTopologyMatchName = "NodeResourceTopologyMatch"
+)
+
+var _ framework.FilterPlugin = &NodeResourceTopologyMatch{}
+
+type nodeTopologyMap map[string]topologyv1alpha1.NodeResourceTopology
+
+type PolicyHandler interface {
```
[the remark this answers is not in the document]
AlexeyPerevalov: Now we have 2 implementation, they are not similar, one is working per all resources of all containers and makes decision for whole pod, another one makes decision per container. And quite different implementation may exist in future, even not related to NUMA, since we may also support https://github.com/intel/cri-resource-manager (by using the same noderesourcetopology-api, it's now flexible enough to represent any kind of hardwares)

Since topologyPolicy is an attribute of the worker node, the idea behind this is to choose appropriate handler for node's policy or skip node if policy is not found (e.g. wasn't provided) for node.
It could be done simpler e.g. with _if_ clause in case of one Filter plugin per all policies we will have. Or one Filter plugin per each policy, but in this case each implementation of Filter plugin will do unnecessary job like skipping nodes if node configured for another policy.

pkg/noderesourcetopology/match.go
AlexeyPerevalov: Do you suggest to do it with map[string]interface{} or just compare policyName and call appropriate handler?

pkg/noderesourcetopology/README.md
```diff
@@ -0,0 +1,224 @@
...
+
+Let us assume we have two nodes in a cluster deployed with sample-device-plugin with the hardware topology described by the diagram below:
+
+![Setup](numa-topology.png)
+
+The hardware topology corresponding to both the nodes is represented by the below CRD instances. These CRD instances are supposed to be created by Node Agents like [Resource Topology Exporter](https://github.com/k8stopologyawareschedwg/resource-topology-exporter) (RTE) or Node feature Discovery (NFD). Please refer to issue [Exposing Hardware Topology through CRDs in NFD](https://github.com/kubernetes-sigs/node-feature-discovery/issues/333) and [Design document](https://docs.google.com/document/d/1Q-4wSu1tzmbOXyGk_2r5_mK6JdXXJA-bOd3cAtBFnwo/edit?ts=5f24171f#) which captures details of enhancing NFD to expose node resource topology through CRDs.
+
+
+ ```yaml
+ # Master Node CRD spec
```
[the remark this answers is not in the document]
AlexeyPerevalov: it's not mandatory, here could be any node. Yes it might look strange since by default kubeadm installation  the master is not available for scheduling. And usually  on typical deployment master note is a VM, and definitely not a bare-metal.

pkg/noderesourcetopology/match.go
```diff
@@ -0,0 +1,358 @@
...
+	topologyPolicyHandlers PolicyHandlerMap
+}
+
+type SingleNUMANodeHandler struct {
+	match *NodeResourceTopologyMatch
+}
+
+type PodLevelResourceHandler struct {
+	match *NodeResourceTopologyMatch
+}
```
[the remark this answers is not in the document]
AlexeyPerevalov: agree, now I reworked this part and removed match * field, but 2 structures still necessary because of 2 different behavior, we discussed it in another comments

pkg/noderesourcetopology/match.go
```diff
@@ -0,0 +1,358 @@
...
+func NewNodeResourceTopologyMatch(args runtime.Object, handle framework.FrameworkHandle) (framework.Plugin, error) {
+	klog.V(5).Infof("creating new NodeResourceTopologyMatch plugin")
+	tcfg, ok := args.(*apiconfig.NodeResourceTopologyMatchArgs)
+	if !ok {
+		return nil, fmt.Errorf("want args to be of type NodeResourceTopologyMatchArgs, got %T", args)
+	}
+
+	topologyMatch := &NodeResourceTopologyMatch{}
+
+	kubeConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
```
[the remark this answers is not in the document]
AlexeyPerevalov: ok, I just didn't think about incluster config, yes we can use clientcmd.BuildConfigFromFlags(), it's even better, thank you.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,348 @@
...
+	"sync"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog/v2"
+	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
+	bm "k8s.io/kubernetes/pkg/kubelet/cm/topologymanager/bitmask"
```
AlexeyPerevalov, line 32: I know about https://docs.google.com/document/d/1WO-ixERpqkCSEXEq30YtEH_z_G-BoLKeCbkRJcKq3xA/edit
in this context using kubernetes/pkg/kubelet not so perfect from maintainability point of view.

pkg/apis/config/types.go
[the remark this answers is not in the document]
Huang-Wei: It sounds to be we should name them as `SingleNUMANodeGeneral` and `SingleNUMANodePodTopology`?

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,348 @@
...
+	nodeTopologyInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc:    topologyMatch.onTopologyAdd,
+			UpdateFunc: topologyMatch.onTopologyUpdate,
+			DeleteFunc: topologyMatch.onTopologyFromDelete,
+		},
+	)
+
+	ctx := context.Background()
+	go nodeTopologyInformer.Informer().Run(ctx.Done())
```
Huang-Wei, line 340: I think L342 actually does the same thing, so this line can be removed.

pkg/noderesourcetopology/match.go
Huang-Wei: Oh, you may misunderstand `xyzLister` which is providerd by Kubernetes client-go SDK. It actually maintains a local cache for you, and xyzLister's methods fetch the xyz object(s) from the client cache in an O(1) manner. It's not polling objects from API server.

In other words, no matter what state of your CR gets changed, the cache (xyzStore) that backs xyzLister is (almost) up-to-date.

IIUC, we can just rely on the xyzLister, because we just look at the up-to-date state of particular xyz objects upon a scheduling cycle (the Filter() hook) - rather than needs to be notified and proactively trigger the scheduling of a Pod.

pkg/noderesourcetopology/match.go
Huang-Wei: We can define a function type:

```go
type PolicyFilter func(*v1.Pod, topologyv1alpha1.ZoneList) *framework.Status
```

And then register each function implementation in the map: `map[apiconfig.TopologyManagerPolicy]PolicyFilter`.

pkg/noderesourcetopology/match.go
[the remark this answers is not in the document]
Huang-Wei: Commented in https://github.com/kubernetes-sigs/scheduler-plugins/pull/143#discussion_r570570966

pkg/apis/config/types.go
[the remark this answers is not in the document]
AlexeyPerevalov: I agree, it worth changing it now, since current names have some historical aspects )

AlexeyPerevalov:
Thank you, I didn't know about cache in Lister, now I see it's a ThreadSafeMap.
The List method returns []*NodeResourceTopology, here need to iterate to find appropriate.
Also it's possible to do it by another way:
`lister.NodeResourceTopology(Namespace).Get(hostName)`
and map keeps with key as "namespace/hostName", so we will provide namespace in config file (AllNamespace doesn't work, since constant is "", and it concatenates to "/hostName",  the storage doesn't know about wildcards )

Huang-Wei:
> Also it's possible to do it by another way:
> `lister.NodeResourceTopology(Namespace).Get(hostName)`
> and map keeps with key as "namespace/hostName"

My understanding is that `NodeResourceTopology` is similar to `Node`, and hence cluster-scoped. However, it's defined as namespaced:

https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/blob/fd566f16210b3eb0e66e1c4f110ad8660b4027eb/manifests/crd.yaml#L62

So does it make more sense to make it a cluster-scoped CRD?

AlexeyPerevalov:
> 
> 
> > Also it's possible to do it by another way:
> > `lister.NodeResourceTopology(Namespace).Get(hostName)`
> > and map keeps with key as "namespace/hostName"
> 
> My understanding is that `NodeResourceTopology` is similar to `Node`, and hence cluster-scoped. However, it's defined as namespaced:
> 
> https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/blob/fd566f16210b3eb0e66e1c4f110ad8660b4027eb/manifests/crd.yaml#L62
> 
> So does it make more sense to make it a cluster-scoped CRD?

Namespace feature is necessary for product purpose, e.g.  to isolate tenants, it's better to keep.

pkg/apis/config/types.go
```diff
@@ -90,3 +101,14 @@ type TargetLoadPackingArgs struct {
...
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// NodeResourceTopologyMatchArgs holds arguments used to configure the NodeResourceTopologyMatch plugin
+type NodeResourceTopologyMatchArgs struct {
+	metav1.TypeMeta
+
+	KubeConfigPath string
+	MasterOverride string
+	Namespace      string
```
Huang-Wei, line 113: Hmm... Why a plugin config is designed to be restricted to one namespace?

pkg/noderesourcetopology/README.md
```diff
@@ -0,0 +1,237 @@
...
+- schedulerName: topo-aware-scheduler
+  plugins:
+    filter:
+      enabled:
+      - name: NodeResourceTopologyMatch
+# optional plugin configs
+  pluginConfig:
+  - name: NodeResourceTopologyMatch
+    args:
+      kubeconfigpath: "/etc/kubernetes/scheduler.conf"
```
Huang-Wei, line 43: make it adhere to the latest code, e.g., namespace (if it's needed).

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.CycleState, pod *v1.Pod, nodeInfo *framework.NodeInfo) *framework.Status {
+	if nodeInfo.Node() == nil {
+		return framework.NewStatus(framework.Error, fmt.Sprintf("Node is nil %s", nodeInfo.Node().Name))
+	}
+	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort {
+		return nil
+	}
+
+	nodeName := nodeInfo.Node().Name
+	nodeTopology, err := tm.lister.NodeResourceTopologies(tm.namespace).Get(nodeName)
```
Huang-Wei, line 205: Again, configuring a namespace in plugin config sounds strange.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+
+	if err != nil {
+		klog.Errorf("Cannot get NodeTopologies from cache: %v", err)
```
Huang-Wei, line 208: "cache" is vague:

```suggestion
		klog.Errorf("Cannot get NodeTopologies from NodeResourceTopologyNamespaceLister: %v", err)
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil
+	}
+
+	klog.V(5).Infof("nodeTopology: %v", nodeTopology)
+	topologyPolicies := getTopologyPolicies(nodeTopology, nodeName)
+	for _, policyName := range topologyPolicies {
```
Huang-Wei, line 214: I think you can direclty use `nodeTopology.TopologyPolicies`? (and cast to the custome string type when searching)

```suggestion
	for _, policyName := range nodeTopology.TopologyPolicies {
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+type NUMANodeList []NUMANode
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (tm *TopologyMatch) Name() string {
+	return Name
+}
+
+// getTopologyPolicies return true if we're working with such policy
+func getTopologyPolicies(nodeTopology *topologyv1alpha1.NodeResourceTopology, nodeName string) []apiconfig.TopologyManagerPolicy {
+	policies := make([]apiconfig.TopologyManagerPolicy, 0)
```
Huang-Wei, line 74: It's conventional in Go to init a slice using:

```suggestion
	var policies[]apiconfig.TopologyManagerPolicy
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+
+type NUMANodeList []NUMANode
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (tm *TopologyMatch) Name() string {
+	return Name
+}
+
+// getTopologyPolicies return true if we're working with such policy
+func getTopologyPolicies(nodeTopology *topologyv1alpha1.NodeResourceTopology, nodeName string) []apiconfig.TopologyManagerPolicy {
```
Huang-Wei, line 73: `nodeName` doesn't seem needed.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+			var numaID int
+			fmt.Sscanf(zone.Name, "node-%d", &numaID)
+			resources := extractResources(zone)
+			nodes = append(nodes, NUMANode{NUMAID: numaID, Resources: resources})
+		}
+	}
+	return nodes
+}
+
+func getNodeTopology(nodes []*topologyv1alpha1.NodeResourceTopology, nodeName string) *topologyv1alpha1.NodeResourceTopology {
```
Huang-Wei, line 186: This function is not used.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+		res[v1.ResourceName(resInfo.Name)] = quantity
+	}
+	return res
+}
+
+func SingleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList) *framework.Status {
+	klog.V(5).Infof("Single NUMA node handler")
+	containers := []v1.Container{}
+	containers = append(pod.Spec.InitContainers, pod.Spec.Containers...)
```
Huang-Wei, line 97: You can directly iterate `append(pod.Spec.InitContainers, pod.Spec.Containers...)` in the foo loop below:

```go
for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
    ...
}
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+	return nil
+
+}
+
+func createNUMANodeList(zones topologyv1alpha1.ZoneList) NUMANodeList {
+	nodes := make(NUMANodeList, 0)
+	for _, zone := range zones {
+		if zone.Type == "Node" {
+			var numaID int
+			fmt.Sscanf(zone.Name, "node-%d", &numaID)
```
Huang-Wei, line 178: Hold on a sec... Are we trying to do scanf here?

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+	for _, policy := range nodeTopology.TopologyPolicies {
+		policies = append(policies, apiconfig.TopologyManagerPolicy(policy))
+	}
+	return policies
+}
+
+func extractResources(zone topologyv1alpha1.Zone) v1.ResourceList {
+	res := make(v1.ResourceList)
+	for _, resInfo := range zone.Resources {
+		quantity, err := resource.ParseQuantity(resInfo.Allocatable.String())
```
Huang-Wei, line 84: I guess you'd like to panic if err is not nil, so probably:

```go
quantity := resource.MustParse()
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+	// prepare NUMANodes list from zoneMap
+	nodes := createNUMANodeList(zones)
+	qos := v1qos.GetPodQOS(pod)
+	for _, container := range containers {
+		bitmask := bm.NewEmptyBitMask()
+		bitmask.Fill()
+
+		checkResourcesForNUMANodes(bitmask, nodes, container.Resources.Requests, qos)
+		if bitmask.IsEmpty() {
+			// definitly we can't align container, so we can't align a pod
```
Huang-Wei, line 108: s/definitly/definitely/

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("Cannot align container: %s", container.Name))
+		}
+	}
+	return nil
+}
+
+func checkResourcesForNUMANodes(bitmask bm.BitMask, nodes NUMANodeList, resources v1.ResourceList, qos v1.PodQOSClass) {
```
Huang-Wei, line 115: I may not be the right person validating the logic here, @swatisehgal could you help review the correctness and ensure it's tested properly? Thanks.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+			}
+		}
+		bitmask.And(resourceBitmask)
+	}
+}
+
+func SingleNUMAPodLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList) *framework.Status {
+	klog.V(5).Infof("Pod Level Resource handler")
+	containers := []v1.Container{}
+	containers = append(pod.Spec.InitContainers, pod.Spec.Containers...)
```
Huang-Wei, line 147: Ditto, you can iterate `append(pod.Spec.InitContainers, pod.Spec.Containers...)` directly in the for loop below.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+
+	resources := make(v1.ResourceList)
+
+	for _, container := range containers {
+		for resource, quantity := range container.Resources.Requests {
+			if quan, ok := resources[resource]; ok {
```
Huang-Wei, line 153: nit: quan -> q is fine.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,263 @@
...
+		}
+	}
+
+	nodes := createNUMANodeList(zones)
+	bitmask := bm.NewEmptyBitMask()
+	bitmask.Fill()
+	checkResourcesForNUMANodes(bitmask, nodes, resources, v1qos.GetPodQOS(pod))
+
+	if bitmask.IsEmpty() {
+		// definitly we can't align container, so we can't align a pod
```
Huang-Wei, line 166: definitly -> definitely

pkg/noderesourcetopology/filter.go
Huang-Wei: It may be more readable to return a `bm.BitMask`, and remove `bitmask bm.BitMask` in the input args, and also we don't need to init the bitmask on the caller side.

pkg/apis/config/types.go
AlexeyPerevalov: Maybe the limitation of only one namespace is not so flexible. I'll provide a list of namespaces this plugin can work with.

manifests/coscheduling/scheduler-config.yaml
```diff
@@ -31,4 +31,4 @@ profiles:
       permitWaitingTimeSeconds: 10
       deniedPGExpirationTimeSeconds: 3
       kubeConfigPath: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
-      kubeMaster: "REPLACE_ME_WIHT_KUBE_MASTER"
+      kubeMaster: "REPLACE_ME_WITH_KUBE_MASTER"
```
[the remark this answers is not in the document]
AlexeyPerevalov: WIHT -> WITH
it's from 33dca048 commit of this PR, but probably it should be in separate PR,  @swatisehgal what do you think?

pkg/apis/config/types.go
```diff
@@ -41,11 +41,22 @@ type CoschedulingArgs struct {
...
 	// Least is the string "Least".
 	Least ModeType = "Least"
 	// Most is the string "Most".
 	Most ModeType = "Most"
+
+	// Constants of type TopologyManagerPolicy represent policy of the worker
+	// node's resource management component. It's TopologyManager in kubele.
+	// SingleNUMANodeContainerLevel represent single-numa-node policy of
+	// the TopologyManager
+	SingleNUMANodeContainerLevel TopologyManagerPolicy = "SingleNUMANodeContainerLevel"
```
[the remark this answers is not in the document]
AlexeyPerevalov: ok, it will be placed in k8stopologyawareschedwg/noderesourcetopology-api

pkg/noderesourcetopology/filter.go
AlexeyPerevalov: yes, not so golangish, to be a more natural for golang it's better to return instead of using *out* argument.

Huang-Wei:
@AlexeyPerevalov I noticed some comments haven't been resolved, so let me know it's ready for another round reivew.

pkg/noderesourcetopology/filter.go
AlexeyPerevalov: yes, because of generalization, the Zone which represents NUMA keeps numaId in the Name field. Worth checking for parsing failures here.

pkg/noderesourcetopology/filter.go
AlexeyPerevalov: no, I think it's better just to skip resource if it was specified incorrectly. As I know kube-apiserver returns BadRequest if yaml is not valid (resource name/value are also checked there)

swatisehgal: I think it should be handled the same way errors in the resource name are handled by the default scheduler i.e. api server does the validation. Lets double check in the scheduler code to see what happens in case the scheduler fails to parse the resources. Does it simply log and continues or panics and we follow the same in our plugin. Will look into the scheduler code to figure out the behavior.

manifests/coscheduling/scheduler-config.yaml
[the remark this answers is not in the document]
swatisehgal: Yep, it was typo I had found and just addressed here. If you both think it is better to be addressed in a separate PR, I can do that. No problem.

pkg/noderesourcetopology/filter.go
swatisehgal: Discussed with @AlexeyPerevalov, since kube-apiserver doesn't apply pod with incorrect resource quantity we won't run into a situation where since kube-scheduler (default) gets it from kube-apiserver which means that in the scheduler plugin the resources in question have already been validated by API server and here too we won't run into a situation where we have invalid resourceNames. But wondering if there could be other reasons that the resource parsing could return an error?

AlexeyPerevalov: in case of NodeResourceTopology CRD both capacity and allocatable have a int-or-string type, one can put there incorrect quantity representation, like "10Gi10", I'm not sure it's possible to represent Quantity object in CRD to rely on verification in generated code, for apiserver it's just a string too. I don't think scheduler is a right place for verification.

pkg/noderesourcetopology/README.md
```diff
@@ -0,0 +1,242 @@
...
+Noderesourcetopology plugin works with namespaces, in this case each CRD could be namespace specific. The default namespace is used if it was omitted in the plugin's configuration.
+
+```yaml
+# Worker Node A CRD spec
+apiVersion: topology.node.k8s.io/v1alpha1
+kind: NodeResourceTopology
+metadata:
+  name: worker-node-A
+  namespace: test-namespace
+topologyPolicies: ["SingleNUMANode"]
```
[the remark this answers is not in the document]
swatisehgal: Yep, will submit a commit shortly with that fix.

manifests/noderesourcetopology/scheduler-configmap.yaml
```diff
@@ -0,0 +1,24 @@
...
+        - schedulerName: topo-aware-scheduler
+          plugins:
+            filter:
+              enabled:
+                - name: NodeResourceTopologyMatch
+          # optional plugin configs
+          pluginConfig:
+          - name: NodeResourceTopologyMatch
+            args:
+              kubeconfigpath: "/etc/kubernetes/scheduler.conf"
```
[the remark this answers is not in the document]
swatisehgal: This config map is already in kube-system namespace in the example.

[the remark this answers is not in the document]
AlexeyPerevalov: If namespace is not specified, the "default" namespace is using.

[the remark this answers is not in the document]
swatisehgal: Added test-namespace in the config

swatisehgal:
@Huang-Wei Thank you so much for all the feedback on this PR so far! We have addressed all the review comments and suggestions and are ready for another round of review.

pkg/noderesourcetopology/README.md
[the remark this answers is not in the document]
Huang-Wei: If I didn't miss anything, it's still outstanding.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,254 @@
...
+			// memory will not be provided in CRD
+			// 2. set numa node as possible node if resource is cpu and it's not guaranteed QoS, since cpu will flow
+			// 3. set numa node as possible node if zero quantity for non existing resource was requested (TODO check topology manaager behaviour)
+			// 4. otherwise check amount of resources
+			if resource == v1.ResourceMemory ||
+				strings.HasPrefix(string(resource), string(v1.ResourceHugePagesPrefix)) ||
+				resource == v1.ResourceCPU && qos != v1.PodQOSGuaranteed ||
+				quantity.Cmp(zeroQuantity) == 0 ||
+				numaQuantity.Cmp(quantity) >= 0 {
+				resourceBitmask.Add(numaNode.NUMAID)
```
Huang-Wei, line 124: IIUC, you can directly use the top-level `bitmask` to manipulate, i.e., eliminate `resourceBitmask`.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,254 @@
...
+			// if can't find requested resource on the node - skip (don't set it as available NUMA node)
+			// if unfound resource has 0 quantity probably this numa node can be considered
+			if !ok && quantity.Cmp(zeroQuantity) != 0 {
+				continue
+			}
+			// Check for the following:
+			// 1. set numa node as possible node if resource is memory or Hugepages (until memory manager will not be merged and
+			// memory will not be provided in CRD
+			// 2. set numa node as possible node if resource is cpu and it's not guaranteed QoS, since cpu will flow
+			// 3. set numa node as possible node if zero quantity for non existing resource was requested (TODO check topology manaager behaviour)
```
Huang-Wei, line 117:
```suggestion
			// 3. set numa node as possible node if zero quantity for non existing resource was requested (TODO check topology manager behaviour)
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return nil
+}
+
+func createNUMANodeList(zones topologyv1alpha1.ZoneList) NUMANodeList {
+	nodes := make(NUMANodeList, 0)
+	for _, zone := range zones {
+		if zone.Type == "Node" {
+			var numaID int
+			_, err := fmt.Sscanf(zone.Name, "node-%d", &numaID)
```
Huang-Wei, line 160: Again, `Sscanf`?

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,254 @@
...
+				continue
+			}
+			// Check for the following:
+			// 1. set numa node as possible node if resource is memory or Hugepages (until memory manager will not be merged and
+			// memory will not be provided in CRD
+			// 2. set numa node as possible node if resource is cpu and it's not guaranteed QoS, since cpu will flow
+			// 3. set numa node as possible node if zero quantity for non existing resource was requested (TODO check topology manaager behaviour)
+			// 4. otherwise check amount of resources
+			if resource == v1.ResourceMemory ||
+				strings.HasPrefix(string(resource), string(v1.ResourceHugePagesPrefix)) ||
```
Huang-Wei, line 120:
```suggestion
				strings.HasPrefix(string(resource), v1.ResourceHugePagesPrefix) ||
```

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,254 @@
...
+		bitmask.And(resourceBitmask)
+	}
+	return bitmask
+}
+
+func SingleNUMAPodLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList) *framework.Status {
+	klog.V(5).Infof("Pod Level Resource handler")
+	resources := make(v1.ResourceList)
+
+	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
```
Huang-Wei, line 136: For PodLevel handling, I'm not quite sure it's correct to simply aggregate InitContainers and InitContainers. The relation in Kubernetes Resource Model to evaluate a pod's request is:

```
resourceRequest = max(sum(podSpec.Containers), podSpec.InitContainers) + overHead
```

So probably we should adapt to KSM accordingly.

Huang-Wei (commented):
Thanks @AlexeyPerevalov and @swatisehgal. Some comments below.

BTW: we also need an integration test.

pkg/noderesourcetopology/filter.go
AlexeyPerevalov: probably yes, but current implementation of TM
https://github.com/kubernetes/kubernetes/blob/9657dd77a3563ffc7dbd061d51f34f714b1f8d5c/pkg/kubelet/cm/topologymanager/scope_pod.go#L57
takes it into account, for both container level and pod level. It's hard for me to image where it could be useful, I mean NUMA/CPU affinity for InitContainer. We'll discuss it with TM authors.

pkg/noderesourcetopology/filter.go
AlexeyPerevalov: It could be done in this way too, e.g. by bitmask.Remove

AlexeyPerevalov:
> 
> 
> Thanks @AlexeyPerevalov and @swatisehgal. Some comments below.
> 
> BTW: we also need an integration test.

We think about it now, but we have only WIP document.
https://github.com/k8stopologyawareschedwg/documentation/pull/2/files

pkg/noderesourcetopology/filter.go
Huang-Wei: The previous version is more readable, sorry for the confusion - I misread `Add` and `And`, so thought `resourceBitmask` was unneeded. Please revert to the previous implementation, and:

- add some comments around `resourceBitmask` like: "for each requested resource, calculate which NUMA slots are good fits, and then AND with the aggregated bitmask".
- we can change signature of `checkResourcesForNUMANodes` to return a boolean value, so that to return early when bitmask is empty - after every iteration it ANDs with `resourceBitmask`. WDYT?

pkg/noderesourcetopology/filter.go
Huang-Wei: SG, let's keep it consistent with TM for now. Good to leave a comment here.

Huang-Wei:
@AlexeyPerevalov in terms of integration test, I meant to create noderesourcetopology-api CR objects, and then test integration with the plugin implementation here. I think this's the last mile to get this PR merged. You can refer to `test/integration/coscheduling_test.go`, which also depends on some custom resource objects.

AlexeyPerevalov:
/hold
waiting for integration test

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,729 @@
...
+	for _, tt := range []struct {
+		name                   string
+		pods                   []*v1.Pod
+		nodeResourceTopologies []*topologyv1alpha1.NodeResourceTopology
+		expectedNodes          []string
+	}{
+		{
+			name: "Filtering out nodes that cannot fit resources on a single numa node in case of Guaranteed pod",
+			pods: []*v1.Pod{
+				withContainer(st.MakePod().Namespace(ns.Name).Name("topology-aware-scheduler-pod").Req(map[v1.ResourceName]string{v1.ResourceCPU: "4", v1.ResourceMemory: "5Gi"}).Obj(), pause, withResList("4", "50Gi"), withResList("4", "50Gi")),
```
AlexeyPerevalov, line 224: One thing I don't like in this code: we need to specify resources twice because Req() just fills Requests, but not Limits.

I added PR for it https://github.com/kubernetes/kubernetes/pull/100658
of course we can introduce our own local helper.

swatisehgal: Sure, but unfortunately one more thing we can't do even with https://github.com/kubernetes/kubernetes/pull/100658 is to be able to define pods of different QoS classes. With your addition of Limit in the PodWrapper we would only able to create Guaranteed pods and still have to create burstable/BE pods using the local helper.  It might worth investing some time to improve the pod wrapper to be able to define pods with various QoS with inputs of request and limit. Adding a similar comment in the PR as well.

pkg/noderesourcetopology/filter.go
Huang-Wei: Bump. Please add some comments for `resourceBitmask`, and rename `checkResourcesForNUMANodes` to `resMatchNUMANodes(...) bool`.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -0,0 +1,461 @@
...
+			if !reflect.DeepEqual(gotStatus, test.wantStatus) {
+				t.Errorf("status does not match: %v, want: %v\n", gotStatus, test.wantStatus)
+			}
+		})
+	}
+}
+
+func TestTopologyRequestsCorrectNodeResourceTopology(t *testing.T) {
+	nodes := nodeTopologyMap{}
+	nodes["default/node1"] = topologyv1alpha1.NodeResourceTopology{
```
Huang-Wei, line 147: It seems only the cpu portions are different, so could you dedupe the code below?

pkg/noderesourcetopology/filter_test.go
```diff
@@ -0,0 +1,461 @@
...
+					}, topologyv1alpha1.ResourceInfo{
+						Name:        nicResourceName,
+						Capacity:    intstr.Parse("30"),
+						Allocatable: intstr.Parse("2"),
+					},
+				},
+			},
+		},
+	}
+	node1Resources := makeResourceListFromZones(nodes["default/node1"].Zones)
```
Huang-Wei, line 278: Ditto. Can you dedupe the code here?

pkg/noderesourcetopology/filter_test.go
```diff
@@ -0,0 +1,461 @@
...
+	node3Resources := makeResourceListFromZones(nodes["default/node3"].Zones)
+
+	node3 := v1.Node{ObjectMeta: metav1.ObjectMeta{Name: "node3"}, Status: v1.NodeStatus{
+		Capacity:    node3Resources,
+		Allocatable: node3Resources,
+	},
+	}
+
+	// Test different QoS Guaranteed/Burstable/BestEffort
+	topologyTests := topologyMatchTests{
```
Huang-Wei, line 303: It's more conventional to make a single struct `topologyMatchTest` and build the slice here `topologyTests := []topologyMatchTest`

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+		KubeConfigPath: kubeConfigPath,
+		Namespaces:     []string{ns.Name},
+	}
+
+	profile := schedapi.KubeSchedulerProfile{
+		SchedulerName: v1.DefaultSchedulerName,
+		Plugins: &schedapi.Plugins{
+			PreBind: &schedapi.PluginSet{
+				Disabled: []schedapi.Plugin{
+					{Name: "*"},
```
Huang-Wei, line 134: This disabling seems irrelevant (also applies to Filter, PreFilter, Reserve).

(correct me if I'm wrong)

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+		testCtx,
+		true,
+		scheduler.WithProfiles(profile),
+		scheduler.WithFrameworkOutOfTreeRegistry(registry),
+	)
+	t.Log("init scheduler success")
+	defer testutils.CleanupTest(t, testCtx)
+
+	// Create a Node.
+	nodeName1 := "fake-node-1"
```
Huang-Wei, line 175: Can you dedupe the code of initing node1 and node2?

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+	node1 := st.MakeNode().Name("fake-node-1").Label("node", nodeName1).Obj()
+	node1.Status.Allocatable = v1.ResourceList{
```
Huang-Wei, line 177: you can simplify this by `.Capacity(xyz)`:

```go
 st.MakeNode().Name("fake-node-1").Label("node", nodeName1).Capacity(xyz).Obj()
```

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+}
+
+func withContainer(pod *v1.Pod, image string) *v1.Pod {
+	pod.Spec.Containers[0].Name = "con0"
+	pod.Spec.Containers[0].Image = image
+	return pod
+}
+
+// Req adds a new container to the inner pod with given resource map.
+func Req(p *st.PodWrapper, resMap map[v1.ResourceName]string) *st.PodWrapper {
```
Huang-Wei, line 621: to make the names consistent, `withReqAndLimit()` might be more descriptive as you set both reqs and limits.

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+									Name:        "cpu",
+									Allocatable: intstr.FromString("2"),
+									Capacity:    intstr.FromString("2"),
+								},
+							},
+						},
+					},
+				},
+			},
+			//both nodes are possible
```
Huang-Wei, line 368: Prefer to remove this as this doesn't add in extra info.

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+						tt.pods[i].Name, tt.expectedNodes, nodeName)
+				}
+
+			}
+			t.Logf("case %v finished", tt.name)
+		})
+	}
+}
+
+func contains(s []string, e string) bool {
```
Huang-Wei, line 489: Shouldn't we use `strings.Contains()`?

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+	pod, err := c.CoreV1().Pods(podNamespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		// This could be a connection error so we want to retry.
+		klog.Errorf("klog error %v", err)
+		return "", err
+	}
+	if pod.Spec.NodeName == "" {
+		return "", fmt.Errorf("Node not assigned")
+	}
+	return pod.Spec.NodeName, nil
```
Huang-Wei, line 509: Just simply do:

```go
if err != nil {
    return "", err
}
return pod.Spec.NodeName, nil
```

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+			for i := range tt.pods {
+				t.Logf("Creating Pod %q", tt.pods[i].Name)
+				_, err := cs.CoreV1().Pods(ns.Name).Create(context.TODO(), tt.pods[i], metav1.CreateOptions{})
+				if err != nil {
+					t.Fatalf("Failed to create Pod %q: %v", tt.pods[i].Name, err)
+				}
+			}
+
+			for i, p := range tt.pods {
+
```
Huang-Wei, line 459: remove the empty line.

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,636 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+			//both nodes are possible
+			expectedNodes: []string{"fake-node-1", "fake-node-2"},
+		},
+		//TODO: Add more test cases
```
Huang-Wei, line 437: Remove this.

test/integration/noderesourcetopology_test.go
AlexeyPerevalov: w/o disabling PreBind VolumeBinding's PreBing plugin is getting called, probably it's due to VolumeBinding in algorithmprovider's registry and it goes to defaultPlugins.
I rechecked, Reserve & PreFilter now plugins don't fail, maybe add volume to our test, in this case VolumeBinding probably should pass.

test/integration/noderesourcetopology_test.go
swatisehgal: Strings.Contains() is to check is a substring is within a string., whereas here we are checking for presence of a string in a slice of strings.

test/integration/noderesourcetopology_test.go
Huang-Wei: I'd expect this plugin to co-exist with default in-tree plugins, otherwise, users would also run into problems as this integration test. So could you update the test to remove all irrelevant disabling?

test/integration/noderesourcetopology_test.go
Huang-Wei: sry, I misread.

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,612 @@
...
+			// Create Pods
+			for i := range tt.pods {
+				t.Logf("Creating Pod %q", tt.pods[i].Name)
+				_, err := cs.CoreV1().Pods(ns.Name).Create(context.TODO(), tt.pods[i], metav1.CreateOptions{})
+				if err != nil {
+					t.Fatalf("Failed to create Pod %q: %v", tt.pods[i].Name, err)
+				}
+			}
+
+			for i, p := range tt.pods {
```
Huang-Wei, line 440: Perfer `for _, p := range tt.pods` and use `p` only in the loop.

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,612 @@
...
+				t.Logf(" p scheduled: %v", p.Name)
+
+				// The other pods should be scheduled on the small nodes.
+				nodeName, err := getNodeName(cs, ns.Name, p.Name)
+				if err != nil {
+					t.Log(err)
+				}
+				if contains(tt.expectedNodes, nodeName) {
+					t.Logf("Pod %q is on a nodes as expected.", p.Name)
+					continue
```
Huang-Wei, line 458: This is the last piece of a single loop, so no need to `continue`.

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,612 @@
...
+	}
+}
+
+func withContainer(pod *v1.Pod, image string) *v1.Pod {
+	pod.Spec.Containers[0].Name = "con0"
+	pod.Spec.Containers[0].Image = image
+	return pod
+}
+
+// Req adds a new container to the inner pod with given resource map.
```
Huang-Wei, line 596:
```suggestion
// withReqAndLimit adds a new container to the inner pod with given resource map.
```

Huang-Wei (commented):
In addition to the comments below, some previous comments have been folded, like https://github.com/kubernetes-sigs/scheduler-plugins/pull/143/files#r604520697, please unfold them and resolve accordingly.

BTW: the unit test can be refactored using fake utilities: https://gist.github.com/Huang-Wei/9852f8a53d47fc683295a0097f5dfd51

pkg/noderesourcetopology/filter.go
Huang-Wei: Bump again.

pkg/noderesourcetopology/filter.go
```diff
@@ -0,0 +1,265 @@
...
+				strings.HasPrefix(string(resource), v1.ResourceHugePagesPrefix) ||
+				resource == v1.ResourceCPU && qos != v1.PodQOSGuaranteed ||
+				quantity.Cmp(zeroQuantity) == 0 ||
+				numaQuantity.Cmp(quantity) >= 0 {
+				// possible to align resources on NUMA node
+				resourceBitmask.Add(numaNode.NUMAID)
+			}
+		}
+		bitmask.And(resourceBitmask)
+	}
```
Huang-Wei, line 136: Add a check above to do early return if possible:

```go
if bitmask.IsEmpty() {
    return false
}
```

test/integration/noderesourcetopology_test.go
Huang-Wei: I don't think it's related to PreBind of VolumeBinding. The conflict seems to hide in some `Filter` plugin.

test/integration/noderesourcetopology_test.go
Huang-Wei: Bump.

Huang-Wei (commented):
Some final comments. Disabling all other plugins doesn't look good still.

test/integration/noderesourcetopology_test.go
Huang-Wei: ok, it's because the memory was built with plain "100", you should use "100Gi".

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,611 @@
...
+			PreFilter: &schedapi.PluginSet{
+				Disabled: []schedapi.Plugin{
+					{Name: "*"},
+				},
+			},
+			Reserve: &schedapi.PluginSet{
+				Disabled: []schedapi.Plugin{
+					{Name: "*"},
+				},
+			},
```
Huang-Wei, line 154: Just enable plugin noderesourcetopology and remove other pieces:

```go
			Filter: &schedapi.PluginSet{
				Enabled: []schedapi.Plugin{
					{Name: noderesourcetopology.Name},
				},
			},
```

test/integration/noderesourcetopology_test.go
```diff
@@ -0,0 +1,611 @@
...
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+
+	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
```
Huang-Wei, line 30: nit: move it to L48.

Huang-Wei (commented):
Some final comments to resolve the integration test issue.

test/integration/noderesourcetopology_test.go
```diff
@@ -172,27 +158,30 @@ func TestTopologyMatchPlugin(t *testing.T) {
...
+		v1.ResourceMemory: "100Gi",
 		v1.ResourcePods:   "32",
 	}
-	node1 := st.MakeNode().Name("fake-node-1").Label("node", nodeName1).Capacity(resList).Obj()
-	n1, err := cs.CoreV1().Nodes().Create(ctx, node1, metav1.CreateOptions{})
-	if err != nil {
-		t.Fatalf("Failed to create Node %q: %v", nodeName1, err)
-	}
+	for _, node := range []struct {
+		name string
```
Huang-Wei, line 167: A nit: as "name" is the only field, we can simply do:

```go
for _, nodeName := []string{"fake-node-1", "fake-node-2"} { 
```

Huang-Wei (commented):
One final nit. LGTM otherwise.

You can address that along with squashing the commits.
Two or three commits are fine - one carries autogen files/docs, the other carries code implemenation.

Huang-Wei:
/lgtm
/approve

Thanks @AlexeyPerevalov and @swatisehgal !

PR #148, opened 2021-02-20
Fix PodGroup CRD definition has no status field schema

penglei:


damemi (commented):
I would think this CRD should be generated from https://github.com/kubernetes-sigs/scheduler-plugins/blob/79882332af9f2635236c8992a439711b4a0cf279/pkg/apis/scheduling/v1alpha1/types.go#L116. Maybe this PR could add those generators + verify scripts to check them too?

Huang-Wei:
> Maybe this PR could add those generators + verify scripts to check them too?

Sounds like a great idea.

@penglei would you like to extend this PR to introduce an automatic script to verify/generatee the missing pieces in the CRD definition yaml?

penglei:
> > Maybe this PR could add those generators + verify scripts to check them too?
> 
> Sounds like a great idea.
> 
> @penglei would you like to extend this PR to introduce an automatic script to verify/generatee the missing pieces in the CRD definition yaml?

Yes, that's a great idea, I'll do the follow up.

I'm not familiar with related technical solutions. Do you have any suggestions on how to verify the missing pieces? (I think we can do a comparison between the newly generated yaml file and the old existence.)

Huang-Wei:
@penglei I guess some tools can be leveraged, e.g., https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen.

A generic approach should be like this:

1. Use a tool to generate a CRD.yaml (e.g., `controller-gen crd <args>`)
2. Diff the CRD.yaml with the in-place yaml (e.g., `diff <output of 1> manifests/coscheduling/crd.yaml`)
3. Return the return code of above `diff` command, also output the diff is they're different.

It's very high-level, some logic may be adjusted. Or, even some repo may already have this verification script.

Huang-Wei:
But I think we can do that in a follow-up PR. @damemi @denkensk WDYT?

cwdsuzhou:
Agree with @Huang-Wei , do this in another PR may be better

denkensk:
+1 
It's common requirements for all CRDs. We can do it in another pr.
And In this PR, Manually add missing fields is ok. @Huang-Wei @damemi

damemi:
Yeah, another PR is fine by me :) Just wanted to suggest that we should make sure these are generated/verified. I agree with @Huang-Wei's approach with [controller-gen](https://github.com/kubernetes-sigs/controller-tools/tree/master/cmd/controller-gen)

Huang-Wei:
/lgtm
/approve

Created a follow-up issue #151.

Issue #161, opened 2021-03-11
Release 0.19.x


seanmalloy:
It's been a while since the last release. The master branch has been updated to be compatible with k8s v1.19. I'm proposing that a scheduler-plugins v0.19.x release be created.

Here is my rough checklist of items to complete for this effort.

- [x] update end user install docs #150 
- [x] update end user install docs for new ARM64 container images
- [x] do the [release process](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/release-guide.md) to create branches and tag

seanmalloy:
@Huang-Wei and @denkensk does this sound reasonable? Am I missing any critical items that must be finished before another release is created?

I can submit a PR soon-ish to update the end user docs with the ARM64 container image details. @xq2005 would you be willing to handle the doc updates to resolve #150?

Huang-Wei:
@seanmalloy Thanks for driving this. The process looks pretty decent.

I just cut a `release-1.19` release and created a tag `v0.19.8`. I suppose the tag push would automatically create a staging image, right? But seems nothing gets generated at gcr.io/k8s-staging-scheduler-plugins/kube-scheduler. Did I miss something?

seanmalloy:
> @seanmalloy Thanks for driving this. The process looks pretty decent.
> 
> I just cut a `release-1.19` release and created a tag `v0.19.8`. I suppose the tag push would automatically create a staging image, right? But seems nothing gets generated at gcr.io/k8s-staging-scheduler-plugins/kube-scheduler. Did I miss something?

@Huang-Wei I'll check it tomorrow.

seanmalloy:
I see the kube-scheduler images is in the staging repo:
<img width="425" alt="Screen Shot 2021-03-11 at 9 13 08 AM" src="https://user-images.githubusercontent.com/685221/110808949-17989b00-824a-11eb-8303-741456478a6b.png">

https://console.cloud.google.com/gcr/images/k8s-staging-scheduler-plugins/GLOBAL/kube-scheduler?gcrImageListsize=30


I see the controller images is in the staging repo:
<img width="369" alt="Screen Shot 2021-03-11 at 9 14 53 AM" src="https://user-images.githubusercontent.com/685221/110809158-43b41c00-824a-11eb-8ac2-5475532a373b.png">


https://console.cloud.google.com/gcr/images/k8s-staging-scheduler-plugins/GLOBAL/controller?gcrImageListsize=30

[...]

PR #163, opened 2021-03-16
Update install doc to match release v0.19.8

### LINKED ISSUE #150: Getting error when trying to enable coscheduler following develop instructions and tutorial
Moved from https://github.com/kubernetes/website/issues/21128,

Using minikube, I compiled the scheduler-plugins image (according to [these instructions](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/doc/develop.md)), and set the image in my `/etc/kubernetes/manifests/kube-scheduler.yaml` file of my master node. I.e:
```
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
  labels:
    component: kube-scheduler
    tier: control-plane
  name: kube-scheduler
  namespace: kube-system
spec:
  containers:
  - command:
    - kube-scheduler
    - --authentication-kubeconfig=/etc/kubernetes/scheduler.conf
    - --authorization-kubeconfig=/etc/kubernetes/scheduler.conf
    - --config=/etc/kubernetes/coscheduling-config.yaml
    - --bind-address=127.0.0.1
    - --kubeconfig=/etc/kubernetes/scheduler.conf
    - --leader-elect=false
    - --port=0
    image: localhost:5000/scheduler-plugins/kube-scheduler:latest   <---- The compiled custom scheduler
    imagePullPolicy: IfNotPresent
    livenessProbe:
      failureThreshold: 8
 .....
```
Which I thought would create the scheduler pod with the custom scheduler image but I get the error message:
```
W0219 17:10:50.073294       1 client_config.go:613] error creating inClusterConfig, falling back to default config: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
couldn't create scheduler using provider "DefaultProvider": initializing profiles: creating profile for scheduler name default-scheduler: error initializing plugin "Coscheduling": failed to init rest.Config: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
```

@Huang-Wei

Huang-Wei:
Part of #161.

Updated the installation doc to match release v0.19.8.

Fixes #150.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -3,15 +3,16 @@
 ## Table of Contents
 
 - [Create a Kubernetes Cluster](#create-a-kubernetes-cluster)
-- [Install release v0.18.9 and use Coscheduling](#install-release-v0189-and-use-coscheduling)
+- [Install release v0.19.8 and use Coscheduling](#install-release-v0198-and-use-coscheduling)
   - [Test Coscheduling](#test-coscheduling)
+- [Instal old-version releases](#install-old-version-releases)
 - [Uninstall Scheduler-plugins](#uninstall-scheduler-plugins)
 
 ## Create a Kubernetes Cluster
 
 Firstly you need to have a Kubernetes cluster, and a `kubectl` command-line tool must be configured to communicate with the cluster.
 
-The Kubernetes version must equal to or greater than **v1.18.0**. To check the version, use `kubectl version --short`.
+The Kubernetes version must equal to or greater than **v1.19.0**. To check the version, use `kubectl version --short`.
 
 If you do not have a cluster yet, create one by using one of the following provision tools:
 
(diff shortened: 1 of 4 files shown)


Huang-Wei:
cc @seanmalloy @xq2005 

/assign @cwdsuzhou @denkensk

doc/install.md
```diff
@@ -3,200 +3,283 @@
...
+      namespace: default
+      resourceVersion: "135603"
+      selfLink: /apis/scheduling.sigs.k8s.io/v1alpha1/namespaces/default/podgroups/pg1
+      uid: b4ac3562-54ab-4c1e-89bb-541a81c6acce
+    spec:
+      minMember: 3
+      scheduleTimeoutSeconds: 10
+    ```
+    
+    << UNSOLVED >>: It's expected to get `status` in the above yaml. Needs debug.
```
Huang-Wei, line 248: @cwdsuzhou @denkensk could you help investigate why the status is not displayed? Better to compare the RBAC/CRD with your production settings.

manifests/install/all-in-one.yaml
```diff
@@ -0,0 +1,83 @@
+# First part
```
Huang-Wei, line 1: @cwdsuzhou @denkensk could you help compare the RBAC/CRD with your production settings? to ensure the correctness and the privileges are not over-exposed. Thanks in advance.

manifests/install/all-in-one.yaml
```diff
@@ -0,0 +1,83 @@
...
+# Apply extra privileges to system:kube-scheduler.
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: system:kube-scheduler:plugins
+rules:
+- apiGroups: ["scheduling.sigs.k8s.io"]
+  resources: ["podgroups", "elasticquotas"]
+  verbs: ["get", "list", "watch", "create", "delete", "patch"]
```
Huang-Wei, line 10: I missed `patch` here; otherwise scheduler reports error:

```
E0317 01:20:22.700494       1 core.go:203] podgroups.scheduling.sigs.k8s.io "pg1" is forbidden: User "system:kube-scheduler" cannot patch resource "podgroups" in API group "schedul
ing.sigs.k8s.io" in the namespace "default"
```

doc/install.md
Huang-Wei: After appending the 2nd commit, a PodGroup with matching Pods can display its `status`, but still empty for a PodGroup without any matching Pods.

manifests/install/all-in-one.yaml
cwdsuzhou: yep, we miss patch here

manifests/install/all-in-one.yaml
```diff
@@ -0,0 +1,83 @@
...
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: scheduler-plugins-controller
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["scheduling.sigs.k8s.io"]
+  resources: ["podgroups", "podgroups/status", "elasticquotas", "elasticquotas/status"]
```
cwdsuzhou, line 47: we have not enabled `status` subresource of podgroup, so `podgroups/status` is not necessary

Huang-Wei: thanks. Could you help verify that when a PodGroup is created (without any matching pods), is its status displayed properly?

cwdsuzhou: > And if I create a deployment with 2 pods, the `phase` gets changed to `PreScheduling`, but if I delete the deployment, its phase stays in `PreScheduling` (I'd expect it to be updated back to `Pending`)

This may be a bug the status would not be back to `Pending`

manifests/install/all-in-one.yaml
cwdsuzhou: Hi, coscheduling works ok in my cluster using the the RBAC(changed in this PR) and CRD

image
  

Since pod group controller is not necessary now, we did not deploy in our production environment.

cwdsuzhou (approved):

manifests/install/all-in-one.yaml
Huang-Wei: > This may be a bug the status would not be back to Pending

I will create an issue tracking this.

BTW: could you help confirm this issue:

> after I updated controller SA's rbac setting (replace Update with Patch), it shows properly. But it's not showing the number:

cwdsuzhou: In my test environment, as I showed https://github.com/kubernetes-sigs/scheduler-plugins/pull/163#discussion_r595711417,  we can see number.

The controller is deployed based on the manifests in this PR

doc/install.md
```diff
@@ -3,200 +3,286 @@
...
+> The following steps are based on a Kubernetes cluster created by Kind.
+
+1. Log into the control plane node 
+   
+    ```bash
+    sudo docker exec -it $(sudo docker ps | grep control-plane | awk '{print $1}') bash
+    ```
   
-2. Backup `kube-scheduler.yaml`
+1. Backup `kube-scheduler.yaml`
```
cwdsuzhou, line 38: seem the sequence is wrong. It is always `1.`

Huang-Wei: It's intentional, it will be rendered properly.

manifests/install/all-in-one.yaml
Huang-Wei: Created #166 to track the UX issues.

Huang-Wei:
Commits squashed and some UX issues are tracked in #166. Please take another look.

/hold cancel

Huang-Wei:
@cwdsuzhou would you minding re-lgtm?

doc/install.md
```diff
@@ -3,200 +3,286 @@
...
+    ```
+    
+> ⚠ NOTE: There are some UX issues need to be addressed in controller side -
+> [#166](https://github.com/kubernetes-sigs/scheduler-plugins/issues/166).
+
+## Install release v0.18.9 (old version) and use Coscheduling
+
+If you're running at v0.18.9, which doesn't depend on PodGroup CRD, you should refer to the
+[install doc](https://github.com/kubernetes-sigs/scheduler-plugins/blob/release-1.19/doc/install.md) in
+branch `release-1.19`.
```
seanmalloy, line 257:
```suggestion
[install doc](https://github.com/kubernetes-sigs/scheduler-plugins/blob/release-1.18/doc/install.md) in
branch `release-1.18`.
```

Huang-Wei: Thanks for the catch. I put this paragraph into a section `Instal lower-version releases`.

seanmalloy:
/lgtm
/hold

Adding a hold so that @cwdsuzhou can review one more time before merge. Please feel free to remove the hold once you are ready for this to merge. Thanks!

Huang-Wei:
Kindly ping @@cwdsuzhou for taking another look.

doc/install.md
```diff
@@ -3,200 +3,287 @@
...
+      scheduleTimeoutSeconds: 10
+      minMember: 3
+    ```
+
+    ```bash
+    $ kubectl apply -f podgroup.yaml
+    ```
+
+1. Create a deploy labelled `pod-group.scheduling.sigs.k8s.io: pg1` to associated with PodGroup
+   `pg1` created in the previous step.
```
cwdsuzhou, line 173: always get `1.`, seems not intentional

cwdsuzhou:
minor hits.  Otherwise, lgtm

doc/install.md
Huang-Wei: When the list is long and frequently modified, it's not easy to always make its numbers ordered accurately :) So keep all as `1.` and make them indented well, markdown will do the rest things.

PR #164, opened 2021-03-18
go sum cleanup

Huang-Wei:
Running `make autogen` to shave off go.sum.

/kind cleanup
/assign @seanmalloy


seanmalloy:
Using Go 1.15.8 I ran the below commands to verify that everything is working as expected.
```
make build.amd64
go mod tidy
git diff # no changes
```

/lgtm

Issue #166, opened 2021-03-19
controller should reconcile the status of PodGroup


Huang-Wei:
In the latest code, the controller doesn't quite reconcile the status of PodGroup:

- [x] Field `status.Scheduled` not displayed - fixed by #166

    If a PodGroup is created, we'd expect all [status fields](https://github.com/kubernetes-sigs/scheduler-plugins/blob/e7ca5ee3ba91f98af11c176f7fec2e2ab9e9cc0c/pkg/apis/scheduling/v1alpha1/types.go#L155-L174) to display properly. But for now it's only `status.phase`.
    
    ```
    $ k get pg pg1 -o yaml 
    ...
    status:
      phase: pending
    ```

- [ ] `status.phase` not reconciled well

    If a PodGroup doesn't have any associated Pods, it should stay in `Pending` state. However, it sometimes stays in `Running` (or other state). We's expect status to be always reconciled.

/kind bug

Huang-Wei:
@cwdsuzhou @denkensk could you take some time to look into the above 2 issues?

cwdsuzhou:
I will open a PR to resolve these issues

cwdsuzhou:
@Huang-Wei  

to  resolve the `1st` issue, we just need remove `omitempty` of these fields.

about the 2nd issue, I have different opinions. If a group has been `Running` , it should not go back to `Pending` again. IMO, we should covert it to `Success`, `Failed` or `Unknown`. If no associated Pods exist, `Unknown` would be better. For other status, conversion should be similar.

WDYT?

Huang-Wei:
> to resolve the `1st` issue, we just need remove `omitempty` of these fields.

Thanks. Please raise a PR fixing that.

> IMO, we should covert it to `Success`, `Failed` or `Unknown`. If no associated Pods exist, `Unknown` would be better. For other status, conversion should be similar.

It sounds strange to use different states to show whether a PodGroup has been scheduled successfully **once** or not. Shouldn't the state just stick to its actual state, in a "stateless" manner?

cwdsuzhou:
> > to resolve the `1st` issue, we just need remove `omitempty` of these fields.
> 
> Thanks. Please raise a PR fixing that.

This does not work. I raise a PR, just add default value to status

cwdsuzhou:
[...]

PR #170, opened 2021-03-31
Add default value to pg status

cwdsuzhou:
Fix https://github.com/kubernetes-sigs/scheduler-plugins/issues/166


Huang-Wei:
Verified it works.

/lgtm
/approve

PR #172, opened 2021-04-01
Bump To Go 1.15.10

seanmalloy:
The k/k repo for 1.19 and 1.20 have been udpated to use Go 1.15.10. See
below PR for reference.

https://github.com/kubernetes/kubernetes/pull/100520

PR #183, opened 2021-04-08
Doc: update doc to fix minor typo

chloejiwon:
Replace `instal` -> `install`

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -5,7 +5,7 @@
 - [Create a Kubernetes Cluster](#create-a-kubernetes-cluster)
 - [Install release v0.19.8 and use Coscheduling](#install-release-v0198-and-use-coscheduling)
   - [Test Coscheduling](#test-coscheduling)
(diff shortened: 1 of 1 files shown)


chloejiwon:
I signed it

seanmalloy (approved):

seanmalloy:
Looks like the integration tests are sometimes timing out. To me it feels like this started happening after the bump to k/k 1.20.

Huang-Wei:
> Looks like the integration tests are sometimes timing out. 

Some coscheduling tests intentionally spawn the Pods interleavingly, and hence cost extra cycles to complete the tests. I think this is the part that takes the most time of our integration tests. This issue cannot be fully resolved without upstream framework changes.

> To me it feels like this started happening after the bump to k/k 1.20.

Without changes newly introduced along with 1.20 bumping, the integration test sometimes was already very close to timeout (10m) - #172's integration test spent 9m57s... So with new integration tests introduced, the test runs overtime frequently than before.

PR #190, opened 2021-04-21
Add PodState plugin test to main_test

yuanchen8911:
A configuration was created for `PodSate` plugin, but it's not included in the test cases.
This PR adds the test to `main_test`.


jasonliu747 (commented):
Thanks! Could you fix the CI error?

PR #195, opened 2021-05-17
Fix nil pointer dereference logic

mcristina422:
Fix nil pointer dereference logic to fix WaitTime calculation


Huang-Wei (commented):
Thanks for the catch.

/lgtm
/approve

PR #196, opened 2021-05-24
Refine coscheduling codes, set waittime to 0

suigh:
Update coscheduling codes with two changes, make the codes easier to read.

change 1: 
Update `UnschedulableAndUnresolvable` to `Unschedulable`. `UnschedulableAndUnresolvable` is used to avoid preemption, while the changed line is at function ` Permit`,  it cannot trigger preemption at all.

The description about `UnschedulableAndUnresolvable` in codes
```
	// UnschedulableAndUnresolvable is used when a (pre-)filter plugin finds a pod unschedulable and
	// preemption would not change anything. ......
```

change 2: 
Update the `waitTime` to `0` if the returned status is not `Wait`, the `waitTime` will only be handled when the return status is `Wait`, or it will be ingored. Returning a `waitTime` may confuse the one who reads the codes.

The related codes to handle returned status:
```
func (f *frameworkImpl) RunPermitPlugins(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string) (status *framework.Status) {
	.....
	statusCode := framework.Success
	for _, pl := range f.permitPlugins {
		status, timeout := f.runPermitPlugin(ctx, pl, state, pod, nodeName)
		if !status.IsSuccess() {
			if status.IsUnschedulable() {  <<=== for `UnschedulableAndUnresolvable` or `Unschedulable`, timeout value is ingored.
				msg := fmt.Sprintf("rejected pod %q by permit plugin %q: %v", pod.Name, pl.Name(), status.Message())
				klog.V(4).Infof(msg)
				return framework.NewStatus(status.Code(), msg)
			}
	}
```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -203,7 +203,7 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 	if err != nil {
 		_, pg := cs.pgMgr.GetPodGroup(pod)
 		if pg == nil {
-			return framework.NewStatus(framework.UnschedulableAndUnresolvable, "PodGroup not found"), waitTime
+			return framework.NewStatus(framework.Unschedulable, "PodGroup not found"), 0
 		}
 		if wait := util.GetWaitTimeDuration(pg, cs.scheduleTimeout); wait != 0 {
 			waitTime = wait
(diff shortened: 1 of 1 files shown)


denkensk:
Thanks for your help.
/lgtm
/assign @Huang-Wei

suigh:
Seems there is not enough resoure to run the test, I will try to re-run the test later.

```
I0524 14:38:47.200900   27343 factory.go:321] "Unable to schedule pod; no fit; waiting" pod="integration-test-7533a003-3a0f-4cac-b6e1-c1fb3a6257f4/t8-p3-2" **err="0/1 nodes are available:** 1 pod with pgName: integration-test-7533a003-3a0f-4cac-b6e1-c1fb3a6257f4/pg8-3 last failed in 3s, deny."
```

PR #198, opened 2021-06-01
Exclude list for Filter plugin

Tal-or:
This update allows the user to specified a list of resources that don't have to be NUMA aligned.
This is useful when the user wants only a subset of resources to be NUMA aligned and doesn't care about other resources.
The exclude list will be provided as one of the scheduler's arguments.

Resource type's names specified in the list should match the names as shown here: https://pkg.go.dev/k8s.io/api/core/v1#ResourceName


pkg/noderesourcetopology/README.md
```diff
@@ -45,6 +45,9 @@ profiles:
         - default
         - production
         - test-namespace
+      excludelist:
```
denkensk, line 48: Can we describe the resources we need instead of excluding what we don't need？
Or if there is reason？

denkensk (commented):
Thanks for your contribution

pkg/noderesourcetopology/README.md
Tal-or: By default, the Filter plugin should check NUMA-alignment for all resources, unless the user wants to exclude a specific resource from that check.

AlexeyPerevalov:
I had another approach in mind:
To avoid scheduler plugin configuration, keep this information at CRD level.
Now every resource comes to plugin in specific Zone (NUMA node), but also we may have a zone for worker node, where we can place a resource list. In this we have resources which don't have NUMA binding.

How topology exporter would define whether it NUMA specific resource or not:
1. This excludeList, but in TopologyExporter. It also doesn't look graceful, since we have to keep information on resource (their properties) in many places (device plugin, topology exporter)
2. It looks like responsibility of device plugin to inform all properties of extended resources. As I know device plugin, may not report any Topology in its Device message (https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/topology_hints.go#L131), so it's better to use this property of particular device plugin.

The only one problem here could be in the *implementation* of the  podresources API, again, since it doesn't distinguish resources with Topology and w/o Topology, (https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/manager.go#L939), so resource w/o topology will be assigned to NUMA 0. But podresources API itself, can provide empty Topology.

sseetharaman6:
I think having a common `excludeList` for the whole cluster could be problematic.  The `single-numa-node` topology manager policy, will consult all hint providers (cpu manager, memory manager, device manager) , but if cpu manager/memory manager are disabled, it would just use the default topology hint provided by the other hint providers.

Let's say for example , I have 2 nodes in the cluster- `nodeA`, with cpu manager enabled and `nodeB`, where its disabled. 
I would need to include `cpu` in my `excludeList` for my pod to ever land on `nodeB`. 
However doing so is problematic, since the scheduler plugin will skip checking `cpu` for `nodeA` as well and when the pod actually lands on the node, the topology manager could possibly hit a `TopologyAffinityError` .

Tal-or:
> Let's say for example , I have 2 nodes in the cluster- `nodeA`, with cpu manager enabled and `nodeB`, where its disabled.
> I would need to include `cpu` in my `excludeList` for my pod to ever land on `nodeB`.
> However doing so is problematic, since the scheduler plugin will skip checking `cpu` for `nodeA` as well and when the pod actually lands on the node, the topology manager could possibly hit a `TopologyAffinityError` .

You're absolutely right, but the Filter plugin was not designed to work w/o  CPUmanager enabled.
Because ```noderesourcetopology ``` CRs which are responsible for the resource accounting, (and are the 'source of truth' for the Filter plugin) won't expose the hardware information properly on a node where CPUmanager disabled.

So although it's an interesting case, I don't think it's a valid one.
Besides, I think having an ```excludeList``` per node is a high configuration burden for the user.

Tal-or:
> I had another approach in mind, since draw back of excludeList is straightforward - how do we know at cluster deploy time which resources are topology specific and which one not. I assume specific device plugin knows about it. And configuration at scheduler level - it's cluster deploy/maintenance burden.
> 
> To avoid scheduler plugin configuration lets keep this information at CRD level or just omit resources w/o topology.
> Now every resource comes to plugin in specific Zone (NUMA node), but also we may have a zone for worker node, where we can place a resource list. In this we have resources which don't have NUMA binding.
> 
> How topology exporter would define whether it NUMA specific resource or not:
> 
> 1. This excludeList, but in TopologyExporter. It also doesn't look graceful, since we have to keep information on resource (their properties) in many places (device plugin, topology exporter), we just move problem to another component.
> 2. Additional Zone (with name whole-worker-node), plugin should handle it, and exporter export resource w/o topology in this zone, but it looks redundant. (since we already have noderesouce fit plugin, and these resources already in the v1.Node...)
> 3. We can just skip exporting resources w/o topology, see comments bellow.
> 
> It looks like responsibility of device plugin to inform all properties of extended resources. As I know device plugin, may not report any Topology in its Device message (omitempty)(https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/topology_hints.go#L131), so it's better to use this property of particular device plugin.
> 
> The only one problem here could be in the _implementation_ of the podresources API, again, since it doesn't distinguish resources with Topology and w/o Topology, (https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/manager.go#L939), so resource w/o topology will be assigned to NUMA 0. But podresources API itself, can provide empty Topology.

This ```excludeList``` meant to address (among other things) the current problem where ```PodresourceAPI``` not exporting memory information: https://github.com/kubernetes/kubernetes/pull/101030 
And due to that, Filter plugin won't allow any Guarantee pod to be scheduled.

About you other suggestions, let's discussed that further on Thursday

Huang-Wei:
(not quite a NUMA expert) My question is: if a Node is NUMA specific, can we infer that **all** resources it employs is also NUMA specific/bound? If so, we can add a nodeSelector in the plugin arguments, so as to match the Nodes (as well as the resources).

AlexeyPerevalov:
> > I had another approach in mind, since draw back of excludeList is straightforward - how do we know at cluster deploy time which resources are topology specific and which one not. I assume specific device plugin knows about it. And configuration at scheduler level - it's cluster deploy/maintenance burden.
> > To avoid scheduler plugin configuration lets keep this information at CRD level or just omit resources w/o topology.
> > Now every resource comes to plugin in specific Zone (NUMA node), but also we may have a zone for worker node, where we can place a resource list. In this we have resources which don't have NUMA binding.
> > How topology exporter would define whether it NUMA specific resource or not:
> > 
> > 1. This excludeList, but in TopologyExporter. It also doesn't look graceful, since we have to keep information on resource (their properties) in many places (device plugin, topology exporter), we just move problem to another component.
> > 2. Additional Zone (with name whole-worker-node), plugin should handle it, and exporter export resource w/o topology in this zone, but it looks redundant. (since we already have noderesouce fit plugin, and these resources already in the v1.Node...)
> > 3. We can just skip exporting resources w/o topology, see comments bellow.
> > 
> > It looks like responsibility of device plugin to inform all properties of extended resources. As I know device plugin, may not report any Topology in its Device message (omitempty)(https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/topology_hints.go#L131), so it's better to use this property of particular device plugin.
> > The only one problem here could be in the _implementation_ of the podresources API, again, since it doesn't distinguish resources with Topology and w/o Topology, (https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/manager.go#L939), so resource w/o topology will be assigned to NUMA 0. But podresources API itself, can provide empty Topology.
> 
> This `excludeList` meant to address (among other things) the current problem where `PodresourceAPI` not exposing memory information: [kubernetes/kubernetes#101030](https://github.com/kubernetes/kubernetes/pull/101030)

we worked around memory, yes, but temporary. Now we have to make up a solution which would fit many requirements.

> And due to that, the Filter plugin won't allow any Guarantee pod to be scheduled.
> 
> About your other suggestions, let's discussed that further on Thursday

AlexeyPerevalov:
> (not quite a NUMA expert) My question is: if a Node is NUMA specific, can we infer that **all** resources it employs is also NUMA specific/bound? If so, we can add a nodeSelector in the plugin arguments, so as to match the Nodes (as well as the resources).

Initially we rely on that case, when node expose only NUMA specific resources and Pod's spec request only NUMA specific resources.
But in general case, I think, no we can't rely, since on the NUMA specific host could be the resources which represents not only real devices (https://kubernetes.slack.com/archives/C012XSGFZQE/p1622129062006900?thread_ts=1622072451.005400&cid=C012XSGFZQE, it's another question why vhost-user requested not by network attachment definition).

Tal-or:
> > > I had another approach in mind, since draw back of excludeList is straightforward - how do we know at cluster deploy time which resources are topology specific and which one not. I assume specific device plugin knows about it. And configuration at scheduler level - it's cluster deploy/maintenance burden.
> > > To avoid scheduler plugin configuration lets keep this information at CRD level or just omit resources w/o topology.
> > > Now every resource comes to plugin in specific Zone (NUMA node), but also we may have a zone for worker node, where we can place a resource list. In this we have resources which don't have NUMA binding.
> > > How topology exporter would define whether it NUMA specific resource or not:
> > > 
> > > 1. This excludeList, but in TopologyExporter. It also doesn't look graceful, since we have to keep information on resource (their properties) in many places (device plugin, topology exporter), we just move problem to another component.
> > > 2. Additional Zone (with name whole-worker-node), plugin should handle it, and exporter export resource w/o topology in this zone, but it looks redundant. (since we already have noderesouce fit plugin, and these resources already in the v1.Node...)
> > > 3. We can just skip exporting resources w/o topology, see comments bellow.
> > > 
> > > It looks like responsibility of device plugin to inform all properties of extended resources. As I know device plugin, may not report any Topology in its Device message (omitempty)(https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/topology_hints.go#L131), so it's better to use this property of particular device plugin.
> > > The only one problem here could be in the _implementation_ of the podresources API, again, since it doesn't distinguish resources with Topology and w/o Topology, (https://github.com/kubernetes/kubernetes/blob/6181a2c7d8efd1d05211f38254e787f4d74f09e9/pkg/kubelet/cm/devicemanager/manager.go#L939), so resource w/o topology will be assigned to NUMA 0. But podresources API itself, can provide empty Topology.
> > 
> > 
> > This `excludeList` meant to address (among other things) the current problem where `PodresourceAPI` not exposing memory information: [kubernetes/kubernetes#101030](https://github.com/kubernetes/kubernetes/pull/101030)
> 
> we worked around memory, yes, but temporary. Now we have to make up a solution which would fit many requirements.
> 
> > And due to that, the Filter plugin won't allow any Guarantee pod to be scheduled.
> > About your other suggestions, let's discussed that further on Thursday
This ```excludeList``` update was intended to solve that by configuring the plugin to skip the memory check, we didn't work around the memory problem yet.

Tal-or:
After further discussion, we decided to implement the ability to exclude specific resources on the NFD side https://github.com/kubernetes-sigs/node-feature-discovery/pull/545 and not in the scheduler plugin itself.

Huang-Wei:
> After further discussion, we decided to implement the ability to exclude specific resources on the NFD side [kubernetes-sigs/node-feature-discovery#545](https://github.com/kubernetes-sigs/node-feature-discovery/pull/545) and not in the scheduler plugin itself.

That sounds like a better idea. Thanks @Tal-or .

Issue #201, opened 2021-06-09
Integration tests are falling 


Tal-or:
**What happened:**
Calling ```make integration-test``` on a fresh clone of the repo end up with failure.

**What you expected to happen:**
The integration test should pass.

**How to reproduce:** 
1. Clone a fresh copy using git 
2. Execute integration-test using ```make integration-test```

The log can be found [here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_scheduler-plugins/198/pull-scheduler-plugins-integration-test-master/1401448322803175424/build-log.txt )

This log was taken from the failure of the integration test of [PR#198](https://github.com/kubernetes-sigs/scheduler-plugins/pull/198)
But the same log with the same error reproduces on a fresh copy of the repo as well.

Huang-Wei:
@Tal-or it's a known timeout issue, and seems https://github.com/kubernetes/test-infra/pull/22212 didn't resolve it.

I'm pinging the test-infra expert to dig why it doesn't work, and update there.

Huang-Wei:
Found the root cause, we should set a go test timeout flag. #202 is working on it.

Tal-or:
Tested it, Everything looks ok.
Closing the issue.
Thanks!

PR #202, opened 2021-06-09
Adjust the timeout to 40m for scheduler-plugins integration test

Huang-Wei:
`go test` default sets a 10m timeout, which is not enough for the integration test.

```
$ go help testflag
...
	-timeout d
	    If a test binary runs longer than duration d, panic.
	    If d is 0, the timeout is disabled.
	    The default is 10 minutes (10m).
```

This PR adjusts the timeout for integration tests to 40m.


Huang-Wei:
Will back-port it to lower versions.

Ref #201.

PR #205, opened 2021-06-24
KEP: Preemption Toleration Plugin

everpeace:
This PR adds a KEP that proposes "Preemption Toleration" plugin which provides a more flexible preemption behavior by introducing a _preemption toleration policy_ in `PriorityClass` resources (defining in the annotation).   

The concept of preemption toleration policy is similar to pod toleration.  The preemption toleration policy defines the condition in which the target priority class can tolerate (prevent from) the preemption by other priority classes or not.  That is a preemptee (a.k.a. victim) side policy that is not covered by Kubernetes official priority and preemption API.

I would be happy if I have community feedbacks.  Once the PR is approved, I will open the implementation and tests PR based on my PoC(https://github.com/everpeace/scheduler-plugins/pull/2).

_Note: I would like to use this plugin in our on-premise Deep Learning focused GPU clusters (N thousands of GPUs and M hundreds of users)._

### TODO

- [x] squash the branch once it was approved

### DIFF
--- a/kep/205-preemption-toleration/README.md
+++ b/kep/205-preemption-toleration/README.md
@@ -1,0 +1,303 @@
+# Preemption Toleration <!-- omit in toc -->
+
+## Table of Contents <!-- omit in toc -->
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+	- [Goals](#goals)
+	- [Non-Goals](#non-goals)
+- [Use Cases](#use-cases)
+	- [Lower priority value but not-being-preempted priority](#lower-priority-value-but-not-being-preempted-priority)
+	- [Conditional preemption with guaranteed running time](#conditional-preemption-with-guaranteed-running-time)
+- [Design Details](#design-details)
+	- [Preemption Toleration API](#preemption-toleration-api)
+	- [Implementing Typical Use Cases by Preemption Toleration APIs](#implementing-typical-use-cases-by-preemption-toleration-apis)
+		- [Lower priority value but not-being-preempted priority](#lower-priority-value-but-not-being-preempted-priority-1)
+		- [Conditional preemption with guaranteed running time](#conditional-preemption-with-guaranteed-running-time-1)
+	- [Plugin implementation](#plugin-implementation)
+		- [PostFilter](#postfilter)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+This plugin provides extended behavior of [Priority and Preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) in Kubernetes Scheduler.
+
+Especially, this plugin enables cluster administrators to define preemption policy from the _victim_ side which is not covered by public `PriorityClass` API. This plugin provides _preemption toleration_ policy in `PriorityClass`, which defines the criteria by which the priority class will be exempt from preemption.
+
+## Motivation
+
+Kubernetes scheduler provides [Pod Priority and Preemption](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/) feature.  Users, typically cluster administrators, can define several `PriorityClass` to classify priority in the cluster.  If the cluster is less elastic (e.g. On-Premise clusters), designing priority classes and preemption rules are very important for computational resource utilization.
+
+`PriorityClass` can have `PreemptionPolicy` configuration for customizing preemptor behavior for the priority class. The allowed values are `PreemptLowerPriority`(default) and `Never`.  If `Never` is set, the priority class becomes [_Non-preempting priority class_](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#non-preempting-priority-class) that can not preempt any pods but may be preempted by higher priority pods.  The `PreemptionPolicy` API is very simple and understandable.  However, this policy only focuses on preemptor side behavior.
+
+This plugin provides more flexible preemption behavior configuration by adding the preemptee (a.k.a victim) side policy in `PriorityClass`. In particular, cluster administrators can define _preemption toleration_ policy that defines the criteria which the priority class will be exempt from preemption.
+
+### Goals
+
+- provide flexible preemption toleration policy API spec and its implementation
+ 
+### Non-Goals
+
+- change Kubernetes public `PriorityClass` API
+
+## Use Cases
+
+### Lower priority value but not-being-preempted priority
+
+Generally speaking, making a job be resumable after preemption, the job would need to preserve its state externally, e.g. persistent volume, object storage, etc., at preemption and restore it at next restart (a.k.a check-pointing). However, when a job uses some proprietary tools that are not modifiable or do not support suspend/resume feature,  users would like to avoid being preempted. 
+
+To achieve this, `PriorityClass` with a high priority value would be enough. But just giving high priority value causes preemption on any pods with lower priority values.  Sometimes, users don't want to do it.  In such a case, the job's priority is essentially low (i.e. the job is expected to run only when the cluster has a vacancy).  However, once scheduled, users don't want the pod to be preempted because the pod is difficult to export/import its state.
+
+On the other hand, from the cluster administrative view, unconditional non-preempted pods are not desirable because there are system-critical, node-critical pods in the cluster. So, the cluster administrator would like to set the minimum priority value for such a priority class.  For example, assume there are three priority classes: system-critical, high and low.  Then, the cluster administrator would like to introduce `low-non-preempted` priority class that can not be preempted by high but can be preempted by system-critical.
+
+```yaml
+  system-critical: 10000
+             high:  9000
+low-non-preempted:  8000 # this priority class can tolerate preemption from priorities less than 10000
+                         # (i.e. high can't preempt this, but system-critical can preempt this)
+              low:  8000 # normal preemption happens on this priority class, i.e. p > 8000 can preempt this priority class.
+```
+
+To realise this feature, the plugin introduces `MinimumPreemptablePriority` in the preemption toleration policy API. See [Preemption Toleration API](#preemption-toleration-api) section and [Implementing Typical Use Cases by Preemption Toleration APIs](#implementing-typical-use-cases-by-preemption-toleration-apis) section below.
+
+### Conditional preemption with guaranteed running time 
+
+This is a typical use-case in machine learning job. As described above, job would need check-pointing. However, most machine learning jobs are iteration-based process, called _epochs_. So, if the cluster is high load and preemption always happens before the first epoch is finished,  the machine learning jobs can never make any progress and computational resources consumed by the job came to nothing in this case.
+
+So, cluster administrators would like to exempt pods that do not finish the first epoch from preemption even if the job has lower priority. But, it is hard to monitor when the first epoch is finished outside of the job. Thus, a minimum running time is useful practically.
+
+For example, cluster administrator would introduce several priority classes with the same priority value for introducing several minimum running time guarantee categories like below. 
+
+```yaml
+        system-critical: 10000  # system-critical can preempt pods in low-non-preempted-{10,30}min priority immediately
+                   high:  9000  # high can preempt pods in low-non-preempted-{10,30}min priority class which elapsed at least {10,30} minutes.
+                    low:  8000  # no minimum running time guarantee
+low-non-preempted-10min:  8000  # 10 min minimum running time guarantee against 8000 < p < 10000
+low-non-preempted-30min:  8000  # 30 min minimum running time guarantee against 8000 < p < 10000
+```
+
+To realise this feature, the plugin also introduces `TolerationSeconds` in the preemption toleration policy API. See [Preemption Toleration API](#preemption-toleration-api) section and [Implementing Typical Use Cases by Preemption Toleration APIs](#implementing-typical-use-cases-by-preemption-toleration-apis) section below.
+
+## Design Details
+
+### Preemption Toleration API
+
+Preemption toleration policy is defined in this struct type.
+
+```go
+type PreemptionToleration struct {
+  // MinimumPreemptablePriority specifies the minimum priority value that can preempt this priority class.
+  // It defaults to the PriorityClass's priority value + 1 if not set, which means pods that have a higher priority value can preempt it.
+  MinimumPreemptablePriority *int32 `json:"minimumPreemptablePriority,omitempty"`
+
+  // TolerationSeconds specifies how long this priority class can tolerate preemption 
+  // by priorities lower than MinimumPreemptablePriority.
+  // It defaults to zero if not set. Zero value means the pod will be preempted immediately. i.e., no toleration at all. 
+  // If it's set to a positive value, the duration will be honored.
+  // If it's set to a negative value, the pod can be tolerated forever - i.e., pods with priority
+  // lower than MinimumPreemptablePriority won't be able to preempt it.
+  // This value affects scheduled pods only (no effect on nominated pods).
+  TolerationSeconds *int64 `json:"tolerationSeconds,omitempty"`
+
+  // more policy will arise in the future.
+}
+```
+
+Users specifies the preemption toleration policy as an annotation with the key `preemption-toleration.scheduling.sigs.k8s.io/<property_name>` in `PriorityClass` as below:
+
+```yaml
+kind: PriorityClass
+metadata:
+  name: toleration-policy-sample
+  annotation:
+    # this key is needed to enable preemption toleration policy for distinguishing 
+    # between no toleration policy and empty toleration policy (all fields will be default)
+    preemption-toleration.scheduling.sigs.k8s.io/enabled: ""
+    preemption-toleration.scheduling.sigs.k8s.io/minimum-preemptable-priority: "10000"
+    preemption-toleration.scheduling.sigs.k8s.io/toleration-seconds: "3600"
+value: 8000
+```
+
+
+### Implementing Typical Use Cases by Preemption Toleration APIs
+
+This section describes how to implement scenarios described in [Use Cases](#use-cases) section by the preemption toleration policy.
+
+#### Lower priority value but not-being-preempted priority
+
+Assume cluster administrator introduces `low-non-preempted` priority class that can not be preempted by high but can be preempted by system-critical.  Then, they would declare `low-non-preempted` `PriorityClass` with preemption toleration policy of `MinimumPreemptablePriority=10000,TolerationSeconds=-1` like below:
+
+```yaml
+# system-critical can preempt low-10min pods immediately 
+# because low-non-preempted-10min can't tolerate the preemption from this priority class
+# (because of MinimumPreemptablePriority=10000)
+  system-critical: 10000
+
+# high pods can not preempt low-non-preempted pods forever
+# because low-non-preempted priority can tolerate the preemption forever 
+# by priority class p < 10000(=MinimumPreemptablePriority)
+# (because of MinimumPreemptablePriority=10000 and TolerationSeconds=-1)
+             high:  9000
+
+# with MinimumPreemptablePriority=10000,TolerationSeconds=-1
+low-non-preempted:  8000 
+
+# low with no preemption toleration policy. normal preemption behavior,
+# i.e. this priority class will be preempted by priority p > 8000
+              low:  8000
+```
+
+Thus, `low-non-preempted` manifest would be like below:
+
+```yaml
+kind: PriorityClass
+metadata:
+  name: low-non-preempted
+  annotation:
+    preemption-toleration.scheduling.sigs.k8s.io/enabled: ""
+    # This priority class can tolerate preemption by priority with p < 10000.
+    preemption-toleration.scheduling.sigs.k8s.io/minimum-preemptable-priority: "10000"
+    # This priority class can tolerate preemption forever by priority with p < 10000(=minimum-preemptable-priority)
+    preemption-toleration.scheduling.sigs.k8s.io/toleration-seconds: "-1"
+value: 8000
+```
+
+#### Conditional preemption with guaranteed running time
+
+Assume cluster administrator introduces `low-non-preempted-10min` priority class that can tolerate the preemption by high priority class for at lease 10 minutes, i.e. this guarantees 10 minutes running time for the priority class.  Then, they would declare `low-non-preempted-10min` `PriorityClass` with preemption toleration policy of `MinimumPreemptablePriority=10000,TolerationSeconds=600` like below:
+
+```yaml
+# system-critical can preempt low-non-preempted-10min pods immediately 
+# because low-non-preempted-10min can't tolerate the preemption from this priority class
+# (because of MinimumPreemptablePriority=10000)
+         system-critical: 10000 
+
+# high pods can preempt low-non-preempted-10min pods which elapsed 
+# at least 10min from being scheduled because toleration expired in 10 minutes
+# (because of MinimumPreemptablePriority=10000 and TolerationSeconds=600)
+                   high:  9000 
+
+# with MinimumPreemptablePriority=10000, TolerationSeconds=600
+low-non-preempted-10min:  8000
+```
+
+Thus, `low-non-preempted-10min` manifest would be like below:
+
+```yaml
+kind: PriorityClass
+metadata:
+  name: low-non-preempted-10min
+  annotation:
+    preemption-toleration.scheduling.sigs.k8s.io/enabled: ""
+    # This priority class can tolerate preemption by priority with p < 10000.
+    preemption-toleration.scheduling.sigs.k8s.io/minimum-preemptable-priority: "10000"
+    # This priority class can tolerate preemption for 10 minutes (600 seconds) 
+    # by priority with p < 10000(=minimum-preemptable-priority)
+    preemption-toleration.scheduling.sigs.k8s.io/toleration-seconds: "600"
+value: 8000
+```
+
+### Plugin implementation
+
+#### PostFilter
+
+This plugin implements the PostFilter extension point because this plugin focuses to extend the scheduler's preemption behavior. Moreover, this plugin just extends victim candidate selection logics in the default preemption behavior. 
+
+The modification would be only in `selectVictimsOnNode` as below:
+
+```golang
+// selectVictimsOnNode finds a minimal set of pods on the given node that should
+// be preempted in order to make enough room for "preemptor" to be scheduled.
+// The algorithm is almost identical to DefaultPreemption plugin's one.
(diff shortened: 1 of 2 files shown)


kep/205-preemption-toleration/kep.yaml
```diff
@@ -0,0 +1,12 @@
+title: Preemption Toleration
+kep-number: 205
+authors:
+  - "@everpeace"
+  - "@ordovicia"
+  - "@shioshiota"
+owning-sig: sig-scheduling
+reviewers: [] # TBD
+approvers: [] # TBD
```
everpeace, line 9: We need sponsors here.  Is anybody interested in this?? 🙇

Huang-Wei: Put me as the reviewer/approver. But I'm just back from vacation and thus a bit swamped, will probably review next week.

everpeace: Thank you so much!

everpeace: I put @Huang-Wei to reviewers/approvers in 5b203db

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+- change Kubernetes public `PriorityClass` API
+
+## Use Cases
+
+### Lower priority value but not-being-preempted priority
+
+Generally speaking, making a job be resumable after preemption, the job would need to preserve its state to outside of the pod, e.g. persistent volume, object storage, etc., at preemption and restore it at next restart (a.k.a check-pointing). However, when a job uses some proprietary tools that are not modifiable or do not support suspend/resume feature,  users would like to avoid being preempted. 
+
+To achieve this, `PriorityClass` with a high priority value would be enough. But just giving high priority value causes preemption on any pods with lower priority values.  Sometimes, users don't want to do it.  In such a case, the job's priority is essentially low (i.e. the job is expected to run only when the cluster has a vacancy).  However, once scheduled, users don't want the pod to be preempted because the pod is difficult to support export/import its state.
```
Huang-Wei, line 48:
```suggestion
To achieve this, `PriorityClass` with a high priority value would be enough. But just giving high priority value causes preemption on any pods with lower priority values.  Sometimes, users don't want to do it.  In such a case, the job's priority is essentially low (i.e. the job is expected to run only when the cluster has a vacancy).  However, once scheduled, users don't want the pod to be preempted because the pod is difficult to export/import its state.
```

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+On the other hand, from the cluster administrative view, unconditional non-preempted pods are not desirable because there are system-critical, node-critical pods in the cluster. So, the cluster administrator would like to set the minimum priority value for such a priority class.  For example, assume there are three priority classes: system-critical, high and low.  Then, the cluster administrator would like to introduce `low-non-preempted` priority class that can not be preempted by high but can be preempted by system-critical.
+
+```yaml
+  system-critical: 10000
+             high:  9000
+              low:  8000
+low-non-preempted:  8000 # with minimumPreemptablePriority=10000
```
Huang-Wei, line 56: If comparing this case with "low-non-preempted=10000 and PreemptionPolicy=Never", is the only difference that you don't want "low-non-preempted" pods to preempt pods with priority value ranging from 8000 to 10000?

If so, is that an alternative is to add another option to PreemptionPolicy (like "only preempt pods with priority value that's less than X")?

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Preemption Toleration API
+
+Preemption toleration policy is defined in this struct type.
+
+```go
+// Please note that this does NOT have ObjectMeta 
+// because this will be embedded in annotation in PriorityClass as unnamed object.
+type PreemptionToleration struct {
+	metav1.TypeMeta `json:",inline"`
```
Huang-Wei, line 85:
```suggestion
  metav1.TypeMeta `json:",inline"`
```

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,126 @@
...
+    scheduling.sigs.k8s.io/preemption-toleration: |
+      # the api is versioned for future enhancement
+      apiVersion: scheduling.sigs.k8s.io/v1beta1
+      kind: PreemptionToleration
+
+      # This priority class can tolerate preemption by priority with 8000 <= p < 10000.
+      minimumPreemptablePriority: 10000
+
+      # And it can tolerate preemption in 1 hour by the pod with priority (8000 <= p < 10000).
+      tolerationSeconds: 3600
```
Huang-Wei, line 113: I'd prefer to exact all the settings to a CustomResource object; otherwise, I don't think it's practical to versioning/defaulting/converting on the annotated string.

Or, you give up support multi-version toleration policies - i.e., define the settings as scheduling.sigs.k8s.io/preemption-toleration-XYZ, individually.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,126 @@
...
+value: 8000
+```
+
+### Plugin implementation
+
+#### PostFilter
+
+This plugin implements the PostFilter extension point because this plugin focuses to extend the scheduler's preemption behavior.  Moreover, this plugin just extends victim candidate selection logics in the default preemption behavior. 
+
```
Huang-Wei, line 122: I think you'd need to add a stanza to describe how to honor the preemption tolerationSeconds setting, e.g., implement the controller logic to monitor/manage the pods with preemption tolerationSeconds settings.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,126 @@
...
+```go
+// Please note that this does NOT have ObjectMeta 
+// because this will be embedded in annotation in PriorityClass as unnamed object.
+type PreemptionToleration struct {
+	metav1.TypeMeta `json:",inline"`
+
+  // MinimumPreemptablePriority specifies the minimum priority value that can preempt this priority class.
+  MinimumPreemptablePriority *int32 `json:"minimumPreemptablePriority,omitempty"`
+
+  // TolerationSeconds specified how many seconds this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.  Null value specifies forever (default).  Duration means the duration from the pod being scheduled to some node.  This value affects only to scheduled pods (no effect on nominated nodes).
```
Huang-Wei, line 90: I hope my understanding is correct:

```suggestion
  // TolerationSeconds specifies how long this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.
  // Nil or zero value means it's preemptable immediately (default).
  // Non-nil value means the duration from the pod being scheduled to some node.
  // This value affects scheduled pods only (no effect on nominated pods).
```

kep/205-preemption-toleration/README.md
everpeace: I see.  If we choose CRD case, `PriorityClass` would need to refer `PreemptionToleration` CR by annotation or a similar way.  Then, I would like to choose the annotation case.

kep/205-preemption-toleration/README.md
everpeace: Thanks. I will document where to how to modify the default preemption plugin.

kep/205-preemption-toleration/README.md
everpeace: I intended to the same semantics with [`tolerationSeconds` in Pod](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#toleration-v1-core):

> By default, it is not set, which means tolerate the taint forever (do not evict). Zero and negative values will be treated as 0 (evict immediately) by the system.

Thus, I would like to fix it to be a more clear explanation like bleow:

```suggestion
  // TolerationSeconds specifies how long this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.
  // Nil value means forever, which is not preempted forever by the priority value lower than MinimumPreeemptable Priority (default)
  // zero and negative value means immediate, which is preempted immediately (it is probably useless)
  // This value affects scheduled pods only (no effect on nominated pods).
```

everpeace (commented):
Thanks for your comment!  I commented back.  PTAL 🙇

kep/205-preemption-toleration/README.md
everpeace: done in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/8484681100265b185b73323495a71813f82bc715

kep/205-preemption-toleration/README.md
everpeace: done in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/335c4118c5e2f2010e4d910eb2524ca4abb0e0fb

kep/205-preemption-toleration/README.md
everpeace: done in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/b162e8de84fad9532fa2b2b6866e924c51f5e066

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+`PriorityClass` can have `PreemptionPolicy` configuration for customizing preemptor behavior for the priority class. The allowed values are `PreemptLowerPriority`(default) and `Never`.  If `Never` is set, the priority class becomes [_Non-preempting priority class_](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#non-preempting-priority-class) that can not preempt any pods but may be preempted by higher priority pods.  The `PreemptionPolicy` API is very simple and understandable.  However, this policy only focuses on preemptor side behavior.
+
+This plugin provides more flexible preemption behavior configuration by adding the preemptee (a.k.a victim) side policy in `PriorityClass`. In particular, cluster administrators can define _preemption toleration_ policy that defines the criteria which the priority class will be exempt from preemption.
+
+### Goals
+
+- provide flexible preemption toleration policy spec
+- and its implementation
```
Huang-Wei, line 36: Make them in one line.

```suggestion
- provide flexible preemption toleration policy API spec and its implementation
```

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+ 
+### Non-Goals
+
+- change Kubernetes public `PriorityClass` API
+
+## Use Cases
+
+### Lower priority value but not-being-preempted priority
+
+Generally speaking, making a job be resumable after preemption, the job would need to preserve its state to outside of the pod, e.g. persistent volume, object storage, etc., at preemption and restore it at next restart (a.k.a check-pointing). However, when a job uses some proprietary tools that are not modifiable or do not support suspend/resume feature,  users would like to avoid being preempted.
```
Huang-Wei, line 46: `to outside of the pod` -> `externally`

kep/205-preemption-toleration/README.md
Huang-Wei: Yes, the tolerationSeconds fits more on the preemptee side.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+Users specifies the preemption toleration policy as an annotation with the key `preemption-toleration.scheduling.sigs.k8s.io/<property_name>` in `PriorityClass` as below:
+
+```yaml
+kind: PriorityClass
+metadata:
+  name: toleration-policy-sample
+  annotation:
+    # this key is needed to enable preemption toleration policy for distinguishing 
+	# between no toleration policy and empty toleration policy (all fields will be default)
```
Huang-Wei, line 104: nit: replace tab with empty characters.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+```
+
+Users specifies the preemption toleration policy as an annotation with the key `preemption-toleration.scheduling.sigs.k8s.io/<property_name>` in `PriorityClass` as below:
+
+```yaml
+kind: PriorityClass
+metadata:
+  name: toleration-policy-sample
+  annotation:
+    # this key is needed to enable preemption toleration policy for distinguishing
```
Huang-Wei, line 103: Hmm... this is another shortage of annotations (vs. CRD). It's ok to start with annotations as an alpha feature, but once we graduate it to beta, do you have any plan to turn to CRD? - which may better manage the specs.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Design Details
+
+### Preemption Toleration API
+
+Preemption toleration policy is defined in this struct type.
+
+```go
+type PreemptionToleration struct {
+  // MinimumPreemptablePriority specifies the minimum priority value that can preempt this priority class.
```
Huang-Wei, line 83: Describe the default value/behavior if it's not set. (like, if not set, pods that have a higher priority value can preempt it)

kep/205-preemption-toleration/README.md
Huang-Wei: Ah, the details of PostFilter is awesome. But what I meant in the original comment was the "controller" logic - like how you want the controller to be implemented, launched and packaged (into scheduler or controller binary?).

(A reference is that Kubernetes uses a nodelifecycle manager to guarantee the tolerationSeconds' semantics)

everpeace (commented):
@Huang-Wei Thanks for the review.  I pushed some fixes and replied to your comments.  PTAL 🙇

kep/205-preemption-toleration/README.md
everpeace: fixed in 8d7f4da

kep/205-preemption-toleration/README.md
everpeace: fixed in b8c0b87

Huang-Wei:
Re https://github.com/kubernetes-sigs/scheduler-plugins/pull/205#discussion_r672752623:

(Github doesn't allow me to comment there)

Oh, I misunderstood it. So it means upon a preemption attempt, the Pods that can be tolerated won't be preempted, also they won't be deleted if the toleration seconds have passed - which is different with Pod's tolerationSeconds where the Pod will be deleted forcibly when tolerationSeconds is used up.

In your case, it's more a concept of "guaranteed minimum duration to run". These Pods can be only preempted when the time passed the minimum duration. Is my understanding correct?

everpeace:
@Huang-Wei 

> In your case, it's more a concept of "guaranteed minimum duration to run". These Pods can be only preempted when the time passed the minimum duration. Is my understanding correct?

Yes, that's right!!

everpeace:
Once I got approved, I would like to squash the commits so far.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+  // If not set, pods that have a higher priority value can preempt it (default).
+  MinimumPreemptablePriority *int32 `json:"minimumPreemptablePriority,omitempty"`
+
+  // TolerationSeconds specifies how long this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.
+  // Nil value means forever, which is not preempted forever (default)
+  // zero and negative value means immediate, which is preempted immediately. i.e. no toleration at all.
```
Huang-Wei, line 88: I don't think we should distinguish nil from zero. How about:

```
// It defaults to zero if not set. Zero value means the pod will be preempted immediately. i.e., no toleration at all.
```

And negative value will be caught during validation phase.

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,215 @@
...
+## Design Details
+
+### Preemption Toleration API
+
+Preemption toleration policy is defined in this struct type.
+
+```go
+type PreemptionToleration struct {
+  // MinimumPreemptablePriority specifies the minimum priority value that can preempt this priority class.
+  // If not set, pods that have a higher priority value can preempt it (default).
```
Huang-Wei, line 83: To standardize the defaulting logic, _ideally_ we should default the value to the PC's priority value. (but not 100% sure it's feasible, let's state it this way for now)

```
// It defaults to the PriorityClass's priority value if not set, which means pods that have a higher priority value can preempt it.
```

kep/205-preemption-toleration/kep.yaml
```diff
@@ -0,0 +1,14 @@
...
+  - "@everpeace"
+  - "@ordovicia"
+  - "@shioshiota"
+owning-sig: sig-scheduling
+reviewers:
+- "@Huang-Wei"
+approvers:
+- "@Huang-Wei"
+creation-date: 2021-06-24
+last-updated: 2021-06-24
```
Huang-Wei, line 13: make it today's date.

Huang-Wei (commented):
Some final comments on defaulting. LGTM otherwise.

kep/205-preemption-toleration/README.md
everpeace: resolved in another thread https://github.com/kubernetes-sigs/scheduler-plugins/pull/205#issuecomment-883822936

kep/205-preemption-toleration/README.md
everpeace: OK. It's a nit but, the default value would be `PriorityClass's priority value + 1` because this is `minimumPreemptableValue` . I fill fix it with this.

kep/205-preemption-toleration/README.md
everpeace: As stated https://github.com/kubernetes-sigs/scheduler-plugins/pull/205#discussion_r662106942:

- `TolerationSeconds: nil` : it can tolerate forever (_perhaps this is a common use case, I think_)
- `TolerationSeconds: 0`: it can't tolerate at all (tolerate in 0 second).

So, how about this??

```
  // TolerationSeconds specifies how long this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.  
  // It defaults to Nil if not set.  Nil value means the pod can tolerate the preemption forever.  Zero value means the pod will be preempted immediately. i.e., no toleration at all.
```

everpeace (commented):
@Huang-Wei sorry for the late reply because it was a holiday week in Japan last week.  I updated the branch and commented back for some. PTAL 🙇  I hope this is the last turn.

kep/205-preemption-toleration/kep.yaml
everpeace: done in f321e6b

kep/205-preemption-toleration/README.md
everpeace: updated in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/671bf6ff7c31eac20f83555497028ccac4852f7a

kep/205-preemption-toleration/README.md
everpeace: done in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/d2ac06178e08980c69073dc4ab34cb1a1dad639e

kep/205-preemption-toleration/README.md
Huang-Wei: @everpeace in terms of API design practices, it's a good practice to use a pointer for the internal-version struct (in pkg/apis/config/types.go); we do that for external versions (in pkg/apis/config/v1XYZ/types.go).

So when the versioned obj is wired in, we can tell if it's set to 0 or left as nil, and then default the value if it's nil (not set). In the internal processing that uses internal version objs, we don't need to worry the fields are set or not - default or user-specified value should have been set.

So the comment "It defaults to Nil if not set" is inaccurate - in internal versions, it should be a concrete value instead of pointer.

kep/205-preemption-toleration/README.md
Huang-Wei: SG.

kep/205-preemption-toleration/README.md
everpeace: @Huang-Wei Ah, I understood.  Then, how about negative value means forever??  I would like to implement a policy that can tolerate _forever_ in some way. How about below??

```
// TolerationSeconds specifies how long this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.  
// It defaults to zero if not set. Zero value means the pod will be preempted immediately. i.e., no toleration at all.
// Negative value means the pod can tolerate forever.
```

done in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/b15a95027b612c9b8c8682d2ce2833bc03fd7f81

everpeace: OR, could we make `Nill` be default (perhaps we can say no default). how about this??

```
// TolerationSeconds specifies how long this priority class can tolerate preemption by priorities lower than MinimumPreemptablePriority.  
// If not set, the pod can tolerate the preemption forever.
```

Huang-Wei: > just leave`Nil` if not set

As I mentioned before, internal API object should be a concrect value (so nil value doesn't work, generally).

You previous comment makes sense to me:

- if it's not set (nil), internally default it to 0 - which means it can be preempted immediately
- if it's set to a positive value - the duration will be honored
- if it's set to a negative value, you _can_ make it semantically mean "don't preempt forever". But I'm not sure this is really what you want? What if some pods set this to a negative value and then the high-priority can never preempt them and pending there forever? It seems to break the semantics of Priority.

everpeace: @Huang-Wei 
> As I mentioned before, internal API object should be a concrect value (so nil value doesn't work, generally).

oh... sorry... I was confused.

>  But I'm not sure this is really what you want? What if some pods set this to a negative value and then the high-priority can never preempt them and pending there forever? It seems to break the semantics of Priority.

`minimumPreemptablePriority` and `tolerationSeconds` will be evaluated in AND-ed.  `tolerationSeconds` is expected to be used with `minimumPreemptablePriority`.   It means `tolerationSeconds` specifies how long this priority class can tolerate preemption by priorities __lower than MinimumPreemptablePriority__ as described in the comment.  For example,

Assume `low-non-preempted-10min` declared with `minimumPreemptablePriority=10000, tolerationSeconds=600`.  Then, `system-critical, high, medium` priorites' preemption behavior will be like below:

```yaml
# system-critical can preempt low-10min pods immediately 
# because low-non-preempted-10min can't tolerate the preemption from this priority class
# (because of minimumPreemptablePriority=10000)
         system-critical: 10000 

# high and medium pods can preempt low-non-preempted-10min pods 
# which elapsed at least 10min from being scheduled because toleration expired in 10 min
# (because of tolerationSeconds=600)
                   high:  9000 
                 medium:  8500 

# assume low-non-preempted-10min has a preemption toleration policy 
# with minimumPreemptablePriority=10000, tolerationSeconds=600
low-non-preempted-10min:  8000
```

updated in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/5ab5cad9eccc3da671bbfecdffa874ee14d1e866

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,218 @@
...
+type PreemptionToleration struct {
+  // MinimumPreemptablePriority specifies the minimum priority value that can preempt this priority class.
+  // It defaults to the PriorityClass's priority value + 1 if not set, which means pods that have a higher priority value can preempt it.
+  MinimumPreemptablePriority *int32 `json:"minimumPreemptablePriority,omitempty"`
+
+  // TolerationSeconds specifies how long this priority class can tolerate preemption 
+  // by priorities lower than MinimumPreemptablePriority.
+  // It defaults to zero if not set. Zero value means the pod will be preempted immediately. i.e., no toleration at all. 
+  // If it's set to a positive value, the duration will be honored.
+  // If it's set to a negative value, the pod can tolerate "forever".
```
Huang-Wei, line 91:
```suggestion
  // If it's set to a negative value, the pod can be tolerated forever - i.e., pods with priority
  // lower than MinimumPreemptablePriority won't be able to preempt it.
```

kep/205-preemption-toleration/README.md
Huang-Wei: I thought `minimumPreemptablePriority` and `tolerationSeconds` work independently. The current semantics looks good - it can only tolerate the pods with priority lower than  `minimumPreemptablePriority`, not always. Can you somehow emphasize it in the example? You did mention it in the latter section (API/implementations), but it'd be better to also emphasize it in the earlier section (the examples).

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+This is a typical use-case in machine learning job. As described above, job would need check-pointing. However, most machine learning jobs are iteration-based process, called _epochs_. So, if the cluster is high load and preemption always happens before the first epoch is finished,  the machine learning jobs can never make any progress and computational resources consumed by the job came to nothing in this case.
+
+So, cluster administrators would like to exempt pods that do not finish the first epoch from preemption even if the job has lower priority. But, it is hard to monitor when the first epoch is finished outside of the job. Thus, a minimum running time is useful practically.
+
+For example, cluster administrator would introduce several priority classes with the same priority value for introducing several minimum running time guarantee categories like below. 
+
+```yaml
+        system-critical: 10000  # system-critical can preempt pods in low-non-preempted-{10,30}min priority immediately
+	               high:  8000  # high can preempt pods in low-non-preempted-{10,30}min priority class which elapsed at least {10,30} minutes.
```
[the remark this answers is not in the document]
everpeace: fixed in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/640fca76068d1f2de2c853850d584e52c91c0e2c

kep/205-preemption-toleration/README.md
```diff
@@ -0,0 +1,218 @@
...
+	                low:  8000  # no minimum running time guarantee
```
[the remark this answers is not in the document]
everpeace: fixed in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/697ab813317d50296f117ce51a4763437f94fb80

kep/205-preemption-toleration/README.md
everpeace: @Huang-Wei 
> Can you somehow emphasize it in the example? You did mention it in the latter section (API/implementations), but it'd be better to also emphasize it in the earlier section (the examples).

Thanks for your suggestion. I agree.  I think it would the earlier "Use Cases" section would be better not to require the readers preemption toleration API knowledge.  So, I added the section "Implementing Typical Use Cases by Preemption Toleration APIs" under the "Preemption Toleration API" section so that it can describe how the API works in detail by examples. And I added the link to the section in "Use Cases" section. 

updated in https://github.com/kubernetes-sigs/scheduler-plugins/pull/205/commits/c50b6e20037a141f1cd9eda22fd7ed2aeaed926f .

Huang-Wei:
@everpeace LGTM right now. Please squash the commits, and I will approve later.

everpeace:
@Huang-Wei Thank you very much for the review! I squashed my branch and rebased it to the latest master. PTAL 🙇

PR #206, opened 2021-06-25
fix: WaitForCacheSync parameter

slaysd:
- duplicate Synced object (`ctrl.pgListerSynced`)

I think that it might be `pgListerSynced` and `podListerSynced`


denkensk:
Thanks @slaysd 
/lgtm
/assign @Huang-Wei

Huang-Wei:
Thanks for the catch.

/approve

PR #207, opened 2021-06-30
standardize error and state of return values of Permit

Huang-Wei:


Huang-Wei:
@cwdsuzhou could you help review this PR?

cwdsuzhou:
/lgtm

thanks @Huang-Wei

PR #213, opened 2021-07-08
update the elastic quota immediately after delete victims

### LINKED ISSUE #211: Capacity Scheduler Strange Preemption Behavior
## Problem brief description
I am currently trying to deploy the capacity scheduler as a secondary Kubernetes scheduler into the AWS EKS cluster. I am able to deploy the scheduler but notices a strange preemption behavior. This behavior does not exist for a cluster with a single node but exists for a cluster with multiple nodes.

## Cluster information
My cluster has 10 machines, with each machine having 4 vCPU. So a total of 40 vCPU in my cluster.

## Scheduler information:
scheduler plugins: v0.19.9

## ElasticQuota Configuration
I have three namespaces, which are called quota1, quota2, quota3.
For the elasticQuota bound to quota1, I have the following configuration:
```
kind: ElasticQuota
metadata:
  name: quota1
  namespace: quota1
spec:
  max:
    cpu: 3
  min:
    cpu: 1
```

For the elasticQuota bound to quota2, I have the following configuration:
```
kind: ElasticQuota
metadata:
  name: quota2
  namespace: quota2
spec:
  max:
    cpu: 3
  min:
    cpu: 3
```

For the elasticQuota bound to quota3, I have the following configuration:
```
kind: ElasticQuota
metadata:
  name: quota3
  namespace: quota3
spec:
  max:
    cpu: 4
  min:
    cpu: 3
```

## Application Configuration
```
Deployment 1 configuration:
apiVersion: apps/v1

kind: Deployment
metadata:
  name: nginx
  namespace: quota1
  labels:
    app: nginx
spec:
  replicas: 4
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      name: nginx
      labels:
        app: nginx
    spec:
      schedulerName: capacityscheduler
      containers:
        - name: nginx
          image: nginx
          resources:
            limits:
              cpu: 1
            `requests:`
              cpu: 1
```
Deployment 2 configuration:
```
Deployment 1 configuration:
apiVersion: apps/v1

kind: Deployment
metadata:
  name: nginx
  namespace: quota2
  labels:
    app: nginx
spec:
  replicas: 4
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      name: nginx
      labels:
        app: nginx
    spec:
      schedulerName: capacityscheduler
      containers:
        - name: nginx
          image: nginx
          resources:
            limits:
              cpu: 1
            `requests:`
              cpu: 1
```
Deployment 3 configuration:
```
Deployment 1 configuration:
apiVersion: apps/v1

kind: Deployment
metadata:
  name: nginx
  namespace: quota3
  labels:
    app: nginx
spec:
  replicas: 4
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      name: nginx
      labels:
        app: nginx
    spec:
      schedulerName: capacityscheduler
      containers:
        - name: nginx
          image: nginx
          resources:
            limits:
              cpu: 1
            `requests:`
              cpu: 1
```

#### Initially, everything behaves normally
Create Deployment 1 in namespace quota1, and 3 pods are running in namespaces quota1, with 1 pending
```
nginx-5b85c78645-2574h   1/1     Running   0          6s
nginx-5b85c78645-8kqcr   1/1     Running   0          6s
nginx-5b85c78645-dn454   0/1     Pending   0          6s
nginx-5b85c78645-qknqx   1/1     Running   0          6s
```

Create Deployment 2 in namespace quota2, and 3 pods are created in namespace quota2 with 1 pending 
```
nginx-5b85c78645-69hl2   1/1     Running   0          9s
nginx-5b85c78645-gkq2t   1/1     Running   0          9s
nginx-5b85c78645-j7k7b   1/1     Running   0          9s
nginx-5b85c78645-lzx6v   0/1     Pending   0          9s
```

#### Wierd behavior starts
Create deployment 3, then all pods in namespace quota1 are terminated

## Preemption Event Logs from namespaces 1
```
pod/nginx-5b85c78645-2574h    Preempted by quota3/nginx-5b85c78645-fn2vv on node1
pod/nginx-5b85c78645-8kqcr    Preempted by quota3/nginx-5b85c78645-npnwb on node2
pod/nginx-5b85c78645-2574h    Preempted by quota3/nginx-5b85c78645-npnwb on node1
pod/nginx-5b85c78645-2574h    Preempted by quota3/nginx-5b85c78645-lgnqp on node1
pod/nginx-5b85c78645-8kqcr    Preempted by quota3/nginx-5b85c78645-lgnqp on node2
pod/nginx-5b85c78645-qknqx    Preempted by quota3/nginx-5b85c78645-lgnqp on node3
pod/nginx-5b85c78645-dn454    Preempted by quota3/nginx-5b85c78645-lgnqp on node2
pod/nginx-5b85c78645-dn454    Preempted by quota3/nginx-5b85c78645-npnwb on node2
```
Note here pod/nginx-5b85c78645-dn454 is preempted later in the log because it gets scheduled when previous running pods in namespace quota1 are preempted.

###### From the above logs we can see that:
1. All previously running pods in quota1 are terminated by deployment 3 on namespace quota3. But since quota1 has min = 1, shouldn't the scheduler at least preserve 1 running pod instead of terminating all of the pods in namespace quota1?
2.  One pod can possibly terminate multiple pods in different nodes. For example, quota3/nginx-5b85c78645-lgnqp has teminated pod/nginx-5b85c78645-8kqcr on node 2,  pod/nginx-5b85c78645-2574h on node1, and pod/nginx-5b85c78645-qknqx on node 3.

###### Other things to note
1. Eventually, quota1 has 1 node, quota2 has 3 nodes, and quota3 has 3 nodes but it takes some time to reach this equilibrium.
2. None of the Pods in namespace 2 get preempted because it does not borrow other's min. However, If we change the previous elasticQuota's max on namespace quota2 from 3 to 4, in this case, quota2 will borrow 1 min from quota3, and when quota3 needs the resources back, all pods in quota2 will be terminated and later new pods for deployment 2 will be created to reach the corresponding min set by the elasticQuota, which is 3 in this case.

## Deployment note:
I deployed the capacity scheduler as a secondary scheduler but didn't deploy a separate controller and instead use the default controller, it this a problem? I did not deploy it because I did not see a section for the controller in the capacity scheduler but there exists a controller section in the coscheduling under the kep directory.
If I have to deploy an extra controller, is there any guideline for setting up the controller for the capacity scheduler? I have tried to modify the provided helm example on coscheduling but the controller will always output a log saying that it is starting coscheduling. 

If anyone can help, that will be great! Thanks!

denkensk:
fix #211 

1. Update the elastic quota immediately after delete victims
2. Consider Nominated Pods in prefilter to avoid repeated preemption

### DIFF
--- a/pkg/apis/config/v1beta1/defaults.go
+++ b/pkg/apis/config/v1beta1/defaults.go
@@ -100,9 +100,7 @@ func SetDefaultsNodeResourcesAllocatableArgs(obj *NodeResourcesAllocatableArgs) 
 
 // SetDefaultsCapacitySchedulingArgs sets the default parameters for CapacityScheduling plugin.
 func SetDefaultsCapacitySchedulingArgs(obj *CapacitySchedulingArgs) {
-	if obj.KubeConfigPath == nil {
-		obj.KubeConfigPath = &defaultKubeConfigPath
-	}
+	// TODO(k/k#96427): get KubeConfigPath and KubeMaster from configuration or command args.
 }
 
 // SetDefaultTargetLoadPackingArgs sets the default parameters for TargetLoadPacking plugin
--- a/pkg/capacityscheduling/capacity_scheduling.go
+++ b/pkg/capacityscheduling/capacity_scheduling.go
@@ -62,7 +62,17 @@ type CapacityScheduling struct {
 
 // PreFilterState computed at PreFilter and used at PostFilter or Reserve.
 type PreFilterState struct {
-	framework.Resource
+	podReq framework.Resource
+
+	// nominatedPodsReqInEQWithPodReq is the sum of podReq and the requested resources of the Nominated Pods
+	// which subject to the same quota(namespace) and is more important than the preemptor.
+	nominatedPodsReqInEQWithPodReq framework.Resource
+
+	// nominatedPodsReqWithPodReq is the sum of podReq and the requested resources of the Nominated Pods
+	// which subject to the all quota(namespace). Generated Nominated Pods consist of two kinds of pods:
+	// 1. the pods subject to the same quota(namespace) and is more important than the preemptor.
+	// 2. the pods subject to the different quota(namespace) and the usage of quota(namespace) does not exceed min.
+	nominatedPodsReqWithPodReq framework.Resource
 }
 
 // Clone the preFilter state.
@@ -195,24 +205,76 @@ func (c *CapacityScheduling) EventsToRegister() []framework.ClusterEvent {
 // 1. Check if the (pod.request + eq.allocated) is less than eq.max.
 // 2. Check if the sum(eq's usage) > sum(eq's min).
 func (c *CapacityScheduling) PreFilter(ctx context.Context, state *framework.CycleState, pod *v1.Pod) *framework.Status {
+	// TODO improve the efficiency of taking snapshot
+	// e.g. use a two-pointer data structure to only copy the updated EQs when necessary.
 	snapshotElasticQuota := c.snapshotElasticQuota()
-	preFilterState := computePodResourceRequest(pod)
+	podReq := computePodResourceRequest(pod)
 
-	state.Write(preFilterStateKey, preFilterState)
 	state.Write(ElasticQuotaSnapshotKey, snapshotElasticQuota)
 
 	elasticQuotaInfos := snapshotElasticQuota.elasticQuotaInfos
 	eq := snapshotElasticQuota.elasticQuotaInfos[pod.Namespace]
 	if eq == nil {
-		return framework.NewStatus(framework.Success, "skipCapacityScheduling")
+		preFilterState := &PreFilterState{
+			podReq: *podReq,
+		}
+		state.Write(preFilterStateKey, preFilterState)
+		return framework.NewStatus(framework.Success)
 	}
 
-	if eq.overUsed(preFilterState.Resource, eq.Max) {
-		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("Pod %v/%v is rejected in Prefilter because ElasticQuota %v is more than Max", pod.Namespace, pod.Name, eq.Namespace))
+	// nominatedPodsReqInEQWithPodReq is the sum of podReq and the requested resources of the Nominated Pods
+	// which subject to the same quota(namespace) and is more important than the preemptor.
+	nominatedPodsReqInEQWithPodReq := &framework.Resource{}
+	// nominatedPodsReqWithPodReq is the sum of podReq and the requested resources of the Nominated Pods
+	// which subject to the all quota(namespace). Generated Nominated Pods consist of two kinds of pods:
+	// 1. the pods subject to the same quota(namespace) and is more important than the preemptor.
+	// 2. the pods subject to the different quota(namespace) and the usage of quota(namespace) does not exceed min.
+	nominatedPodsReqWithPodReq := &framework.Resource{}
+
+	nodeList, err := c.fh.SnapshotSharedLister().NodeInfos().List()
+	if err != nil {
+		return framework.NewStatus(framework.Error, fmt.Sprintf("Error getting the nodelist: %v", err))
 	}
 
-	if elasticQuotaInfos.aggregatedMinOverUsedWithPod(preFilterState.Resource) {
-		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("Pod %v/%v is rejected in Prefilter because total ElasticQuota used is more than min", pod.Namespace, pod.Name))
+	for _, node := range nodeList {
+		nominatedPods := c.fh.NominatedPodsForNode(node.Node().Name)
+		for _, p := range nominatedPods {
+			if p.Pod.UID == pod.UID {
+				continue
+			}
+			ns := p.Pod.Namespace
+			info := c.elasticQuotaInfos[ns]
+			if info != nil {
+				pResourceRequest := computePodResourceRequest(p.Pod).ResourceList()
+				// If they are subject to the same quota(namespace) and p is more important than pod,
+				// p will be added to the nominatedResource and totalNominatedResource.
+				// If they aren't subject to the same quota(namespace) and the usage of quota(p's namespace) does not exceed min,
+				// p will be added to the totalNominatedResource.
+				if ns == pod.Namespace && corev1helpers.PodPriority(p.Pod) >= corev1helpers.PodPriority(pod) {
+					nominatedPodsReqInEQWithPodReq.Add(pResourceRequest)
+					nominatedPodsReqWithPodReq.Add(pResourceRequest)
+				} else if ns != pod.Namespace && !info.usedOverMin() {
+					nominatedPodsReqWithPodReq.Add(pResourceRequest)
+				}
+			}
+		}
+	}
+
+	nominatedPodsReqInEQWithPodReq.Add(podReq.ResourceList())
+	nominatedPodsReqWithPodReq.Add(podReq.ResourceList())
+	preFilterState := &PreFilterState{
+		podReq:                         *podReq,
+		nominatedPodsReqInEQWithPodReq: *nominatedPodsReqInEQWithPodReq,
+		nominatedPodsReqWithPodReq:     *nominatedPodsReqWithPodReq,
+	}
+	state.Write(preFilterStateKey, preFilterState)
+
+	if eq.usedOverMaxWith(nominatedPodsReqInEQWithPodReq) {
+		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("Pod %v/%v is rejected in PreFilter because ElasticQuota %v is more than Max", pod.Namespace, pod.Name, eq.Namespace))
+	}
+
+	if elasticQuotaInfos.aggregatedUsedOverMinWith(*nominatedPodsReqWithPodReq) {
+		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("Pod %v/%v is rejected in PreFilter because total ElasticQuota used is more than min", pod.Namespace, pod.Name))
 	}
 
 	return framework.NewStatus(framework.Success, "")
@@ -316,7 +378,7 @@ func (c *CapacityScheduling) preempt(ctx context.Context, state *framework.Cycle
 	}
 
 	// 1) Ensure the preemptor is eligible to preempt other pods.
-	if !defaultpreemption.PodEligibleToPreemptOthers(pod, nodeLister, m[pod.Status.NominatedNodeName]) {
+	if !c.PodEligibleToPreemptOthers(pod, nodeLister, m[pod.Status.NominatedNodeName], state) {
 		klog.V(5).InfoS("Pod is not eligible for more preemption.", "pod", klog.KObj(pod))
 		return "", nil
 	}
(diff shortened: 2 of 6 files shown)


pkg/apis/config/v1beta1/defaults.go
```diff
@@ -95,9 +95,7 @@ func SetDefaultsNodeResourcesAllocatableArgs(obj *NodeResourcesAllocatableArgs)
 
 // SetDefaultsCapacitySchedulingArgs sets the default parameters for CapacityScheduling plugin.
 func SetDefaultsCapacitySchedulingArgs(obj *CapacitySchedulingArgs) {
-	if obj.KubeConfigPath == nil {
-		obj.KubeConfigPath = &defaultKubeConfigPath
-	}
+	// TODO(k/k#96427): get KubeConfigPath and KubeMaster from configuration or command args.
```
denkensk, line 98: use the inClusterConfig instead of default KubeConfig Path then we can deploy the scheduler in deployment with serviceaccount.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -605,6 +641,66 @@ func selectVictimsOnNode(
...
+// - Reject the victim pods if they are in waitingPod map
+// - Clear the low-priority pods' nominatedNodeName status if needed
+func (c *CapacityScheduling) PrepareCandidate(candidate Candidate, fh framework.Handle, cs kubernetes.Interface, pod *v1.Pod) error {
+	for _, victim := range candidate.Victims().Pods {
+		if err := util.DeletePod(cs, victim); err != nil {
+			klog.Errorf("Error preempting pod %v/%v: %v", victim.Namespace, victim.Name, err)
+			return err
+		}
+
+		c.deletePodIfPresent(pod)
```
denkensk, line 656: Apart from this line, it is no different from upstream in `PrepareCandidate` and `getLowerPriorityNominatedPods`

Huang-Wei (commented):
I'd like a UT or integration test to reproduce the issue.

pkg/capacityscheduling/elasticquota.go
```diff
@@ -52,20 +52,22 @@ func (e ElasticQuotaInfos) aggregatedMinOverUsedWithPod(podRequest framework.Res
...
 // Each namespace can only have one ElasticQuota.
 type ElasticQuotaInfo struct {
-	Namespace string
-	pods      sets.String
-	Min       *framework.Resource
-	Max       *framework.Resource
-	Used      *framework.Resource
+	Namespace    string
+	pods         sets.String
+	deletingPods map[string]*v1.Pod
```
Huang-Wei, line 57: Not quite a fan of using this variable. Could you explain the reasons? Why not just update the cache upon receiving Pod Deleting events.

Huang-Wei (commented):
Still, I strongly recommend we craft a test to trigger the original issue.

Huang-Wei:
@denkensk let me know when it's ready for another round of review.

pkg/apis/config/v1beta1/defaults.go
denkensk: If the user wants to deploy scheduler-plugins as a unified scheduler binary, he can set the kubeconfig path in `KubeSchedulerConfiguration`.  
This issue is that if we set the default KubeConfig Path, we can't set up the scheduler as deployment in https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/install/charts, because the default path is different with inClusterConfig.

pkg/capacityscheduling/elasticquota.go
denkensk: I delete the variable to store the deleting pods.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -201,11 +222,50 @@ func (c *CapacityScheduling) PreFilter(ctx context.Context, state *framework.Cyc
...
+		for _, p := range nominatedPods {
+			ns := p.Namespace
+			info := c.elasticQuotaInfos[ns]
+			if info != nil {
+				pResourceRequest := computePodResourceRequest(p).ResourceList()
+				// If they are subject to the same quota(namespace) and p is more important than pod,
+				// p will be added to the nominatedResource and totalNominatedResource.
+				// If they aren't subject to the same quota(namespace) and the usage of quota(p's namespace) does not exceed min,
+				// p will be added to the totalNominatedResource.
+				if ns == pod.Namespace && p.UID != pod.UID && corev1helpers.PodPriority(p) >= corev1helpers.PodPriority(pod) {
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -84,6 +85,24 @@ func (s *ElasticQuotaSnapshotState) Clone() framework.StateData {
...
+}
+
+type TotalNominatedResourceWithPodState struct {
+	framework.Resource
+}
+
+// Clone the TotalNominatedResourceWithPodState state.
+func (t *TotalNominatedResourceWithPodState) Clone() framework.StateData {
+	return t
+}
```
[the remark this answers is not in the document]
denkensk: Done

pkg/capacityscheduling/capacity_scheduling_test.go
```diff
@@ -123,8 +123,26 @@ func TestPreFilter(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			var registerPlugins []st.RegisterPluginFunc
```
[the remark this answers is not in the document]
denkensk: I add this as an integration test.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -577,7 +655,7 @@ func selectVictimsOnNode(
 			klog.V(5).Infof("Pod %v/%v is a potential preemption victim on node %v.", p.Namespace, p.Name, nodeInfo.Node().Name)
 		}
 
-		if preemptorWithElasticQuota && (preemptorElasticQuotaInfo.overUsed(preFilterState.Resource, preemptorElasticQuotaInfo.Max) || elasticQuotaInfos.aggregatedMinOverUsedWithPod(preFilterState.Resource)) {
+		if preemptorWithElasticQuota && (preemptorElasticQuotaInfo.overUsed(nominatedResourceWithPod, preemptorElasticQuotaInfo.Max) || elasticQuotaInfos.aggregatedMinOverUsedWithPod(totalNominatedResourceWithPod)) {
```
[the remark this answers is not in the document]
denkensk: the nominated pods isn't the running pods in node. I think we may needn't to update the nominated pods when addPods/deletePods

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -190,23 +196,66 @@ func New(obj runtime.Object, handle framework.Handle) (framework.Plugin, error)
...
+	nominatedPodsReqWithPodReq := &framework.Resource{}
+
+	nodeList, err := c.frameworkHandle.SnapshotSharedLister().NodeInfos().List()
+	if err != nil {
+		return framework.NewStatus(framework.Error, fmt.Sprintf("Error getting the nodelist: %v", err))
 	}
 
-	if elasticQuotaInfos.aggregatedMinOverUsedWithPod(preFilterState.Resource) {
-		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("Pod %v/%v is rejected in Prefilter because total ElasticQuota used is more than min", pod.Namespace, pod.Name))
+	for _, node := range nodeList {
```
denkensk, line 220: The root cause is that I didn't think about the nominated pods before. This will lead to repeated over preemption.
@Huang-Wei

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -201,11 +222,50 @@ func (c *CapacityScheduling) PreFilter(ctx context.Context, state *framework.Cyc
...
+		}
+	}
+
+	nominatedResource.Add(preFilterState.Resource.ResourceList())
+	totalNominatedResource.Add(preFilterState.Resource.ResourceList())
+
+	state.Write(NominatedResourceWithPodStateKey, &NominatedResourceWithPodState{Resource: *nominatedResource})
+	state.Write(TotalNominatedResourceWithPodStateKey, &TotalNominatedResourceWithPodState{Resource: *totalNominatedResource})
+
+	if eq.overUsed(*nominatedResource, eq.Max) {
```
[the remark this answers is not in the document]
denkensk: nominatedResourceWithPods is nominatedResource + podReq

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -201,11 +202,46 @@ func (c *CapacityScheduling) PreFilter(ctx context.Context, state *framework.Cyc
 		return framework.NewStatus(framework.Success, "skipCapacityScheduling")
 	}
 
-	if eq.overUsed(preFilterState.Resource, eq.Max) {
+	// nominatedResource is the sum of the requested resources of the Nominated Pods
```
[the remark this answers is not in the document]
denkensk: The dryRunPreemption checks the whether a preemption strategy works or not in func`reprievePod` before.

denkensk:
@Huang-Wei 
Please take a look again. Thanks

denkensk:
ping @Huang-Wei again~
Thanks

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -64,7 +64,13 @@ type CapacityScheduling struct {
 
 // PreFilterState computed at PreFilter and used at PostFilter or Reserve.
 type PreFilterState struct {
-	framework.Resource
+	podReq framework.Resource
+	// nominatedPodsReqInEQWithPodReq is the sum of the requested resources of the Nominated Pods
+	// which subject to the same quota(namespace) and podReq.
+	nominatedPodsReqInEQWithPodReq framework.Resource
+	// nominatedPodsReqWithPodReq is the sum of the requested resources of the Nominated Pods
+	// which subject to the all quota(namespace) and podReq.
```
Huang-Wei, line 72: Actually, you should also mention only the "guaranteed" nominated pod that don't exceed their min quota are counted here. Because the logic checks `!info.usedOverMin()`.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -342,6 +393,83 @@ func (c *CapacityScheduling) preempt(ctx context.Context, state *framework.Cycle
...
+		if err != nil {
+			klog.Errorf("error reading %q from cycleState: %v", ElasticQuotaSnapshotKey, err)
+			return true
+		}
+		preemptorElasticQuotaInfo, preemptorWithElasticQuota := elasticQuotaSnapshotState.elasticQuotaInfos[pod.Namespace]
+
+		var moreThanMinWithPreemptor bool
+		// Check if there is elastic quota in the preemptor's namespace.
+		if preemptorWithElasticQuota {
+			moreThanMinWithPreemptor = preemptorElasticQuotaInfo.overUsed(preFilterState.nominatedPodsReqWithPodReq, preemptorElasticQuotaInfo.Min)
```
Huang-Wei, line 432: Over used `nominatedPodsReqWithPodReq` or `nominatedPodsReqInEQWithPodReq`? Please add some comments here.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -342,6 +393,83 @@ func (c *CapacityScheduling) preempt(ctx context.Context, state *framework.Cycle
...
+				if p.Pod.DeletionTimestamp != nil {
+					pWithElasticQuotaInfo, pWithElasticQuota := elasticQuotaSnapshotState.elasticQuotaInfos[p.Pod.Namespace]
+					if !pWithElasticQuota {
+						continue
+					}
+					if p.Pod.Namespace == pod.Namespace && corev1helpers.PodPriority(p.Pod) < podPriority {
+						// There is a terminating pod on the nominated node.
+						return false
+					} else if p.Pod.Namespace != pod.Namespace && !moreThanMinWithPreemptor && pWithElasticQuotaInfo.usedOverMin() {
+						// There is a terminating pod on the nominated node.
```
Huang-Wei, line 451: This comment needs to be updated.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -342,6 +393,83 @@ func (c *CapacityScheduling) preempt(ctx context.Context, state *framework.Cycle
...
+			for _, p := range nodeInfo.Pods {
+				if p.Pod.DeletionTimestamp != nil {
+					pWithElasticQuotaInfo, pWithElasticQuota := elasticQuotaSnapshotState.elasticQuotaInfos[p.Pod.Namespace]
+					if !pWithElasticQuota {
+						continue
+					}
+					if p.Pod.Namespace == pod.Namespace && corev1helpers.PodPriority(p.Pod) < podPriority {
+						// There is a terminating pod on the nominated node.
+						return false
+					} else if p.Pod.Namespace != pod.Namespace && !moreThanMinWithPreemptor && pWithElasticQuotaInfo.usedOverMin() {
```
Huang-Wei, line 450: `!moreThanMinWithPreemptor && pWithElasticQuotaInfo.usedOverMin()` indicates its own EQ's usage is over min, while also globally nominated pods' total request is less than min? Could you elaborate more here?

Huang-Wei (commented):
Sry for the late response.

pkg/capacityscheduling/capacity_scheduling.go
denkensk: The nominated pods are not used only for the preparation of preemption. They are also used to determine whether there are enough quota to schedule the pod in prefilter.

denkensk:
@Huang-Wei 
Please help to review it again. Sorry for the delay.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -347,6 +407,87 @@ func (c *CapacityScheduling) preempt(ctx context.Context, state *framework.Cycle
...
+						// There is a terminating pod on the nominated node.
+						// If the terminating pod is in the same namespace with preemptor
+						// and it is less important than preemptor,
+						// return false to avoid preempting more pods.
+						return false
+					} else if p.Pod.Namespace != pod.Namespace && !moreThanMinWithPreemptor && pWithElasticQuotaInfo.usedOverMin() {
+						// There is a terminating pod on the nominated node.
+						// The terminating pod isn't in the same namespace with preemptor.
+						// If moreThanMinWithPreemptor is false, it indicates that preemptor can preempt the pods belong to the namespace whose quota is over used than min.
+						return false
```
Huang-Wei, line 471: Why return false?

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -478,7 +623,9 @@ func selectVictimsOnNode(
 	var moreThanMinWithPreemptor bool
 	// Check if there is elastic quota in the preemptor's namespace.
 	if preemptorWithElasticQuota {
-		moreThanMinWithPreemptor = preemptorElasticQuotaInfo.overUsed(preFilterState.Resource, preemptorElasticQuotaInfo.Min)
+		nominatedPodsReqInEQWithPodReq = preFilterState.nominatedPodsReqInEQWithPodReq
+		nominatedPodsReqWithPodReq = preFilterState.nominatedPodsReqWithPodReq
+		moreThanMinWithPreemptor = preemptorElasticQuotaInfo.overUsed(nominatedPodsReqInEQWithPodReq, preemptorElasticQuotaInfo.Min)
```
Huang-Wei, line 628: Ditto. This can be moved below L635.

Huang-Wei (commented):
The core logic looks good to me - ensure (1) the preemptor EQ's `used` not exceed its `max`, including the nominated pods with higher priority in this EQ (b/c preemptor cannot preempt them), and (2) aggregated `used` not exceed `min`, including the nominated pods in case (1) and the nominated pods in other EQs that don't fully utilized their `min` (so the preemtor cannot preempt them either).

In addition to the comments below. Some logic can be du-duplicated like https://gist.github.com/Huang-Wei/9ad93a67f3de31f094253f99b0a516e9.

pkg/capacityscheduling/capacity_scheduling.go
denkensk: I add more comments in the code.
1. The terminating pod isn't in the same namespace with preemptor.
2. If moreThanMinWithPreemptor is false, it indicates that preemptor can preempt the pods belong to the namespace whose quota is over used than min.
3. And if the used of terminating pod's quota is over min, so the space released by terminating pod on the nominated node can be used by the preemptor.
4. return false to avoid preempting more pods.

pkg/capacityscheduling/elasticquota.go
```diff
@@ -103,6 +103,23 @@ func (e *ElasticQuotaInfo) overUsed(podRequest framework.Resource, resource *fra
...
+	if e.Used.MilliCPU > e.Min.MilliCPU {
+		return true
+	}
+
+	if e.Used.Memory > e.Min.Memory {
+		return true
+	}
+
+	for rName, rQuant := range e.Min.ScalarResources {
+		if rQuant > e.Min.ScalarResources[rName] {
```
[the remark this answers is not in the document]
denkensk: Yes.. thanks for your catch

denkensk:
@Huang-Wei 
Thanks for your serious comments. I modified the code based on your comments.
Please review it again. Thanks so much.

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: This is still outstanding. `moreThanMinWithPreemptor` can be moved to its most close usage, which is a few lines below that did another `preemptorWithEQ` check.

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: thanks for the explanation.

```
// If moreThanMinWithPreemptor is false, it indicates that preemptor can preempt the pods belong to the namespace whose quota is over used than min.
```

Can we reword to:

```
// If moreThanMinWithPreemptor is false, it indicates that preemptor can preempt the pods in other EQs whose used is over min.
```

pkg/capacityscheduling/capacity_scheduling.go
Huang-Wei: L626 and L627 should also be moved.

Huang-Wei (commented):
Thanks @denkensk.  LGTM except for some final nits commented above.

denkensk:
@Huang-Wei 
Please review it again. Thanks so much.
I will squash commits if everything is ok.

Huang-Wei:
LGTM. Please squash the commits.

denkensk:
@Huang-Wei 
Squashed. Thanks.

Huang-Wei:
/lgtm
/approve

Thanks @denkensk !

PR #215, opened 2021-07-19
Add comments for NodeResourceTopologyMatchArgs

AlexeyPerevalov:
It mostly code cleanup, we missed it in our initial commit, but other plugins have such comments.


swatisehgal:
Looks good to me!

Issue #199, opened 2021-06-04
Scheduler plugin takes Nodes' Pods CIDR in into account


lianghao208:
Is that possible to take nodes' Pods CIDR in into account in scheduling phase?
I have met this situation multiple times:
Pods were bond to Nodes without enough Pod IPs to assign.
It leads to a fatal problem that Pods can't run normally since there was no IP to assign.
So is that possible to check if Nodes has enough Pod IPs to assign in scheduler plugin?

denkensk:
> Pods were bond to Nodes without enough Pod IPs to assign.
---

Yes. This is common when the density of pods in a node is relatively high. We set the MaxPod of node as the limit of pods   to address this issue. But I think we can also try to solve it in scheduler.

lianghao208:
> > Pods were bond to Nodes without enough Pod IPs to assign.
> 
> Yes. This is common when the density of pods in a node is relatively high. We set the MaxPod of node as the limit of pods to address this issue. But I think we can also try to solve it in scheduler.

@denkensk Can I solve this issue by implementing a PreFilter plugin to ignore the nodes that have no IP to allocate( Simply determine if the pod already used up all the PodCIDR field from nodes resource).

denkensk:
Yes. I think it makes sense.  You can start with a simple proposal for your idea.

wangyanghack:
There is situation that some cni plugin using ENI network interface to get the performance equivalent to the underlying network. Quantity of pods in a node depending on quantity of ENI and quantity of ip every ENI can be attached. So is that possible to check if Nodes has enough ENIs and free IPs on ENI  to assign in scheduler plugin?

lianghao208:
> So is that possible to check if Nodes has enough ENIs and free IPs on ENI to assign in scheduler plugin?
Yes, I think it needs some places to store each node's allocatable IPs for pods, either in node's `status.allocatable` field or in a CRD, WDYT?

wangyanghack:
> > So is that possible to check if Nodes has enough ENIs and free IPs on ENI to assign in scheduler plugin?
> 
> @wangyanghack Yes, I think it needs some places to store each node's allocatable IPs for pods, either in node's `status.allocatable` field or in a CRD, WDYT?

@lianghao208 Yes, I agree with you. Checking nodes in cluster who use ENIs cannot use .spec.PodCIDR to calculate node's allocatable IPs for pods.

PR #219, opened 2021-07-24
KEP: Pod Allocatable IP Plugin

lianghao208:
According to the discussion in #199 , I submit this kep.

### DIFF
--- a/kep/219-pod-allocatable-ip-scheduling/README.md
+++ b/kep/219-pod-allocatable-ip-scheduling/README.md
@@ -1,0 +1,49 @@
+# Pod Allocatable IP<!-- omit in toc -->
+
+## Table of Contents <!-- omit in toc -->
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Use Cases](#use-cases)
+- [Design Details](#design-details)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+This plugin takes node's `.spec.PodCIDR` into account in scheduling cycle. When the density of pods in a node is relatively high, nodes might not have sufficient allocatable Pod IP to allocate.
+Even though kubelet has its own `--max-pods` option flag to limit the total number of Pods on nodes, we can filter those nodes who have insufficient Pod IP to allocate in scheduling cycle. 
+
+## Motivation
+[kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) provides the `--allocate-node-cidrs` options to mark Pod CIDR in each node's `.spec.PodCIDR` field.
+When a node has already allocated all the Pod IPs from `.spec.PodCIDR`, a scheduling Pod will ignore this node in scheduling cycle. It avoids the situation that pods are binding to insufficient Pod IP nodes,
+which is useful when a Pod cannot be scheduled due to the limited Pod IP allocation constraint in one node. This constraint is likely to result in Pods stucking in `ContainerCreating` phase.
+
+### Goals
+
+1. Implement a Filter plugin to filter out the nodes who have already allocated all the Pod IPs from `.spec.PodCIDR`.
+2. This plugin only works for those clusters who use standard CNI plugins and enable the `--allocate-node-cidrs` options in [kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/), which means all pod IPs are strictly allocated from each node's `.spec.PodCIDR`.
+
+### Non-Goals
+
+1. Considering each node's Pod Allocatable IP in those clusters who use Elastic Network Interface(ENI) for Pod.
+2. Considering each node's Pod Allocatable IP in those clusters whose `--allocate-node-cidrs` options of kube-controller-manager is disabled, which means the PodCIDR is not recorded into node resources.
+
+## Use Cases
+
+Make sure to set the `--allocate-node-cidrs=true --cluster-cidr=<cidr>` options in kube-controller-manager and use standard [CNI plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/network-plugins/). Each node would be assigned an IP subnet through the `--cluster-cidr=<cidr>` configuration.
+For the node resources, `.spec.PodCIDR` field will be set as well.
+
+## Design Details
+
+We implement a filter plugin based on the scheduler framework. 
+
+After nodes were joining into the cluster, their Pod IP subnet was assigned, and the `.spec.PodCIDR` field in node resources was set as well.
+
+When a Pod is in scheduling cycle, the Pod Allocatable IP Filter Plugin will check if nodes has sufficient allocatable Pod IP to assign according its `.spec.PodCIDR` field.
+
+Comparing each node's `.spec.PodCIDR` field and Pod IPs that have already been assigned, the Plugin calculates the number of allocatable Pod IP, filters out those nodes who have insufficient allocatable Pod IP to assign.
+
+## Implementation History
--- a/kep/219-pod-allocatable-ip-scheduling/README.md
+++ b/kep/219-pod-allocatable-ip-scheduling/README.md
@@ -14,17 +14,16 @@
 ## Summary
 
 This plugin takes a node's `.spec.PodCIDR` into account in a scheduling cycle. When the density of pods in a node is relatively high, the node might not have sufficient allocatable Pod IPs to allocate.
-Even though kubelet has its own `--max-pods` option flag to limit the total number of Pods on nodes, we can filter those nodes who have insufficient Pod IP to allocate in scheduling cycle. 
+Even though kubelet has its own `--max-pods` option flag to limit the total number of Pods on nodes, we can filter out those nodes who have insufficient Pod IPs to allocate in Pods' `ContainerCreating` phase. 
 
 ## Motivation
 [kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) provides the `--allocate-node-cidrs` options to mark Pod CIDR in each node's `.spec.PodCIDR` field.
 When a node has already allocated all the Pod IPs from `.spec.PodCIDR`, a scheduling Pod will ignore this node in scheduling cycle. It avoids the situation that pods are bound to nodes without insufficient Pod IPs,
-which is useful when a Pod cannot be scheduled due to the limited Pod IP allocation constraint in one node. This constraint is likely to result in Pods stucking in `ContainerCreating` phase.
+which is useful when a Pod cannot be scheduled due to the limited Pod IP allocation constraint in one node. Without it, a pod can be scheduled and then stuck in `ContainerCreating` phase.
 
 ### Goals
 
-1. Implement a Filter plugin to filter out the nodes who have already allocated all the Pod IPs from `.spec.PodCIDR`.
-2. This plugin only works for those clusters who use standard CNI plugins and enable the `--allocate-node-cidrs` options in [kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/), which means all pod IPs are strictly allocated from each node's `.spec.PodCIDR`.
+- Implement a Filter plugin to filter out the nodes who have already allocated all the Pod IPs from `.spec.PodCIDR`.
 
 ### Non-Goals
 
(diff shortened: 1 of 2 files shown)


kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Design Details
+
+We implement a filter plugin based on the scheduler framework. 
+
+After nodes were joining into the cluster, their Pod IP subnet was assigned, and the `.spec.PodCIDR` field in node resources was set as well.
+
+When a Pod is in scheduling cycle, the Pod Allocatable IP Filter Plugin will check if nodes has sufficient allocatable Pod IP to assign according its `.spec.PodCIDR` field.
+
+Comparing each node's `.spec.PodCIDR` field and Pod IPs that have already been assigned, the Plugin calculates the number of allocatable Pod IP, filters out those nodes who have insufficient allocatable Pod IP to assign.
```
yuanchen8911, line 47: How to calculate the number of allocatable IPs? Please add details.

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,49 @@
...
+<!-- /toc -->
+
+## Summary
+
+This plugin takes node's `.spec.PodCIDR` into account in scheduling cycle. When the density of pods in a node is relatively high, nodes might not have sufficient allocatable Pod IP to allocate.
+Even though kubelet has its own `--max-pods` option flag to limit the total number of Pods on nodes, we can filter those nodes who have insufficient Pod IP to allocate in scheduling cycle. 
+
+## Motivation
+[kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) provides the `--allocate-node-cidrs` options to mark Pod CIDR in each node's `.spec.PodCIDR` field.
+When a node has already allocated all the Pod IPs from `.spec.PodCIDR`, a scheduling Pod will ignore this node in scheduling cycle. It avoids the situation that pods are binding to insufficient Pod IP nodes,
```
yuanchen8911, line 21:
```suggestion
When a node has already allocated all the Pod IPs from `.spec.PodCIDR`, a scheduling Pod will ignore this node in scheduling cycle. It avoids the situation that pods are bound to nodes without insufficient Pod IPs,
```

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,49 @@
...
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Use Cases](#use-cases)
+- [Design Details](#design-details)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+This plugin takes node's `.spec.PodCIDR` into account in scheduling cycle. When the density of pods in a node is relatively high, nodes might not have sufficient allocatable Pod IP to allocate.
```
yuanchen8911, line 16:
```suggestion
This plugin takes a node's `.spec.PodCIDR` into account in a scheduling cycle. When the density of pods in a node is relatively high, the node might not have sufficient allocatable Pod IPs to allocate.
```

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,49 @@
...
+Even though kubelet has its own `--max-pods` option flag to limit the total number of Pods on nodes, we can filter those nodes who have insufficient Pod IP to allocate in scheduling cycle.
```
yuanchen8911, line 17: a scheduling cycle?

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Summary
+
+This plugin takes node's `.spec.PodCIDR` into account in scheduling cycle. When the density of pods in a node is relatively high, nodes might not have sufficient allocatable Pod IP to allocate.
+Even though kubelet has its own `--max-pods` option flag to limit the total number of Pods on nodes, we can filter those nodes who have insufficient Pod IP to allocate in scheduling cycle. 
+
+## Motivation
+[kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) provides the `--allocate-node-cidrs` options to mark Pod CIDR in each node's `.spec.PodCIDR` field.
+When a node has already allocated all the Pod IPs from `.spec.PodCIDR`, a scheduling Pod will ignore this node in scheduling cycle. It avoids the situation that pods are binding to insufficient Pod IP nodes,
+which is useful when a Pod cannot be scheduled due to the limited Pod IP allocation constraint in one node. This constraint is likely to result in Pods stucking in `ContainerCreating` phase.
```
yuanchen8911, line 22: How about "Without it, a pod can be scheduled and then stuck in `ContainerCreating` phase."

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Goals
+
+1. Implement a Filter plugin to filter out the nodes who have already allocated all the Pod IPs from `.spec.PodCIDR`.
+2. This plugin only works for those clusters who use standard CNI plugins and enable the `--allocate-node-cidrs` options in [kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/), which means all pod IPs are strictly allocated from each node's `.spec.PodCIDR`.
```
yuanchen8911, line 27: How about clarifying it in the `Limitations` section?

kep/219-pod-allocatable-ip-scheduling/README.md
yuanchen8911: Questions. 
1. When and how is a node's `spec.PodCIDR` updated? 
2. Will race conditions be possible?  
 - A node is considered feasible in filtering/scoring phases, but its allocatable IPs is updated in binding and late phases. 
 - A node is chosen for a pod in a scheduling cycle.  Before the pods is bound or created, `spec.PodCIDR` will not be updated and its value won't reflect the accurate allocatable IPs on the node. The next scheduling cycle will mis-calculate the allocatable IPs on the node . 

Do you need additional plugins (Permit/Reserve/Unreserve/PostBind) to handle IP state?

yuanchen8911 (changes requested):
Additional design/implementation details are needed.

kep/219-pod-allocatable-ip-scheduling/README.md
lianghao208: > When and how is a node's spec.PodCIDR updated?

I think the node's `spec.PodCIDR` updated only when users manually modify the field to change the ip pool for pod, or `--cluster-cidr` has been modified in kube-controller-manager. Under both circumstance, pods needed to be re-created and pod IPs needed to be re-assigned as well.

lianghao208: > 2. Will race conditions be possible?
> 
> * A node is considered feasible in filtering/scoring phases, but its allocatable IPs is updated in binding and late phases.
> * A node is chosen for a pod in a scheduling cycle.  Before the pods is bound or created, `spec.PodCIDR` will not be updated and its value won't reflect the accurate allocatable IPs on the node. The next scheduling cycle will mis-calculate the allocatable IPs on the node .
> 
> Do you need additional plugins (Permit/Reserve/Unreserve/PostBind) to handle IP state?

Yes, this is a good suggestion, I think we need an extra PreBind plugin to double check the allocatable IPs. This will make sure the number of allocatable IPs  is accurate before the binding.
@yuanchen8911

lianghao208: > How to calculate the number of allocatable IPs? Please add details.

done. Thanks for the suggestions and reviews!
@yuanchen8911

yuanchen8911: > > When and how is a node's spec.PodCIDR updated?
> 
> @yuanchen8911 I think the node's `spec.PodCIDR` updated only when users manually modify the field to change the ip pool for pod, or `--cluster-cidr` has been modified in kube-controller-manager. Under both circumstance, pods needed to be re-created and pod IPs needed to be re-assigned as well.

1. When is a new pod's `spec.PodCIDR` updated,  binding or creating phase? Which controller or scheduler does it? 
2. When an existing pod is deleted, which controller is responsible for updating the field, e.g., reclaiming the IP and making it allocatable?

lianghao208: > When is a new pod's spec.PodCIDR updated, binding or creating phase? Which controller or scheduler does it?

`spec.PodCIDR` is a static field in node resources to record Pod IP subnet, it won't be updated after assigning IPs for pod. 

> When an existing pod is deleted, which controller is responsible for updating the field, e.g., reclaiming the IP and making it allocatable?

As what I was mentioned in ` Design Details ` section, the plugin starts a client to list all the Pods that has been scheduled in this node, which can calculate the accurate allocatable IPs. ( There is no need to concern about the  ip reclaiming or assigning mechanism)

@yuanchen8911

yuanchen8911: OK，I guess I misunderstood the meaning of the field. It specifies the maximum number of allocable IPs of a node, which is set statically. Right?

lianghao208: Yes, it is set statically.
@yuanchen8911

yuanchen8911: I'm afraid a node's allocated IPs is not equivalent to the number of running pods from scheduling perspective. The scheduler should also consider pods that are successfully scheduled but not created yet. Otherwise, potential race condition can happen as mentioned in a previous comment. 

Once the scheduler successfully schedules a pod and moves it to binding phase, allocated IPs should be deducted by 1 because the a new scheduling cycle will start. 

Once a pod is created and starts running in a node, the scheduler needs to update the allocated IPs. Otherwise, allocated IPs will be double counted. 

Another case is if a scheduled pod is stuck in containerCreating, When should its allocated IP be recycled? What about if it's deleted before being created? 

So the scheduler needs to watch for pod addition and deletion events and update allocated IP count accordingly.

lianghao208: > I'm afraid a node's allocated IPs is not equivalent to the number of running pods from scheduling perspective. The scheduler should also consider pods that are successfully scheduled but not created (running) yet. Otherwise, a race condition can happen as mentioned in a previous comment.
> 

The plugin is not just the considering the running pod. Actually, all the pods that has been bound (the pods' `nodeName` field were set ) should be taken into account.

> After the scheduler successfully schedules a pod and moves it to the binding phase, allocatable IPs should be deducted by 1 because the a new scheduling cycle will start and the scheduled pod is not running yet. Once the pod starts running, the scheduler needs to update the node's allocatable IPs too (plus 1). Otherwise, the allocated IPs will be double counted (i.e., IP leak).
> 

That's why we need a preBind plugin to check the node's allocatable IPs right before the pods' binding phase in the binding cycle.

> Another scenario is if a scheduled pod is stuck in containerCreating, when will its allocated IP be recycled? What about if the pod is deleted before it's created?

In this scenario, we consider this Pod has already taken a pod IP assignment quota from node (no need to recycle the allocated IP). Even though Pod IP has not been assigned to this pod yet, we assume the IP assignment will happen eventually as soon as other Pods are deleted. 
If a node has pods stuck in containerCreating for the insufficient ip reasons, no other pods can be scheduled to this node either.

> I think the scheduler will need to watch for pod addition and deletion events and update allocated or allocatable IP count accordingly.

If counting the number of those pods who have the same `nodeName` field value can prove how many pods have already bound to the same nodes, I don't think this is necessary.

@yuanchen8911

yuanchen8911: Assume a node's IP capacity is 8 with 4 pods running and a new pod is submitted (will be placed on the node). Here are the allocated IP changes.

1. Before pod scheduling: 4
2. After pod scheduling: 5 (!=#running pods; update must be done before prebind because a new schedule cycle will start after that)
3. If the scheduled pod is deleted before running: 4;
If the pod is created and running: 5

Without tracking pod creation and deletion events, how can you tell which case it is and hence update allocated IPs differently for the next pod?

lianghao208: > After pod scheduling and before pod creation: 5 (!= num. of running pods; 

If this pod has finished its binding cycle, the allocatable IP will be 3 (8-5=3).
If this pod has not finished binding to this node, the allocatable IP will still be 4 (8-4=4).
It doesn't matter if this pod is running or not, as long as this pod has finished bound to a node after a binding cycle, num of node's allocatable IPs will minus 1.

> update must be done before prebind (e.g. reservation) because a new schedule cycle will start after that)
> Without tracking pods' creation and deletion events, how can you tell which case it is and hence update allocated IPs differently?

we don't need a place to store each node's allocated IP, the preBind plugin checks the number of pods that have already been bound to the node ( only when a pod finished its binding cycle that can have its`.nodeName` field be set) to determine if this node has allocatable IPs to assign.

@yuanchen8911

yuanchen8911: Maybe we need a discussion offline. 

Additional detail with examples about how to track IP count in the KEP would be helpful anyway.

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+We implement filter plugin and preBind plugin based on the scheduler framework.
+
+After nodes were joining into the cluster, their Pod IP subnet was assigned, and the `.spec.PodCIDR` field in node resources was set as well.
+
+When a Pod is in scheduling cycle, the Plugin will check if nodes has sufficient allocatable Pod IP to assign according its `.spec.PodCIDR` field.
+
+In the scheduling cycle, the filter plugin compares each node's `.spec.PodCIDR` field and Pod IPs that have already been assigned, the Plugin calculates the number of allocatable Pod IP, filters out those nodes who have insufficient allocatable Pod IPs to assign.
+
+In the binding cycle, the preBind plugin double check the number of allocatable Pod IPs in the chosen node to avoid the race conditions (e.g., A node is chosen for Pod A in a scheduling cycle. Before the Pod A's binding cycle, Pod B has been bound to this node in its own binding cycle, this could result in mis-calculation of allocatable IPs on the node)
```
denkensk, line 53: Why we need to double check the number of allocatable Pod IP? 
PodA is assumed in reserve plugins. It will be calculated in filter plugin of podB scheduling cycle. Should we default that PodA will use an IP address here？

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+The allocatable Pod IP calculation method is shown as follows:
+
+1. Obtain the node's Pod IP subnet from its own `.spec.PodCIDR` field (e.g., subnet `192.168.1.0/29` has 6 valid IP to assign).
+2. Obtain the number of Pods that have already been bound to this node. To achieve that, the plugin starts a client to list all the Pods that has been scheduled in this node, and then filter out those Pods in hostNetwork (e.g., there are 3 Pods have been bound to this node).
```
denkensk, line 58: I think It is better not to request the api server during scheduling cycle?
Consider maintaining a cache？

denkensk (commented):
Thanks 
Some comments

kep/219-pod-allocatable-ip-scheduling/README.md
lianghao208: > PodA is assumed in reserve plugins. It will be calculated in filter plugin of podB‘s scheduling cycle. Should we default that PodA will use an IP address here？

I agree, implementing a reserve plugin is a more reasonable way to avoid the race conditions. 
@denkensk

kep/219-pod-allocatable-ip-scheduling/README.md
lianghao208: What if we maintaining a cache to store the number of allocated IPs including:
1. pods in reserve plugin
2. pods already bound to the nodes
@denkensk

kep/219-pod-allocatable-ip-scheduling/README.md
```diff
@@ -0,0 +1,61 @@
...
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Use Cases](#use-cases)
+- [Design Details](#design-details)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+## Summary
+
+This plugin takes a node's `.spec.PodCIDR` into account in a scheduling cycle. When the density of pods in a node is relatively high, the node might not have sufficient allocatable Pod IPs to allocate.
```
Huang-Wei, line 16: For now, the defeault scheduler only takes `.status.Allocatable["pods"]` into consideration when scheduling. Is this KEP's main purpose is to switch to use `min(num of .spec.PodCIDR, .status.Allocatable["pods"])`?

lianghao208: Yes, that's right. Any suggestions on the implementation details?

Huang-Wei: I think that's a generic improvement and we can achieve it in the default scheduler. Do you want to pursue it in k/k?

lianghao208: Yes, I would like to do it. What should I do to pursue this in k/k? @Huang-Wei

Huang-Wei: You can raise an issue describing the idea, and involve sig-scheduling and sig-node to see if it makes sense. No kep is needed as the change is small. After we get to a consensus, you can start the implementation.

PR #220, opened 2021-07-26
Score Plugin for Topology Aware Scheduler

Tal-or:
Score plugin allows the topology-aware scheduler to have better pod scheduling decisions.
By ranking all suitable nodes based on a certain scoring strategy, the scheduler will be able to select the finest node as the candidate for the pod assignment.

Available strategies are:
- "Least Allocated": Favors node with the least amount of available resource.
- "Balanced Allocation": Favors nodes with balanced resource usage rate.
- "Most Allocated": Favors node with the most amount of available resource

All strategies are a simplified version of the in-tree Kubernetes noderesources scheduling plugins
https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/framework/plugins/noderesources

Proposal doc:
https://drive.google.com/file/d/1XqCqRvvTLdt7uFzfCkLxPW139ojJm3AM/view?usp=sharing

KEP:
https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/119-node-resource-topology-aware-scheduling


cmd/scheduler/main.go
```diff
@@ -48,7 +48,7 @@ func main() {
 		app.WithPlugin(coscheduling.Name, coscheduling.New),
 		app.WithPlugin(loadvariationriskbalancing.Name, loadvariationriskbalancing.New),
 		app.WithPlugin(noderesources.AllocatableName, noderesources.NewAllocatable),
-		app.WithPlugin(noderesourcetopology.Name, noderesourcetopology.New),
+		app.WithPlugin(filter.Name, filter.New),
```
Huang-Wei, line 53: I understand you want to tease apart the filter and score logic, but it doesn't look quite necessary as they share some commons structs and should semantically be placed into the same package IMO.

A reference is the upstream [podtoplogyspread](https://github.com/kubernetes/kubernetes/tree/master/pkg/scheduler/framework/plugins/podtopologyspread) filter and score plugins. The folder can be kept as noderesourcetopology, and you can put the logic into different files: plugin.go, filter.go, score.go, common.go, etc.

manifests/noderesourcetopology/scheduler-config.yaml
```diff
@@ -19,3 +22,11 @@ profiles:
       kubeconfigpath: "/etc/kubernetes/scheduler.conf"
       namespaces:
       - "test-namespace"
+  - name: NodeResourceTopologyResourceAllocationScore
```
Huang-Wei, line 25: Again, we can one plugin to serve as both filter and score plugins, and if so the args can be merged into NodeResourceTopologyMatchArgs.

pkg/apis/config/types.go
```diff
@@ -143,3 +143,16 @@ type NodeResourceTopologyMatchArgs struct {
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// NodeResourceTopologyResourceAllocationScoreArgs holds arguments used to configure the ResourceAllocationScore plugin
+type NodeResourceTopologyResourceAllocationScoreArgs struct {
+	metav1.TypeMeta
+
+	KubeConfigPath          string
+	MasterOverride          string
+	Namespaces              []string
+	ScoreSchedulingStrategy string
```
Huang-Wei, line 156: `ScoringStrategy` - to make it consistent with https://github.com/kubernetes/kubernetes/blob/ea3bcbc2050b89d11f005a2735c8808e45b8aa21/pkg/scheduler/apis/config/types_pluginargs.go#L93.

Also make it an enum, and define the possible strategies.

pkg/apis/config/types.go
```diff
@@ -143,3 +143,16 @@ type NodeResourceTopologyMatchArgs struct {
...
+	Resources               []schedulerconfig.ResourceSpec
```
Huang-Wei, line 157: Can you add some comments to explain the new fields? (Resources and ScoringStrategy)

pkg/noderesourcetopology/README.md
```diff
@@ -21,7 +21,7 @@ In case the cumulative count of node resource allocatable appear to be the same
 
 ### Config
 
- Enable the "NodeResourceTopologyMatch" Fitler plugin via SchedulerConfigConfiguration.
+ Enable the "NodeResourceTopologyMatch" Filter plugin and NodeResourceTopologyResourceAllocationScore Score plugin via SchedulerConfigConfiguration.
```
Huang-Wei, line 24:
```suggestion
 Enable the "NodeResourceTopologyMatch" Filter and Score plugins via SchedulerConfigConfiguration.
```

pkg/noderesourcetopology/README.md
```diff
@@ -45,6 +48,15 @@ profiles:
...
         - test-namespace
+  - name: NodeResourceTopologyResourceAllocationScore
+    args:
+      kubeconfigpath: "/etc/kubernetes/scheduler.conf"
+      namespaces:
+        - default
+        - production
+        - test-namespace
+      # other strategies are most-allocatable and balanced-allocation
+      scoreschedulingstrategy: "least-allocatable"
```
Huang-Wei, line 59:
```suggestion
      scoringStrategy: "LeastAllocatable"
```

Huang-Wei (commented):
The package moving makes the changes a bit difficult to track, so would you mind updating the PR per the first comment, and then I continue the reviewing?

cmd/scheduler/main.go
Tal-or: Thank you @Huang-Wei for taking the time to review this, I see your point but, I tried to implement that in a way that keeps the modularity of the package as much as possible and I'll explain:
The score plugin depends on the filter plugin.
There is no reasonable/realistic scenario to use the score plugin without the filter plugin, so combining the two into one plugin is the right way (as you were suggesting).

On the other hand, the filter plugin not depends on the score plugin. 
A user might use the filter plugin without the score plugin, because, for example, the user prefers not to use the scoring strategies that we offer, but use the default scheduler score plugin instead.

So by combining the two plugins together we enforce using our score strategies, and IMHO it might be less flexible unless you have an idea, or you know a way to avoid that? 

Please let me know what you think

pkg/noderesourcetopology/README.md
Tal-or: Done

pkg/apis/config/types.go
Tal-or: Done

pkg/apis/config/types.go
Tal-or: Done

cmd/scheduler/main.go
Huang-Wei: Good question. Basically, upon the plugin initialization, it still serves as a combination of Filter and Score plugins. However, users can disable the Score functionalities via disabling it in the ComponentConfig. Here is an example how to disable the upstream NodeResourceFit score plugin:

```yaml
...
profiles:
- plugins:
    score:
      disabled:
      - name: NodeResourcesFit
```

Huang-Wei: BTW: we need to disable one particular score plugin is because there is a default list of enabled plugins. For scheduler-plugins, the enabled plugins need to be always explicitly enabled, so we just specify which Filter/Score plugins to enable in the component config yaml.

Tal-or: I changed the package structure per your request.
Please continue with the review when possible and let me know if additional changes are needed.

pkg/noderesourcetopology/plugin.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+	apiconfig "sigs.k8s.io/scheduler-plugins/pkg/apis/config"
+
+	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+)
+
+// TopologyMatch plugin which run simplified version of TopologyManager's admit handler
+type TopologyMatch struct {
+	nodeResTopologyPlugin
```
Huang-Wei, line 17: I'd prefer to unroll `nodeResTopologyPlugin` here, and remove noderesourcetopology/types.go.

pkg/noderesourcetopology/plugin.go
```diff
@@ -0,0 +1,74 @@
...
+		return nil, err
+	}
+
+	resToWeightMap := make(resourceToWeightMap)
+	for _, resource := range tcfg.Resources {
+		resToWeightMap[v1.ResourceName(resource.Name)] = resource.Weight
+	}
+
+	// TBD what strategy should be provided by default
+	var scoringStrategy = leastAllocatableScoreStrategy
```
Huang-Wei, line 55: This should be defaulted and registered in pkg/apis/config/v1beta1/zz_generated.defaults.go.

Also, we should validate and abort on illegal scoring strategy.

pkg/noderesourcetopology/types.go
```diff
@@ -0,0 +1,19 @@
+package noderesourcetopology
```
Huang-Wei, line 1: As mentioned above, this doesn't look necessary to be extracted into a single file.

pkg/noderesourcetopology/plugin.go
```diff
@@ -0,0 +1,74 @@
...
+	apiconfig "sigs.k8s.io/scheduler-plugins/pkg/apis/config"
+
+	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+)
+
+// TopologyMatch plugin which run simplified version of TopologyManager's admit handler
+type TopologyMatch struct {
+	nodeResTopologyPlugin
+	policyHandlers      PolicyHandlerMap
+	scoreStrategy       scoreStrategy
```
Huang-Wei, line 19: Given it's a function, `scorer` or `scorerFn` maybe read better.

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+	switch strategy {
+	case apiconfig.LeastAllocatable:
+		return leastAllocatableScoreStrategy
+	case apiconfig.MostAllocatable:
+		return mostAllocatableScoreStrategy
+	case apiconfig.BalancedAllocation:
+		return balancedAllocationScoreStrategy
+	// TBD what strategy should be provided by default
+	default:
+		return leastAllocatableScoreStrategy
```
Huang-Wei, line 114: The strategy should have defaulted earlier, so let's throw an error here.

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+		return defaultWeight
+	}
+
+	return w
+}
+
+func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string) (int64, *framework.Status) {
+	// calculates the fraction of requested to capacity per each numa-node.
+	// return the numa-node with the minimal score as the node's total score
+	return tm.score(pod, nodeName)
```
Huang-Wei, line 40: Why not directly unroll `score()` here?

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+func (tm *TopologyMatch) score(pod *v1.Pod, nodeName string) (int64, *framework.Status) {
+	klog.V(5).Infof("Call score for node %v", nodeName)
+	nodeTopology := findNodeTopology(nodeName, &tm.nodeResTopologyPlugin)
+
+	if nodeTopology == nil {
+		return 0, nil
+	}
+
+	klog.V(5).Infof("nodeTopology: %v", nodeTopology)
+	calculateResourceAllocatableRequest(pod, nodeTopology)
```
Huang-Wei, line 57: remove this line.

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+		numaScores[numa.NUMAID] = score(requested, numa.Resources, resourceToWeightMap)
+	}
+	klog.V(5).Infof("numaScores: %v", numaScores)
+	minScore := findMinScore(numaScores)
+	klog.V(5).Infof("node score: %v", minScore)
+	return minScore, nil
+}
+
+func calculateResourceAllocatableRequest(pod *v1.Pod, nodeTopology *topologyv1alpha1.NodeResourceTopology) (v1.ResourceList, NUMANodeList) {
+	containers := append(pod.Spec.InitContainers, pod.Spec.Containers...)
```
Huang-Wei, line 76: I recalled I asked the same question to @AlexeyPerevalov before: is it intended to be semantically different from other res calculations? https://github.com/kubernetes/kubernetes/blob/9ff3b7e744b34c099c1405d9add192adbef0b6b1/pkg/scheduler/framework/plugins/noderesources/fit.go#L162

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -0,0 +1,175 @@
...
+	topoclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
+	topologyinformers "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/informers/externalversions"
+	listerv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/listers/topology/v1alpha1"
+)
+
+// topologyListerInstance is a Singleton object
+// Should not be accessed directly.
+// Should be accessed via getNodeTopologyInformer function
+var (
+	topologyListerInstance *listerv1alpha1.NodeResourceTopologyLister
```
Huang-Wei, line 24: -1 on maintaining a global variable. IIRC, you want to ensure the informer to be instantiated once, but isn't that already the case (instantiated in New())?

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -0,0 +1,175 @@
...
+	ctx := context.Background()
+	topologyInformerFactory.Start(ctx.Done())
+	topologyInformerFactory.WaitForCacheSync(ctx.Done())
+
+	return &nodeResourceTopologyLister, nil
+}
+
+// getNodeTopologyLister will init v1alpha1.NodeResourceTopologyInformer once and return it.
+// if v1alpha1.NodeResourceTopologyInformer already initialized, the same instance will be return
+func getNodeTopologyLister(masterOverride, kubeConfigPath *string) (*listerv1alpha1.NodeResourceTopologyLister, error) {
```
Huang-Wei, line 73: This is only called by New(). We don't need to introduce once() and a global variable to achieve this, do we?

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+
+// scoreForEachNUMANode will iterate over all NUMA zones of the node and invoke the scoreStrategy func for every zone.
+// it will return the minimal score of all the calculated NUMA's score, in order to avoid edge cases.
+func scoreForEachNUMANode(requested v1.ResourceList, numaList NUMANodeList, score scoreStrategy, resourceToWeightMap resourceToWeightMap) (int64, *framework.Status) {
+	numaScores := make([]int64, len(numaList))
+	for _, numa := range numaList {
+		numaScores[numa.NUMAID] = score(requested, numa.Resources, resourceToWeightMap)
+	}
+	klog.V(5).Infof("numaScores: %v", numaScores)
+	minScore := findMinScore(numaScores)
```
Huang-Wei, line 70: Can we get this done when iterating `numaList`?

Huang-Wei (commented):
Haven't looked into tests.

pkg/apis/config/types.go
```diff
@@ -142,4 +158,11 @@ type NodeResourceTopologyMatchArgs struct {
 
 	// Namespaces to be considered by NodeResourceTopologyMatch plugin
 	Namespaces []string
+
+	// ScoringStrategy a scoring model that determine how the plugin will score the nodes.
+	ScoringStrategy *ScoringStrategy
+
+	// Resources a list of pairs <resource, weight> to be considered while scoring
+	// allowed weights start from 1.
+	Resources       []schedulerconfig.ResourceSpec
```
Huang-Wei, line 167: The formatting doesn't look right. Did you have a "run gofmt on every save" setup on your editor?

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+	defaultWeight = int64(1)
+)
+
+type scoreStrategy func(v1.ResourceList, v1.ResourceList, resourceToWeightMap) int64
+
+// resourceToWeightMap contains resource name and weight.
+type resourceToWeightMap map[v1.ResourceName]int64
+
+// weight return the weight of the resource and defaultWeight if weight not specified
+func (rw *resourceToWeightMap) weight(r v1.ResourceName) int64 {
```
Huang-Wei, line 24: As `map` is already a pointer-based structure, we can just use the concrete value as the method receiver.

```suggestion
func (rw resourceToWeightMap) weight(r v1.ResourceName) int64 {
```

So on L25, we just call `w, ok := rw[r]`

pkg/noderesourcetopology/README.md
```diff
@@ -21,7 +21,7 @@ In case the cumulative count of node resource allocatable appear to be the same
 
 ### Config
 
- Enable the "NodeResourceTopologyMatch" Fitler plugin via SchedulerConfigConfiguration.
+Enable the "NodeResourceTopologyMatch" Filter plugin and NodeResourceTopologyResourceAllocationScore Score plugin via SchedulerConfigConfiguration.
```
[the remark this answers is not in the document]
Tal-or: My mistake. it slipped somehow during the last changes.

pkg/noderesourcetopology/pluginhelpers.go
Tal-or: Yes. will remove that. it was useful when the plugins were separated. not needed anymore

Tal-or:
I tried pushing every review's comment on a separate commit, in order to make the next review cycles easier.
I Will squash once you'll be happy with the final version.

About the other comments, which wasn't addressed yet, they are dependent on https://github.com/kubernetes-sigs/scheduler-plugins/pull/220#discussion_r681598114 
Please let me know what do you think about it and I'll update everything accordingly.

pkg/noderesourcetopology/score.go
```diff
@@ -0,0 +1,117 @@
...
+	nodeTopology := findNodeTopology(nodeName, &tm.nodeResTopologyPlugin)
+
+	if nodeTopology == nil {
+		return 0, nil
+	}
+
+	klog.V(5).Infof("nodeTopology: %v", nodeTopology)
+	calculateResourceAllocatableRequest(pod, nodeTopology)
+	resources, allocatablePerNUMA := calculateResourceAllocatableRequest(pod, nodeTopology)
+	return scoreForEachNUMANode(resources, allocatablePerNUMA, tm.scoreStrategy, tm.resourceToWeightMap)
```
AlexeyPerevalov: score you calculated here is not applicable for ContainerLevel TM policy, we need to take it into account here.

pkg/noderesourcetopology/score.go
AlexeyPerevalov: yes, we followed the way TopologyManager is working
but here in score.go I see only PodLevel behavior of TM, but don't see ContainerLevel.

AlexeyPerevalov: and we don't considering pod.Spec.Overhead, but we should.

AlexeyPerevalov: > I recalled I asked the same question to @AlexeyPerevalov before: is it intended to be semantically different from other res calculations? https://github.com/kubernetes/kubernetes/blob/9ff3b7e744b34c099c1405d9add192adbef0b6b1/pkg/scheduler/framework/plugins/noderesources/fit.go#L162

@Huang-Wei I'm not sure it reasonable to generalize both topology and none topology resource calculation, since topology is a hierarchical, currently we use only NUMA, but in future it would be hyper threaded cores and LLC groups of CPU.

Huang-Wei: As long as it's adherent with TM, I'm fine. (it's just every time I look at similar code, I cannot help getting confused...)

Would you mind commenting here, and if possible, paste the link to how TopologyManager calculates the res quest? That would de-confuse other readers as well.

pkg/apis/config/types.go
```diff
@@ -128,6 +128,22 @@ type LoadVariationRiskBalancingArgs struct {
...
+	LeastAllocatable ScoringStrategyType = "LeastAllocatable"
+	// BalancedAllocation strategy favors nodes with balanced resource usage rate
+	BalancedAllocation ScoringStrategyType = "BalancedAllocation"
+	// MostAllocatable strategy favors node with the most amount of available resource
+	MostAllocatable ScoringStrategyType = "MostAllocatable"
+)
+
+type ScoringStrategy struct {
+	Type ScoringStrategyType
+}
```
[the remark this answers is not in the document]
Huang-Wei: > Would you prefer to follow the same convention?

Sounds good as the `[]ResourceSpec` will be used in scoring **only**.

pkg/noderesourcetopology/score.go
Huang-Wei: @Tal-or could you put the above comment into the code?

Tal-or: Sure will add that

pkg/apis/config/types.go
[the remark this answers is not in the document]
Tal-or: Great. Will work on that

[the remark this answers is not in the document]
Tal-or: Done

pkg/apis/config/v1beta1/zz_generated.defaults.go
```diff
@@ -28,39 +28,5 @@ import (
 // Public to allow building arbitrary schemes.
 // All generated defaulters are covering - they call all nested defaulters.
 func RegisterDefaults(scheme *runtime.Scheme) error {
-	scheme.AddTypeDefaultingFunc(&CapacitySchedulingArgs{}, func(obj interface{}) { SetObjectDefaultsCapacitySchedulingArgs(obj.(*CapacitySchedulingArgs)) })
```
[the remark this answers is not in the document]
Tal-or: Done

pkg/noderesourcetopology/score.go
Tal-or: Done

pkg/noderesourcetopology/plugin.go
```diff
@@ -0,0 +1,109 @@
...
+	lister     *listerv1alpha1.NodeResourceTopologyLister
+	namespaces []string
+}
+
+type tmScopeHandler interface {
+	TopologyScopedFilter(pod *v1.Pod, zones topologyv1alpha1.ZoneList) *framework.Status
+	TopologyScopedScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList, scorerFn scoreStrategy, resourceToWeightMap resourceToWeightMap) (int64, *framework.Status)
+}
+
+type podScopeHandler struct{}
```
Huang-Wei, line 33: An empty struct doesn't look necessary to me. Can we switch the interface to a struct, with two functions as fields? Something like https://gist.github.com/Huang-Wei/16daaf5352b1bc576de653c3cc660f29.

pkg/noderesourcetopology/filter.go
```diff
@@ -21,65 +21,18 @@ import (
...
-		if err != nil {
-			klog.Errorf("Failed to parse %s", resInfo.Allocatable.String())
-			continue
-		}
-		res[v1.ResourceName(resInfo.Name)] = quantity
-	}
-	return res
-}
-
 func SingleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList) *framework.Status {
```
Huang-Wei, line 36: nit: make this function along with SingleNUMAPodLevelHandler() private.

pkg/noderesourcetopology/least_allocatable.go
```diff
@@ -0,0 +1,38 @@
+package noderesourcetopology
```
Huang-Wei, line 1: Add coding template. (applies elsewhere)

pkg/noderesourcetopology/least_allocatable.go
```diff
@@ -0,0 +1,38 @@
...
+// The less allocatable resources the node has, the higher the score is.
+func leastAllocatableScore(requested, numaCapacity resource.Quantity) int64 {
+	if numaCapacity.CmpInt64(0) == 0 {
+		return 0
+	}
+	if requested.Cmp(numaCapacity) > 0 {
+		return 0
+	}
+
+	return (requested.Value() * framework.MaxNodeScore) / numaCapacity.Value()
```
Huang-Wei, line 37: Apologize that I should have spotted it earlier.

Should we use the term "Allocated" (aligned with the upstream plugins) or "Allocatable"? My 2 cents is "Allocated". WDYT? If you agree, the logic here should be `capacity - requested`, and vice versa for "mostAllocated". Also the string liternals in types.go should be updated.

Tal-or: Actually, I was trying to be aligned with the naming convention of this repo's plugin:
For example here https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/pkg/noderesources we're also using the terms "Allocatable", "leastAllocatable" etc. so I thought it would be more accurate to keep the same convention. 

But again, if you think it would be better to follow the upstream plugins convention, I'll do so.

pkg/apis/config/types.go
```diff
@@ -142,4 +158,11 @@ type NodeResourceTopologyMatchArgs struct {
 
 	// Namespaces to be considered by NodeResourceTopologyMatch plugin
 	Namespaces []string
+
+	// ScoringStrategy a scoring model that determine how the plugin will score the nodes.
+	ScoringStrategy *ScoringStrategy
```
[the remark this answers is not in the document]
Tal-or: Thank you again for your review.
>If we just specify ScoringStrategyType here, we can use the concrete type (instead of pointer) in both internal and external types.go.

But now we also have the ```ResourceSpec[]``` [inside](https://github.com/kubernetes-sigs/scheduler-plugins/pull/220/commits/06ec7cd6189322449fb7f2e4edc643a84e5b8dd4) as well and not just the ```ScoringStrategyType```. this is why I kept it as a pointer.

pkg/apis/config/v1beta1/defaults.go
```diff
@@ -139,4 +144,22 @@ func SetDefaultsNodeResourceTopologyMatchArgs(obj *NodeResourceTopologyMatchArgs
 	if len(obj.Namespaces) == 0 {
 		obj.Namespaces = []string{metav1.NamespaceDefault}
 	}
+
+	if obj.ScoringStrategy == nil {
```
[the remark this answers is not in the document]
Tal-or: Please refer to my last comment here: https://github.com/kubernetes-sigs/scheduler-plugins/pull/220#discussion_r687411452

pkg/noderesourcetopology/least_allocatable.go
Huang-Wei: Oh, this repo's **allocatable** plugin was an supplement to the upstream plugins, please see the motivations in its KEP. And this PR tries to introduce regular plugins for topology aware scenarios, so I'd prefer to rename to "xyzallocated". Thanks.

pkg/apis/config/types.go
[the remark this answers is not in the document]
Huang-Wei: My 2 cents is that if a field is required, the definition in the internal version should be a concrete type. So I prefer to make it a concrete type, no matter it has a slice inside or not.

But I also noticed the upstream [version](https://github.com/kubernetes/kubernetes/blob/1f8b1b84fbfbf88bf50429b1f5673fd34ae30050/pkg/scheduler/apis/config/types_pluginargs.go#L93) also specified it as a pointer, so @alculquicondor could you confirm it's intended or not?

Huang-Wei (commented):
There are some compilation error. Please figure out.

pkg/apis/config/types.go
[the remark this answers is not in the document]
alculquicondor: It's not necessarily intended and we are not very consistent on this. Both would behave just the same from most users' perspective.

It just becomes extra cumbersome when it's a pointer to a native type such as `*int32`.

But +1 to concrete type for simplicity.

Tal-or:
@Huang-Wei all comments were addressed.
Please let me know if you're happy with the changes and then I'll squash everything.
Also, I'll update the commit's and PR's content regarding the strategies' names.

pkg/apis/config/v1beta1/types.go
```diff
@@ -130,13 +130,32 @@ type LoadVariationRiskBalancingArgs struct {
...
 type NodeResourceTopologyMatchArgs struct {
 	metav1.TypeMeta `json:",inline"`
 
-	KubeConfigPath *string  `json:"kubeconfigpath,omitempty"`
-	MasterOverride *string  `json:"masteroverride,omitempty"`
-	Namespaces     []string `json:"namespaces,omitempty"`
+	KubeConfigPath  *string         `json:"kubeconfigpath,omitempty"`
+	MasterOverride  *string         `json:"masteroverride,omitempty"`
+	Namespaces      []string        `json:"namespaces,omitempty"`
+	ScoringStrategy ScoringStrategy `json:"scoringStrategy,omitempty"`
```
Huang-Wei, line 160:
```suggestion
	ScoringStrategy *ScoringStrategy `json:"scoringStrategy,omitempty"`
```

As it's the external versioned type.

pkg/noderesourcetopology/least_allocated.go
```diff
@@ -0,0 +1,55 @@
...
+func leastAllocatedScore(requested, numaCapacity resource.Quantity) int64 {
+	if numaCapacity.CmpInt64(0) == 0 {
+		return 0
+	}
+	if requested.Cmp(numaCapacity) > 0 {
+		return 0
+	}
+	numaValue := numaCapacity.Value()
+	requestedValue := requested.Value()
+	return ((numaValue - requestedValue) * framework.MaxNodeScore) / numaCapacity.Value()
```
Huang-Wei, line 54: nit:

```suggestion
	return (numaValue - requestedValue) * framework.MaxNodeScore / numaCapacity.Value()
```

pkg/noderesourcetopology/least_allocated.go
```diff
@@ -0,0 +1,55 @@
...
+		weightSum += weight
+	}
+
+	return numaNodeScore / weightSum
+}
+
+// The used capacity is calculated on a scale of 0-MaxNodeScore (MaxNodeScore is
+// constant with value set to 100).
+// 0 being the lowest priority and 100 being the highest.
+// The more allocatable resources the node has, the higher the score is.
```
Huang-Wei, line 44: to align with the file name:

```suggestion
// The less allocated resources the node has, the higher the score is.
```

pkg/noderesourcetopology/most_allocated.go
```diff
@@ -0,0 +1,54 @@
...
+		weightSum += weight
+	}
+
+	return numaNodeScore / weightSum
+}
+
+// The used capacity is calculated on a scale of 0-MaxNodeScore (MaxNodeScore is
+// constant with value set to 100).
+// 0 being the lowest priority and 100 being the highest.
+// The less allocatable resources the node has, the higher the score is.
```
Huang-Wei, line 44: to align with the file name:

```suggestion
// The more allocated resources the node has, the higher the score is.
```

pkg/noderesourcetopology/most_allocated.go
```diff
@@ -0,0 +1,54 @@
...
+func mostAllocatedScore(requested, numaCapacity resource.Quantity) int64 {
+	if numaCapacity.CmpInt64(0) == 0 {
+		return 0
+	}
+	if requested.Cmp(numaCapacity) > 0 {
+		return 0
+	}
+
+	return (requested.Value() * framework.MaxNodeScore) / numaCapacity.Value()
```
Huang-Wei, line 53: nit:

```suggestion
	return requested.Value() * framework.MaxNodeScore / numaCapacity.Value()
```

test/integration/noderesourcetopology_test.go
```diff
@@ -20,10 +20,11 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology"
```
Huang-Wei, line 23: nit: move it to L47.

Huang-Wei (commented):
Some nits. LGTM otherwise.

BTW: I'm bumping to deps to k8s 1.21. Would you mind holding it until #226 gets merged?

If it's desired to get it in the scheduler-plugins [0.20.X release](https://github.com/kubernetes-sigs/scheduler-plugins/releases), later you can cherry-pick it to the `release-1.20` branch. If not, this will be published along the next 0.21.X release.

Tal-or:
> Some nits. LGTM otherwise.
> 
> BTW: I'm bumping to deps to k8s 1.21. Would you mind holding it until #226 gets merged?
> 
> If it's desired to get it in the scheduler-plugins [0.20.X release](https://github.com/kubernetes-sigs/scheduler-plugins/releases), later you can cherry-pick it to the `release-1.20` branch. If not, this will be published along the next 0.21.X release.

Yes. it can hold until #226 gets merged, nothing urgent.

Huang-Wei:
#226 gets merged. You can start to rebase and squash the commits (maybe into 2: one for codegen/doc, the other for implementation)

pkg/apis/config/v1beta1/types.go
Tal-or: Again thank you for taking the time and review everything.
But after this change, running ```hack/update-codegen.sh``` ended up with:
```
E0820 08:11:19.961016 2283072 conversion.go:755] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/scheduler-plugins/pkg/apis/config/v1beta1.NodeResourceTopologyMatchArgs -> sigs.k8s.io/scheduler-plugins/pkg/apis/config.NodeResourceTopologyMatchArgs
E0820 08:11:19.961063 2283072 conversion.go:756]   the following fields need manual conversion:
E0820 08:11:19.961067 2283072 conversion.go:758]       - ScoringStrategy
E0820 08:11:19.961338 2283072 conversion.go:755] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/scheduler-plugins/pkg/apis/config.NodeResourceTopologyMatchArgs -> sigs.k8s.io/scheduler-plugins/pkg/apis/config/v1beta1.NodeResourceTopologyMatchArgs
E0820 08:11:19.961351 2283072 conversion.go:756]   the following fields need manual conversion:
E0820 08:11:19.961356 2283072 conversion.go:758]       - ScoringStrategy
```
Maybe it's an additional bug in the code generator?

Huang-Wei: I _think_ it's as expected. The defaulter logic will be able to cover it. Could you add a UT in cmd/scheduler/main_test.go to ensure an argument without specifying the scoringStrategy field can be instantiated properly?

BTW: the codegen script somehow removes two functions `Convert_v1beta1_NodeResourceTopologyMatchArgs_To_config_NodeResourceTopologyMatchArgs` and `Convert_config_NodeResourceTopologyMatchArgs_To_v1beta1_NodeResourceTopologyMatchArgs`, which causes a compilation error. Not quite sure it's a bug. I tried to add them back and it looks good. FYI: https://github.com/Huang-Wei/scheduler-plugins/commit/a91ae79ae578f63462aecc45da1b1634a9adf190.

Tal-or: > I _think_ it's as expected. The defaulter logic will be able to cover it. Could you add a UT in cmd/scheduler/main_test.go to ensure an argument without specifying the scoringStrategy field can be instantiated properly?

Without adding logic for manual conversion of scoringStrategy field in https://github.com/kubernetes-sigs/scheduler-plugins/pull/220/commits/563a9ed8e2d437aa858cfcc9fddd4b7665de6fc7#diff-9c6b4d879381c71601a34af877d2fa07eb444b866564a0e9e6050e97bcfd4557R274 the UT  in cmd/scheduler/main_test.go are falling. 
I tried stick into the conversion as it shown in [noderesourcefit](https://github.com/kubernetes/kubernetes/blob/b9565beef015df9ae8e73b408d106531bea5ba77/pkg/scheduler/apis/config/v1beta2/zz_generated.conversion.go#L515)
the only difference is that here the conversion is concrete <-> pointer

Tal-or:
The integration test falling with:
```
+++ [0823 12:29:20] Checking etcd is on PATH
+++ [0823 12:29:20] Starting etcd instance
unable to identify if etcd is bound to port 2379. unable to find ss or netstat utilities.
+++ [0823 12:29:20] Cleaning up etcd
+++ [0823 12:29:20] Integration test cleanup complete
make: *** [Makefile:123: integration-test] Error 1
``` 
I tested it locally on my laptop, but it didn't reproduce.

Huang-Wei:
I recently bumped the golang version to 1.16, and it seems the official golang docker images removed `ss` utility:

```shell
root@wei-dev:~# docker run -it golang:1.16.7 bash
root@8cae4123bab8:/go# which ss

root@wei-dev:~# docker run -it golang:1.15.10 bash
root@085c4ba5d7d9:/go# which ss
/bin/ss
```

that causes the etcd script cannot use `ss` to verify if the etcd port is in use.

Huang-Wei:
A solution is to use `gcr.io/k8s-testimages/kubekins-e2e:latest-1.21`, but it occupies ~3G. Sounds like an overkill to just get `ss` installed:

```script
root@wei-dev:~# docker images | grep e2e
gcr.io/k8s-testimages/kubekins-e2e          latest-1.21    7499620c52a6   2 weeks ago     3.31GB
```

Huang-Wei:
/lgtm
/approve

Thanks @Tal-or !

Issue #151, opened 2021-02-25
Script to verify incorrect CRD yaml


Huang-Wei:
Per https://github.com/kubernetes-sigs/scheduler-plugins/pull/148#issuecomment-783950924, it's favorable to have a script verifying incorrect/missing specs in individual CRD yaml, and automate this in CI.

/kind feature
/assign @penglei

Huang-Wei:
@penglei I cannot assign it to you. But you can `/assign` to get it self-assigned.

pravarag:
@Huang-Wei I'd like to work on this issue

/assign

pravarag:
@Huang-Wei I'm guessing these are the steps I can take while trying to write a script for verifying CRD yamls as mentioned here: https://github.com/kubernetes-sigs/scheduler-plugins/pull/148#issuecomment-783950924

Huang-Wei:
@pravarag yes, that's correct. Moreover, you can also get inspiring ideas from the ["hack"](https://github.com/kubernetes/kubernetes/tree/master/hack) scripts under the k/k repo.

pravarag:
@Huang-Wei @damemi I've created a WIP PR https://github.com/kubernetes-sigs/scheduler-plugins/pull/222 though I'm still looking for ways to test the script. Kindly suggest if any changes/updates are required in the code.

Huang-Wei:
Fixed by #222.

PR #222, opened 2021-08-02
Add script to verify incorrect CRD yaml

pravarag:
This PR extends: https://github.com/kubernetes-sigs/scheduler-plugins/issues/151

I've taken references from existing kubernetes/kubernetes `verify-` scripts to identify incorrect or missing specs while generating CRD yaml.

Creating WIP PR for initial reviews and suggestions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,7 @@ update-vendor:
 .PHONY: unit-test
 unit-test: autogen
 	hack/unit-test.sh
+	hack/verify-crdgen.sh
 
 .PHONY: install-etcd
 install-etcd:
@@ -125,7 +126,6 @@ integration-test: install-etcd autogen
 .PHONY: verify-gofmt
 verify-gofmt:
 	hack/verify-gofmt.sh
-	hack/verify-crdgen.sh
 
 .PHONY: clean
 clean:
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,7 @@ integration-test: install-etcd autogen
 .PHONY: verify-gofmt
 verify-gofmt:
 	hack/verify-gofmt.sh
+	hack/verify-crdgen.sh
 
 .PHONY: clean
 clean:
--- a/hack/verify-crdgen.sh
+++ b/hack/verify-crdgen.sh
@@ -18,35 +18,28 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCHED_PLUGIN_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
-source "${SCHED_PLUGIN_ROOT}/hack/lib/init.sh"
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${SCRIPT_ROOT}/hack/lib/init.sh"
 
 kube::golang::verify_go_version
-
-cd "${SCHED_PLUGIN_ROOT}"
 
 CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
 
 # Download controller-gen locally
-CONTROLLER_GEN="$(shell pwd)/bin/controller-gen"
-$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen)
+CONTROLLER_GEN="${GOPATH}/bin/controller-gen"
+go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest
 
-# Create a temp directory
-_tmpdir="$(mktemp -d "${SCHED_PLUGIN_ROOT}/_tmp")"
 # Generate CRD
-pushd "${_tmpdir}" &> /dev/null
-$(CONTROLLER_GEN) ${CRD_OPTIONS} path="./..." output:crd:artifacts:config=${_tmpdir}/crd/bases
+api_paths="./pkg/apis/scheduling/v1alpha1/...;./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/..."
 
-popd &> /dev/null
+${CONTROLLER_GEN} ${CRD_OPTIONS} paths="${api_paths}" output:dir="./manifests/crds"
 
-pushd "${SCHED_PLUGIN_ROOT}" > /dev/null 2>&1
-if ! _out="$(diff -Naupr manifests/coscheduling/ "${_crdtmp}/crd/bases/}")"; then
-    echo "Generated output differs:" >&2
+if ! _out="$(git --no-pager diff -I"edited\smanually" --exit-code)"; then
+    echo "Generated output differs" >&2
     echo "${_out}" >&2
-    echo "Verification failed."
+    echo "Verification for CRD generators failed."
     exit 1
 fi
-popd &> /dev/null
 
-rm -r ./_tmpdir
 echo "Controllers Gen for CRD verified."
--- a/hack/verify-crdgen.sh
+++ b/hack/verify-crdgen.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCHED_PLUGIN_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${SCHED_PLUGIN_ROOT}/hack/lib/init.sh"
+
+kube::golang::verify_go_version
+
+cd "${SCHED_PLUGIN_ROOT}"
+
+CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Download controller-gen locally
+CONTROLLER_GEN="$(shell pwd)/bin/controller-gen"
+$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen)
+
+# Create a temp directory
+_tmpdir="$(mktemp -d "${SCHED_PLUGIN_ROOT}/_tmp")"
+# Generate CRD
+pushd "${_tmpdir}" &> /dev/null
+$(CONTROLLER_GEN) ${CRD_OPTIONS} path="./..." output:crd:artifacts:config=${_tmpdir}/crd/bases
+
+popd &> /dev/null
+
+pushd "${SCHED_PLUGIN_ROOT}" > /dev/null 2>&1
+if ! _out="$(diff -Naupr manifests/coscheduling/ "${_crdtmp}/crd/bases/}")"; then
+    echo "Generated output differs:" >&2
+    echo "${_out}" >&2
+    echo "Verification failed."
+    exit 1
+fi
+popd &> /dev/null
+
+rm -r ./_tmpdir
+echo "Controllers Gen for CRD verified."
--- a/manifests/capacityscheduling/crd.yaml
+++ b/manifests/capacityscheduling/crd.yaml
@@ -1,54 +1,87 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  name: elasticquotas.scheduling.sigs.k8s.io
   annotations:
-    "api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/scheduler-plugins/pull/52"
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/52 # edited manually
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: elasticquotas.scheduling.sigs.k8s.io
 spec:
   group: scheduling.sigs.k8s.io
   names:
+    kind: ElasticQuota
+    listKind: ElasticQuotaList
     plural: elasticquotas
     singular: elasticquota
-    kind: ElasticQuota
-    shortNames:
-    - eq
-    - eqs
+    shortNames: # edited manually
+    - eq  # edited manually
+    - eqs # edited manually
   scope: Namespaced
   versions:
-    - name: v1alpha1
-      served: true
-      storage: true
-      schema:
-        openAPIV3Schema:
-          type: object
-          properties:
-            spec:
-              type: object
-              properties:
-                min:
-                  type: object
-                  additionalProperties:
-                    anyOf:
-                      - type: integer
-                      - type: string
-                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                    x-kubernetes-int-or-string: true
-                max:
-                  type: object
-                  additionalProperties:
-                    anyOf:
-                      - type: integer
-                      - type: string
-                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                    x-kubernetes-int-or-string: true
-            status:
-              type: object
-              properties:
-                used:
-                  type: object
-                  additionalProperties:
-                    anyOf:
-                      - type: integer
-                      - type: string
-                    pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
-                    x-kubernetes-int-or-string: true
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: ElasticQuota sets elastic quota restrictions per namespace
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: ElasticQuotaSpec defines the Min and Max for Quota.
+            properties:
+              max:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: Max is the set of desired max limits for each named resource.
+                  The usage of max is based on the resource configurations of successfully
+                  scheduled pods.
+                type: object
+              min:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: Min is the set of desired guaranteed limits for each
+                  named resource.
+                type: object
+            type: object
+          status:
+            description: ElasticQuotaStatus defines the observed use.
+            properties:
+              used:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: Used is the current observed total usage of the resource
+                  in the namespace.
+                type: object
+            type: object
+        type: object
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
--- a/manifests/coscheduling/crd.yaml
+++ b/manifests/coscheduling/crd.yaml
@@ -1,59 +1,106 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
-  name: podgroups.scheduling.sigs.k8s.io
   annotations:
-    "api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/scheduler-plugins/pull/50"
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/50 # edited manually
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: podgroups.scheduling.sigs.k8s.io
 spec:
   group: scheduling.sigs.k8s.io
   names:
     kind: PodGroup
+    listKind: PodGroupList
     plural: podgroups
     singular: podgroup
-    shortNames:
-    - pg
-    - pgs
+    shortNames: # edited manually
+    - pg # edited manually
+    - pgs # edited manually
   scope: Namespaced
   versions:
-  - name: "v1alpha1"
-    served: true
-    storage: true
+  - name: v1alpha1
     schema:
       openAPIV3Schema:
-        type: object
+        description: PodGroup is a collection of Pod; used for batch workload.
         properties:
-          spec:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
             type: object
+          spec:
+            description: Specification of the desired behavior of the pod group.
             properties:
               minMember:
-                type: integer
-                minimum: 1
-              scheduleTimeoutSeconds:
+                description: MinMember defines the minimal number of members/tasks
+                  to run the pod group; if there's not enough resources to start all
+                  tasks, the scheduler will not start anyone.
+                format: int32
                 type: integer
               minResources:
-                type: object
                 additionalProperties:
-                  type: string
-          status:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: MinResources defines the minimal resource of members/tasks
+                  to run the pod group; if there's not enough resources to start all
+                  tasks, the scheduler will not start anyone.
+                type: object
+              scheduleTimeoutSeconds:
+                description: ScheduleTimeoutSeconds defines the maximal time of members/tasks
+                  to wait before run the pod group;
+                format: int32
+                type: integer
             type: object
+          status:
+            description: Status represents the current information about a pod group.
+              This data may not be up to date.
             properties:
-              phase:
-                type: string
+              failed:
+                description: The number of pods which reached phase Failed.
+                format: int32
+                type: integer
               occupiedBy:
+                description: OccupiedBy marks the workload (e.g., deployment, statefulset)
+                  UID that occupy the podgroup. It is empty if not initialized.
+                type: string
+              phase:
+                description: Current phase of PodGroup.
+                type: string
+              running:
+                description: The number of actively running pods.
+                format: int32
+                type: integer
+              scheduleStartTime:
+                description: ScheduleStartTime of the group
+                format: date-time
                 type: string
               scheduled:
+                description: The number of actively running pods.
+                format: int32
                 type: integer
-                default: 0
-              running:
-                type: integer
-                default: 0
               succeeded:
+                description: The number of pods which reached phase Succeeded.
+                format: int32
                 type: integer
-                default: 0
-              failed:
-                type: integer
-                default: 0
-              scheduleStartTime:
-                type: string
-                format: date-time
-
+            type: object
+        type: object
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
--- a/manifests/noderesourcetopology/crd.yaml
+++ b/manifests/noderesourcetopology/crd.yaml
@@ -1,12 +1,13 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
-    controller-gen.kubebuilder.io/version: v0.6.0
+    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870 # edited manually
+    controller-gen.kubebuilder.io/version: v0.6.2
   creationTimestamp: null
   name: noderesourcetopologies.topology.node.k8s.io
-  namespace: ""
 spec:
   group: topology.node.k8s.io
   names:
--- a/manifests/crds/scheduling.sigs.k8s.io_elasticquotas.yaml
+++ b/manifests/crds/scheduling.sigs.k8s.io_elasticquotas.yaml
@@ -1,0 +1,1 @@
+../capacityscheduling/crd.yaml
--- a/manifests/crds/scheduling.sigs.k8s.io_podgroups.yaml
+++ b/manifests/crds/scheduling.sigs.k8s.io_podgroups.yaml
@@ -1,0 +1,1 @@
+../coscheduling/crd.yaml
--- a/manifests/crds/topology.node.k8s.io_noderesourcetopologies.yaml
+++ b/manifests/crds/topology.node.k8s.io_noderesourcetopologies.yaml
@@ -1,0 +1,1 @@
+../noderesourcetopology/crd.yaml


damemi (commented):
/ok-to-test
You could add this to one of the `Makefile` targets so this runs automatically, maybe just under `unit-test` for now. Putting it under its own target will require updating k8s.io/test-infra to run the new tests

hack/verify-crd-gen.sh
```diff
@@ -0,0 +1,37 @@
...
+
+kube::golang::verify_go_version
+
+cd "${SCHED_PLUGIN_ROOT}"
+
+CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Download controller-gen locally
+CONTROLLER_GEN="$(shell pwd)/bin/controller-gen"
+$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0)
```
Huang-Wei, line 18: What if we don't specify the version? I guess it will auto fetch the latest stable semantic version? for this moment, v0.6.2?

hack/verify-crd-gen.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Create a temp directory
+_tmpdir="$(mktemp -d "${SCHED_PLUGIN_ROOT}/_tmp")"
+_crdtmp="${_tmpdir}"
```
Huang-Wei, line 22: Can we just use `_tmpdir` everywhere?

hack/verify-crd-gen.sh
```diff
@@ -0,0 +1,37 @@
...
+# Generate CRD
+pushd "${_crdtmp}" > /dev/null 2>&1
```
Huang-Wei, line 24: nit: All occurrences of `2>&1` can be shortened to `&>`:

```shell
pushd "${_crdtmp}" &> /dev/null
```

hack/verify-crd-gen.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
Huang-Wei, line 1: nit: `#!/usr/bin/env bash` is more portable.

hack/verify-crd-gen.sh
```diff
@@ -0,0 +1,37 @@
...
+pushd "${_crdtmp}" > /dev/null 2>&1
+$(CONTROLLER_GEN) ${CRD_OPTIONS} path="./..." output:crd:artifacts:config=${_crdtmp}/crd/bases
+
+popd > /dev/null 2>&1
+
+pushd "${SCHED_PLUGIN_ROOT}" > /dev/null 2>&1
+if ! _out="$(diff -Naupr manifests/coscheduling/ "${_crdtmp}/crd/bases/}")"; then
+    echo "Generated output differs:" >&2
+    echo "${_out}" >&2
+    echo "Verification failed."
```
Huang-Wei, line 33: Prefer to explicitly add `exit 1` below; otherwise, this script (if I understand correctly) will always return 0 as you catch the potential error in L30.

hack/verify-crd-gen.sh
```diff
@@ -0,0 +1,37 @@
...
+cd "${SCHED_PLUGIN_ROOT}"
+
+CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Download controller-gen locally
+CONTROLLER_GEN="$(shell pwd)/bin/controller-gen"
+$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.5.0)
+
+# Create a temp directory
+_tmpdir="$(mktemp -d "${SCHED_PLUGIN_ROOT}/_tmp")"
```
Huang-Wei, line 21: Should we clean this folder up in all cases?

Huang-Wei (commented):
Thanks @pravarag for driving this!

Huang-Wei:
> You could add this to one of the `Makefile` targets so this runs automatically, maybe just under `unit-test` for now. Putting it under its own target will require updating k8s.io/test-infra to run the new tests

Maybe hold off linking it to Makefile for now. I can hook it up with the CI job later.

damemi:
> Maybe hold off linking it to Makefile for now. I can hook it up with the CI job later.

We could link it to one of the Makefile targets that's already run by CI, with a todo to move it to its own job later. That can be done in this PR and it will validate that the test actually runs and passes in CI before we merge it

Huang-Wei:
> We could link it to one of the Makefile targets that's already run by CI, with a todo to move it to its own job later. That can be done in this PR and it will validate that the test actually runs and passes in CI before we merge it

SG. Then the CI can look over our shoulders since the very beginning.

hack/verify-crd-gen.sh
pravarag: Removed the version, have changed to `cmd/controller-gen`

hack/verify-crd-gen.sh
pravarag: Updated `2>&1` as suggested.

hack/verify-crd-gen.sh
pravarag: Updated the path as suggested with the latest push.

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,52 @@
...
+
+kube::golang::verify_go_version
+
+cd "${SCHED_PLUGIN_ROOT}"
+
+CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Download controller-gen locally
+CONTROLLER_GEN="$(shell pwd)/bin/controller-gen"
+$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen)
```
Huang-Wei, line 32: These 2 commands don't work on either my Mac or Linux env. Did you verify it works?

pravarag: Okay, I think I somehow configuring the `controller-gen` in the environment first that's why the commands might not be working. I'll update the code.

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
pravarag: Updated

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,52 @@
...
+pushd "${SCHED_PLUGIN_ROOT}" > /dev/null 2>&1
+if ! _out="$(diff -Naupr manifests/coscheduling/ "${_crdtmp}/crd/bases/}")"; then
+    echo "Generated output differs:" >&2
+    echo "${_out}" >&2
+    echo "Verification failed."
+    exit 1
+fi
+popd &> /dev/null
+
+rm -r ./_tmpdir
```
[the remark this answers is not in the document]
pravarag: I've updated the step to delete _tmp directory using `trap` call here: https://github.com/kubernetes-sigs/scheduler-plugins/pull/222/files#diff-4e3ddc507e7011467ddc0a1168a0eb8355821fe18f934b3a9042fb6ceef37cacR36

hack/verify-crdgen.sh
Huang-Wei: This is still outstanding.

pravarag:
While trying to run `controller-gen` in my local, I'm getting below error, not sure if this could be related to any missing packages in my local:

```
 ->  ./hack/verify-crdgen.sh
mkdir /crd: read-only file system
../../../../pkg/mod/k8s.io/client-go@v0.20.5/transport/round_trippers.go:25:2: missing go.sum entry for module providing package golang.org/x/oauth2 (imported by k8s.io/kubernetes/test/integration/util); to add:
	go get k8s.io/kubernetes/test/integration/util@v1.20.5
../../../../pkg/mod/github.com/prometheus/client_golang@v1.9.0/prometheus/desc.go:24:2: missing go.sum entry for module providing package github.com/golang/protobuf/proto (imported by k8s.io/client-go/discovery); to add:
	go get k8s.io/client-go/discovery@v0.20.5
../../../../pkg/mod/github.com/googleapis/gnostic@v0.4.1/openapiv2/OpenAPIv2.pb.go:9:2: missing go.sum entry for module providing package github.com/golang/protobuf/ptypes/any (imported by github.com/googleapis/gnostic/openapiv2); to add:
	go get github.com/googleapis/gnostic/openapiv2@v0.4.1
../../../../pkg/mod/github.com/prometheus/client_golang@v1.9.0/prometheus/value.go:24:2: missing go.sum entry for module providing package github.com/golang/protobuf/ptypes (imported by github.com/prometheus/client_golang/prometheus); to add:
	go get github.com/prometheus/client_golang/prometheus@v1.9.0
../../../../pkg/mod/k8s.io/apimachinery@v0.20.5/pkg/util/diff/diff.go:27:2: missing go.sum entry for module providing package github.com/google/go-cmp/cmp (imported by k8s.io/kubernetes/pkg/scheduler); to add:
	go get k8s.io/kubernetes/pkg/scheduler@v1.20.5
../../../../pkg/mod/github.com/prometheus/client_model@v0.2.0/go/metrics.pb.go:9:2: missing go.sum entry for module providing package github.com/golang/protobuf/ptypes/timestamp (imported by github.com/prometheus/client_model/go); to add:
	go get github.com/prometheus/client_model/go@v0.2.0
../../../../pkg/mod/k8s.io/apiserver@v0.20.5/pkg/server/egressselector/egress_selector.go:32:2: missing go.sum entry for module providing package google.golang.org/grpc (imported by k8s.io/apiserver/pkg/server/egressselector); to add:
	go get k8s.io/apiserver/pkg/server/egressselector@v0.20.5
../../../../pkg/mod/sigs.k8s.io/apiserver-network-proxy/konnectivity-client@v0.0.15/proto/client/client.pb.go:26:2: missing go.sum entry for module providing package google.golang.org/grpc/codes (imported by k8s.io/apiserver/pkg/endpoints/handlers); to add:
	go get k8s.io/apiserver/pkg/endpoints/handlers@v0.20.5
../../../../pkg/mod/sigs.k8s.io/apiserver-network-proxy/konnectivity-client@v0.0.15/proto/client/client.pb.go:27:2: missing go.sum entry for module providing package google.golang.org/grpc/status (imported by k8s.io/apiserver/pkg/endpoints/handlers); to add:
	go get k8s.io/apiserver/pkg/endpoints/handlers@v0.20.5
../../../../pkg/mod/k8s.io/apiserver@v0.20.5/pkg/authentication/token/cache/cached_token_authenticator.go:33:2: missing go.sum entry for module providing package golang.org/x/sync/singleflight (imported by k8s.io/apiserver/pkg/authentication/token/cache); to add:
	go get k8s.io/apiserver/pkg/authentication/token/cache@v0.20.5
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/connectivity/connectivity.go:21:2: missing go.sum entry for module providing package go.uber.org/zap (imported by k8s.io/component-base/logs/json); to add:
	go get k8s.io/component-base/logs/json@v0.20.5
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/connectivity/connectivity.go:22:2: missing go.sum entry for module providing package google.golang.org/grpc/connectivity (imported by go.etcd.io/etcd/clientv3/balancer); to add:
	go get go.etcd.io/etcd/clientv3/balancer@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/picker/roundrobin_balanced.go:22:2: missing go.sum entry for module providing package go.uber.org/zap/zapcore (imported by k8s.io/component-base/logs/json); to add:
	go get k8s.io/component-base/logs/json@v0.20.5
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/picker/err.go:20:2: missing go.sum entry for module providing package google.golang.org/grpc/balancer (imported by go.etcd.io/etcd/clientv3/balancer); to add:
	go get go.etcd.io/etcd/clientv3/balancer@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/picker/picker.go:22:2: missing go.sum entry for module providing package google.golang.org/grpc/resolver (imported by go.etcd.io/etcd/clientv3/balancer); to add:
	go get go.etcd.io/etcd/clientv3/balancer@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/balancer.go:30:2: missing go.sum entry for module providing package google.golang.org/grpc/resolver/dns (imported by go.etcd.io/etcd/clientv3/balancer); to add:
	go get go.etcd.io/etcd/clientv3/balancer@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/balancer/balancer.go:31:2: missing go.sum entry for module providing package google.golang.org/grpc/resolver/passthrough (imported by go.etcd.io/etcd/clientv3/balancer); to add:
	go get go.etcd.io/etcd/clientv3/balancer@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/credentials/credentials.go:27:2: missing go.sum entry for module providing package google.golang.org/grpc/credentials (imported by go.etcd.io/etcd/clientv3); to add:
	go get go.etcd.io/etcd/clientv3@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/pkg/logutil/discard_logger.go:20:2: missing go.sum entry for module providing package google.golang.org/grpc/grpclog (imported by k8s.io/kubernetes/test/integration/framework); to add:
	go get k8s.io/kubernetes/test/integration/framework@v1.20.5
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/client.go:39:2: missing go.sum entry for module providing package google.golang.org/grpc/keepalive (imported by go.etcd.io/etcd/clientv3); to add:
	go get go.etcd.io/etcd/clientv3@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/go.etcd.io/etcd@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489/clientv3/ctx.go:23:2: missing go.sum entry for module providing package google.golang.org/grpc/metadata (imported by go.etcd.io/etcd/clientv3); to add:
	go get go.etcd.io/etcd/clientv3@v0.5.0-alpha.5.0.20200910180754-dd1b699fc489
../../../../pkg/mod/k8s.io/kube-openapi@v0.0.0-20201113171705-d219536bb9fd/pkg/validation/strfmt/format.go:25:2: missing go.sum entry for module providing package github.com/mitchellh/mapstructure (imported by k8s.io/kube-openapi/pkg/validation/strfmt); to add:
	go get k8s.io/kube-openapi/pkg/validation/strfmt@v0.0.0-20201113171705-d219536bb9fd
../../../../pkg/mod/k8s.io/legacy-cloud-providers@v0.20.5/azure/azure_fakes.go:24:2: missing go.sum entry for module providing package github.com/golang/mock/gomock (imported by k8s.io/legacy-cloud-providers/azure); to add:
	go get k8s.io/legacy-cloud-providers/azure@v0.0.0
../../../../pkg/mod/k8s.io/legacy-cloud-providers@v0.20.5/gce/gce.go:34:2: missing go.sum entry for module providing package cloud.google.com/go/compute/metadata (imported by k8s.io/legacy-cloud-providers/gce); to add:
	go get k8s.io/legacy-cloud-providers/gce@v0.0.0
../../../../pkg/mod/github.com/!google!cloud!platform/k8s-cloud-provider@v0.0.0-20200415212048-7901bc822317/pkg/cloud/meta/meta.go:22:2: missing go.sum entry for module providing package google.golang.org/api/compute/v0.alpha (imported by github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud); to add:
	go get github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud@v0.0.0-20200415212048-7901bc822317
../../../../pkg/mod/github.com/!google!cloud!platform/k8s-cloud-provider@v0.0.0-20200415212048-7901bc822317/pkg/cloud/meta/meta.go:23:2: missing go.sum entry for module providing package google.golang.org/api/compute/v0.beta (imported by github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud); to add:
	go get github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud@v0.0.0-20200415212048-7901bc822317
../../../../pkg/mod/github.com/!google!cloud!platform/k8s-cloud-provider@v0.0.0-20200415212048-7901bc822317/pkg/cloud/meta/meta.go:24:2: missing go.sum entry for module providing package google.golang.org/api/compute/v1 (imported by github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud); to add:
	go get github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud@v0.0.0-20200415212048-7901bc822317
../../../../pkg/mod/github.com/!google!cloud!platform/k8s-cloud-provider@v0.0.0-20200415212048-7901bc822317/pkg/cloud/gce_projects.go:26:2: missing go.sum entry for module providing package google.golang.org/api/googleapi (imported by github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud); to add:
	go get github.com/GoogleCloudPlatform/k8s-cloud-provider/pkg/cloud@v0.0.0-20200415212048-7901bc822317
../../../../pkg/mod/k8s.io/legacy-cloud-providers@v0.20.5/gce/gce.go:36:2: missing go.sum entry for module providing package golang.org/x/oauth2/google (imported by k8s.io/legacy-cloud-providers/gce); to add:
	go get k8s.io/legacy-cloud-providers/gce@v0.0.0
../../../../pkg/mod/k8s.io/legacy-cloud-providers@v0.20.5/gce/gce.go:40:2: missing go.sum entry for module providing package google.golang.org/api/container/v1 (imported by k8s.io/legacy-cloud-providers/gce); to add:
	go get k8s.io/legacy-cloud-providers/gce@v0.0.0
../../../../pkg/mod/k8s.io/legacy-cloud-providers@v0.20.5/gce/gce.go:41:2: missing go.sum entry for module providing package google.golang.org/api/option (imported by k8s.io/legacy-cloud-providers/gce); to add:
	go get k8s.io/legacy-cloud-providers/gce@v0.0.0
../../../../pkg/mod/k8s.io/legacy-cloud-providers@v0.20.5/gce/gce_tpu.go:29:2: missing go.sum entry for module providing package google.golang.org/api/tpu/v1 (imported by k8s.io/legacy-cloud-providers/gce); to add:
	go get k8s.io/legacy-cloud-providers/gce@v0.0.0
Error: not all generators ran successfully
run `controller-gen crd:trivialVersions=true,preserveUnknownFields=false paths=./... output:crd:artifacts:config=/crd/bases -w` to see all available markers, or `controller-gen crd:trivialVersions=true,preserveUnknownFields=false paths=./... output:crd:artifacts:config=/crd/bases -h` for usage
```

hack/verify-crdgen.sh
pravarag: I've updated the code and fixed the controller-gen command but facing the error as mentioned here: https://github.com/kubernetes-sigs/scheduler-plugins/pull/222#issuecomment-896978942

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,53 @@
...
+source "${SCHED_PLUGIN_ROOT}/hack/lib/init.sh"
+
+kube::golang::verify_go_version
+
+cd "${SCHED_PLUGIN_ROOT}"
+
+CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Create a temp directory
+_tmpdir="$(mkdir -p "${SCHED_PLUGIN_ROOT}/_tmp")"
```
Huang-Wei, line 31: Try to leverage `mktemp`:

```bash
_tmpdir="$(mktemp -d /tmp/sched-plugins.XXXXX)"
```

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,53 @@
...
+trap "rm -rf ${_tmpdir}" EXIT
+
+# Download controller-gen locally
+CONTROLLER_GEN="${GOPATH}/bin/controller-gen"
+$(go get sigs.k8s.io/controller-tools/cmd/controller-gen)
```
Huang-Wei, line 36: `go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest` works

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,53 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCHED_PLUGIN_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+source "${SCHED_PLUGIN_ROOT}/hack/lib/init.sh"
+
+kube::golang::verify_go_version
+
+cd "${SCHED_PLUGIN_ROOT}"
```
Huang-Wei, line 26: all the `cd` and `pushd` operations can be deleted as this script is assumed to be run under the project root folder.

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,53 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCHED_PLUGIN_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
```
Huang-Wei, line 21: Let's name it `SCRIPT_ROOT` - to be consistent with other scripts.

go.mod
```diff
@@ -3,26 +3,28 @@ module sigs.k8s.io/scheduler-plugins
...
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.0.8
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/paypal/load-watcher v0.1.1
 	github.com/spf13/pflag v1.0.5
-	github.com/stretchr/testify v1.6.1
-	k8s.io/api v0.20.5
-	k8s.io/apiextensions-apiserver v0.0.0
-	k8s.io/apimachinery v0.20.5
+	github.com/stretchr/testify v1.7.0
+	k8s.io/api v0.21.3
```
Huang-Wei, line 13: We should keep it with version v0.20.X

pravarag: Updated with latest push

hack/verify-crdgen.sh
pravarag: Changed the name as suggested.

pravarag:
I've updated the changes as per @Huang-Wei 's suggestions and after running in my local, I'm getting the below output:

```
 ->  ./hack/verify-crdgen.sh
SCRIPT_ROOT: ./hack/..
PWD: /Users/pravaragrawal/Work/go_work/src/github.com/pravarag/scheduler-plugins
Generated output differs:
diff -Naupr manifests/coscheduling/crd.yaml /tmp/sched-plugins.pTLF/crd/bases/crd.yaml
--- manifests/coscheduling/crd.yaml	2021-08-02 15:41:06.000000000 +0530
+++ /tmp/sched-plugins.pTLF/crd/bases/crd.yaml	1970-01-01 05:30:00.000000000 +0530
@@ -1,59 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1
-kind: CustomResourceDefinition
-metadata:
-  name: podgroups.scheduling.sigs.k8s.io
-  annotations:
-    "api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/scheduler-plugins/pull/50"
-spec:
-  group: scheduling.sigs.k8s.io
-  names:
-    kind: PodGroup
-    plural: podgroups
-    singular: podgroup
-    shortNames:
-    - pg
-    - pgs
-  scope: Namespaced
-  versions:
-  - name: "v1alpha1"
-    served: true
-    storage: true
-    schema:
-      openAPIV3Schema:
-        type: object
-        properties:
-          spec:
-            type: object
-            properties:
-              minMember:
-                type: integer
-                minimum: 1
-              scheduleTimeoutSeconds:
-                type: integer
-              minResources:
-                type: object
-                additionalProperties:
-                  type: string
-          status:
-            type: object
-            properties:
-              phase:
-                type: string
-              occupiedBy:
-                type: string
-              scheduled:
-                type: integer
-                default: 0
-              running:
-                type: integer
-                default: 0
-              succeeded:
-                type: integer
-                default: 0
-              failed:
-                type: integer
-                default: 0
-              scheduleStartTime:
-                type: string
-                format: date-time
-
diff -Naupr manifests/coscheduling/scheduler-config.yaml /tmp/sched-plugins.pTLF/crd/bases/scheduler-config.yaml
--- manifests/coscheduling/scheduler-config.yaml	2021-08-02 15:41:06.000000000 +0530
+++ /tmp/sched-plugins.pTLF/crd/bases/scheduler-config.yaml	1970-01-01 05:30:00.000000000 +0530
@@ -1,32 +0,0 @@
-apiVersion: kubescheduler.config.k8s.io/v1beta1
-kind: KubeSchedulerConfiguration
-leaderElection:
-  leaderElect: false
-clientConnection:
-  kubeconfig: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
-profiles:
-- schedulerName: default-scheduler
-  plugins:
-    queueSort:
-      enabled:
-        - name: Coscheduling
-      disabled:
-        - name: "*"
-    preFilter:
-      enabled:
-        - name: Coscheduling
-    permit:
-      enabled:
-        - name: Coscheduling
-    reserve:
-      enabled:
-        - name: Coscheduling
-    postBind:
-      enabled:
-        - name: Coscheduling
-  pluginConfig:
-  - name: Coscheduling
-    args:
-      permitWaitingTimeSeconds: 10
-      deniedPGExpirationTimeSeconds: 3
-      kubeConfigPath: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
diff -Naupr manifests/coscheduling/scheduling.sigs.k8s.io_elasticquotas.yaml /tmp/sched-plugins.pTLF/crd/bases/scheduling.sigs.k8s.io_elasticquotas.yaml
--- manifests/coscheduling/scheduling.sigs.k8s.io_elasticquotas.yaml	1970-01-01 05:30:00.000000000 +0530
+++ /tmp/sched-plugins.pTLF/crd/bases/scheduling.sigs.k8s.io_elasticquotas.yaml	2021-08-12 19:48:13.000000000 +0530
@@ -0,0 +1,83 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: elasticquotas.scheduling.sigs.k8s.io
+spec:
+  group: scheduling.sigs.k8s.io
+  names:
+    kind: ElasticQuota
+    listKind: ElasticQuotaList
+    plural: elasticquotas
+    singular: elasticquota
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: ElasticQuota sets elastic quota restrictions per namespace
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: ElasticQuotaSpec defines the Min and Max for Quota.
+            properties:
+              max:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: Max is the set of desired max limits for each named resource.
+                  The usage of max is based on the resource configurations of successfully
+                  scheduled pods.
+                type: object
+              min:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: Min is the set of desired guaranteed limits for each
+                  named resource.
+                type: object
+            type: object
+          status:
+            description: ElasticQuotaStatus defines the observed use.
+            properties:
+              used:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: Used is the current observed total usage of the resource
+                  in the namespace.
+                type: object
+            type: object
+        type: object
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
diff -Naupr manifests/coscheduling/scheduling.sigs.k8s.io_podgroups.yaml /tmp/sched-plugins.pTLF/crd/bases/scheduling.sigs.k8s.io_podgroups.yaml
--- manifests/coscheduling/scheduling.sigs.k8s.io_podgroups.yaml	1970-01-01 05:30:00.000000000 +0530
+++ /tmp/sched-plugins.pTLF/crd/bases/scheduling.sigs.k8s.io_podgroups.yaml	2021-08-12 19:48:13.000000000 +0530
@@ -0,0 +1,102 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: podgroups.scheduling.sigs.k8s.io
+spec:
+  group: scheduling.sigs.k8s.io
+  names:
+    kind: PodGroup
+    listKind: PodGroupList
+    plural: podgroups
+    singular: podgroup
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    schema:
+      openAPIV3Schema:
+        description: PodGroup is a collection of Pod; used for batch workload.
+        properties:
+          apiVersion:
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
+            type: string
+          kind:
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            type: string
+          metadata:
+            type: object
+          spec:
+            description: Specification of the desired behavior of the pod group.
+            properties:
+              minMember:
+                description: MinMember defines the minimal number of members/tasks
+                  to run the pod group; if there's not enough resources to start all
+                  tasks, the scheduler will not start anyone.
+                format: int32
+                type: integer
+              minResources:
+                additionalProperties:
+                  anyOf:
+                  - type: integer
+                  - type: string
+                  pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                  x-kubernetes-int-or-string: true
+                description: MinResources defines the minimal resource of members/tasks
+                  to run the pod group; if there's not enough resources to start all
+                  tasks, the scheduler will not start anyone.
+                type: object
+              scheduleTimeoutSeconds:
+                description: ScheduleTimeoutSeconds defines the maximal time of members/tasks
+                  to wait before run the pod group;
+                format: int32
+                type: integer
+            type: object
+          status:
+            description: Status represents the current information about a pod group.
+              This data may not be up to date.
+            properties:
+              failed:
+                description: The number of pods which reached phase Failed.
+                format: int32
+                type: integer
+              occupiedBy:
+                description: OccupiedBy marks the workload (e.g., deployment, statefulset)
+                  UID that occupy the podgroup. It is empty if not initialized.
+                type: string
+              phase:
+                description: Current phase of PodGroup.
+                type: string
+              running:
+                description: The number of actively running pods.
+                format: int32
+                type: integer
+              scheduleStartTime:
+                description: ScheduleStartTime of the group
+                format: date-time
+                type: string
+              scheduled:
+                description: The number of actively running pods.
+                format: int32
+                type: integer
+              succeeded:
+                description: The number of pods which reached phase Succeeded.
+                format: int32
+                type: integer
+            type: object
+        type: object
+    served: true
+    storage: true
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: []
+  storedVersions: []
Verification failed.
```
I'm not sure if this is correct validation, but I'm still looking for other scenarios to validate the output.

Huang-Wei:
@pravarag the output looks promising now. I think we can use the output generated by `controller-gen` as the base crd manifest, and add some customized ones like "annotations.api-approved.kubernetes.io" and "spec.names.shortNames". (please help to verify if other things are missed). So you can modifies `manifests/*/crd.yaml` accordingly.

The other thing is that we may have to verify the crd yamls individually - unlike the current logic that compares all in a batch. So the desired logic maybe sth. like:

```
controller-gen to generate all crd base yamls
for each crd.yaml under folder "manifests":
    compare it with the one generated by controller-gen
done
```

Huang-Wei:
Another abnormal symptom I spotted is: by running `controller-gen`:

```
controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" paths="./..." output:crd:artifacts:config=/tmp/crd/bases
```

It only generates 2 crd base yamls:

```
⇒  ls /tmp/crd/bases
scheduling.sigs.k8s.io_elasticquotas.yaml  scheduling.sigs.k8s.io_podgroups.yaml
```

But actually we have 3 in place:

```
⇒  ls manifests/*/crd.yaml
manifests/capacityscheduling/crd.yaml  manifests/coscheduling/crd.yaml  manifests/noderesourcetopology/crd.yaml
```

Need to figure out why the types of noderesourcetopology cannot be detected by `controller-gen`.

pravarag:
> Another abnormal symptom I spotted is: by running `controller-gen`:
> 
> ```
> controller-gen "crd:trivialVersions=true,preserveUnknownFields=false" paths="./..." output:crd:artifacts:config=/tmp/crd/bases
> ```
> 
> It only generates 2 crd base yamls:
> 
> ```
> ⇒  ls /tmp/crd/bases
> scheduling.sigs.k8s.io_elasticquotas.yaml  scheduling.sigs.k8s.io_podgroups.yaml
> ```
> 
> But actually we have 3 in place:
> 
> ```
> ⇒  ls manifests/*/crd.yaml
> manifests/capacityscheduling/crd.yaml  manifests/coscheduling/crd.yaml  manifests/noderesourcetopology/crd.yaml
> ```
> 
> Need to figure out why the types of noderesourcetopology cannot be detected by `controller-gen`.

@Huang-Wei I did find a slightly relative issue with controller-gen not picking up all the resources reported here, https://github.com/kubernetes-sigs/kubebuilder/issues/2270. But since the other issue is for a particular api, I'm not sure if we can make a relative change 🤔

damemi:
@pravarag `controller-gen` isn't picking up the noderesourcetopology types because there is no `topology.node.k8s.io` API group defined under the API paths passed to that command, and it doesn't look like those types are defined in this repo, are they?

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,51 @@
...
+#CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+CRD_OPTIONS="crd:trivialVersions=true,preserveUnknownFields=false"
+
+# Create a temp directory
+_tmpdir="$(mktemp -d "/tmp/sched-plugins.XXXX")"
+# trap "rm -rf ${_tmpdir}" EXIT
+
+# Download controller-gen locally
+CONTROLLER_GEN="${GOPATH}/bin/controller-gen"
+$(go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest)
```
Huang-Wei, line 37: Why embed it here? We should let the error to be yielded if there is any.

Huang-Wei:
> because there is no `topology.node.k8s.io` API group defined under the API paths passed to that command

Ah, that's true. We should explicitly specify the API paths.

@prachirp you can specify a list of the API paths explicitly, and then run the `controller-gen` command:

```bash
api_paths="./pkg/apis/scheduling/v1alpha1/...;./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/..."
controller-gen crd:trivialVersions=true,preserveUnknownFields=false paths="$api_paths" output:crd:stdout

# compare the above output with `cat manifests/*/crd.yaml`. (some special characters can be omitted)
```

pravarag:
> @pravarag you can specify a list of the API paths explicitly, and then run the `controller-gen` command:
> 
> ```shell
> api_paths="./pkg/apis/scheduling/v1alpha1/...;./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/..."
> controller-gen crd:trivialVersions=true,preserveUnknownFields=false paths="$api_paths" output:crd:stdout
> 
> # compare the above output with `cat manifests/*/crd.yaml`. (some special characters can be omitted)
> ```
I'm somehow getting this error while trying to use the `noderesourcetopology-api` path as below,

```
controller-gen crd:trivialVersions=true,preserveUnknownFields=false paths=./pkg/apis/scheduling/v1alpha1/...;./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/... output:crd:stdout -w
-bash: ./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/...: No such file or directory
```
Is there a way I can confirm the exact path name?

damemi:
@pravarag @Huang-Wei `-bash: ./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/...: No such file or directory` is probably because we don't commit the vendor directory to this repo

I would be +1 to start committing the vendor directory here. The original argument against it was that this was meant to be more of a toolbox of plugins, but now that this repo is being published as its own scheduler I think committing an up-to-date vendor directory is important (eg, for development, offline builds, and now trying to generate CRDs from vendored api packages).

pravarag:
> I would be +1 to start committing the vendor directory here. The original argument against it was that this was meant to be more of a toolbox of plugins, but now that this repo is being published as its own scheduler I think committing an up-to-date vendor directory is important (eg, for development, offline builds, and now trying to generate CRDs from vendored api packages).
> 
> The other option is to clone that repo and reference the local path instead, but this would be inconsistent among different users.

Thanks @damemi for mentioning the failure reason and though I'm a little new around on how to commit a different directory in the existing repo but what I'm guessing is to basically use `vendor` directory as a submodule under `scheduler-plugins` and is this the correct directory: https://github.com/kubernetes/kubernetes/tree/master/vendor/github.com ? 
If so, I'm not able to locate the following sub-directory named as `k8stopologyawareschedwg/noderesourcetopology-api/` under the above mentioned

Huang-Wei:
@damemi yes, I've been thinking of that for a while. Especially this project gains more interested and we've quite a lot of questions asking about configuring a dev env.

I will work on that.

@pravarag Before I commit the whole vendor directory, you can run `go mod vendor && hack/update-generated-openapi.sh` to generate the vendor dir properly.

damemi:
> and is this the correct directory: https://github.com/kubernetes/kubernetes/tree/master/vendor/github.com

No, what you need to do is run `go mod vendor` from this repo, like @Huang-Wei said, which should create a new `vendor` directory with all of this repo's dependencies. This does not have anything to do with the `vendor` directory in `k8s.io/kubernetes` (though side note, that directory exists from the same `go mod vendor` command, except run there).

You'll also need to remove the `vendor/` line from the `.gitignore` file

pravarag:
Thanks @Huang-Wei @damemi for your suggestions, I'm now able to generate all the files under `/tmp/sched-plugins.XXX/` directory as below,

```
->  ls -latr /tmp/sched-plugins.P0mF/crd/bases/
scheduling.sigs.k8s.io_elasticquotas.yaml         scheduling.sigs.k8s.io_podgroups.yaml             topology.node.k8s.io_noderesourcetopologies.yaml
```
Now, one thing to note here is these files are getting generated with different names and to accommodate the suggestions in this [comment](https://github.com/kubernetes-sigs/scheduler-plugins/pull/222#issuecomment-897816284) to verify the files individually, I'm trying to implement something like this:

```
# compare crds individualy
crd_files="$(ls ./manifests/*/crd.yaml)"

for file in ${crd_files}; do
    if $file == *"capacityscheduling"*; then
        if ! _out="$(diff -au "${file}" "${_tmpdir}/crd/bases/scheduling.sigs.k8s.io_elasticquotas.yaml}")"; then
        echo "Generated output differs:" >&2
        echo "${_out}" >&2
        echo "Verification failed."
        exit 1
        fi
    fi
done
```
Please correct if this is a wrong approach for the same.

Huang-Wei:
@damemi 's proposal looks pretty neat. To add up to simplify the comparing, we can create a dedicated folder called `manifests/crds`, and place the crd yamls. Then use `controller-gen schemapatch` and `git diff` to verify.

To avoid crd yamls' duplication, we can create symbolic links in their corresponding folder:

- manifests/capacityscheduling/crd.yaml -> manifests/crds/scheduling.sigs.k8s.io_elasticquotas.yaml
- manifests/coscheduling/crd.yaml -> manifests/crds/scheduling.sigs.k8s.io_podgroups.yaml
- manifests/noderesourcetopology/crd.yaml -> manifests/crds/topology.node.k8s.io_noderesourcetopologies.yaml

pravarag:
I tried to make it work with `schemapath` changes but it's still generates only one `crd.yaml` under the output directory. I've reverted back to the point where we are evaluating all the crds individually, opening it for review. @Huang-Wei @damemi @seanmalloy please suggest for any changes that I can accommodate in the current code 🙂

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Generate CRD
+api_paths="./pkg/apis/scheduling/v1alpha1/...;./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/..."
+
+${CONTROLLER_GEN} ${CRD_OPTIONS} paths="${api_paths}" output:crd:artifacts:config=${_tmpdir}/crd/bases
+
+# compare crds individualy
+crd_files="$(ls ./manifests/*/crd.yaml)"
+
+for file in ${crd_files}; do
```
Huang-Wei, line 47: It'd be easier if we do this way:

1. Soft-link the *crd.yaml to a specific folder `./manifests/crds/`, and name them with the exact name that `controller-gen` would generate.
    ```
    $ ll manifests/crds
    total 0
    lrwxr-xr-x 1 wei.huang1 30 Sep  3 14:05 scheduling.sigs.k8s.io_elasticquotas.yaml -> ../capacityscheduling/crd.yaml
    lrwxr-xr-x 1 wei.huang1 24 Sep  3 14:05 scheduling.sigs.k8s.io_podgroups.yaml -> ../coscheduling/crd.yaml
    lrwxr-xr-x 1 wei.huang1 32 Sep  3 14:06 topology.node.k8s.io_noderesourcetopologies.yaml -> ../noderesourcetopology/crd.yaml
    ```
2. Just specify the output folder to `./manifests/crds`, so the generate output will overwrite.
3. Then use `git diff` to verify if there are any significant incompatible diffs.
4. In this PR, check in the generated crds, but keeps the original manual-maintained fields:
    - annotations.api-approved.kubernetes.io
    - spec.names.shortNames

I verified locally, which should work.

hack/verify-crdgen.sh
pravarag: Here, did you mean remove the `$()` enclosing for the `go install` command and let it run as it is?

pravarag:
The following PR test: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_scheduler-plugins/222/pull-scheduler-plugins-verify-gofmt-master/1435199796162859008 is failing, I'm guessing this could be due to missing `vendor` packages in the directory?

hack/verify-crdgen.sh
```diff
@@ -0,0 +1,46 @@
...
+# Download controller-gen locally
+CONTROLLER_GEN="${GOPATH}/bin/controller-gen"
+$(go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest)
+
+# Generate CRD
+api_paths="./pkg/apis/scheduling/v1alpha1/...;./vendor/github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/..."
+
+${CONTROLLER_GEN} ${CRD_OPTIONS} paths="${api_paths}" output:dir="./manifests/crds"
+
+if ! _out="$(git diff ./manifests)"; then
```
Huang-Wei, line 39: Firstly, by default, `git diff` would always return 0, so this line would always succeed. We can add `--exit-code` to the git command to fix it.

Secondly, use `--no-pager` to ensure the diff is captured properly.
Thirdly, as we meniotned earlier, in some cases, we need to manually provide some fields, and we don't want to include those in the diff result. So we should (1) define a common pattern to detect, e.g., simply add a `# edited manually` comment on each line, and then (2) use a `-I` option to exclude the keywords.

More details, please check the suggested diff.

manifests/capacityscheduling/crd.yaml
```diff
@@ -1,12 +1,19 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: elasticquotas.scheduling.sigs.k8s.io
   annotations:
     "api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/scheduler-plugins/pull/52"
```
Huang-Wei, line 8: nit:

```suggestion
    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/52
```

Huang-Wei (commented):
We're close to merging. I drafted a diff to make the whole script work: https://gist.github.com/Huang-Wei/81d46d393b1339878287bb631339490f. See below for detailed explanations.

Makefile
```diff
@@ -125,6 +125,7 @@ integration-test: install-etcd autogen
 .PHONY: verify-gofmt
 verify-gofmt:
 	hack/verify-gofmt.sh
+	hack/verify-crdgen.sh
```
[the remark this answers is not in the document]
pravarag: Added

manifests/capacityscheduling/crd.yaml
```diff
@@ -1,12 +1,19 @@
...
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: elasticquotas.scheduling.sigs.k8s.io
```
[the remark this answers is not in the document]
pravarag: Done

manifests/coscheduling/crd.yaml
```diff
@@ -1,59 +1,107 @@
...
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: podgroups.scheduling.sigs.k8s.io
   annotations:
     "api-approved.kubernetes.io": "https://github.com/kubernetes-sigs/scheduler-plugins/pull/50"
+    controller-gen.kubebuilder.io/version: v0.6.2
+  creationTimestamp: null
+  name: podgroups.scheduling.sigs.k8s.io
```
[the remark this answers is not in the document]
pravarag: Done

manifests/noderesourcetopology/crd.yaml
```diff
@@ -1,9 +1,11 @@
...
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
-    controller-gen.kubebuilder.io/version: v0.6.0
+    controller-gen.kubebuilder.io/version: v0.6.2
   creationTimestamp: null
   name: noderesourcetopologies.topology.node.k8s.io
   namespace: ""
```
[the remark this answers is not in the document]
pravarag: Done

hack/verify-crdgen.sh
pravarag: Done.

manifests/capacityscheduling/crd.yaml
```diff
@@ -1,54 +1,88 @@
...
 spec:
   group: scheduling.sigs.k8s.io
   names:
+    kind: ElasticQuota
+    listKind: ElasticQuotaList
     plural: elasticquotas
     singular: elasticquota
     kind: ElasticQuota
```
Huang-Wei, line 18: This is duplicated.

manifests/noderesourcetopology/crd.yaml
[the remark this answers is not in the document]
Huang-Wei: it's not removed.

manifests/noderesourcetopology/crd.yaml
```diff
@@ -1,9 +1,10 @@
+
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
-    controller-gen.kubebuilder.io/version: v0.6.0
```
Huang-Wei, line 6: restore this line, and replace v0.6.0 with v0.6.2

manifests/capacityscheduling/crd.yaml
pravarag: Removed the duplicated entry.

manifests/noderesourcetopology/crd.yaml
pravarag: I've restored the line and put v0.6.2 as new value

manifests/capacityscheduling/crd.yaml
Huang-Wei: Well, the comment was meant to delete the latter `kind: ElasticQuota` so as to keep the fields sorted alphabetically; otherwise the codegren diffs. Also `listKind` should be restored.

(you can take a look at the CI failure log: click the "raw build-log.txt" link on https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_scheduler-plugins/222/pull-scheduler-plugins-unit-test-master/1435806643223793664)

manifests/noderesourcetopology/crd.yaml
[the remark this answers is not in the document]
Huang-Wei: bump.

manifests/capacityscheduling/crd.yaml
pravarag: Oh my bad, I misinterpreted that. I've updated the code with latest push.

manifests/noderesourcetopology/crd.yaml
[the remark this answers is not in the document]
pravarag: Again missed it, changed with latest push.

manifests/noderesourcetopology/crd.yaml
pravarag: The last falling piece, I again misread the change. Fixed it and I can see PR checks have also passed now. Thanks @Huang-Wei for your valuable reviews 🙂

Huang-Wei:
Pretty neat! Thanks @pravarag .

/lgtm
/approve

PR #223, opened 2021-08-03
volume topology spread

mruck:


seanmalloy:
@mruck were you able to sign the CLA?

Huang-Wei:
@mruck thanks for the PR. 

Before jumping into the implementation, we typically create an issue to ensure a requirement is thoroughly discussed first, and then decide where we can hold the implementation. If you don't mind, could you discuss the idea first?

Issue #236, opened 2021-08-24
Missing scheme_test.go


Huang-Wei:
We have main_test.go to ensure the plugins are instantiated properly, but we didn't cover whether other portions such as plugin arguments are decoded as expected.

We need a pkg/apis/config/scheme/scheme_test.go to cover this. Sample can be found in k/k: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/scheme/scheme_test.go.

lianghao208:
/assign
I can fix this.

Huang-Wei:
Fixed by #237.

PR #237, opened 2021-08-25
Add missing scheme_test.go

lianghao208:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

Adds missing scheme_test.go to ensure plugin arguments are decoded as expected.
 
 **Which issue(s) this PR fixes**:

#236

### DIFF
--- a/pkg/apis/config/scheme/scheme_test.go
+++ b/pkg/apis/config/scheme/scheme_test.go
@@ -1,0 +1,423 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheme
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


pkg/apis/config/scheme/scheme_test.go
```diff
@@ -0,0 +1,336 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	v1 "k8s.io/kube-scheduler/config/v1"
+	defaultconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
+	"sigs.k8s.io/scheduler-plugins/pkg/apis/config"
```
Huang-Wei, line 30: nit: add an empty line above

pkg/apis/config/scheme/scheme_test.go
```diff
@@ -0,0 +1,336 @@
...
+// into their appropriate internal types and defaults are applied.
+func TestCodecsDecodePluginConfig(t *testing.T) {
+	testCPUQuantity, _ := resource.ParseQuantity("1000m")
+	testCases := []struct {
+		name         string
+		data         []byte
+		wantErr      string
+		wantProfiles []defaultconfig.KubeSchedulerProfile
+	}{
+		//v1beta1 tests
```
Huang-Wei, line 51:
```suggestion
		// v1beta1 tests
```

pkg/apis/config/scheme/scheme_test.go
```diff
@@ -0,0 +1,336 @@
...
+									Type:    config.Prometheus,
+									Address: "http://prometheus-k8s.monitoring.svc.cluster.local:9090",
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
Huang-Wei, line 143: Can you add more cases?

- leave some fields unspecified, and then verify they get defaulted properly
- provide some illegal args, and verify if we get the validation error

lianghao208: done, thanks for the review!

PR #238, opened 2021-08-25
Fix access to null pointer

yangtau:
What this PR does:
- fix access to nil pointer
- lowercase the first letter of reasons for `framework.Status`
   (I think it is better to lowercase the reason string, and the reason is similar to [error strings](https://github.com/golang/go/wiki/CodeReviewComments#error-strings))


yangtau:
@Huang-Wei could you review this PR, thanks

Huang-Wei:
Thanks @yangtau !

Would you mind also back-porting it to 1.20 and 1.19 ?

PR #248, opened 2021-09-02
Fix podgroup controller UT issue

JaneLiuL:
Fix podgroup controller UT issue, I found the source code line 157 if err is not nil, would not deal with.
So I try to add some debug log, the result as below
```
--- FAIL: Test_Run (2.02s)
    --- FAIL: Test_Run/Group_status_convert_from_scheduling_to_succeed#01 (0.20s)
        podgroup_test.go:163: pg.Status.Phase not as expect pg6, now the pg.Status.Phase is Finished
    --- FAIL: Test_Run/Group_status_convert_from_pending_to_prescheduling (0.20s)
        podgroup_test.go:163: pg.Status.Phase not as expect pg7, now the pg.Status.Phase is  Finished
```

### DIFF
--- a/pkg/controller/podgroup_test.go
+++ b/pkg/controller/podgroup_test.go
@@ -96,7 +96,7 @@ func Test_Run(t *testing.T) {
 			podNames:          []string{"pod1", "pod2"},
 			podPhase:          v1.PodPending,
 			previousPhase:     v1alpha1.PodGroupScheduling,
-			desiredGroupPhase: v1alpha1.PodGroupFailed,
+			desiredGroupPhase: v1alpha1.PodGroupFinished,
 			podNextPhase:      v1.PodSucceeded,
 		},
 		{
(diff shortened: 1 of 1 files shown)


JaneLiuL:
/assign @Huang-Wei

@Huang-Wei could you review this PR, thanks

pkg/controller/podgroup_test.go
```diff
@@ -154,7 +163,7 @@ func Test_Run(t *testing.T) {
 				return true, nil
 			})
 			if err != nil {
-
+				t.Fatal("Unexpected error", err)
```
Huang-Wei, line 166: Thanks for the catch. A bit surprised nobody spotted this before...

`ErrorS` should suffice:

```suggestion
				t.ErrorS(err, "Unexpected phase for podGroup")
```

Let's follow the structured logging format - use t.

JaneLiuL: Hi, t have no ErrorS method, it provide Fatal or Errorf or Error, do you mean klog.ErrorS?
`t.ErrorS undefined (type *"testing".T has no field or method ErrorS)`
I still think `t.Fatal` is better :)

Huang-Wei: I misread... Apology for that.

Issue #249, opened 2021-09-02
Noderesourcetopology plugin disqualifies pods that request memory


Tal-or:
**What happened:**
I'm trying to deploy a pod that requests some memory and has a guaranteed Qos, using the noderesourcetopology plugin.
```
kind: Pod
...
  schedulerName: topo-aware-scheduler
  resources:
       limits:
         memory: "100Mi"
         cpu: "1"
       requests:
         memory: "100Mi"
         cpu: "1"
```
Although there are nodes in the cluster that are able to address the pod's request, the pod is getting qualified.

**What you expected to happen:**
The pod should be admitted by the scheduler plugin.

**How to reproduce it (as minimally and precisely as possible):**
1. On a Kind cluster, deploy the scheduler-plugin using the installation instruction from here: https://github.com/killianmuldoon/documentation/tree/installguide/install-guides 
2. Create a pod with the spec as described above

**Why it happened?:**
NFD, which is the resource-exporter component for the scheduler plugin, doesn't hold information about memory distribution between NUMA zones.

On the other hand, the noderesourcetopology plugin is looking for the memory information in order to decide if the node can address the pod's request.

When noderesourcetopology plugin queries NFD for memory information (which NFD doesn't have) it disqualified the node.
The same happens for all nodes, thus no pod which asks for memory can be deployed.

**How it can be resolved?**
NFD should provide the memory information.
The memory information already exported by the PodreasourceAPI: https://github.com/kubernetes/kubernetes/pull/101030 
Once NFD will have the ability to account for the memory resources, we should update the plugin to work with the fixed NFD version and the issue will get resolved.

AlexeyPerevalov:
Description little bit incorrect.
Now plugin works it around https://github.com/kubernetes-sigs/scheduler-plugins/blob/9487dd73fb5b38b7ac81076260eb638e56340e41/pkg/noderesourcetopology/filter.go#L79
by counting memory as available resource always, this is not accurate.
Yes we have to remove this condition.

Huang-Wei:
@AlexeyPerevalov could you help check if NFD version is compatible with the plugin logic? for both master and other release branches?

Tal-or:
> Description little bit incorrect.
> Now plugin works it around
> 
> https://github.com/kubernetes-sigs/scheduler-plugins/blob/9487dd73fb5b38b7ac81076260eb638e56340e41/pkg/noderesourcetopology/filter.go#L79
> 
> 
[...]

PR #256, opened 2021-09-09
noderesourcetopology: crosschecks NUMA node resources with node resources

Tal-or:
This patch intended to solve https://github.com/kubernetes-sigs/scheduler-plugins/issues/249

In the current implementation, in case that pod asks for a resource that can't be found on the NUMA node, the filter plugin would filter out the NUMA node.

This behavior is not always true, because sometimes there are resources that are not NUMA aligned (like specific devices)
or not yet supported by the topology exporter (like memory).

We also need to make sure that the requested resource does not exist at the node level.

Signed-off-by: Talor Itzhak <titzhak@redhat.com>

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -33,7 +33,7 @@ import (
 
 type PolicyHandler func(pod *v1.Pod, zoneMap topologyv1alpha1.ZoneList) *framework.Status
 
-func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList) *framework.Status {
+func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nodeInfo *framework.NodeInfo) *framework.Status {
 	klog.V(5).InfoS("Single NUMA node handler")
 
 	// prepare NUMANodes list from zoneMap
@@ -43,7 +43,7 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
 	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
 	// and normal containers in the one scope
 	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
-		if resMatchNUMANodes(nodes, container.Resources.Requests, qos) {
+		if resMatchNUMANodes(nodes, container.Resources.Requests, qos, nodeInfo) {
 			// definitely we can't align container, so we can't align a pod
 			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align container: %s", container.Name))
 		}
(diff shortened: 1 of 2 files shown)


Huang-Wei:
/ok-to-test

LGTM. But I will leave the official review to @swatisehgal and @AlexeyPerevalov.

Huang-Wei:
`pull-scheduler-plugins-verify-crdgen-master` failed due to missing the latest Makefile rule. You can rebase master to get it resolved.

Tal-or:
> `pull-scheduler-plugins-verify-crdgen-master` failed due to missing the latest Makefile rule. You can rebase master to get it resolved.

I see that this one https://github.com/kubernetes-sigs/scheduler-plugins/pull/258 is adding the missing Make target.
I'll rebase again after it will get merged.

pkg/noderesourcetopology/filter.go
```diff
@@ -64,11 +64,13 @@ func resMatchNUMANodes(nodes NUMANodeList, resources v1.ResourceList, qos v1.Pod
...
+		for _, numaNode := range numaNodes {
 			numaQuantity, ok := numaNode.Resources[resource]
-			// if can't find requested resource on the node - skip (don't set it as available NUMA node)
-			// if unfound resource has 0 quantity probably this numa node can be considered
-			if !ok && quantity.Cmp(zeroQuantity) != 0 {
+			// if the requested resource can't be found on the NUMA node, we still need to check
+			// if the resource can be found at the node itself, because there are resources which are not NUMA aligned
+			// or not supported by the topology exporter - if resource was not found at both checks - skip (don't set it as available NUMA node).
+			// if the un-found resource has 0 quantity probably this numa node can be considered.
+			if !ok && !resourceFoundOnNode(resource, quantity, nodeInfo) && quantity.Cmp(zeroQuantity) != 0 {
```
AlexeyPerevalov, line 73: ok, we're skipping every resource which is not on the worker, IOW we don't skip if it's on the worker node and its quantity is enough.

pkg/noderesourcetopology/filter.go
```diff
@@ -143,3 +145,15 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
 	}
 	return nil
 }
+
+// resourceFoundOnNode checks whether a given resource exist at the node level
+// and whether the given quantity is big enough
+func resourceFoundOnNode(wantResource v1.ResourceName, wantQuantity resource.Quantity, nodeInfo *framework.NodeInfo) bool {
```
AlexeyPerevalov, line 151: I'm not sure is it worth iterating every time, maybe just create a map in Filter, and pass then it instead of NodeInfo, WDYT?

AlexeyPerevalov (commented):
in general I'm ok, let think, worth reducing iterations?

Huang-Wei:
> I see that this one #258 is adding the missing Make target.
> I'll rebase again after it will get merged.

It's merged now.

pkg/noderesourcetopology/filter.go
Tal-or: SGTM, I'll do that

Tal-or: My mistake ```nodeInfo.Allocatable.ResourceList()``` is already a map so don't need to iterate at all.

swatisehgal:
/lgtm.
@AlexeyPerevalov Are you happy enough to merge this?

swatisehgal:
/lgtm
@AlexeyPerevalov Are you happy enough to merge this?

Huang-Wei:
/approve
/hold
for @AlexeyPerevalov to take a final look.

PR #276, opened 2021-10-20
NodeResourceTopology: Return error if no NRT instances found

swatisehgal:
In case no NRT instances are present in any of the namespaces
specified in the Topology aware Scheduler plugin config, we should
return an error with the information rather than silently admitting
the pod as that pod could end with Topology Affinity Error.

For more context please refer to the slack thread here:
https://kubernetes.slack.com/archives/C01A82FKSQK/p1633514787010400?thread_ts=1628171809.003900&cid=C01A82FKSQK

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -21,7 +21,7 @@ import (
 	"fmt"
 	"strings"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
 	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
@@ -127,7 +127,11 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
 	}
 
 	nodeName := nodeInfo.Node().Name
-	nodeTopology := findNodeTopology(nodeName, &tm.nodeResTopologyPlugin)
+	nodeTopology, err := findNodeTopology(nodeName, &tm.nodeResTopologyPlugin)
+
+	if err != nil {
+		return framework.NewStatus(framework.Error, err.Error())
+	}
 
 	if nodeTopology == nil {
 		return nil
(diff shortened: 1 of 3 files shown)


catblade:
@swatisehgal  what is the timeline on the NRT object update?  Is that tracked somewhere?

fromanirh:
> I'm not against of this change, but would like to mention, before this PR we had behavior as follows: if NRT wasn't found for appropriate namespace, it means topology check is not needed. This PR changes it and now we have to have NRT for every node involved in scheduling, so now we definitely need to narrow the worker node scopes by node label selector/node affinity mechanism, otherwise we will skip worker node w/o NRD. [...]

We were considering an additional usecase on which indeed we would experimental with a different behaviour: given a subset of nodes (which brings extra complexity as you described), the NRT data _must_ be there, and if not this should be caught by the plugin, which should return error - or in general avoid scheduling without the NRT data.

So i'd like to propose another possible approach: add another, trivial, filter (disabled by default) which passes iff the node has the corresponding NRT data available. This way:
- consumers not interested in this new behaviour can just keep things as usual and see no changes
- consumers interested in this new behaviour can enable the new filter. According to [the docs](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/#filter) we should be able to make the new filter run before the NodeResourceTopologyMatch, but even if it runs afterwards is such big of a deal IMO, as this behaviour is opt-in anyway
- we leverage the composability of the scheduling framework.

Would this approach be acceptable for you? There is no code yet, I'm just asking to explore the options.

Huang-Wei:
> Would this approach be acceptable for you? There is no code yet, I'm just asking to explore the options.

If we were to implement this in the upstream (k/k), usually we introduce a feature gate to model it - disabled by default until NRT objects are available on all nodes. However, I'm not 100% how a feature gate's semantics can be extended/leveraged in scheduler plugins. My guess is no.

In that case, a viable option is to abstract this in the plugin args. And to make the semantic obvious and backwards compatible, we'd better not default this option to false/true - users must specify it explicitly.

AlexeyPerevalov:
I agree to make it optional through plugin arg.

PR #277, opened 2021-10-20
NodeResourceTopology: Make `default` as the default namespace in scheduler config

swatisehgal:
In order to enable Topology aware scheduling in a Kuberetes cluster, we need to
deploy a producer which exposes resource topology information as CRs and the
NodeResourceTopology scheduler plugin.

An upstream producer of the NodeResourceTopology instances is Node Feature Discovery.
The associated PR where this feature is enabled in NFD is here:
https://github.com/kubernetes-sigs/node-feature-discovery/pull/525

The default namespace where NFD deploys CR instances is `default` as can be seen here:
 https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/cmd/nfd-master/main.go#L108
so in order to have a better user experience and to make sure that both
these components work together seamlessly we change the default namespace
in the scheduler plugin config to be the same as in NFD i.e. `default`.

### DIFF
--- a/manifests/noderesourcetopology/ns.yaml
+++ b/manifests/noderesourcetopology/ns.yaml
@@ -1,5 +1,0 @@
----
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: test-namespace # Test namespace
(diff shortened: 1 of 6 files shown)


fromanirh:
This is a good idea and a nice little changes that streamlines the initial experience, a nice complement to https://github.com/kubernetes-sigs/scheduler-plugins/pull/276 . In general, however, to really have a seamless experience, a higher level orchestrator agent (human, operator, controller...) should ensure the consistency of the namespace. 
Of course this is out of scope for scheduler-plugins: this just means the options to mitigate configuration issues in the context of the plugin are limited.

PR #280, opened 2021-11-01
Trimaran: bug fixes for unpopulated watcher metrics

zorro786:
This PR fixes bugs in Trimaran plugins when metrics from watcher are never populated in Trimaran watcher cache since the scheduler started. This could happen due to misconfigurations (incorrect) provider address, token etc., and if no metrics provider is running.

Also adds missing default handling for TargetLoadPackingArgs

### DIFF
--- a/pkg/apis/config/v1beta1/defaults.go
+++ b/pkg/apis/config/v1beta1/defaults.go
@@ -115,22 +115,14 @@ func SetDefaultTargetLoadPackingArgs(args *TargetLoadPackingArgs) {
 	if args.TargetUtilization == nil || *args.TargetUtilization <= 0 {
 		args.TargetUtilization = &DefaultTargetUtilizationPercent
 	}
-	metricProviderType := string(args.MetricProvider.Type)
-	validMetricProviderType := metricProviderType == string(pluginConfig.KubernetesMetricsServer) ||
-		metricProviderType == string(pluginConfig.Prometheus) ||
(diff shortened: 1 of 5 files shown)


pkg/apis/config/v1beta1/defaults.go
```diff
@@ -115,6 +115,13 @@ func SetDefaultTargetLoadPackingArgs(args *TargetLoadPackingArgs) {
 	if args.TargetUtilization == nil || *args.TargetUtilization <= 0 {
 		args.TargetUtilization = &DefaultTargetUtilizationPercent
 	}
+	metricProviderType := string(args.MetricProvider.Type)
```
Huang-Wei, line 118: Silencing the error doesn't sound like a good idea, as the user may be specifying the type with a typo.

I'd suggest assigning the type to DefaultMetricProviderType when it's empty, and error it out in terms of illegal type, in the validation logic.

You can refer to how NodeResourcesAllocatableArgs.mode is defaulted and validated.

Huang-Wei: And I also realized you're following the way how `LoadVariationRiskBalancingArgs` is handled. I's suggest updating that as well.

zorro786: Fixed

Huang-Wei:
@zorro786 some UTs failed.

zorro786:
@Huang-Wei Fixed thanks

Issue #260, opened 2021-09-17
There are no latency/network-aware scheduling plugins available,  but applications like IoT and video services need these!


jpedro1992:
**Issues**

Recent applications are demanding lower latency. Scheduling pods only focused on reducing costs are not suitable for applications where low latency plays a major role. Applications such as the **Internet of Things (IoT)** and **video services** would benefit the most from plugins where **latency** and **bandwidth** are considered in the decision process and not only resource usages (e.g., CPU and RAM). 

Several organizations face latency when using cloud services. Distance from servers is usually the primary culprit.  The best strategy is to reduce the latency between services belonging to the same application. This work is inspired by [Service Function Chaining](https://www.sciencedirect.com/science/article/pii/S1084804516301989). Also, bandwidth plays an important role since overloaded nodes would degrade performance. 

**Possible Solution**

We propose a **Network-Aware plugin** focused on delivering low latency and respecting bandwidth constraints during pod allocations. 

**Scopes**

- Provide a networkAware plugin to extend scheduling features of Kubernetes by considering latency and bandwidth.
- Consider different pods as an application to establish a service dependency graph through pod affinities.
- Map the infrastructure into zones and nodes to consider network topology information (e.g., network weights) in the scheduling process.  
- The advertising of the nodes (physical) bandwidth capacity as [extended resources](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/).
- Evaluate the risk of allocating pods (Filter) on certain nodes based on their current demand (network bandwidth).
- Near-optimal scheduling decisions (Scoring) focused on low latency based on the [Dijkstra Shortest Path Calculation](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm).

**Next Steps**

We are planning to submit a KEP in the next few weeks. 
We are adjusting our design based on the feedback received from the community on September 9th. 
Presentation recording available [here](https://youtu.be/D9jSqUiaq1Q). 

**Contact Team**

@jpedro1992
@wangchen615

PR #281, opened 2021-11-04
KEP: Network-Aware framework

jpedro1992:
According to the proposal in #260, I submit this KEP.


wangchen615:
> According to the proposal in #260, I submit this KEP.

@jpedro1992 , could you squash all your commits?

PR #283, opened 2021-11-04
bump k8s deps to 1.21.6

Huang-Wei:
I will do a release cut once it's merged.


Huang-Wei:
@seanmalloy could you /lgtm this?

Issue #271, opened 2021-10-11
Cannot list resources "ConfigMap" errors persists in scheduler-plugin logs when running as a secondary scheduler


wangchen615:
I found the following error messages kept popping up when I already have all RBAC rules defined https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml when running scheduler-plugins as a secondary scheduler.

```
E1010 15:43:05.868788       1 reflector.go:138] k8s.io/apiserver/pkg/server/dynamiccertificates/configmap_cafile_content.go:206: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "extension-apiserver-authentication" is forbidden: User "system:serviceaccount:trimaran:trimaran-scheduler" cannot list resource "configmaps" in API group "" in the namespace "kube-system"
```

It only impacts the installation with a secondary scheduler

I believe we should probably add the following ClusterRole to https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml .

```
rules:
- apiGroups:
  - ""
  resources:
  - configmaps
  verbs:
  - get
  - list
  - watch
```

PR #272, opened 2021-10-11
install: create a RoleBinding for obtain delegated authentication

Huang-Wei:
Fixes #271.

This PRs adds a RoleBinding to obtain delegated authentication which is defined in `extension-apiserver-authentication-reader/kube-system`.

PR #285, opened 2021-11-05
Update rbac.yaml to avoid "Failed to watch" error

wangchen615:
This is to resolve and close issue #271 .


denkensk:
/lgtm
Thanks @wangchen615

Huang-Wei:
@wangchen615 the root cause is the RBAC rules applied to the secondary scheduler doesn't fully "copy" the ones applied to default scheduler, in particular, access to configmap "extension-apiserver-authentication". This was fixed in the upstream by creating a RoleBinding: https://github.com/kubernetes/kubernetes/pull/72491

So I think we should follow the same way here, instead of imposing a direct configmap access. 

I raised PR #272 earlier, could you verify if it works?

wangchen615:
Ok, I will close this PR then.
> same way here, instead of imposing a direct con



> #272

PR #290, opened 2021-11-14
Introduce v1beta2 plugin args

Huang-Wei:
This PR does the following changes:

- copy v1beta1 plugin args to v1beta2 plugin args
- remove `kubeConfigPath` in all plugin args - it's not needed anymore as we're vendoring k/k 1.22.
    - NOTE: this implies old scheduler config (`CapacitySchedulingArgs`) that with _bare_ `kubeConfigPath` is totally removed. So upon migration the scheduler config.
- updated tests and cleaned up utility code
- updated the documents


Huang-Wei:
@denkensk could you help review?

manifests/noderesourcetopology/scheduler-config.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: kubescheduler.config.k8s.io/v1beta1
+apiVersion: kubescheduler.config.k8s.io/v1beta2
```
denkensk, line 1: The pluginConfig for `NodeResourceTopologyMatch' also needs to be updated to remove the kubeconfigpath.

manifests/noderesourcetopology/scheduler-configmap.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   namespace: kube-system
 data:
   scheduler-config.yaml: |
-      apiVersion: kubescheduler.config.k8s.io/v1beta1
+      apiVersion: kubescheduler.config.k8s.io/v1beta2
```
denkensk, line 8: same in this file

pkg/apis/config/v1beta2/defaults_test.go
```diff
@@ -0,0 +1,162 @@
...
+				TargetUtilization:         pointer.Int64Ptr(40),
+				MetricProvider: MetricProviderSpec{
+					Type: "KubernetesMetricsServer",
+				},
+			},
+		},
+		{
+			name: "set non default TargetLoadPackingArgs",
+			config: &TargetLoadPackingArgs{
+				DefaultRequests:           v1.ResourceList{v1.ResourceCPU: resource.MustParse("100m")},
```
denkensk, line 98: Use a different value than the default to make this test case validated？

denkensk (commented):
@Huang-Wei 
Thanks so much. Some small comments.

pkg/apis/config/v1beta2/defaults_test.go
Huang-Wei: This is non-default value, isn't it? (default value is `1000m`)

manifests/noderesourcetopology/scheduler-configmap.yaml
Huang-Wei: done.

manifests/noderesourcetopology/scheduler-config.yaml
Huang-Wei: done.

pkg/apis/config/v1beta2/defaults_test.go
denkensk: Yes, I misread it before.

denkensk:
/lgtm
/approve
Thanks  @Huang-Wei

PR #291, opened 2021-11-15
NodeResourceTopology scheduler plugin cleanup

swatisehgal:
Minor clean up to the NoderesourceTopology scheduler plugin manifests, configs, deployment specs and documentation.

Updates include:
    - removal of scheduler-name from the arguments
    - README updates to add pointer to NFD
    - test-deployment.yaml updates to make sure that it is consistent with its representation in the README

Signed-off-by: Swati Sehgal <swati.tcd@gmail.com>

### DIFF
--- a/manifests/noderesourcetopology/deploy.yaml
+++ b/manifests/noderesourcetopology/deploy.yaml
@@ -27,7 +27,6 @@ spec:
           imagePullPolicy: Never
           command:
           - /bin/kube-scheduler
-          - --scheduler-name=topo-aware-scheduler
           - --authentication-kubeconfig=/etc/kubernetes/scheduler.conf
           - --authorization-kubeconfig=/etc/kubernetes/scheduler.conf
           - --config=/etc/kubernetes/scheduler-config/scheduler-config.yaml
--- a/manifests/noderesourcetopology/test-deployment.yaml
+++ b/manifests/noderesourcetopology/test-deployment.yaml
@@ -21,11 +21,11 @@ spec:
         resources:
           limits:
             cpu: 1
-            memory: 0
+            memory: 100Mi
             example.com/deviceA: 1
             example.com/deviceB: 1
           requests:
             cpu: 1
-            memory: 0
+            memory: 100Mi
             example.com/deviceA: 1
             example.com/deviceB: 1
--- a/pkg/noderesourcetopology/README.md
+++ b/pkg/noderesourcetopology/README.md
@@ -64,6 +64,8 @@ Let us assume we have two nodes in a cluster deployed with sample-device-plugin 
 
 The hardware topology corresponding to both the nodes is represented by the below CRD instances. These CRD instances are supposed to be created by Node Agents like [Resource Topology Exporter](https://github.com/k8stopologyawareschedwg/resource-topology-exporter) (RTE) or Node feature Discovery (NFD). Please refer to issue [Exposing Hardware Topology through CRDs in NFD](https://github.com/kubernetes-sigs/node-feature-discovery/issues/333) and [Design document](https://docs.google.com/document/d/1Q-4wSu1tzmbOXyGk_2r5_mK6JdXXJA-bOd3cAtBFnwo/edit?ts=5f24171f#) which captures details of enhancing NFD to expose node resource topology through CRDs.
 
+For configuring your cluster with [NFD-topology updater](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/get-started/introduction.md#nfd-topology-updater), a software component in Node Feature Discovery which creates NodeResourceTopology CRs corresponding to nodes in the cluster follow the Quick Start guide [here]( https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/get-started/quick-start.md#additional-optional-installation-steps).
+
 ```yaml
 # Worker Node A CRD spec
 apiVersion: topology.node.k8s.io/v1alpha1
@@ -148,7 +150,7 @@ zones:
              $ kubectl get noderesourcetopologies.topology.node.k8s.io
             ```
 
-         1. Deploy the CRs representative of the hardware topology of the worker-node-A and worker-node-B:
+         1. Deploy the CRs representative of the hardware topology of the worker-node-A and worker-node-B if CRs haven't been created using RTE or NFD as mentioned above:
 
             ```bash
              $ kubectl create -f worker-node-A.yaml
@@ -157,7 +159,6 @@ zones:
             NOTE: In case you are testing this demo by creating CRs manually, ensure that the names of the nodes in the cluster match the CR names.
 
 - Copy cluster kubeconfig file to /etc/kubernetes/scheduler.conf
-
 - Build the image locally
 
     ```bash
@@ -218,8 +219,10 @@ spec:
       schedulerName: topo-aware-scheduler
       containers:
       - name: test-deployment-1-container-1
-        image: nginx:1.7.9
+        image: quay.io/fromani/numalign
         imagePullPolicy: IfNotPresent
+        command: ["/bin/sh", "-c"]
+        args: [ "while true; do numalign; sleep 100000; done;" ]
         resources:
           limits:
             cpu: 1


manifests/noderesourcetopology/cluster-role.yaml
```diff
@@ -14,10 +14,10 @@ rules:
   verbs: ["*"]
 - apiGroups: [""]
   resources: ["nodes"]
-  verbs: ["get", "list", "patch"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: [""]
   resources: ["pods"]
-  verbs: ["get","list","watch","update"]
+  verbs: ["get", "list", "watch", "update", "delete"]
```
fromanirh, line 20: do we really need to delete pods?

manifests/noderesourcetopology/cluster-role.yaml
```diff
@@ -32,7 +32,7 @@ subjects:
     namespace: kube-system
 roleRef:
   kind: ClusterRole
-  name: noderesourcetoplogy-handler
+  name: system:kube-scheduler
```
fromanirh, line 35: I think I know from where this comes from, we experimented with secondary-scheduler-operator which is doing something similar here. Not sure this (= these manifests) is the right place for this change, though

manifests/noderesourcetopology/cluster-role.yaml
swatisehgal: I borrowed it from the RBAC permissions that I saw in the repo here: https://github.com/kubernetes-sigs/scheduler-plugins/blob/6e58632f57f80274dc5e0f97df7f1d1907cc7c25/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml#L28 but can remove it.

manifests/noderesourcetopology/cluster-role.yaml
swatisehgal: The secondary-scheduler-operator is a way of automating the deployment of scheduler plugins but here we want to provide all the artifacts that the user needs to play around with, configure and deploy this scheduler plugin in their cluster. We provide manifests and the steps we have provided in the README to support that. Without these changes, the scheduler plugin is not going to work as expected hence its important to update these manifests to allow the user to deploy NRT scheduler plugin successfully as a secondary scheduler and to deploy a test pod/deployment with this scheduler.

manifests/noderesourcetopology/cluster-role.yaml
swatisehgal: Removed the `delete` verb.

manifests/noderesourcetopology/deploy.yaml
```diff
@@ -27,25 +27,17 @@ spec:
           imagePullPolicy: Never
           command:
           - /bin/kube-scheduler
-          - --scheduler-name=topo-aware-scheduler
-          - --authentication-kubeconfig=/etc/kubernetes/scheduler.conf
-          - --authorization-kubeconfig=/etc/kubernetes/scheduler.conf
```
Huang-Wei, line 32: I think these two args are still needed ([doc](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-scheduler/)). They're neither deprecated nor derive the settings from scheduler CC:

```
kubeconfig  file pointing at the 'core' kubernetes server with enough rights to  create tokenreviews.authentication.k8s.io. This is optional. If empty,  all token requests are considered to be anonymous and no client CA is  looked up in the cluster.
```

manifests/noderesourcetopology/scheduler-config.yaml
```diff
@@ -2,9 +2,6 @@ apiVersion: kubescheduler.config.k8s.io/v1beta1
 kind: KubeSchedulerConfiguration
 leaderElection:
   leaderElect: false
-clientConnection:
-  # kubeconfig: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
-  kubeconfig: "/etc/kubernetes/scheduler.conf"
```
Huang-Wei, line 7: Nope, this is still needed.

manifests/noderesourcetopology/scheduler-config.yaml
```diff
@@ -18,8 +15,6 @@ profiles:
   pluginConfig:
   - name: NodeResourceTopologyMatch
     args:
-      # kubeconfig: "REPLACE_ME_WITH_KUBE_CONFIG_PATH"
-      kubeconfigpath: "/etc/kubernetes/scheduler.conf"
```
Huang-Wei, line 22: I have one PR opened tracking this already: #290.

Huang-Wei (commented):
Some changes overlapped with #290. Let's wait for #290 to be merged first and rebase here.

manifests/noderesourcetopology/deploy.yaml
damemi: The doc says it's optional, so those settings must be accessible from somewhere else too right? Are they available if you just use the `--kubeconfig` flag?

swatisehgal: But in the documentation I see using `--kubeconfig` flag is deprecated?
```
--kubeconfig string
DEPRECATED: path to kubeconfig file with authorization and master location information. This parameter is ignored if a config file is specified in --config.
```
and we are have worked to avoid accessing the kubeconfig in the favour of obtaining it from the scheduler framework handler.  I am a little confused here to be honest.

manifests/noderesourcetopology/scheduler-config.yaml
swatisehgal: Ack, Thanks!

manifests/noderesourcetopology/scheduler-config.yaml
swatisehgal: Hmm, okay. I didn't realize that. Thanks for letting me know.

manifests/noderesourcetopology/deploy.yaml
damemi: > we still need to mount in the kubeconfig into the scheduler plugin deployment? I thought that kubernetes/kubernetes#100644 was proposed to overcome the need to specify the kubeconfig path in the plugin arguments?

kubernetes/kubernetes#100644 does remove the need to specify the kubeconfig path in the plugin arguments. These flags aren't plugin arguments, they're core to the scheduler itself

Huang-Wei: > @Huang-Wei If we still need `--authentication-kubeconfig` and `--authorization-kubeconfig` then it means we still need to mount in the kubeconfig into the scheduler plugin deployment?

That's correct, just think of scheduler-plugins as a variant of the default scheduler. You need to at least provide a kubeConfig :-)

Huang-Wei: What kubernetes/kubernetes#100644 tries to solve is in the Go SDK's perspective, the out-of-tree plugins don't need to specify the kubeConfig again.

swatisehgal: Okay, so why don't we have it specified here: https://github.com/kubernetes-sigs/scheduler-plugins/blob/6e58632f57f80274dc5e0f97df7f1d1907cc7c25/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml#L44-L48?

swatisehgal: Also, we experimented with deployment of scheduler plugin without these flags and it worked fine and did not complain. I wonder why it is working for us if it shouldn't :thinking:

Huang-Wei: > Okay, so why don't we have it specified here?

That's a mistake, we should specify it in that deployment as well.

> Also, we experimented with deployment of scheduler plugin without these flags and it worked fine and did not complain.

As the doc says, it's optional. In rare cases, we can use another kubeconfig to serve the token requests. The hack script (hack/local-up-cluster.sh) in k/k also specifies it in controller-manager and kube-scheduler's startup.

Huang-Wei:
@swatisehgal you can rebase now.

swatisehgal:
@Huang-Wei Apologies, this PR somehow got neglected. It has been rebased and updated. It is now much smaller thanks to all the feedback. PTAL when you have some time. Thanks in advance!

manifests/noderesourcetopology/cluster-role.yaml
swatisehgal: This has been removed from the PR now.

Huang-Wei:
@swatisehgal it looks neat!

/lgtm
/approve

PR #292, opened 2021-11-17
NodeResourceTopology: Bump NRT API to version v0.0.12

swatisehgal:
The NodeResourceTopology API has been made cluster
scoped as in the current context a CR corresponds to
a Node and since Node is a cluster scoped resource it
makes sense to make NRT cluster scoped as well.

Ref: https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/18

### DIFF
--- a/manifests/noderesourcetopology/crd.yaml
+++ b/manifests/noderesourcetopology/crd.yaml
@@ -17,7 +17,7 @@ spec:
     shortNames:
     - node-res-topo
     singular: noderesourcetopology
-  scope: Namespaced
+  scope: Cluster
   versions:
   - name: v1alpha1
     schema:
--- a/pkg/noderesourcetopology/plugin.go
+++ b/pkg/noderesourcetopology/plugin.go
@@ -36,11 +36,6 @@ type NUMANode struct {
 
 type NUMANodeList []NUMANode
 
-type nodeResTopologyPlugin struct {
-	lister     *listerv1alpha1.NodeResourceTopologyLister
-	namespaces []string
-}
-
 type tmScopeHandler struct {
 	filter func(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nodeInfo *framework.NodeInfo) *framework.Status
 	score  func(pod *v1.Pod, zones topologyv1alpha1.ZoneList, scorerFn scoreStrategy, resourceToWeightMap resourceToWeightMap) (int64, *framework.Status)
@@ -64,7 +59,7 @@ type PolicyHandlerMap map[topologyv1alpha1.TopologyManagerPolicy]tmScopeHandler
 
 // TopologyMatch plugin which run simplified version of TopologyManager's admit handler
 type TopologyMatch struct {
-	nodeResTopologyPlugin
+	lister              listerv1alpha1.NodeResourceTopologyLister
 	policyHandlers      PolicyHandlerMap
 	scorerFn            scoreStrategy
 	resourceToWeightMap resourceToWeightMap
@@ -107,10 +102,7 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	}
 
 	topologyMatch := &TopologyMatch{
-		nodeResTopologyPlugin: nodeResTopologyPlugin{
-			lister:     lister,
-			namespaces: tcfg.Namespaces,
-		},
+		lister:              lister,
 		policyHandlers:      newPolicyHandlerMap(),
 		scorerFn:            scoringFunction,
 		resourceToWeightMap: resToWeightMap,
--- a/pkg/noderesourcetopology/pluginhelpers.go
+++ b/pkg/noderesourcetopology/pluginhelpers.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"fmt"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
@@ -31,25 +31,17 @@ import (
 	listerv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/listers/topology/v1alpha1"
 )
 
-func findNodeTopology(nodeName string, nodeResTopoPlugin *nodeResTopologyPlugin) *topologyv1alpha1.NodeResourceTopology {
-	klog.V(5).InfoS("Namespaces for nodeResTopoPlugin", "namespaces", nodeResTopoPlugin.namespaces)
-	for _, namespace := range nodeResTopoPlugin.namespaces {
-		klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", nodeResTopoPlugin.lister)
-		// NodeTopology couldn't be placed in several namespaces simultaneously
-		lister := nodeResTopoPlugin.lister
-		nodeTopology, err := (*lister).NodeResourceTopologies(namespace).Get(nodeName)
-		if err != nil {
-			klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyNamespaceLister")
-			continue
-		}
-		if nodeTopology != nil {
-			return nodeTopology
-		}
+func findNodeTopology(nodeName string, lister listerv1alpha1.NodeResourceTopologyLister) *topologyv1alpha1.NodeResourceTopology {
+	klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", lister)
+	nodeTopology, err := lister.Get(nodeName)
+	if err != nil {
+		klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyLister")
+		return nil
 	}
-	return nil
+	return nodeTopology
 }
 
-func initNodeTopologyInformer(kubeConfig *restclient.Config) (*listerv1alpha1.NodeResourceTopologyLister, error) {
+func initNodeTopologyInformer(kubeConfig *restclient.Config) (listerv1alpha1.NodeResourceTopologyLister, error) {
 	topoClient, err := topoclientset.NewForConfig(kubeConfig)
 	if err != nil {
 		klog.ErrorS(err, "Cannot create clientset for NodeTopologyResource", "kubeConfig", kubeConfig)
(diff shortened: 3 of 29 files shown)


pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -32,19 +32,15 @@ import (
...
-			return nodeTopology
-		}
+	klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", nodeResTopoPlugin.lister)
+	// NodeTopology couldn't be placed in several namespaces simultaneously
+	lister := nodeResTopoPlugin.lister
+	nodeTopology, err := (*lister).Get(nodeName)
+	if err != nil {
+		klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyLister")
+	}
+	if nodeTopology != nil {
```
fromanirh, line 42: or just
```
return nodeTopology
```
:)

swatisehgal: Ah yeah! Will fix now

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -32,21 +32,14 @@ import (
...
-		}
-		if nodeTopology != nil {
-			return nodeTopology
-		}
+	klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", nodeResTopoPlugin.lister)
+	// NodeTopology couldn't be placed in several namespaces simultaneously
+	lister := nodeResTopoPlugin.lister
+	nodeTopology, err := (*lister).Get(nodeName)
+	if err != nil {
+		klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyLister")
```
Huang-Wei, line 40: Could you check if it's possible to return a non-nil `nodeTopology` along with a non-nil `err`? If so, we should explicitly `return nil` below this line.

manifests/noderesourcetopology/crd.yaml
```diff
@@ -4,7 +4,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870 # edited manually
```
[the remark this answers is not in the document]
swatisehgal: Ack, done!

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -32,21 +32,14 @@ import (
...
-			klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyNamespaceLister")
-			continue
-		}
-		if nodeTopology != nil {
-			return nodeTopology
-		}
+	klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", nodeResTopoPlugin.lister)
+	// NodeTopology couldn't be placed in several namespaces simultaneously
+	lister := nodeResTopoPlugin.lister
+	nodeTopology, err := (*lister).Get(nodeName)
```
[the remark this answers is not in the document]
swatisehgal: Fixed

pkg/noderesourcetopology/pluginhelpers.go
swatisehgal: No there is no case where a non-nil nodeTopology is returned with a non-nil err. I have updated as per your recommendation.

swatisehgal:
Going to rebase.

pkg/noderesourcetopology/plugin.go
```diff
@@ -64,7 +59,7 @@ type PolicyHandlerMap map[topologyv1alpha1.TopologyManagerPolicy]tmScopeHandler
 
 // TopologyMatch plugin which run simplified version of TopologyManager's admit handler
 type TopologyMatch struct {
-	nodeResTopologyPlugin
+	lister              *listerv1alpha1.NodeResourceTopologyLister
```
fromanirh, line 62: I think this, being an interface, should be passed around as value, not as pointer

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -31,22 +31,14 @@ import (
...
-		lister := nodeResTopoPlugin.lister
-		nodeTopology, err := (*lister).NodeResourceTopologies(namespace).Get(nodeName)
-		if err != nil {
-			klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyNamespaceLister")
-			continue
-		}
-		if nodeTopology != nil {
-			return nodeTopology
-		}
+func findNodeTopology(nodeName string, lister listerv1alpha1.NodeResourceTopologyLister) *topologyv1alpha1.NodeResourceTopology {
```
fromanirh, line 34: this function is now so simple that I'd inline in the calling site, if IIRC is called in just one place

fromanirh (commented):
noted some possible improvements

pkg/noderesourcetopology/plugin.go
swatisehgal: Ack

pkg/noderesourcetopology/pluginhelpers.go
swatisehgal: Ack

swatisehgal: Took a look at the code. It is being used both in the `Filter()` function and the `Score()` function so maybe it is still better to keep it as a function rather than inlining.

AlexeyPerevalov:
LGTM

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: yup, if this is the case let's not inline it.

fromanirh (commented):
LGTM

swatisehgal:
Thanks @AlexeyPerevalov and @fromanirh!

/assign @Huang-Wei 

for approval

Huang-Wei:
@swatisehgal could you keep the v1beta1 config spec unchanged?

swatisehgal:
> @swatisehgal could you keep the v1beta1 config spec unchanged? LGTM otherwise.

I have added v1beta1 scheduler config specs. 
Please note that both versions (v1beta1 and v1beta2) should no longer support `namespace` field as the CRD is now cluster scoped.

Huang-Wei:
> Please note that both versions (v1beta1 and v1beta2) should no longer support `namespace` field as the CRD is now cluster scoped.

We should make v1beta1 backward-compatible. If the user is currently using a v1beta1 config against 1.22, namespaces' absence would fail the scheduler'startup; while if we keep v1beta1 as-is, the old v1beta1 config would still work - internal it's converted to the latest v1beta2 config.

fromanirh:
> > Please note that both versions (v1beta1 and v1beta2) should no longer support `namespace` field as the CRD is now cluster scoped.
> 
> We should make v1beta1 backward-compatible. If the user is currently using a v1beta1 config against 1.22, namespaces' absence would fail the scheduler'startup; while if we keep v1beta1 as-is, the old v1beta1 config would still work - internal it's converted to the latest v1beta2 config.

In this case we are dropping information (v1beta2-compliant plugin just doesn't care anymore about the `namespaces`) so I guess it should be sufficient to emit a line in the logs informing users that the `namespaces` value, if given, would be ignored, right?

I can't think of any better way to handle this case in this circumstance.

swatisehgal:
> > Please note that both versions (v1beta1 and v1beta2) should no longer support `namespace` field as the CRD is now cluster scoped.
> 
> We should make v1beta1 backward-compatible. If the user is currently using a v1beta1 config against 1.22, namespaces' absence would fail the scheduler'startup; while if we keep v1beta1 as-is, the old v1beta1 config would still work - internal it's converted to the latest v1beta2 config.

In v1beta1, `namespace` was always an optional field:
```
Namespaces []string `json:"namespaces,omitempty"`
```
Keeping it in the spec when it later gets ignored is not ideal in my opinion. 

The concern about someone using an older version of v1beta1 scheduler config (with `namespace` field) resulting in error is totally valid but can be easily tackled with documentation. Given that this is a relatively new scheduler plugin (and feature) and pretty much all the stakeholders/consumers of this feature involved are in the loop we won't be causing too much damage. Having said that, I acknowledge that this is a risk and the approach suggested by @fromanirh is definitely a less risky way of handling it.

@Huang-Wei if you feel strongly feel about keeping the `namespace` field in the v1beta1 config, I will revert the v1beta1 changes and add a log message indicating that `namespace` field will be ignored. Alternatively, on top of the current changes we can add additional information in the README that the `namespace` field has been deprecated from the v1beta1 config.

Huang-Wei:
> In v1beta1, namespace was always an optional field:

Optional tag doesn't make a difference. We're talking about whether the v1beta1 API is still compatible.

I re-read the API design principles here: [api_changes.md#incompatible-api-changes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#incompatible-api-changes)

> Any API call (e.g. a structure POSTed to a REST endpoint) that succeeded before your change must succeed after your change.
> Any API call that does not use your change must behave the same as it did before your change.

I believe the current change violates these rules.

However, given the 1.22 release vendors the cluster-scoped CRD, and uses v1beta2 as the default internal version, it implies the v1beta1 config is already incompatible - the behavior with []namespaces are already removed. So we have to admit that the v1beta1 is incompatible due to the incompatibility of the dependent CRD. Based on that, I'm ok with the current change.

/approve
/lgtm

/hold
(in case you want to do extra efforts to support both namespace and cluster-scoped CRD in 1.22)

swatisehgal:
Thanks @Huang-Wei. Given that we don't have a strong reason or use case that requires us to support both namespace and cluster scoped CRD and taking into consideration the current scope and use case at hand, I feel comfortable proceeding without the namespace support.  @fromanirh @AlexeyPerevalov Do you both agree?  

I will unhold the PR once I get an ACK to proceed.

fromanirh:
it's unfortunate that we cannot have any form of backward compatibility, but indeed the root cause is the CRD change, which I don't know how can be made compatble - the namespace scoped -> cluster scoped seems to be an unvodiable breaking.

Because of this I see no real option to improve the compatibility side besides documenting, perhaps with a followup doc PR which makes explicit which version of the plugin is compatible to which versions of the CRD.
On the bright side, due to the nature of this plugin, the damage is very limited.

For all these reasons here's my ACK.

AlexeyPerevalov:
I give my ACK too
NRT was used in product, but the old version of NRT, so during the process of updating it's not a problem to not use a namespaces.

PR #299, opened 2021-11-18
feat: activate sibling pods of a PodGroup

Huang-Wei:
This PR leverage the upstream feature https://github.com/kubernetes/kubernetes/pull/103383 to proactively move pods of a PodGroup back to internal scheduling activeQ, and thus boost the overall co-scheduling efficiency a lot <sup>1</sup>.

Here is the test before and after this PR:

Before:

```
master ⇒  go test -timeout=40m ./test/integration/... -run TestCoschedulingPlugin
ok  	sigs.k8s.io/scheduler-plugins/test/integration	727.235s
```

After:

```
activate-podgroup-pods ⇒  go test -timeout=40m ./test/integration/... -run TestCoschedulingPlugin
ok  	sigs.k8s.io/scheduler-plugins/test/integration	23.949s
```

[1]: The above test improves the integration test by 30X. According to the observation of @mateuszlitwin on the OpenAI production workloads, it can increase the efficiency by 10X~100X.


Huang-Wei:
/cc @mateuszlitwin

Could you help review the changes?

pkg/coscheduling/coscheduling.go
```diff
@@ -228,6 +228,21 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 		waitTime = 0
 	}
 
+	// We will also request to move the sibling pods back to activeQ.
```
mateuszlitwin, line 231: I think it is fine.

One might consider doing activation only when numWaitingPods < minAvailable - if group is already permitted maybe there is no need to activate remaining pods? I do not have strong opinions here, but for simplicity I think it is better to always activate pods.

pkg/coscheduling/coscheduling.go
```diff
@@ -228,6 +228,21 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
...
+	if podsToMove := cs.pgMgr.SiblingPods(pod); len(podsToMove) != 0 {
+		c, err := state.Read(framework.PodsToActivateKey)
+		if err == nil {
+			if s, ok := c.(*framework.PodsToActivate); ok {
+				s.Lock()
+				for _, pod := range podsToMove {
+					namespacedName := core.GetNamespacedName(pod)
+					s.Map[namespacedName] = pod
```
mateuszlitwin, line 239: Should we ignore pods that are already waiting?

mateuszlitwin (approved):

pkg/coscheduling/coscheduling.go
Huang-Wei: > One might consider doing activation only when numWaitingPods < minAvailable - if group is already permitted maybe there is no need to activate remaining pods?

Are you suggesting move the logic to `case core.Wait` only?

pkg/coscheduling/coscheduling.go
Huang-Wei: We can, but IMO it doesn't pay much benefits back. Activate a waiting pod is a no-op, while filtering waiting pods have to iterate the whole waitingPodsMap.

Huang-Wei:
Off the topic, in my local testing env, the `podLister` sometimes doesn't return the created pods accurately:

```
01 I1119 13:17:44.265418   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p1-1"
02 I1119 13:17:44.269206   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p2-1"
03 E1119 13:17:44.269298   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p1-1 cannot find enough sibling pods, current pods number: 1, minMember of group: 3" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p1-1"
04 I1119 13:17:44.272646   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p3-1"
05 E1119 13:17:44.272901   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p2-1 cannot find enough sibling pods, current pods number: 0, minMember of group: 4" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p2-1"
06 I1119 13:17:44.276089   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p1-2"
07 E1119 13:17:44.276860   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p3-1 cannot find enough sibling pods, current pods number: 0, minMember of group: 4" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p3-1"
08 E1119 13:17:44.280974   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p1-2 cannot find enough sibling pods, current pods number: 1, minMember of group: 3" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p1-2"
09 I1119 13:17:44.281315   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p2-2"
10 I1119 13:17:44.285119   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p3-2"
11 E1119 13:17:44.285295   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p2-2 cannot find enough sibling pods, current pods number: 2, minMember of group: 4" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p2-2"
12 I1119 13:17:44.289412   46136 coscheduling_test.go:384] "Creating pod " podName="t7-p1-3"
13 E1119 13:17:44.289619   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p3-2 cannot find enough sibling pods, current pods number: 2, minMember of group: 4" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p3-2"
14 E1119 13:17:44.294081   46136 coscheduling.go:143] "PreFilter failed" err="pre-filter pod t7-p1-3 cannot find enough sibling pods, current pods number: 2, minMember of group: 3" pod="integration-test-6379e655-8c12-46d8-a5b1-61f5d72619a9/t7-p1-3"
```

In the above log:

- at L05, existing pods num should be 1 instead of 0
- at L07, existing pods num should be 1 instead of 0
- at L08, existing pods num should be 2 instead of 1
- at L14, existing pods num should be 3 instead of 2

Not sure you noticed similar symptoms in your production workloads. It may cause unexpected preemption and unneeded scheduling cycles.

pkg/coscheduling/coscheduling.go
mateuszlitwin: Possibly, I think it kind of depends how important is to schedule pods above the minAvailable limit.

pkg/coscheduling/coscheduling.go
```diff
@@ -228,6 +228,21 @@ func (cs *Coscheduling) Permit(ctx context.Context, state *framework.CycleState,
 		waitTime = 0
 	}
 
+	// We will also request to move the sibling pods back to activeQ.
+	if podsToMove := cs.pgMgr.SiblingPods(pod); len(podsToMove) != 0 {
```
mateuszlitwin, line 232: Another suggestion: Maybe this whole functionality should be in `pgMgr.ActivateSiblings(state)` and inline `SiblingPods` inside that function?

mateuszlitwin:
Weird, maybe it is because you are using podLister from different SharedInformerFactory() than scheduler? Any reason we do not use SharedInformerFactory() provided by the framework?

pkg/coscheduling/coscheduling.go
Huang-Wei: Yup, that makes sense to me.

Huang-Wei:
Bingo, that's it. I do remember I raised the concern in the original review of coscheduling PR, but somehow it slipped.

cc/ @cwdsuzhou @denkensk I believe this is a bug and should be back-ported.

pkg/coscheduling/coscheduling.go
denkensk: In this case, we shouldn't need to exclude the pod itself from SiblingPods either, right?
```go
	for i := range pods {
		if pods[i].UID == pod.UID {
			return append(pods[:i], pods[i+1:]...)
		}
	}
```

pkg/coscheduling/core/core.go
```diff
@@ -105,6 +107,26 @@ func NewPodGroupManager(pgClient pgclientset.Interface, snapshotSharedLister fra
 	return pgMgr
 }
 
+// SiblingPods return pods that belong to the same PodGroup of the given pod.
```
denkensk, line 110:
```suggestion
// SiblingPods returns pods that belong to the same PodGroup of the given pod.
```

Huang-Wei:
Comments addressed. PTAL.

denkensk:
/lgtm
@Huang-Wei  Thanks.

pkg/coscheduling/core/core.go
```diff
@@ -105,6 +106,42 @@ func NewPodGroupManager(pgClient pgclientset.Interface, snapshotSharedLister fra
...
+		}
+	}
+
+	if len(pods) != 0 {
+		if c, err := state.Read(framework.PodsToActivateKey); err == nil {
+			if s, ok := c.(*framework.PodsToActivate); ok {
+				s.Lock()
+				for _, pod := range pods {
+					namespacedName := GetNamespacedName(pod)
+					s.Map[namespacedName] = pod
```
cmssczy, line 137: do we need a lock when we write to map?

PR #302, opened 2021-11-20
avoid instantiate core resource's informers

### LINKED ISSUE #301: racing conditions upon multiple pod informers
It's an anti-pattern to have more than one pod informers as each informer maintains its own cache, which can cause racing conditions. We should have one single source of truth across the code base.

Here is one [symptom](https://github.com/kubernetes-sigs/scheduler-plugins/pull/299#issuecomment-974484806) I observed in the coscheduling integration test.

We should prevent instantiating core resource informer in our plugins; instead, reuse the scheduler frameworkHandle's SharedInformerFactory.

/kind bug

Huang-Wei:
Core resource's (like Pod/Node) informers should be obtained from frameworkHandle's SharedInformerFactory.

Fixes #301.

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -19,15 +19,11 @@ package coscheduling
 import (
 	"context"
 	"fmt"
-	"os"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/informers"
 	"k8s.io/client-go/tools/cache"
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
@@ -70,17 +66,7 @@ func New(obj runtime.Object, handle framework.Handle) (framework.Plugin, error) 
 	pgClient := pgclientset.NewForConfigOrDie(handle.KubeConfig())
 	pgInformerFactory := pgformers.NewSharedInformerFactory(pgClient, 0)
 	pgInformer := pgInformerFactory.Scheduling().V1alpha1().PodGroups()
-
-	fieldSelector, err := fields.ParseSelector(",status.phase!=" + string(v1.PodSucceeded) + ",status.phase!=" + string(v1.PodFailed))
-	if err != nil {
-		klog.ErrorS(err, "ParseSelector failed")
-		os.Exit(1)
-	}
-	informerFactory := informers.NewSharedInformerFactoryWithOptions(handle.ClientSet(), 0, informers.WithTweakListOptions(func(opt *metav1.ListOptions) {
-		opt.LabelSelector = util.PodGroupLabel
-		opt.FieldSelector = fieldSelector.String()
-	}))
-	podInformer := informerFactory.Core().V1().Pods()
+	podInformer := handle.SharedInformerFactory().Core().V1().Pods()
 
 	scheduleTimeDuration := time.Duration(args.PermitWaitingTimeSeconds) * time.Second
 	deniedPGExpirationTime := time.Duration(args.DeniedPGExpirationTimeSeconds) * time.Second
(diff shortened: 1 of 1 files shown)


pkg/coscheduling/coscheduling.go
```diff
@@ -70,17 +66,7 @@ func New(obj runtime.Object, handle framework.Handle) (framework.Plugin, error)
 	pgClient := pgclientset.NewForConfigOrDie(handle.KubeConfig())
 	pgInformerFactory := pgformers.NewSharedInformerFactory(pgClient, 0)
 	pgInformer := pgInformerFactory.Scheduling().V1alpha1().PodGroups()
-
-	fieldSelector, err := fields.ParseSelector(",status.phase!=" + string(v1.PodSucceeded) + ",status.phase!=" + string(v1.PodFailed))
```
lianghao208, line 74: We don't need to filter out the pods in succeeded and failed phase any more?

Huang-Wei: The informer carried over from the scheduler framework has already done this.

mateuszlitwin: Correct me if I am wrong: if we filter out succeeded and failed pods then we will potentially not allow scheduling of pods that are still pending in case some pods succeeded or failed earlier because number of pods will be smaller than minAvailable.

Huang-Wei: This is why we're relying on the "snaphost" instead of podLister to calculate the number of "available" pods:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/417f2f6ea22035e72e8afb095b9af5d236427137/pkg/coscheduling/core/core.go#L267-L269

(Note that assigned and assumed pods can be obtained from the "snapshot").

Huang-Wei: For now, coscheduling obtained the assigned + assumed pods from scheduler's snapshot. In this case, succeeded & failed pods will be both excluded. As you mentioned, this might be problematic, in particular with excluded succeeded pods.

According to your exp., how practical this problem may exist in a real-world env?

denkensk: I think it depends entirely on how the operator/controller handles failed pods. It may recreate pods after some pods fail. If we don't exclude failed or succeeded pods. These new pods will always meet the minimum number limit in the co-scheduling. But there may not be enough resources in the cluster to allow all newly created pods to be scheduled. So we need to exclude the failed or successful pods as well.

mateuszlitwin: I think we should include succeeded and failed pods when comparing against `.spec.minAvailable` in Permit. 

Imagine scenario with a job where every pod needs to succeed. If one of the pods is preempted and other pods from the same group succeeded, recreated pod will never be scheduled, but it should pass `minAvailable` check in Permit. Only solution for this would be some kind of custom control mechanism updating pod group resource which is pretty complicated from user perspective.

> But there may not be enough resources in the cluster to allow all newly created pods to be scheduled. So we need to exclude the failed or successful pods as well.

I do not understand this comment. Are you talking whether to include or exclude finished pods in PreFilter check regarding resources available in the cluster?

What exactly are the benefits of ignoring succeed/failed pods?

mateuszlitwin: Anyways, we should not be blocking the review on this, we can move this discussion to an issue.

mateuszlitwin (approved):

pkg/coscheduling/coscheduling.go
Huang-Wei: I'd like to pursue embedding a non-filtered indexInformer in the framework handle, and this has to be implemented in the upstream.

Huang-Wei: It seems in sharedInformerFactory, you cannot associate two informers for one API resource:

https://github.com/kubernetes/kubernetes/blob/9a75e7b0fd1b567f774a3373be640e19b33e7ef1/staging/src/k8s.io/client-go/informers/factory.go#L179-L180

denkensk: >  If one of the pods is preempted and other pods from the same group succeeded, recreated pod will never be scheduled, but it should pass minAvailable check in Permit.

Please correct me if I have misrepresented. @mateuszlitwin  @Huang-Wei 

Does this scenario mean that the job can still run even if some workers are preempted or fail?  Or can a newly created Pod continue to get the job done without having to cooperate with other Pods(other succeed pods)?

I think we should be clear about the concepts expressed in the minAvailable. 
1. the minimum runnable instance that satisfies the job run.
2. or succeed + failed + running？

denkensk: > Anyways, we should not be blocking the review on this, we can move this discussion to an issue.
+1

denkensk:
/lgtm
Thanks @Huang-Wei

pkg/coscheduling/coscheduling.go
Huang-Wei: Let me give an example. A Job requires 3 replicas to run altogether. And `Pod1`, `Pod2`, `Pod3` all get scheduled and running. Now let's say `Pod1` and `Pod2` succeeded, but `Pod3` failed. Then the controller spins up a replacement pod `Pod3'`. In this case, `Pod3'` wouldn't be scheduled as `Pod1` and `Pod3` have been excluded from the scheduler snapshot as well as the "filtered" podinformer.

@mateuszlitwin I suppose this is your concern, right?

mateuszlitwin: Yes. If Pod3 is deleted (e.g. Pod3 was preempted due to node going down or was lower priority than other pod) and then recreated it will not schedule.

PR #304, opened 2021-11-24
NodeResourceTopology: Update README with NRT CR dependency information

swatisehgal:
Capture the dependency information and explain that for Scheduler
plugin version > v0.21.6 depends on NRT CRD v0.0.12 and the
namespace field has been deprecated.


swatisehgal:
README update as a follow up from discussion in https://github.com/kubernetes-sigs/scheduler-plugins/pull/292

/cc @Huang-Wei @fromanirh @AlexeyPerevalov

PR #307, opened 2021-12-03
coscheduling: PostFilter handles both Filter and PreFilter

mateuszlitwin:
Currently, coscheduling.PostFilter rejects waiting pods only if coscheduling.PreFilter did not fail.

I think pod group pods should be rejected regardless. For example, if `minResources` check failed in PreFilter we should reject waiting pods in PostFilter.

Thread on slack: https://kubernetes.slack.com/archives/C09TP78DV/p1638478990123200

CC @denkensk @cwdsuzhou @Huang-Wei


Huang-Wei (commented):
Given the logic of proactively activate a group of pods is in place, I'm fine with this aggressive strategy to reject a group of pods when default preemption doesn't help.

/approve
/hold

Leaving /lgtm to @denkensk and @cwdsuzhou .

mateuszlitwin:
Yeah, with pod activation being in place I think we can be more aggressive with pod group rejections. I also think it will make plugin easier to understand and more predictable.

denkensk:
/lgtm
/unhold
Thanks @mateuszlitwin

PR #308, opened 2021-12-07
Fix controller reconcile PodGroup status

### LINKED ISSUE #166: controller should reconcile the status of PodGroup

lianghao208:
Fix #166 

It will be 3 final state of pod group: Failed, Finished, Pending.

Pending means: 
1. The pod group has been accepted by the system, but scheduler can not allocate enough resources to it.
2. The PodGroup doesn't have any associated Pods or associated Pods scaled down to 0.

Failed means: 
at least one of `spec.minMember` pods is failed.

Finished means:
all of `spec.minMember` pods are successfully.

/cc @Huang-Wei @cwdsuzhou @denkensk

### DIFF
--- a/pkg/controller/podgroup.go
+++ b/pkg/controller/podgroup.go
@@ -243,6 +243,11 @@ func (ctrl *PodGroupController) syncHandler(key string) error {
 		pgCopy.Status.Succeeded = succeeded
 		pgCopy.Status.Running = running
 
+		if len(pods) == 0 {
+			pgCopy.Status.Phase = schedv1alpha1.PodGroupPending
+			break
+		}
+
 		if pgCopy.Status.Scheduled >= pgCopy.Spec.MinMember && pgCopy.Status.Phase == schedv1alpha1.PodGroupScheduling {
 			pgCopy.Status.Phase = schedv1alpha1.PodGroupScheduled
 		}
--- a/pkg/controller/podgroup_test.go
+++ b/pkg/controller/podgroup_test.go
@@ -110,7 +110,7 @@ func Test_Run(t *testing.T) {
 			podNextPhase:      v1.PodSucceeded,
 		},
 		{
-			name:               "Group group should not enqueue, created too long",
+			name:               "Group should not enqueue, created too long",
 			pgName:             "pg8",
 			minMember:          2,
 			podNames:           []string{"pod1", "pod2"},
@@ -120,7 +120,7 @@ func Test_Run(t *testing.T) {
 			podGroupCreateTime: &createTime,
 		},
 		{
-			name:              "Group group min member more than Pod number",
+			name:              "Group min member more than Pod number",
 			pgName:            "pg9",
 			minMember:         3,
 			podNames:          []string{"pod91", "pod92"},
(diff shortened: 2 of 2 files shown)


pkg/controller/podgroup.go
```diff
@@ -238,6 +238,9 @@ func (ctrl *PodGroupController) syncHandler(key string) error {
 					failed++
 				}
 			}
+		} else {
+			pgCopy.Status.Phase = schedv1alpha1.PodGroupPending
+			break
 		}
 		pgCopy.Status.Failed = failed
```
denkensk, line 245: If the number of pods of the group is scaled to 0, we also need to update this information.
```go
pgCopy.Status.Failed = failed
pgCopy.Status.Succeeded = succeeded
pgCopy.Status.Running = running
```

pkg/controller/podgroup_test.go
```diff
@@ -144,6 +153,13 @@ func Test_Run(t *testing.T) {
 
 			pgInformerFactory.Start(ctx.Done())
 			informerFactory.Start(ctx.Done())
+			// pod number scales down to 0
```
denkensk, line 156: We don't need to add this delete operation here. In this case, we can set 
```go
podNames: []string{}
```
Add this in line 143 or line 144
```go
if len(c.podsNames) != 0 { ... }
```

pkg/controller/podgroup_test.go
```diff
@@ -128,6 +128,15 @@ func Test_Run(t *testing.T) {
 			previousPhase:     v1alpha1.PodGroupPending,
 			desiredGroupPhase: v1alpha1.PodGroupPreScheduling,
 		},
+		{
+			name:              "Pod number scale down to 0",
```
denkensk, line 132: Group status convert from running to pending ?

denkensk (commented):
Thank you for your help. Added some comments.
@lianghao208

pkg/controller/podgroup_test.go
lianghao208: Yes, if the associated Pods is always 0, the PodGroup will stay in Pending phase:
```go
	switch pgCopy.Status.Phase {
	case "":
		pgCopy.Status.Phase = schedv1alpha1.PodGroupPending
	case schedv1alpha1.PodGroupPending:
		if len(pods) >= int(pg.Spec.MinMember) {
			pgCopy.Status.Phase = schedv1alpha1.PodGroupPreScheduling
			fillOccupiedObj(pg, pods[0])
		}
	default:
...
```

I think the only situation that may cause problem is when associated Pods scale down to 0, which means PodGroup will convert Running to Pending.

pkg/controller/podgroup_test.go
lianghao208: done

pkg/controller/podgroup.go
lianghao208: done

pkg/controller/podgroup_test.go
lianghao208: done

lianghao208:
@denkensk Thank you for the review, I've updated the changes, PTAL.

pkg/controller/podgroup.go
```diff
@@ -238,6 +238,12 @@ func (ctrl *PodGroupController) syncHandler(key string) error {
 					failed++
 				}
 			}
+		} else {
```
denkensk, line 241: It may be a little duplicate of L248-L250. 
I think it would be better to rewrite it like this, WDYT?
```go
	for _, pod := range pods {
		switch pod.Status.Phase {
		case v1.PodRunning:
			running++
		case v1.PodSucceeded:
			succeeded++
		case v1.PodFailed:
			failed++
		}
	}

       pgCopy.Status.Failed = failed
       pgCopy.Status.Succeeded = succeeded
       pgCopy.Status.Running = running

       if len(pods) == 0 {
            pgCopy.Status.Phase = schedv1alpha1.PodGroupPending
            break
       }
```

denkensk (commented):
Just a small comment, please take a look.
@lianghao208

pkg/controller/podgroup.go
lianghao208: That makes sense to me. Thanks.

lianghao208: done.

denkensk:
/lgtm
@lianghao208  Thanks for your help

denkensk:
/assign @huang-wei

Pls take a look again

PR #312, opened 2021-12-20
fix bugs of resource calculation in node resource topology

Garrybest:
Signed-off-by: Garrybest <garrybest@foxmail.com>

According to the [doc](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources), the Pod's effective request is the higher of:

- the sum of all app containers(spec.Containers) request for a resource.
- the effective init containers(spec.InitContainers) request for a resource.

The effective init containers request is the highest request on all init containers.

And that's what the cpumanager of kubelet does when caculating resources. See [here](https://github.com/kubernetes/kubernetes/blob/b1479da6ad7d4afeb5056ce1a17312a39e06a888/pkg/kubelet/cm/cpumanager/policy_static.go#L357-L383).

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -98,18 +98,8 @@ func resMatchNUMANodes(numaNodes NUMANodeList, resources v1.ResourceList, qos v1
 
 func singleNUMAPodLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nodeInfo *framework.NodeInfo) *framework.Status {
 	klog.V(5).InfoS("Pod Level Resource handler")
-	resources := make(v1.ResourceList)
 
-	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
-	// and normal containers in the one scope
-	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
-		for resource, quantity := range container.Resources.Requests {
-			if q, ok := resources[resource]; ok {
-				quantity.Add(q)
-			}
-			resources[resource] = quantity
-		}
-	}
+	resources := util.GetPodEffectiveRequest(pod)
 
 	if resMatchNUMANodes(createNUMANodeList(zones), resources, v1qos.GetPodQOS(pod), nodeInfo) {
 		// definitely we can't align container, so we can't align a pod
--- a/pkg/util/resource_test.go
+++ b/pkg/util/resource_test.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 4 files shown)


pkg/noderesourcetopology/filter.go
```diff
@@ -156,3 +146,34 @@ func resourceFoundOnNode(resName v1.ResourceName, wantQuantity resource.Quantity
 	}
 	return false
 }
+
+func getPodResource(pod *v1.Pod) v1.ResourceList {
```
denkensk, line 150: Better to move the function to `https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/util/resource.go`. Then it can be used by other plugins if needed.

fromanirh:
/cc @swatisehgal @AlexeyPerevalov 
will review myself ASAP

swatisehgal:
Thanks for the PR @Garrybest and for the reminder @fromanirh. This PR somehow slipped through the cracks. I will take a look at it ASAP.

swatisehgal:
This is a good catch @Garrybest and I appreciate your contribution. I think it makes sense to handle the initcontainers and app containers the way they have been implemented in the CPU manager and to reflect the documentation [here ](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#resources) to determine the effective resource request and limit.

/lgtm

/hold for @AlexeyPerevalov and @fromanirh to remove the hold after their review

Huang-Wei:
There is an upstream issue talking about having a central place to calculate pod resources: https://github.com/kubernetes/kubernetes/issues/106769.

We can leverage that once it's completed.

Huang-Wei:
/approve
Will leave the /unhold to @AlexeyPerevalov and @fromanirh.

fromanirh (commented):
the code looks good to me, but  I think we should add few unit tests. This code is very easy to test, so adding them should be quick and painless.

pkg/util/resource_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestGetPodEffectiveRequest(t *testing.T) {
+	type args struct {
+		containerRequest     []v1.ResourceList
+		initContainerRequest []v1.ResourceList
+	}
+	tests := []struct {
+		name string
+		args args
```
Huang-Wei, line 25: just flatten args here

pkg/util/resource_test.go
```diff
@@ -0,0 +1,126 @@
+package util
```
Huang-Wei, line 1: Missing copyright.

pkg/util/resource_test.go
```diff
@@ -0,0 +1,126 @@
...
+			args: args{
+				containerRequest: []v1.ResourceList{
+					makeResourceList(1, 1),
+				},
+				initContainerRequest: nil,
+			},
+			want: makeResourceList(1, 1),
+		},
+		{
+			name: "2 container",
```
Huang-Wei, line 39: 2 containers (applies elsewhere)

swatisehgal:
/unhold
as unit tests were added as per @fromanirh's request.

PR #224, opened 2021-08-07
PreemptionToleration Plugin and its tests

everpeace:
This PR is an implementation of "Preemption Toleration" plugin proposed in (#205).

In this plugin, many code is borrowed from default preemption plugin in kube-scheduler because this plugin requires almost identical logics with this but most of them are not exposed. I believe kubernetes/kubernetes#95131 resolves it.  I isolated a commit of coping of borrowed code in a2b8f0a.

- [x] squash the commits

PR #313, opened 2021-12-21
Add PreemptionToleration to Plugin List in README

everpeace:
It just added [Preemption Toleration](pkg/preemptiontoleration/README.md) plugin to the list in README which is merged in #224 .


denkensk:
@everpeace 
/lgtm
/approve
Thanks for your contribution.
Another suggestion: it will be better to add an example in the readme.

PR #314, opened 2021-12-21
eliminate versioned scheduler config types in internal config types (pkg/apis/types.go)

### LINKED ISSUE #309: eliminate the "versioned" import in pkg/apis/config
Currently, we're using _versioned_ import in `pkg/apis/config/types.go`:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/apis/config/types.go#L22

This should be eliminated.

ref: https://github.com/kubernetes-sigs/scheduler-plugins/pull/224#discussion_r712521093, https://github.com/kubernetes-sigs/scheduler-plugins/pull/224#discussion_r767310629

everpeace:
because internal config types should only use/depend on unversioned types and they should be converted from versioned config types.

fixes #309


pkg/apis/config/zz_generated.deepcopy.go
```diff
@@ -22,9 +22,9 @@ limitations under the License.
 package config
 
 import (
-	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	runtime "k8s.io/apimachinery/pkg/runtime"
-	v1 "k8s.io/kube-scheduler/config/v1"
+	apisconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
```
denkensk, line 27: `apisconfig` is the generated abbreviate. Maybe we can use it to replace the `unversioned` in other files to keep consistent.

denkensk (commented):
Small comment.
Thanks @everpeace

pkg/apis/config/zz_generated.deepcopy.go
denkensk: And the defaultconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"

everpeace:
I close this #344 already resolved the issue: 
https://github.com/kubernetes-sigs/scheduler-plugins/pull/334/files#diff-a18720136db46ef660133ca97a9e1962a52e8219d3eb688daa5a9a81c64deb2fL22-L25

PR #317, opened 2022-01-16
register custom resource events for efficient pod enqueuing

Huang-Wei:
Along with https://github.com/kubernetes/kubernetes/pull/101394, the downstream plugins can register custom resource event(s) so as to trigger efficient pod queueing.

### DIFF
--- a/pkg/capacityscheduling/capacity_scheduling.go
+++ b/pkg/capacityscheduling/capacity_scheduling.go
@@ -42,6 +42,7 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/defaultpreemption"
 	schedutil "k8s.io/kubernetes/pkg/scheduler/util"
 
+	"sigs.k8s.io/scheduler-plugins/pkg/apis/scheduling"
 	"sigs.k8s.io/scheduler-plugins/pkg/apis/scheduling/v1alpha1"
 	"sigs.k8s.io/scheduler-plugins/pkg/generated/clientset/versioned"
 	schedinformer "sigs.k8s.io/scheduler-plugins/pkg/generated/informers/externalversions"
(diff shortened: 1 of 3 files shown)


pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -184,9 +185,12 @@ func New(obj runtime.Object, handle framework.Handle) (framework.Plugin, error)
 }
 
 func (c *CapacityScheduling) EventsToRegister() []framework.ClusterEvent {
+	// To register a custom event, follow the naming convention at:
+	// https://git.k8s.io/kubernetes/pkg/scheduler/eventhandlers.go#L403-L410
+	eqGVK := fmt.Sprintf("elasticquotas.v1alpha1.%v", scheduling.GroupName)
 	return []framework.ClusterEvent{
 		{Resource: framework.Pod, ActionType: framework.Delete},
-		// TODO: once bump the dependency to k8s 1.22, addd custom object events.
+		{Resource: framework.GVK(eqGVK), ActionType: framework.Add | framework.Update},
```
denkensk, line 193: We also need to include the Delete event for elasticquota. Pod may be scheduled successfully if the eq is deleted.

Huang-Wei: Can you elaborate in which circumstance a deleted eq would potentially make a pod (failed by this plugin) schedulable? My understanding is that the unschedulable pods needs to be get its own label updated, and it's irrelevant with EQ's change.

denkensk: I wonder if there is a scenario.
Scheduling of PodA in namespace A fails because the used in eq of namespace A will exceed the max rather than a lack of cluster resources. Then we remove the eq in namespace A so that it is no longer restricted to pod in namespace A. At this point we may need to have the pod in namespace A rescheduled.

But in fact we are not allowed to delete eq in a production environment if there are pods in the namespace A, unless the pods are deleted first.

Huang-Wei: sg, I will update the event to `All`.

pkg/noderesourcetopology/plugin.go
```diff
@@ -117,8 +118,12 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 // should be registered for this plugin since a Pod update may free up resources
 // that make other Pods schedulable.
 func (tm *TopologyMatch) EventsToRegister() []framework.ClusterEvent {
+	// To register a custom event, follow the naming convention at:
+	// https://git.k8s.io/kubernetes/pkg/scheduler/eventhandlers.go#L403-L410
+	nrtGVK := fmt.Sprintf("noderesourcetopologies.v1alpha1.%v", topologyapi.GroupName)
 	return []framework.ClusterEvent{
 		{Resource: framework.Pod, ActionType: framework.Delete},
 		{Resource: framework.Node, ActionType: framework.Add | framework.UpdateNodeAllocatable},
+		{Resource: framework.GVK(nrtGVK), ActionType: framework.Add | framework.Update},
```
Huang-Wei, line 127: FYI: @swatisehgal @AlexeyPerevalov: this change is aimed to give a pod failed by NRT filter plugin to be retried immediately. I think only NRT's add and update events are correlated.

denkensk:
/lgtm

Thanks @Huang-Wei

PR #318, opened 2022-01-19
NodeResourceTopology: handle best-effort pods in score plugin

Tal-or:
Scheduling best-effort pods with score plugin leads to scheduler pod crashing.
Best-effort pods are not requesting any numeric value of resources so all resources are 0, thus the program panics with a division by zero operation.

For every Best-effort pod we should return a score of 100 since from the NodeResourceTopology perspective, every node is a good fit for best-effort pods and this is what this patch is aiming to do.

### DIFF
--- a/pkg/noderesourcetopology/score.go
+++ b/pkg/noderesourcetopology/score.go
@@ -24,6 +24,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
+	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	apiconfig "sigs.k8s.io/scheduler-plugins/pkg/apis/config"
 
@@ -55,6 +56,11 @@ func (rw resourceToWeightMap) weight(r v1.ResourceName) int64 {
 
 func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string) (int64, *framework.Status) {
 	klog.V(5).InfoS("Scoring node", "nodeName", nodeName)
+	// if it's a non-guaranteed pod, every node is considered to be a good fit
+	if v1qos.GetPodQOS(pod) != v1.PodQOSGuaranteed {
+		return framework.MaxNodeScore, nil
+	}
+
 	nodeTopology := findNodeTopology(nodeName, tm.lister)
 
 	if nodeTopology == nil {
--- a/test/integration/noderesourcetopology_test.go
+++ b/test/integration/noderesourcetopology_test.go
@@ -553,6 +553,32 @@ func TestTopologyMatchPlugin(t *testing.T) {
 			},
 			expectedNodes: []string{"fake-node-1"},
 		},
+		// in the following "Scheduling Best-Effort pod with least-allocatable/balanced-allocation/most-allocatable strategy" tests,
+		// both nodes are expected to get the maximum score.
+		// in case of multiple nodes with the same highest score, the chosen node selection might vary.
+		// thus, we can't tell exactly which node will get selected, so the only option is to accept both of them,
+		// and it's still good enough for making sure that pod not stuck in pending state and/or the program isn't crashing
+		{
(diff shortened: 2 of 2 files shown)


pkg/noderesourcetopology/score.go
```diff
@@ -61,6 +62,11 @@ func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState,
 		return 0, nil
 	}
 
+	// if it's a non-guaranteed pod, every node is account to be a good fit
+	if v1qos.GetPodQOS(pod) != v1.PodQOSGuaranteed {
+		return 100, nil
```
fromanirh, line 67: the intent of the code is well explained and looks fine to me, I just wonder if the right value to use here (100) is documented somewhere and if we should use a named constant for it.

test/integration/noderesourcetopology_test.go
```diff
@@ -553,6 +553,59 @@ func TestTopologyMatchPlugin(t *testing.T) {
 			},
 			expectedNodes: []string{"fake-node-1"},
 		},
+		{
+			name: "Scheduling Best-Effort pod with most-allocatable strategy scheduler",
```
fromanirh, line 557: I think we should run a test like this also for `leastAllocatable` and `balancedAllocation`.

fromanirh (commented):
thanks for the updates, LGTM now :)

pkg/noderesourcetopology/score.go
```diff
@@ -61,6 +62,11 @@ func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState,
 		return 0, nil
 	}
 
+	// if it's a non-guaranteed pod, every node is account to be a good fit
+	if v1qos.GetPodQOS(pod) != v1.PodQOSGuaranteed {
```
Huang-Wei, line 66: I think you can even move it above `findNodeTopology()`.

Huang-Wei (commented):
Thanks @Tal-or for the catch!

A nit. LGTM otherwise.

Tal-or:
> Thanks @Tal-or for the catch!

I'm not the one who found it, I'm only the one who fix it  :)

Huang-Wei:
@Tal-or could you take a look at the ci failure?

test/integration/noderesourcetopology_test.go
```diff
@@ -553,6 +553,32 @@ func TestTopologyMatchPlugin(t *testing.T) {
...
+		{
+			name: "Scheduling Best-Effort pod with least-allocatable strategy scheduler",
+			pods: []*v1.Pod{
+				st.MakePod().Namespace(ns.Name).Name("topology-aware-scheduler-pod").SchedulerName(leastAllocatableScheduler).Container(pause).Obj(),
+			},
+			// in this test, both nodes are expected to get the maximum score.
+			// in case of multiple nodes with the same highest score the selection done randomly.
+			// thus, we can't tell exactly which node will get selected, so the only option is to accept both of them,
+			// and it's still good enough for making sure that pod not stuck in pending state and/or the program isn't crashing
+			// same goes to the other balanced-allocation and most-allocatable strategy tests below
```
fromanirh, line 565: minor: this is a good comment, but it applies to all the three cases (leastAllocatable, mostAllocatable, balancedAllocation), right? moving it between lines 555 and 556 and explicitely mention the three strategies can be clearer.

fromanirh (commented):
good catch on the test fixes!

Tal-or:
> @Tal-or could you take a look at the ci failure?

Test failed because when there are multiple nodes with the highest score the selection might vary:
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/generic_scheduler.go#L158 

I modify the test a bit in order to handle that and document the rationale in both the commit and comment in the code.

test/integration/noderesourcetopology_test.go
Tal-or: sure. done

Huang-Wei:
The introduced tests can cover the crash issue upon best-efforts pods, so LGTM.

/lgtm
/approve

Tal-or:
> The introduced tests can cover the crash issue upon best-efforts pods, so LGTM.
> 
> /lgtm /approve

Thanks for the quick response.
Much appreciated

PR #319, opened 2022-01-19
cleanup: "allocable" -> "allocated" to adapt to upstream naming convension

Huang-Wei:
Change the literal "allocable" to "allocated" so as to adapt to upstream naming convension.

Before (which doesn't read well IMO):

```go
leastAllocatableScheduler = fmt.Sprintf("%v-scheduler", string(scheconfig.MostAllocated))
```

After:

```go
mostAllocatedScheduler = fmt.Sprintf("%v-scheduler", string(scheconfig.MostAllocated))
```

### DIFF
--- a/test/integration/noderesourcetopology_test.go
+++ b/test/integration/noderesourcetopology_test.go
@@ -59,9 +59,9 @@ const (
 )
 
 var (
-	leastAllocatableScheduler   = fmt.Sprintf("%v-scheduler", string(scheconfig.MostAllocated))
+	mostAllocatedScheduler      = fmt.Sprintf("%v-scheduler", string(scheconfig.MostAllocated))
 	balancedAllocationScheduler = fmt.Sprintf("%v-scheduler", string(scheconfig.BalancedAllocation))
-	mostAllocatableScheduler    = fmt.Sprintf("%v-scheduler", string(scheconfig.LeastAllocated))
+	leastAllocatedScheduler     = fmt.Sprintf("%v-scheduler", string(scheconfig.LeastAllocated))
 )
 
 func TestTopologyMatchPlugin(t *testing.T) {
@@ -126,7 +126,7 @@ func TestTopologyMatchPlugin(t *testing.T) {
 	cfg.Profiles = append(cfg.Profiles,
 		// a profile with both the filter and score enabled and score strategy is MostAllocated
 		makeProfileByPluginArgs(
-			leastAllocatableScheduler,
+			mostAllocatedScheduler,
 			makeResourceAllocationScoreArgs(ns.Name, &scheconfig.ScoringStrategy{Type: scheconfig.MostAllocated}),
 		),
 		// a profile with both the filter and score enabled and score strategy is BalancedAllocation
@@ -136,7 +136,7 @@ func TestTopologyMatchPlugin(t *testing.T) {
 		),
 		// a profile with both the filter and score enabled and score strategy is LeastAllocated
 		makeProfileByPluginArgs(
-			mostAllocatableScheduler,
+			leastAllocatedScheduler,
 			makeResourceAllocationScoreArgs(ns.Name, &scheconfig.ScoringStrategy{Type: scheconfig.LeastAllocated}),
 		),
 	)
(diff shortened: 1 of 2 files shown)


Huang-Wei:
/hold
to merge #318 first.

Tal-or:
I found this change very reasonable, but if we're going in this direction, don't we need to change the naming convention in both testing and implementation?

swatisehgal:
Thanks @Huang-Wei for the PR! Aligning with upstream naming conventions makes a lot of sense and prevents unnecessary confusion so I am in the favor of this change. 
I agree wit @Tal-or comment above. Furthermore, (in addition to updating tests and implementation) we need to make sure that the configs [here](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/manifests/noderesourcetopology) and [documentation](https://github.com/kubernetes-sigs/scheduler-plugins/blame/master/pkg/noderesourcetopology/README.md#L56) is updated appropriately as well.

Huang-Wei:
> I found this change very reasonable, but if we're going in this direction, don't we need to change the naming convention in both testing and implementation?

@Tal-or I don't mean to eliminate all literals of "allocatable"; instead, just align the naming of the scoring strategies to be "xyzAllocated". The literal "allocatable" stands for a (fixed) amount of resources that a node has, so both upstream and here are all using it correctly. If I miss something, please let me know.

@swatisehgal yup, the README needs an update. However, the configs look good to me - the NRT's allocatable is named right, which is just like a regular Node's allocatable, which as I mentioned above, is a fixed value and thus don't need to be changed.

Tal-or:
> > I found this change very reasonable, but if we're going in this direction, don't we need to change the naming convention in both testing and implementation?
> 
> @Tal-or I don't mean to eliminate all literals of "allocatable"; instead, just align the naming of the scoring strategies to be "xyzAllocated". The literal "allocatable" stands for a (fixed) amount of resources that a node has, so both upstream and here are all using it correctly. If I miss something, please let me know.

Thank you for clarifying the scope of this change.
I went over the package files`, and yes, we already have all scoring strategies as "xyzAllocated" so I'm ok with this PR.

test/integration/noderesourcetopology_test.go
```diff
@@ -553,15 +553,15 @@ func TestTopologyMatchPlugin(t *testing.T) {
 			},
 			expectedNodes: []string{"fake-node-1"},
 		},
-		// in the following "Scheduling Best-Effort pod with least-allocatable/balanced-allocation/most-allocatable strategy" tests,
+		// in the following "Scheduling Best-Effort pod with most-allocated/balanced-allocation/lease-allocated strategy" tests,
```
Tal-or, line 556: typo: ```least-allocated``` instead ```lease-allocated```

test/integration/noderesourcetopology_test.go
```diff
@@ -573,9 +573,9 @@ func TestTopologyMatchPlugin(t *testing.T) {
 			expectedNodes: []string{"fake-node-1", "fake-node-2"},
 		},
 		{
-			name: "Scheduling Best-Effort pod with most-allocatable strategy scheduler",
+			name: "Scheduling Best-Effort pod with lease-allocated strategy scheduler",
```
Tal-or, line 576: typo: ```least-allocated``` instead ```lease-allocated```

Tal-or (commented):
Small typos, otherwise LGTM

test/integration/noderesourcetopology_test.go
Huang-Wei: updated. thanks!

swatisehgal:
We need to update naming in the configs in the https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/manifests/noderesourcetopology directory. Other than that LGTM!

Huang-Wei:
@swatisehgal there is nothing we need to change in `manifests/noderesourcetopology` - the literals "allocatable" there are not related with the Score strategy, just a node's allocable resoures:

```
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins|rename-ScoringStrategyType ⇒  ag allocatable manifests/noderesourcetopology
manifests/noderesourcetopology/crd.yaml
90:                      allocatable:
94:                        description: Allocatable quantity of the resource, corresponding
95:                          to allocatable in node status, i.e. total amount of this
104:                          available for new (to be scheduled) pods, i.e. Allocatable
121:                    - allocatable

manifests/noderesourcetopology/worker-node-B.yaml
13:        allocatable: 3
16:        allocatable: 3
22:        allocatable: 3
25:        allocatable: 3

manifests/noderesourcetopology/worker-node-A.yaml
13:        allocatable: 3
16:        allocatable: 1
19:        allocatable: 2
25:        allocatable: 3
28:        allocatable: 2
31:        allocatable: 1
```

swatisehgal:
Ah, yeah my bad 🤦 the names corresponding in the score strategy name in the config files are already correct. I assumed that we had to update those as well.

/lgtm

Issue #322, opened 2022-01-26
move PodGroupLabel to pkg/apis


zw0610:
`PodGroupLabel` (`"pod-group.scheduling.sigs.k8s.io"`) is exposed to users: https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/coscheduling/README.md?plain=1#L111

It would be better to PodGroupLabel under pkg/apis.

PR #323, opened 2022-01-27
move PodGroup label to pkg/api

denkensk:
Recently we are trying to integrate the coscheduling with [training-operator](https://github.com/kubeflow/training-operator).
https://github.com/kubeflow/training-operator/issues/1518
https://github.com/kubeflow/training-operator/pull/1526
https://github.com/kubeflow/common/pull/185

It's better to move PodGroup label to pkg/api for other projects to import. 

Fix https://github.com/kubernetes-sigs/scheduler-plugins/issues/322


pkg/apis/scheduling/v1alpha1/types.go
```diff
@@ -107,6 +107,9 @@ const (
 
 	// PodGroupFailed means at least one of `spec.minMember` pods is failed.
 	PodGroupFailed PodGroupPhase = "Failed"
+
+	// PodGroupLabel is the default label of coscheduling
+	PodGroupLabel = "pod-group.scheduling.sigs.k8s.io"
```
Huang-Wei, line 112: Maybe reuse the literal defined in https://github.com/kubernetes-sigs/scheduler-plugins/blob/ec632c3d7e04b7b372f9a6f4338b0dbc53ef3d46/pkg/apis/scheduling/register.go#L21?

denkensk: Sg thanks.
Done

Huang-Wei:
/lgtm

Thanks @denkensk .

PR #326, opened 2022-02-01
enhancement: improve the helm chart implementation to configure multi plugins

### LINKED ISSUE #321: How to add a new scheduler profile
Hi,
is it possible to add documentation on how to add new scheduler, scheduler profile and integrate it into the deployment with Helm?
When I look at `manifests` folder, I see that every profile has a folder with at least a `scheduler-config.yaml`, but I don't understand how and where to customize the various charts so that I can include my scheduler plugin in the build+deploy pipeline.
Also, my scheduler plugin pod needs a sidecar container and I would also provide a `deploy.yaml`, which I guess should go in `manifests/the-new-plugin` folder along with the `scheduler-config.yaml`.
Any help is appreciated.

Thanks

helayoty:
In this PR:
- Give users the ability to configure to enable plugins as a list.
- Give users the ability to configure to set list of plugins args.
Fixes #321

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/Chart.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/Chart.yaml
@@ -15,9 +15,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.21.6
+version: 0.22.6
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.21.6
+appVersion: 0.22.6
--- a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.plugins.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
@@ -5,35 +6,43 @@ metadata:
   namespace: {{ .Values.scheduler.namespace }}
 data:
   scheduler-config.yaml: |
-    apiVersion: kubescheduler.config.k8s.io/v1beta1
+    apiVersion: kubescheduler.config.k8s.io/v1beta2
     kind: KubeSchedulerConfiguration
     leaderElection:
       leaderElect: false
     profiles:
+    # Compose all plugins in one profile
     - schedulerName: scheduler-plugins-scheduler
       plugins:
+      # Ensure only the first queueSort plugin is rendered.
+
+ {{- $enabledPlugins := .Values.plugins.enabled -}}
+
+  {{- if and (.Values.global.queueSort) (has (index .Values.global.queueSort 0) $enabledPlugins) }}
         queueSort:
           enabled:
-          - name: Coscheduling
+          - name: {{ index .Values.global.queueSort 0 }}
           disabled:
           - name: "*"
-        preFilter:
+  {{- end }}
+   
+    {{- $hasExtension := true -}}
+    
+      {{- range $extension, $plugins := .Values.global.extensions }}
+       {{- $hasExtension = true -}}
+           {{- range $plugins }}
+             {{- if and $hasExtension (has . $enabledPlugins) }}
+        {{ $extension }}:
           enabled:
-          - name: Coscheduling
-        postFilter:
-          enabled:
-          - name: Coscheduling
-        permit:
-          enabled:
-          - name: Coscheduling
-        reserve:
-          enabled:
-          - name: Coscheduling
-        postBind:
-          enabled:
-          - name: Coscheduling
-      pluginConfig:
-      - name: Coscheduling
-        args:
-          permitWaitingTimeSeconds: 10
-          deniedPGExpirationTimeSeconds: 3
+           {{- $hasExtension = false -}}
+              {{- end}}
+            {{- if has . $enabledPlugins }}
+           - name: {{ title . }} 
+            {{- end}}
+         
+        {{- end }}
+      {{- end }}
+
+ {{- end }}
+
+  {{- /* TODO: wire CRD installation with enabled plugins. */}}
--- a/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
@@ -3,6 +3,9 @@ kind: ClusterRole
 metadata:
   name: scheduler-plugins-scheduler
 rules:
+- apiGroups: [""]
+  resources: ["namespaces"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: ["", "events.k8s.io"]
   resources: ["events"]
   verbs: ["create", "patch", "update"]
@@ -22,10 +25,10 @@ rules:
   verbs: ["get", "update"]
 - apiGroups: [""]
   resources: ["nodes"]
-  verbs: ["get", "list", "watch"]
+  verbs: ["get", "list", "watch", "patch"]
 - apiGroups: [""]
   resources: ["pods"]
-  verbs: ["delete", "get", "list", "watch"]
+  verbs: ["delete", "get", "list", "watch", "update"]
 - apiGroups: [""]
   resources: ["bindings", "pods/binding"]
   verbs: ["create"]
@@ -54,8 +57,12 @@ rules:
   resources: ["subjectaccessreviews"]
   verbs: ["create"]
 - apiGroups: ["storage.k8s.io"]
-  resources: ["csinodes", "storageclasses"]
+  resources: ["csinodes", "storageclasses" , "csidrivers" , "csistoragecapacities"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["topology.node.k8s.io"]
+  resources: ["noderesourcetopologies"]
+  verbs: ["*"]
+# resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
   resources: ["podgroups", "elasticquotas"]
   verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
@@ -81,6 +88,13 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["get", "list", "watch"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["get", "list", "watch", "patch"]
+- apiGroups: ["topology.node.k8s.io"]
+  resources: ["noderesourcetopologies"]
+  verbs: ["*"]
+# resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
   resources: ["podgroups", "elasticquotas"]
   verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
@@ -114,4 +128,3 @@ subjects:
 - kind: ServiceAccount
   name: {{ .Values.controller.name }}
   namespace: {{ .Values.controller.namespace }}
-
--- a/manifests/install/charts/as-a-second-scheduler/values.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/values.yaml
@@ -1,15 +1,35 @@
+
 # Default values for scheduler-plugins-as-a-second-scheduler.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
 scheduler:
   name: scheduler-plugins-scheduler
-  image: k8s.gcr.io/scheduler-plugins/kube-scheduler:v0.21.6
+  image: k8s.gcr.io/scheduler-plugins/kube-scheduler:v0.22.6
   namespace: scheduler-plugins
   replicaCount: 1
 
 controller:
   name: scheduler-plugins-controller
-  image: k8s.gcr.io/scheduler-plugins/controller:v0.21.6
+  image: k8s.gcr.io/scheduler-plugins/controller:v0.22.6
   namespace: scheduler-plugins
   replicaCount: 1
+
+# LoadVariationRiskBalancing and TargetLoadPacking are not enabled by default
+# as they need extra RBAC privileges on metrics.k8s.io.
+
+plugins:
+  enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch", "NodeResourcesAllocatable"]
+
+global:  
+   # queueSort is not indented under extensions 
+   # as it needs to be globally enabled once. 
+  queueSort: ["Coscheduling"]
+  extensions: 
+    preFilter: ["Coscheduling", "CapacityScheduling"] 
+    filter: ["NodeResourceTopologyMatch"] 
+    postFilter: ["Coscheduling", "CapacityScheduling"] 
+    score: ["NodeResourceTopologyMatch", "NodeResourcesAllocatable"] 
+    permit: ["Coscheduling"] 
+    reserve: ["Coscheduling", "CapacityScheduling"] 
+    postBind: ["Coscheduling"]
--- a/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
@@ -45,7 +45,6 @@ spec:
         - --address=0.0.0.0
         - --leader-elect=false
         - --config=/etc/kubernetes/scheduler-config.yaml
-        - --scheduler-name=scheduler-plugins-scheduler
         image: {{ .Values.scheduler.image }}
         livenessProbe:
           httpGet:


manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -54,10 +57,21 @@ rules:
...
   verbs: ["create"]
 - apiGroups: ["storage.k8s.io"]
-  resources: ["csinodes", "storageclasses"]
+  resources: ["*"]
   verbs: ["get", "list", "watch"]
+# resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
-  resources: ["podgroups", "elasticquotas"]
+  resources: [
+  {{- if .Values.podgroups.enable }}
```
Huang-Wei, line 65: we may define some sub-fields under scheduler plugin name in values.yaml to indicate which CRDs are associated.

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -54,10 +57,21 @@ rules:
   resources: ["subjectaccessreviews"]
   verbs: ["create"]
 - apiGroups: ["storage.k8s.io"]
-  resources: ["csinodes", "storageclasses"]
+  resources: ["*"]
```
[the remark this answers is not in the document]
helayoty: I'm getting exceptions related to not being able to access `csistoragecapacities`, `csidrivers`. Let me add them only instead.

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -3,6 +3,9 @@ kind: ClusterRole
 metadata:
   name: scheduler-plugins-scheduler
 rules:
+- apiGroups: [""]
+  resources: ["namespaces", "configmaps"]
+  verbs: ["get", "list", "watch"]
```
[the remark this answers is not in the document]
helayoty: removed

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -10,13 +10,10 @@ data:
     leaderElection:
       leaderElect: false
     profiles:
-    - schedulerName: scheduler-plugins-scheduler
+    # user can add as many profiles here. Adding two profiles as an example
+{{- if .Values.podgroups.enable}}
+    - schedulerName: Coscheduling-plugins-scheduler
```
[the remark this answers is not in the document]
helayoty: Because we will have multiple profiles, we need to have different name for each one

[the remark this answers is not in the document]
Huang-Wei: how about give this option to the user? like `--set profiles.prof1.plugins="PluginX,PluginY" profiles.prof2.plugins="PluginA" ...`

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -10,13 +10,11 @@ data:
     leaderElection:
       leaderElect: false
     profiles:
-    - schedulerName: scheduler-plugins-scheduler
+    - schedulerName: default-scheduler
```
Huang-Wei, line 13: Note that `- schedulerName: default-scheduler` is not compatible with a secondary scheduler configuration as the default scheduler is also running out there.

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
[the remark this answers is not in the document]
helayoty: If you would like to give the user the ability to customize names, user has to pass the whole object -profile -

We can do that later once we upgrade to `v1.23` which profile only contains the name of scheulder

manifests/capacityscheduling/crd.yaml
```diff
@@ -1,4 +1,5 @@
 
+{{- if has "capacityScheduling" .Values.plugins.enable}}
```
Huang-Wei, line 2: I think this will fail at https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/hack/verify-crdgen.sh.

Can we find another way to keep this yaml intact?

helayoty: It will work if I removed it but the crd will be deployed even we don't explicitly add it to the enable list. what do you think?

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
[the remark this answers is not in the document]
helayoty: As we agreed, we will take care of this later.

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -10,13 +10,10 @@ data:
...
-    - schedulerName: scheduler-plugins-scheduler
+    # user can add as many profiles here. Adding two profiles as an example
+{{- if has "coscheduling" .Values.plugins.enable}}
+    - schedulerName: Coscheduling-plugins-scheduler
       plugins:
-        queueSort:
-          enabled:
-          - name: Coscheduling
-          disabled:
-          - name: "*"
```
Huang-Wei, line 19: These are needed for `Coscheduling`.

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -10,13 +10,10 @@ data:
     leaderElection:
       leaderElect: false
     profiles:
-    - schedulerName: scheduler-plugins-scheduler
+    # user can add as many profiles here. Adding two profiles as an example
+{{- if has "coscheduling" .Values.plugins.enable}}
+    - schedulerName: Coscheduling-plugins-scheduler
```
Huang-Wei, line 15: I'd suggest starting with a one-profile to include all plugins:

```yaml
- schedulerName: scheduler-plugins-scheduler
```

and just inject the Helm if directives around the particular plugins.

Otherwise, the test pod has to specify its `schedulerName` differently.

manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -8,6 +8,10 @@ scheduler:
   namespace: scheduler-plugins
   replicaCount: 1
 
+plugins:
+    enable: [ "coscheduling", "capacityScheduling"]
+    resources: [ "podgroups", "elasticquotas" ]
```
Huang-Wei, line 13: let's keep it 2 spaces instead of 4.

manifests/capacityscheduling/crd.yaml
Huang-Wei: Given these crds serve multiple purposes. I think removing the helm directives outweigh the benefits it brings in. So yes, I'm fine with registering the CRDs even if the corresponding plugin is not enabled.

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -10,8 +10,10 @@ data:
     leaderElection:
       leaderElect: false
     profiles:
+    # user can add as many profiles here. Adding two profiles as an example
```
Huang-Wei, line 13: The current templating needs some improvement. I played with Golang/Helm templating a bit, here is my suggestion:

```yaml
---
plugins:
{{- /* Ensure only the first queueSort plugin is rendered. */}}
{{- /* Note: the trick is b/c no "break" support in Go templating. */}}

{{- $queueSortFound := false }}
{{- range $p := .Values.plugins }}
{{- if and (has (title $p) $.Values.queueSort) (not $queueSortFound) }}
{{- $queueSortFound =  true }}
queueSort:
  enabled:
  - name: {{ title $p }}
  disabled:
  - name: "*"
{{- end }}
{{- end }}

{{- range $extension, $plugins := .Values.extensions }}
{{- $pluginFound := false }}
  {{- range $p := $.Values.plugins }}
  {{- if has (title $p) $plugins }}
{{- if not $pluginFound }}
{{ $extension }}:
  enabled:
{{- end }}
  - name: {{ title $p }}
  {{- $pluginFound = true }}
  {{- end }}
  {{- end }}
{{- end }}

{{- /* TODO: support pluginConfig. */}}
{{- /* TODO: tire CRD installation with enabled plugins. */}}
```

And the `values.yaml` can be like this:

```yaml
---
plugins: [""]

# queueSort is not indented under extensions
# as it needs to be globally enabled once.
queueSort: ["Coscheduling"]

extensions:
  preFilter: ["Coscheduling", "CapacityScheduling"]
  filter: ["NodeResourceTopologyMatch"]
  postFilter: ["Coscheduling", "CapacityScheduling", "PreemptionToleration"]
  score: ["NodeResourceTopologyMatch", "LoadVariationRiskBalancing", "TargetLoadPacking", "NodeResourcesAllocatable"]
  permit: ["Coscheduling"]
  reserve: ["Coscheduling", "CapacityScheduling"]
  postBind: ["Coscheduling"]
```

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -54,10 +54,15 @@ rules:
...
-  resources: ["csinodes", "storageclasses"]
+  resources: ["csinodes", "storageclasses" , "csidrivers" , "csistoragecapacities"]
   verbs: ["get", "list", "watch"]
+# resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
-  resources: ["podgroups", "elasticquotas"]
+  resources: [
+    {{- range .Values.plugins.resources }}
+    "{{.}}",
+    {{- end}}
```
Huang-Wei, line 64: let's hard code the CRDs for now.

manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -8,6 +8,10 @@ scheduler:
   namespace: scheduler-plugins
   replicaCount: 1
 
+plugins:
```
Huang-Wei, line 11: See my previous comment for a updated values.yaml.

manifests/install/charts/as-a-second-scheduler/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.21.6
+version: 0.21.7
```
Huang-Wei, line 18: Can we make it 0.21.6-1? (indicating the scheduler image isn't bumped). Also change appVersion to 0.21.6-1?

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -9,31 +9,39 @@ data:
     kind: KubeSchedulerConfiguration
     leaderElection:
       leaderElect: false
+    clientConnection:
+        kubeconfig: "/etc/kubernetes/scheduler.conf"
     profiles:
+    # user can add as many profiles here. Adding two profiles as an example
```
Huang-Wei, line 15: Let's just remove this line of comments, or just say:

```
Compose all plugins in one profile.
```

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -9,31 +9,39 @@ data:
...
     - schedulerName: scheduler-plugins-scheduler
       plugins:
+      # Ensure only the first queueSort plugin is rendered.
+
+  {{- if gt (len $.Values.queueSort) 0 }}
```
Huang-Wei, line 20: Wait a second... Does that mean the user has to provide a full values.yaml to override the defaults?

It's a good UX, right? I'd pursue that the user just provides a plugin list, and optionally extra plugin configs. WDYT?

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -81,10 +85,64 @@ rules:
...
+  verbs: ["*"]
+# resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
   resources: ["podgroups", "elasticquotas"]
   verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: topo-aware-scheduler-as-kube-scheduler
```
Huang-Wei, line 102: Why we need to treat NodeResourceTopology separately?

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -9,31 +9,39 @@ data:
...
-          permitWaitingTimeSeconds: 10
-          deniedPGExpirationTimeSeconds: 3
+      {{- range $index, $config := .Values.pluginsConfig }}
+        - name: {{ $config.name }}
+          args: 
+           {{- toYaml $config.args | nindent 12 }}
+      {{- end }}
+  {{- end}}
+
+  {{- /* TODO: tire CRD installation with enabled plugins. */}}
```
Huang-Wei, line 47: I made a typo... tire -> wire.

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
helayoty: That means that if the user specifies a value in the array(queueSort and rest of extensions apart), it will be applied, otherwise, the default will be used. So the user doesn't have to provide all values in the `values.yaml`

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
helayoty: I wanted to indicate that this binding is only used for the `NodeResourceTopology`, otherwise, it can be removed

manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
Huang-Wei: If a user wants to enable `CapacityScheduling` **only**, what's the Helm command to achieve that?

helayoty: Yes, by running the following:

`helm install scheduler-plugins ./as-a-second-scheduler/ --set extensions.preFilter={"CapacityScheduling"} --set extensions.postFilter={"CapacityScheduling"} --set extensions.reserve={"CapacityScheduling"} --set queueSort=null --set extensions.filter=null --set extensions.score=null --set extensions.permit=null --set extensions.postBind=null`

Huang-Wei: I'm afraid this is a bit verbose... Can we just provide a "plugins" option to make it oriented to users? End-user is better not aware of the extensions settings:

- plugins.enabled="CapacityScheduling" - this should be all the user needs to provide
- (optionally) plugins.disabled="<any default plugins>" - user can also disable plugins enabled by default in the vanilla scheduler

helayoty: > I'm afraid this is a bit verbose... Can we just provide a "plugins" option to make it oriented to users? End-user is better not aware of the extensions settings:
> 
> * plugins.enabled="CapacityScheduling" - this should be all the user needs to provide
> * (optionally) plugins.disabled="" - user can also disable plugins enabled by default in the vanilla scheduler

Alright, now the user can set only one variable `plugins.enabled` :

`helm install scheduler-plugins ./as-a-second-scheduler/ --set "plugins.enabled={"Coscheduling","CapacityScheduling","NodeResourceTopologyMatch","LoadVariationRiskBalancing","TargetLoadPacking","NodeResourcesAllocatable"}"`

Huang-Wei: Kudos! Thanks for the efforts!

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
Huang-Wei: Could you please try remove it? I don't think `NodeResourceTopology` needs a dedicated RBAC setting.

Huang-Wei (commented):
Some final nits.

Per our offline talk, the lack of support on deriving kubeConfig imposed extra burden of crafting a helm chart. Let's wait for a v0.22.6 image to be out and thus we can eliminate kubeConfigPath from the pluginConfig, which is tracked in https://github.com/kubernetes-sigs/scheduler-plugins/issues/330.

Huang-Wei:
I tested this PR with the following changes on the 0.22.6 staing(temporary) build, which works great! FYI.

Once https://github.com/kubernetes/k8s.io/pull/3367 gets merged, you should be able to try it out.

```diff
diff --git a/manifests/install/charts/as-a-second-scheduler/Chart.yaml b/manifests/install/charts/as-a-second-scheduler/Chart.yaml
index 646b131..7979922 100644
--- a/manifests/install/charts/as-a-second-scheduler/Chart.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/Chart.yaml
@@ -15,9 +15,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.21.6-1
+version: 0.22.6
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.21.6-1
+appVersion: 0.22.6
diff --git a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
index e9e83c7..14c7dc8 100644
--- a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
@@ -6,12 +6,10 @@ metadata:
   namespace: {{ .Values.scheduler.namespace }}
 data:
   scheduler-config.yaml: |
-    apiVersion: kubescheduler.config.k8s.io/v1beta1
+    apiVersion: kubescheduler.config.k8s.io/v1beta2
     kind: KubeSchedulerConfiguration
     leaderElection:
       leaderElect: false
-    clientConnection:
-        kubeconfig: "/etc/kubernetes/scheduler.conf"
     profiles:
     # Compose all plugins in one profile
     - schedulerName: scheduler-plugins-scheduler
diff --git a/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml b/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
index cb75347..26a4594 100644
--- a/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
@@ -44,10 +44,7 @@ spec:
         - /bin/kube-scheduler
         - --address=0.0.0.0
         - --leader-elect=false
-        - --authentication-kubeconfig=/etc/kubernetes/scheduler.conf
-        - --authorization-kubeconfig=/etc/kubernetes/scheduler.conf
-        - --config=/etc/kubernetes/scheduler-config/scheduler-config.yaml
-        - --scheduler-name=scheduler-plugins-scheduler
+        - --config=/etc/kubernetes/scheduler-config.yaml
         image: {{ .Values.scheduler.image }}
         livenessProbe:
           httpGet:
@@ -66,16 +63,11 @@ spec:
           privileged: false
         volumeMounts:
         - name: scheduler-config
-          mountPath: /etc/kubernetes/scheduler-config
-        - mountPath: /etc/kubernetes
-          name: kubeconfig
+          mountPath: /etc/kubernetes/
+          readOnly: true
       hostNetwork: false
       hostPID: false
       volumes:
       - name: scheduler-config
         configMap:
           name: scheduler-config
-      - hostPath:
-          path: /etc/kubernetes
-          type: Directory
-        name: kubeconfig
diff --git a/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml b/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
index 920d909..f63ee8a 100644
--- a/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
@@ -33,7 +33,7 @@ rules:
   resources: ["pods/status"]
   verbs: ["patch", "update"]
 - apiGroups: [""]
-  resources: ["replicationcontrollers", "services"]
+  resources: ["namespaces", "replicationcontrollers", "services"]
   verbs: ["get", "list", "watch"]
 - apiGroups: ["apps", "extensions"]
   resources: ["replicasets"]
diff --git a/manifests/install/charts/as-a-second-scheduler/values.yaml b/manifests/install/charts/as-a-second-scheduler/values.yaml
index d1b73de..0700f9c 100644
--- a/manifests/install/charts/as-a-second-scheduler/values.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/values.yaml
@@ -5,64 +5,31 @@
 
 scheduler:
   name: scheduler-plugins-scheduler
-  image: k8s.gcr.io/scheduler-plugins/kube-scheduler:v0.21.6
+  image: k8s.gcr.io/scheduler-plugins/kube-scheduler:v0.22.6
   namespace: scheduler-plugins
   replicaCount: 1
 
 controller:
   name: scheduler-plugins-controller
-  image: k8s.gcr.io/scheduler-plugins/controller:v0.21.6
+  image: k8s.gcr.io/scheduler-plugins/controller:v0.22.6
   namespace: scheduler-plugins
   replicaCount: 1
 
+# Disable LoadVariationRiskBalancing and TargetLoadPacking as they need
+# extra RBAC privileges on metrics.k8s.io.
+
 plugins:
-  enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch","LoadVariationRiskBalancing", "TargetLoadPacking", "NodeResourcesAllocatable"]
+  enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch", "NodeResourcesAllocatable"]
 
-global:  
-   # queueSort is not indented under extensions 
-   # as it needs to be globally enabled once. 
+global:
+   # queueSort is not indented under extensions
+   # as it needs to be globally enabled once.
   queueSort: ["Coscheduling"]
-  extensions: 
-    preFilter: ["Coscheduling", "CapacityScheduling"] 
-    filter: ["NodeResourceTopologyMatch"] 
-    postFilter: ["Coscheduling", "CapacityScheduling"] 
-    score: ["NodeResourceTopologyMatch", "LoadVariationRiskBalancing", "TargetLoadPacking", "NodeResourcesAllocatable"] 
-    permit: ["Coscheduling"] 
-    reserve: ["Coscheduling", "CapacityScheduling"] 
+  extensions:
+    preFilter: ["Coscheduling", "CapacityScheduling"]
+    filter: ["NodeResourceTopologyMatch"]
+    postFilter: ["Coscheduling", "CapacityScheduling"]
+    score: ["NodeResourceTopologyMatch", "NodeResourcesAllocatable"]
+    permit: ["Coscheduling"]
+    reserve: ["Coscheduling", "CapacityScheduling"]
     postBind: ["Coscheduling"]
-
-  pluginsConfig:
-    - name: CapacityScheduling
-      args:
-        kubeConfigPath: "/etc/kubernetes/scheduler.conf"
-    - name: Coscheduling
-      args:
-        kubeConfigPath: "/etc/kubernetes/scheduler.conf"
-        permitWaitingTimeSeconds: 10
-        deniedPGExpirationTimeSeconds: 3
-    - name: NodeResourceTopologyMatch
-      args:
-        kubeconfigpath: "/etc/kubernetes/scheduler.conf"
-        namespaces:
-          - default
-        scoringStrategy:
-          type: "LeastAllocated"
-    - name: NodeResourcesAllocatable
-      args:
-        mode: Least
-        resources:
-        - name: cpu
-          weight: 1000000
-        - name: memory
-          weight: 1
-    - name: TargetLoadPacking
-      args: 
-        targetUtilization: 60 
-        defaultRequests:
-          cpu: "1000m" 
-        defaultRequestsMultiplier: "1.8"
-        watcherAddress: https://google.com
-    - name: LoadVariationRiskBalancing
-      args:
-        safeVarianceMargin: 1
-        safeVarianceSensitivity: 2

```

manifests/install/charts/as-a-second-scheduler/Chart.yaml
```diff
@@ -15,9 +15,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.21.6
+version: 0.21.6-1
```
Huang-Wei, line 18: 0.22.6

manifests/install/charts/as-a-second-scheduler/Chart.yaml
```diff
@@ -15,9 +15,9 @@ type: application
...
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 0.21.6
+appVersion: 0.21.6-1
```
Huang-Wei, line 23: 0.22.6

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -3,6 +3,9 @@ kind: ClusterRole
 metadata:
   name: scheduler-plugins-scheduler
 rules:
+- apiGroups: [""]
+  resources: ["namespaces", "configmaps"]
```
Huang-Wei, line 7: remove configmaps as it's obtained via delegated authentication #272.

manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -1,15 +1,32 @@
...
   namespace: scheduler-plugins
   replicaCount: 1
 
 controller:
   name: scheduler-plugins-controller
-  image: k8s.gcr.io/scheduler-plugins/controller:v0.21.6
+  image: k8s.gcr.io/scheduler-plugins/controller:v0.22.6
   namespace: scheduler-plugins
   replicaCount: 1
+
```
Huang-Wei, line 17: Add a comment here:

```
# LoadVariationRiskBalancing and TargetLoadPacking are not enabled by default
# as they need extra RBAC privileges on metrics.k8s.io.
```

manifests/capacityscheduling/elasticquota-example.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: scheduling.sigs.k8s.io/v1alpha1
 kind: ElasticQuota
 metadata:
   name: test
-  namespace: test
+  namespace: default
```
Huang-Wei, line 5: Revert this as I suppose this is intended in some document.

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -81,10 +88,47 @@ rules:
...
+subjects:
+- kind: User
+  name: system:kube-scheduler
+  namespace: kube-system
+  apiGroup: rbac.authorization.k8s.io
+roleRef:
+  kind: ClusterRole
+  name: scheduler-plugins-controller
+  apiGroup: rbac.authorization.k8s.io
+---
```
Huang-Wei, line 131: I think we can delete these lines.

Huang-Wei (commented):
Good to be merged. Some final comments.

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
helayoty: I can still see errors in the logs that can't list namespaces

Huang-Wei:
@helayoty Pretty neat! Could you squash the commits? It's ready to be merged.

helayoty:
> @helayoty Pretty neat! Could you squash the commits? It's ready to be merged.

Thanks! done :)

PR #329, opened 2022-02-06
missing capacitySchedulingConfig config test

Huang-Wei:
capacitySchedulingConfig is defined but not tested. This PR adds the missing test.


denkensk:
/lgtm

Thanks. @Huang-Wei

PR #332, opened 2022-02-08
update the documents to adapt to 0.22.6

Huang-Wei:
/assign @denkensk


denkensk:
/lgtm

Thanks. @Huang-Wei

PR #337, opened 2022-02-10
define the purpose of the repo clearly

### LINKED ISSUE #278: Add explanation of the difference between in-tree plugins and plugins on this repo
Hello team. 

I saw https://github.com/kubernetes/kubernetes/issues/105802
And @Huang-Wei said
>  it's a good timing to consider onboard the long-requested coscheduling feature in the upstream.

I'm curious as to why this is about to be approved. (It doesn't mean I'm against it.)

It would be good to add an explanation of the difference between in-tree plugins and plugins on this repo, at least I want to know that :)

I, personally, think there is no reason to prevent adding useful plugins to in-tree, because users can disable them if they don't need them. Are the plugins in this repository developed as experimental and will be added to in-tree after they mature like `coscheduling`? Or, is `coscheduling` a special case, and will plugins that are considered unnecessary for most users be placed in this repository?

Sorry if this has already been explained somewhere.

/kind documentation

sanposhiho:
#### What type of PR is this?

/kind documentation

This PR aims to define the purpose of the repo in README. 

I am writing this based on the discussion in https://github.com/kubernetes/kubernetes/issues/105802.
Since there is no clear conclusion in that issue, we will need to discuss which criteria are really needed and which additional criteria need to be added.

Fixes #278 
Related https://github.com/kubernetes/kubernetes/issues/105802

I'm not a native English speaker and not good at English, so please review English expressions as well.


sanposhiho:
The failing should be resolved in #336.

README.md
```diff
@@ -2,6 +2,15 @@
 
 Repository for out-of-tree scheduler plugins based on the [scheduler framework](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/).
 
+This repository includes some incubating ideas of solveing real problems which is hard to be solved with only in-tree scheduler plugins. 
+You can use it directly via our providing images, use it as the library to build your custom scheduler, or use it as an incubator to build new plugins for your ideas.
+
+Whether a plugin is an in-tree plugin or a plugin here should be determined by the following points:
```
denkensk, line 8: I think this part should not appear in the README here.
Whether a plugin is an in-tree, it is needed to decide on a case-by-case basis. Even though the following conditions are met, it does not determine whether it can be placed in the in-tree.

README.md
```diff
@@ -2,6 +2,15 @@
 
 Repository for out-of-tree scheduler plugins based on the [scheduler framework](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/).
 
+This repository includes some incubating ideas of solveing real problems which is hard to be solved with only in-tree scheduler plugins.
```
denkensk, line 5: As you said, there is no clear criteria. So we can't call them incubating ideas.

I prefer to call it an enhanced scheduler based on the scheduler framework

denkensk (commented):
Thanks. 
I think REAEME is very important. 
It is to illustrate how we positioned the project. So do not want to have unclear descriptions.

And the relationship with the out-tree and in-tree should not be listed here, right? Maybe it should be in the introduction of scheduler framework?

Huang-Wei (commented):
Generally, I'm reluctant to put the statements on the top of README.md, especially given they're sort of subjective and optioned.

I'd expect we just describe the facts instead of judging which plugins should be in-tree, something like:

```
This repo provides scheduler plugins that are exercised in large companies.
These plugins can be vendored as Golang SDK libraries or used out-of-box via the pre-built images or Helm charts.
Additionally, this repo incorporates best practices and utilities to compose a high-quality scheduler plugin.
```

sanposhiho:
Thanks for reviewing @Huang-Wei @denkensk.
I applied @Huang-Wei's suggestion. 

So, we need to decide whether to in-tree or leave it here for each new plugin. It would be nice if the reason why new plugins are made as in-tree/scheduler-plugins are described somewhere. They may be useful if we decide the criteria in the future...

PR #338, opened 2022-02-14
bump go version to 1.17 in Dockerfiles

Huang-Wei:
#### What type of PR is this?

/kind failing-test

With the recent Go version bump in source code, the Go version in Dockerfile needs to upgraded as well; otherwise, it will throw an error: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-scheduler-plugins-push-images/1492937069842927616

```
go mod tidy: go.mod file indicates go 1.17, but maximum supported version is 1.16
```

Fixes #

### DIFF
--- a/build/controller/Dockerfile
+++ b/build/controller/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ARG ARCH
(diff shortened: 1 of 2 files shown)


Huang-Wei:
@seanmalloy could you kindly review this change?

PR #344, opened 2022-02-22
Polish integration tests to use envtest

### LINKED ISSUE #339: Polish integration tests
The current integration test relies on the upstream apiservertesting package, and due to that, we had to bear with some extra work/hack:

- add `update-generated-openapi.sh`
- introduce `hack/testdata` folder and a hack to [soft-link ](https://github.com/kubernetes-sigs/scheduler-plugins/blob/9ed1891abcd07dc6d8cc9631cc57e548e2250865/hack/integration-test.sh#L50) to the vendor folder

I'm proposing to leverage [setup-envtest](https://pkg.go.dev/sigs.k8s.io/controller-runtime/tools/setup-envtest) to achieve the same goal, while simplifying the the codebase as well as Makefile entries.

yibozhuang:
#### What type of PR is this?

/kind cleanup

This change removes the dependency on
`apiservertesting` testing package and leverage
the `envtest` framework from controller-runtime
https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/envtest
to spin up `apiserver` and `etcd` in integration test

Using this framework will simplify the code
and remove the need for `hack/testdata` and
the autogen target that uses `update-generated-openapi.sh`
The bring up of `apiserver` and `etcd` is done in the
TestMain so that we only need to do this once and
the integration tests will not need to spin up
`apiserver` each time which was the case before.

As part of the this rework, ensured that all pods
are properly cleaned up and removed at the end
of each integration test.

Signed-off By: Yibo Zhuang <yibzhuang@gmail.com>

Fixes #339


hack/integration-test.sh
```diff
@@ -21,42 +21,16 @@ set -o pipefail
...
-}
-
 runTests() {
-  kube::log::status "Starting etcd instance"
-  CLEANUP_REQUIRED=1
-  kube::etcd::start
+  kube::log::status "Configuring envtest"
+  TEMP_DIR=${TMPDIR-/tmp}
+  source ${TEMP_DIR}/setup-envtest
+  rm -rf ${TEMP_DIR}/setup-envtest
```
Huang-Wei, line 28: Can we make this part of `cleanup` and use `trap cleanup EXIT` to guarantee it?

go.mod
```diff
@@ -21,12 +19,114 @@ require (
...
-	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65
 	k8s.io/kube-scheduler v0.23.3
 	k8s.io/kubernetes v1.23.3
 	k8s.io/utils v0.0.0-20211116205334-6203023598ed
-	sigs.k8s.io/yaml v1.2.0
+	sigs.k8s.io/controller-runtime v0.11.1
+	sigs.k8s.io/yaml v1.3.0
+)
+
+require (
```
Huang-Wei, line 29: This new block is due to go 1.17?

test/integration/main_test.go
```diff
@@ -17,11 +17,32 @@ limitations under the License.
...
+	"log"
+	"path/filepath"
 	"testing"
 
-	"k8s.io/kubernetes/test/integration/framework"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
 )
 
+var testCfg *rest.Config
```
Huang-Wei, line 28: Add a comment above, and maybe rename to globalKubeConfig

```
// Global kubeConfig to connect to the testing API server.
```

test/integration/utils.go
```diff
@@ -143,3 +154,102 @@ func (e *eqWrapper) Used(used v1.ResourceList) *eqWrapper {
 func (e *eqWrapper) Obj() *v1alpha1.ElasticQuota {
 	return e.ElasticQuota
 }
+
+type testContext struct {
```
Huang-Wei, line 158: any particular reason to compose one instead of reusing the upstream TestContext?

test/integration/utils.go
```diff
@@ -143,3 +154,102 @@ func (e *eqWrapper) Used(used v1.ResourceList) *eqWrapper {
...
+	if testCtx.DynInformerFactory != nil {
+		testCtx.DynInformerFactory.Start(testCtx.Ctx.Done())
+	}
+	testCtx.InformerFactory.WaitForCacheSync(testCtx.Ctx.Done())
+	if testCtx.DynInformerFactory != nil {
+		testCtx.DynInformerFactory.WaitForCacheSync(testCtx.Ctx.Done())
+	}
+}
+
+func cleanupPods(t *testing.T, testCtx *testContext, pods []*v1.Pod) {
```
Huang-Wei, line 211: Same here, any particular reason to duplicate the code here instead of reusing the k/k version?

Huang-Wei (commented):
Thanks @yibozhuang ! This PR is super neat!

Some nits below.

go.mod
yibozhuang: Right I'm not sure why these indirect requires are being added after running
```
go mod tidy
```

I took the first one `github.com/Azure/go-ansiterm` for example and did
```
$ go mod why -m github.com/Azure/go-ansiterm
# github.com/Azure/go-ansiterm
sigs.k8s.io/scheduler-plugins/cmd/scheduler
k8s.io/kubernetes/cmd/kube-scheduler/app
k8s.io/component-base/term
github.com/moby/term
github.com/moby/term/windows
github.com/Azure/go-ansiterm
```

test/integration/main_test.go
yibozhuang: Sure thing, done.

go.mod
Huang-Wei: > Right I'm not sure why these indirect requires are being added after running

It seems to be a result of Go 1.17: https://stackoverflow.com/questions/69432877/why-there-are-two-require-blocks-in-go-mod-since-go-1-17

> Because in Go 1.17 the module graph has been changed to enable pruning and lazy loading. The second require block contains indirect dependencies.

Huang-Wei:
/lgtm
/approve

Thanks @yibozhuang !

Makefile
```diff
@@ -107,23 +107,19 @@ update-vendor:
...
+unit-test: update-vendor
 	hack/unit-test.sh
 
-.PHONY: install-etcd
-install-etcd:
-	hack/install-etcd.sh
-
-.PHONY: autogen
-autogen: update-vendor
-	hack/update-generated-openapi.sh
```
zorro786, line 119: @yibozhuang Removing this script is causing the following error for me when I try to run intg tests:

```
+++ [0308 22:31:18] Configuring envtest
+++ [0308 22:31:18] Running integration test cases
# k8s.io/kubernetes/cmd/kube-apiserver/app
vendor/k8s.io/kubernetes/cmd/kube-apiserver/app/server.go:401:80: undefined: "k8s.io/kubernetes/pkg/generated/openapi".GetOpenAPIDefinitions
```

Earlier running `make autogen` was used to fix this. I tried running `update-vendor` but it is not helping.

CC: @Huang-Wei

yibozhuang: See https://github.com/kubernetes-sigs/scheduler-plugins/pull/344#discussion_r813126976

Basically if your integration test is importing `k8s.io/kubernetes/test/integration/util` then you will hit this problem.
So I suggest taking a look at the `utils.go` at https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/test/integration/utils.go and update your test to remove any dependency on the k/k `test/integration/util` package.

zorro786: Neat!

Huang-Wei: @zorro786 now we get rid of hack/update-generated-openapi.sh :)

PR #347, opened 2022-02-25
implement strict alignment check for the noderesourcetopology filter plugin

### LINKED ISSUE #346: noderesourcetopology: require NUMA alignment for memory and hugepages
The noderesourcetopology filter plugin used to skip memory and hugepages when checking alignment requirements because the reporting of these resources wasn't readily available. That's no longer the case with changes landed in kubernetes >= 1.23 and with fixes to topology updating agents.

Considering the platform got up to speed, and considering the goals of the filter plugin, we should start doing stricter alignment checks for memory and hugepages, like the plugin already does for CPUs and devices.

fromanirh:
Add one of the following kinds:
/kind cleanup

The noderesourcetopology filter plugin used to skip memory and hugepages when checking alignment requirements because the reporting of these resources wasn't readily available. That's no longer the case, so we can now implement a stricter alignment check.

Fixes #346

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -91,18 +91,22 @@ func resMatchNUMANodes(numaNodes NUMANodeList, resources v1.ResourceList, qos v1
 func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, numaQuantity resource.Quantity) bool {
 	// Check for the following:
 	// 1. set numa node as possible node if resource is memory or Hugepages
-	// 2. set numa node as possible node if resource is cpu and it's not guaranteed QoS, since cpu will flow
-	// 3. set numa node as possible node if zero quantity for non existing resource was requested
-	// 4. otherwise check amount of resources
-	if resource == v1.ResourceMemory ||
-		strings.HasPrefix(string(resource), v1.ResourceHugePagesPrefix) ||
-		resource == v1.ResourceCPU && qos != v1.PodQOSGuaranteed ||
-		quantity.IsZero() ||
-		numaQuantity.Cmp(quantity) >= 0 {
-		// possible to align resources on NUMA node
+	if resource == v1.ResourceMemory {
 		return true
 	}
-	return false
+	if strings.HasPrefix(string(resource), v1.ResourceHugePagesPrefix) {
+		return true
+	}
+	// 2. set numa node as possible node if resource is cpu and it's not guaranteed QoS, since cpu will flow
+	if resource == v1.ResourceCPU && qos != v1.PodQOSGuaranteed {
+		return true
+	}
+	// 3. set numa node as possible node if zero quantity for non existing resource was requested
+	if quantity.IsZero() {
+		return true
+	}
+	// 4. otherwise check amount of resources
+	return numaQuantity.Cmp(quantity) >= 0
 }
 
 func singleNUMAPodLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nodeInfo *framework.NodeInfo) *framework.Status {
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -73,19 +73,12 @@ func resMatchNUMANodes(numaNodes NUMANodeList, resources v1.ResourceList, qos v1
 			if !ok && !resourceFoundOnNode(resource, quantity, nodeInfo) && !quantity.IsZero() {
 				continue
 			}
-			// Check for the following:
-			// 1. set numa node as possible node if resource is memory or Hugepages
-			// 2. set numa node as possible node if resource is cpu and it's not guaranteed QoS, since cpu will flow
-			// 3. set numa node as possible node if zero quantity for non existing resource was requested
-			// 4. otherwise check amount of resources
-			if resource == v1.ResourceMemory ||
-				strings.HasPrefix(string(resource), v1.ResourceHugePagesPrefix) ||
-				resource == v1.ResourceCPU && qos != v1.PodQOSGuaranteed ||
-				quantity.IsZero() ||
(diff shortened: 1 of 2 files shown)


fromanirh:
I'd love to add even more tests but the PR is now reviewable
@AlexeyPerevalov @swatisehgal @Tal-or

pkg/noderesourcetopology/filter.go
```diff
@@ -71,22 +70,15 @@ func resMatchNUMANodes(numaNodes NUMANodeList, resources v1.ResourceList, qos v1
...
-				resourceBitmask.Add(numaNode.NUMAID)
+
+			if !isNUMANodeSuitable(qos, resource, quantity, numaQuantity) {
+				continue
 			}
+
+			resourceBitmask.Add(numaNode.NUMAID)
 		}
 		bitmask.And(resourceBitmask)
 		if bitmask.IsEmpty() {
```
shajmakh, line 84: IIUC this function returns true if there is a matching numa to fit the resources. 
this code returns once the numa matching is found for one source (the first), don't we want to loop over the rest?

fromanirh: the function returns true if we need to discard the set of NUMA zones (pertaining to a compute node), false if at least a NUMA zone among all the provided ones can accomodate the workload, as defined as by its resources footprint (`resources` argument)

swatisehgal (approved):
Thanks for the PR, looks good to me. The refactor made the code much more readable!

Given that the memory support is enabled in the exporters, it makes sense to handle it appropriately in the scheduler plugin. 
/lgtm

fromanirh:
I decided to add more tests to the current PR

fromanirh:
@swatisehgal could you please have another look? :)

swatisehgal:
> @swatisehgal could you please have another look? :)

Sure, will take a look ASAP.

swatisehgal (approved):
/lgtm

pkg/noderesourcetopology/filter_test.go
```diff
@@ -35,111 +35,115 @@ import (
...
-		TopologyPolicies: []string{string(topologyv1alpha1.SingleNUMANodeContainerLevel)},
-		Zones: topologyv1alpha1.ZoneList{
-			{
-				Name: "node-0",
-				Type: "Node",
-				Resources: topologyv1alpha1.ResourceInfoList{
-					MakeTopologyResInfo(cpu, "20", "4"),
-					MakeTopologyResInfo(memory, "8Gi", "8Gi"),
-					MakeTopologyResInfo(nicResourceName, "30", "10"),
+func TestNodeResourceTopologyFilterPlugin(t *testing.T) {
```
Huang-Wei, line 44: nit: `TestNodeResourceTopologyFilter` should suffice.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -35,111 +35,115 @@ import (
...
-					MakeTopologyResInfo(memory, "8Gi", "4Gi"),
-					MakeTopologyResInfo(nicResourceName, "30", "2"),
+				{
+					Name: "node-75",
+					Type: "Node",
+					Resources: topologyv1alpha1.ResourceInfoList{
+						MakeTopologyResInfo(cpu, "30", "4"),
+						MakeTopologyResInfo(memory, "8Gi", "4Gi"),
+						MakeTopologyResInfo(nicResourceName, "30", "2"),
+					},
```
Huang-Wei, line 140: In the real world, is it common to have both numa and non-numa nodes in the same cluster? If so, maybe it's necessary to add a non-numa node as the fifth node, and check if the logic is as expected?

Also, the test's style can be improved. Per each sub-test, we can run through the nodes via Filter(), and compare the status on each node.

(btw: we can pursue these in a follow-up)

pkg/noderesourcetopology/filter_test.go
fromanirh: sounds good, will drop this commit.

pkg/noderesourcetopology/filter_test.go
fromanirh: I totally want to send even more test coverage in a followup PR.

From my experience, we can have NUMA and non-NUMA boxes in the same cluster, but nodes are expected to be partitioned in such a way that the scheduler never has to choose between these two groups (e.g. NUMA and non-NUMA nodes should never be compared by the scheduler).
Nevertheless, we should indeed have a test which cover this cases, if nothing else to document the expectations.

pkg/noderesourcetopology/filter.go
```diff
@@ -96,6 +88,32 @@ func resMatchNUMANodes(numaNodes NUMANodeList, resources v1.ResourceList, qos v1
...
+	}
+	// 3. set numa node as possible node if zero quantity for non existing resource was requested
+	if quantity.IsZero() {
+		return true
+	}
+	// 4. otherwise check amount of resources
+	if numaQuantity.Cmp(quantity) >= 0 {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
fromanirh: thanks, fixed!

pkg/noderesourcetopology/filter_test.go
Huang-Wei: Thanks. We should track them in a follow-up, along with https://github.com/kubernetes-sigs/scheduler-plugins/pull/347#discussion_r821944445

fromanirh: sure thing, done: https://github.com/kubernetes-sigs/scheduler-plugins/issues/351

pkg/noderesourcetopology/filter.go
fromanirh: filed https://github.com/kubernetes-sigs/scheduler-plugins/issues/351 - I'll evaluate if sending one or two PRs (testing + refactoring vs all in one)

PR #348, opened 2022-02-28
Initial commit Network-Aware framework

jpedro1992:
#### What type of PR is this?

This is the initial PR for the network-aware framework approved in this [KEP](https://github.com/kubernetes-sigs/scheduler-plugins/pull/282).

The PR includes two custom resources (AppGroup and NetworkTopology), three plugins (Networkaware), and one controller (AppGroup).

/kind feature

This proposal introduces an end-to-end solution to model/weight a cluster's network latency and topological information, and leverage that to better schedule latency- and bandwidth-sensitive workloads. Further information is available in our merged [KEP](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/260-network-aware-scheduling).

After several interactions in the KEP PR, we have updated our design and we submit the first version of the proposed network-aware framework. All feedback is welcome. Thanks.

Cc: @Huang-Wei


wangchen615:
cc @ravisantoshgudimetla @ingvagabund and @damemi

Huang-Wei:
> So, how should I proceed to take care of this issue for this PR since both CRDs are not approved yet? What is the proper way to add novel CRDs and to solve the issue raised by "api-approved.kubernetes.io"? Thanks.

@jpedro1992 you can just check in changes to include the "api-approved.kubernetes.io" fields, just like other CRDs do. Don't forget to append `# edited manually` in the end.

jpedro1992:
> > So, how should I proceed to take care of this issue for this PR since both CRDs are not approved yet? What is the proper way to add novel CRDs and to solve the issue raised by "api-approved.kubernetes.io"? Thanks.
> 
> @jpedro1992 you can just check in changes to include the "api-approved.kubernetes.io" fields, just like other CRDs do. Don't forget to append `# edited manually` in the end.

@Huang-Wei Thank you for getting back to me. 

Currently, I have the following "api-approved.kubernetes.io" field in the AppGroup and NetworkTopology CRDs: 

`api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/348 # edited manually`

I run `hack/verify-crdgen.sh` and it still fails. Then, if I manually edit the CRDs in the `manifests/crds` folder with the "api-approved.kubernetes.io" fields, the integration tests pass: 

image

Should I amend my commit with this version, or there is any other way to solve this issue?

wangchen615:
@jpedro1992 I got some suggestions from other reviewers to split the big PR into several commits. E.g. CRD/data type definitions + clientsets, Sorting + Filtering + Scoring implementation, examples, etc.

Huang-Wei:
> Should I amend my commit with this version

yes, please do.

@wangchen615 's suggestion is valid. It'd be very helpful for reviewing if you can split this PR into 2 smaller ones:

- APIs, including versioned/unversioned type definitions, validation, defaulting, conversion and tests (if needed)
- Pugin implementations and tests (UT and integration)

jpedro1992:
/remove-lifecycle stale

Please let me know if I can provide further help/info on the PR revision.

pkg/controller/appgroup.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+	}
+
+	// Sort workload data by Selector
+	sort.Sort(util.BySelector(agCopy.Spec.Workloads))
+
+	for id, workloadName := range order {
+		index := int32(id + 1)
+		// Find workload data by name (Binary Search)
+		w := util.FindWorkloadByName(agCopy.Spec.Workloads, workloadName)
```
ingvagabund, line 382: Worth checking whether `w is actually empty v1alpha1.AppGroupWorkloadInfo{}`. For that case it's more preferable to either return a pointer to the workload instead of an empty struct. Or, index in the list and check for -1. E.g.
```go
func FindWorkloadByName(workloadList v1alpha1.AppGroupWorkloadList, name string) *v1alpha1.AppGroupWorkloadInfo {
	low := 0
	high := len(workloadList) - 1

	for low <= high {
		mid := (low + high) / 2
		if workloadList[mid].Workload.Name == name {
			return &workloadList[mid].Workload // Return the WorkloadInfo
		} else if workloadList[mid].Workload.Name < name {
			low = mid + 1
		} else if workloadList[mid].Workload.Name > name {
			high = mid - 1
		}
	}
	// Workload Info was not found
	return nil
}
...
w := util.FindWorkloadByName(agCopy.Spec.Workloads, workloadName)
if w == nil {
	err := fmt.Errorf("unable to find workload by name")
	klog.Error(err, "Binary search failed")
	return topologyList, err
}
topologyList = append(topologyList, v1alpha1.AppGroupTopologyInfo{
	Workload: *w,
	Index: index,
})
```

or

```go
func FindWorkloadByName(workloadList v1alpha1.AppGroupWorkloadList, name string) int {
	low := 0
	high := len(workloadList) - 1

	for low <= high {
		mid := (low + high) / 2
		if workloadList[mid].Workload.Name == name {
			return mid
		} else if workloadList[mid].Workload.Name < name {
			low = mid + 1
		} else if workloadList[mid].Workload.Name > name {
			high = mid - 1
		}
	}
	// Workload Info was not found
	return -1
}
...
idx := util.FindWorkloadByName(agCopy.Spec.Workloads, workloadName)
if idx == -1 {
	err := fmt.Errorf("unable to find workload by name")
	klog.Error(err, "Binary search failed")
	return topologyList, err
}
topologyList = append(topologyList, v1alpha1.AppGroupTopologyInfo{
	Workload: w[idx],
	Index: index,
})
```

pkg/controller/appgroup.go
```diff
@@ -0,0 +1,423 @@
...
+				Selector:   w.Selector,
+				APIVersion: w.APIVersion,
+				Namespace:  w.Namespace,
+			},
+			Index: index,
+		})
+	}
+
+	// Sort TopologyList by Selector
+	klog.V(5).Infof("Sort Topology List by workload name... ")
```
ingvagabund, line 397: s/workload name/workload selector name

pkg/controller/appgroup.go
```diff
@@ -0,0 +1,423 @@
...
+	klog.V(5).Info("topologyList: ", topologyList)
+	return topologyList, nil
+}
+
+// defaultTopologyOrder : retrieves a default sequence order for workload deployment
+func defaultTopologyOrder(workloadList v1alpha1.AppGroupWorkloadList) v1alpha1.AppGroupTopologyList {
+	var topologyList v1alpha1.AppGroupTopologyList
+	var i int32
+	i = 1
+	for _, w := range workloadList {
```
ingvagabund, line 409: You can drop `i` in favor of the for indexing:
```go
for i, w := range workloadList {
	topologyList = append(topologyList, v1alpha1.AppGroupTopologyInfo{
		Workload: w.Workload,
		Index:    i+1,
	})
}
```

pkg/util/appgroup.go
```diff
@@ -0,0 +1,316 @@
...
+	return v1alpha1.AppGroupWorkloadInfo{}
+}
+
+// GetPodAppGroupLabel : get AppGroup from pod annotations
+func GetPodAppGroupLabel(pod *v1.Pod) string {
+	return pod.Labels[v1alpha1.AppGroupLabel]
+}
+
+// GetPodAppGroupSelector : get Workload Selector from pod annotations
+func GetPodAppGroupSelector(pod *v1.Pod) string {
```
ingvagabund, line 83: This function is currently unused

pkg/util/appgroup.go
```diff
@@ -0,0 +1,316 @@
+/*
```
ingvagabund, line 1: All this code is currently coupled with `pkg/controller/appgroup.go`. What's the likelihood the code gets reused some place other than `pkg/controller/appgroup.go`?

Also, worth introducing `appgroup_test.go` testing individual sorting algorithm. `TestAppGroupController_Run` is testing all the bits from the top level making it harder to test specific use cases for e.g. the sorting algorithms.

ingvagabund (commented):
`NetworkTopology` type is currently not used at all. Worth introducing the type later alongside its controller. Or, if there's only an external controller then alongside the filter/score plugin(s).

ingvagabund:
> It'd be very helpful for reviewing if you can split this PR into 2 smaller ones:
>
> APIs, including versioned/unversioned type definitions, validation, defaulting, conversion and tests (if needed)
> Plugin implementations and tests (UT and integration)

Both points are relevant until a plugin is introduced given this PR currently introduces only CRDs and a controller. Since CRD types under https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/apis/scheduling have their own validation/defaulting.

@Huang-Wei is there any policy against introducing new CRDs and their corresponding controllers ahead of plugins? The AppGroup type and its controller will probably have no use beside recomputing the topological sorting. On the other hand, is the idea of this repository to always provide fully functioning plugins? Or, rather allowing the community to also continuously accept controllers which provide selected/pre-computed view of the world? So consumers of the repository can build the scheduling framework as combination of already existing plugins/controllers + other custom plugins/controllers not yet merged in the repository.

Huang-Wei:
> Both points are not relevant until a plugin is introduced given this PR currently introduces only CRDs and a controller. Since CRD types under [`master`/apis/scheduling](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/apis/scheduling?rgh-link-date=2022-08-11T13%3A01%3A25Z) have their own validation/defaulting.

IIRC, that comment was made when the PR was combining all things together.

> is there any policy against introducing new CRDs and their corresponding controllers ahead of plugins?

To make this repo sustainable (given limited reviewers' cycles), we have to use our bandwidth efficiently and focus on scheduling bits. So, I'm a bit inclined to have AppGroup and its controllers resided temporarily in its own repo, and moved back (if necessary) once it gets one or two release cycles to soak. Even if it's hosted outside, we can still compile it into our controller image. (it's just we need to be cautious to avoid cyclic dependency) In this way, @jpedro1992 you can evolve API and controller at your own pace. And in this repo, we focus on scheduling portion for now so that we won't be the blocker to see this plugin's debut in scheduler-plugins. WDYT?

jpedro1992:
Hi @ingvagabund and @Huang-Wei, 

Thank you for your revision! I will be working on the suggested code modifications. 

For me it's fine if we keep the CRDs + controllers outside of sig-scheduling for now. Is there any guide / example I could follow from another plugin to know exactly how to host the components outside of sig-scheduling and build them inside? 

I cannot just simply remove the CRDs + controllers and add the plugins since they require the AppGroup CRD. 

Thanks in advance!

apis/scheduling/v1alpha1/types.go
```diff
@@ -192,3 +193,283 @@ type PodGroupList struct {
...
+	Status AppGroupStatus `json:"status,omitempty" protobuf:"bytes,3,opt,name=status"`
+}
+
+// Constants for AppGroup
+const (
+	// AppGroupLabel is the default label of the AppGroup for the network-aware framework
+	AppGroupLabel = "app-group." + scheduling.GroupName
+
+	// AppGroupSelectorLabel is the default selector label for Pods belonging to a given Workload (e.g., workload = App-A)
+	AppGroupSelectorLabel = "workload"
```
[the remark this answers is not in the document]
jpedro1992: Indeed. my new suggestion:

`appGroupSelectorLabel = AppGroupLabel + ".workload"

pkg/controller/appgroup.go
```diff
@@ -0,0 +1,423 @@
...
+	// Sort workload data by Selector
+	sort.Sort(util.BySelector(agCopy.Spec.Workloads))
+
+	for id, workloadName := range order {
+		index := int32(id + 1)
+		// Find workload data by name (Binary Search)
+		w := util.FindWorkloadByName(agCopy.Spec.Workloads, workloadName)
+
+		topologyList = append(topologyList, v1alpha1.AppGroupTopologyInfo{
+			Workload: v1alpha1.AppGroupWorkloadInfo{
```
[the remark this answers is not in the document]
jpedro1992: Exactly. I will update the code. Thanks!

pkg/controller/appgroup.go
```diff
@@ -0,0 +1,423 @@
...
+	for _, w := range workloadList {
+		for _, dependency := range w.Dependencies {
+			tree[w.Workload.Name] = append(tree[w.Workload.Name], dependency.Workload.Name)
+		}
+	}
+
+	klog.V(5).Info("Service Dependency Tree: ", tree)
+
+	// Calculate order based on the specified algorithm
+	switch algorithm {
```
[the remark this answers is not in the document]
jpedro1992: Thanks! I will update the code.

pkg/util/appgroup.go
jpedro1992: This function is used by the plugins to get the AppGroup label from pods. Indeed, it is unused since the plugins are not included in this PR...

manifests/appgroup/redis-appGroup-example.yaml
```diff
@@ -0,0 +1,29 @@
...
+  name: redis-cluster
+spec:
+  numMembers: 2
+  topologySortingAlgorithm: AlternateKahn
+  workloads:
+    - workload:
+        kind: "Deployment"
+        name: "redis-leader-deployment"
+        selector: "redis-leader"
+        apiVersion: "appsv1"
```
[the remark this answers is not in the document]
jpedro1992: Yes, correct. Sorry for the typo...

pkg/controller/appgroup.go
```diff
@@ -0,0 +1,423 @@
...
+		runtime.HandleError(err)
+		return
+	}
+	ctrl.podAdded(obj)
+}
+
+// pgUpdated : reacts to a pod update
+func (ctrl *AppGroupController) podUpdated(old, new interface{}) {
+	ctrl.podAdded(new)
+}
```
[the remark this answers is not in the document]
jpedro1992: In the previous design, we were updating the CR via the controller when pods were rescheduled to other nodes. In the current design, perhaps it does not make sense to sync the AppGroup when pods are updated. So, it might be indeed removed to avoid multiples sync without a clear purpose...

jpedro1992:
Hi @ingvagabund and @Huang-Wei,

We decided to host the multiple components in the following repo: [Diktyo-io](https://github.com/diktyo-io)

Regarding the PR, do you prefer to do the modifications in this PR (remove CRDs + APIs and replace it with plugins) or open a new PR with the plugins? 

Thanks in advance!

jpedro1992:
> /close in favor of #432

I agree.

PR #354, opened 2022-03-15
noderesourcetopology: revamp the filter logging

fromanirh:
#### What type of PR is this?

/kind cleanup

Improve the logging in the noderesourcetopology filter plugin, reducing the logs but making them more dense and aiming to better clarify the decision points in the code.

Fixes N/A

With these changes applied, the logs now look like
```
I0329 14:59:16.519008       1 scheduling_queue.go:928] "About to try and schedule pod" pod="e2e-test-workload-placement-tmpol-4922/testpod"
I0329 14:59:16.519038       1 scheduler.go:516] "Attempting to schedule pod" pod="e2e-test-workload-placement-tmpol-4922/testpod"
I0329 14:59:16.519702       1 pluginhelpers.go:40] "Lister for nodeResTopoPlugin" lister=&{indexer:0xc000900f90}
I0329 14:59:16.519776       1 filter.go:174] "Found NodeResourceTopology" nodeTopology="nodeB.test.net"
I0329 14:59:16.519795       1 filter.go:129] "Pod Level Resource handler"
I0329 14:59:16.519703       1 pluginhelpers.go:40] "Lister for nodeResTopoPlugin" lister=&{indexer:0xc000900f90}
I0329 14:59:16.519880       1 pluginhelpers.go:83] "extracted NUMA resources" logKey="node-0" cpu="14" memory="15 GiB"
I0329 14:59:16.519900       1 filter.go:174] "Found NodeResourceTopology" nodeTopology="nodeA.test.net"
I0329 14:59:16.519910       1 filter.go:129] "Pod Level Resource handler"
I0329 14:59:16.519914       1 pluginhelpers.go:83] "extracted NUMA resources" logKey="node-1" cpu="10" memory="16 GiB"
I0329 14:59:16.519935       1 pluginhelpers.go:160] "pod handler NUMA resources" logKey="nodeB.test.net/node-0" cpu="14" memory="15 GiB"
I0329 14:59:16.519941       1 pluginhelpers.go:83] "extracted NUMA resources" logKey="node-0" cpu="2" memory="2.0 GiB"
I0329 14:59:16.519952       1 pluginhelpers.go:160] "pod handler NUMA resources" logKey="nodeB.test.net/node-1" cpu="10" memory="16 GiB"
I0329 14:59:16.519970       1 filter.go:149] "target resources" logKey="e2e-test-workload-placement-tmpol-4922/testpod" cpu="14" memory="16 GiB"
I0329 14:59:16.520005       1 filter.go:92] "feasible" logKey="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeB.test.net" NUMA=0 resource="cpu"
I0329 14:59:16.520030       1 filter.go:92] "feasible" logKey="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeB.test.net" NUMA=1 resource="memory"
I0329 14:59:16.519953       1 pluginhelpers.go:83] "extracted NUMA resources" logKey="node-1" cpu="14" memory="20 GiB"
I0329 14:59:16.520059       1 filter.go:96] "early verdict" logKey="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeB.test.net" resource="memory" suitable="false"
I0329 14:59:16.520070       1 pluginhelpers.go:160] "pod handler NUMA resources" logKey="nodeA.test.net/node-0" cpu="2" memory="2.0 GiB"
I0329 14:59:16.520081       1 pluginhelpers.go:160] "pod handler NUMA resources" logKey="nodeA.test.net/node-1" cpu="14" memory="20 GiB"
I0329 14:59:16.520090       1 filter.go:149] "target resources" logKey="e2e-test-workload-placement-tmpol-4922/testpod" cpu="14" memory="16 GiB"
I0329 14:59:16.520107       1 filter.go:92] "feasible" logKey="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeA.test.net" NUMA=1 resource="cpu"
I0329 14:59:16.520116       1 filter.go:92] "feasible" logKey="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeA.test.net" NUMA=1 resource="memory"
I0329 14:59:16.520126       1 filter.go:101] "final verdict" logKey="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeA.test.net" suitable=true
I0329 14:59:16.520397       1 default_binder.go:52] "Attempting to bind pod to node" pod="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeA.test.net"
I0329 14:59:16.536251       1 eventhandlers.go:166] "Delete event for unscheduled pod" pod="e2e-test-workload-placement-tmpol-4922/testpod"
I0329 14:59:16.536305       1 eventhandlers.go:186] "Add event for scheduled pod" pod="e2e-test-workload-placement-tmpol-4922/testpod"
I0329 14:59:16.536754       1 round_trippers.go:454] POST https://172.30.0.1:443/api/v1/namespaces/e2e-test-workload-placement-tmpol-4922/pods/testpod/binding 201 Created in 16 milliseconds
I0329 14:59:16.536907       1 cache.go:384] Finished binding for pod 943a6a56-3e83-44f3-94c7-94081404e3ed. Can be expired.
I0329 14:59:16.536994       1 scheduler.go:672] "Successfully bound pod to node" pod="e2e-test-workload-placement-tmpol-4922/testpod" node="nodeA.test.net" evaluatedNodes=5 feasibleNodes=1
```

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/scheduler-plugins
 go 1.17
 
 require (
+	github.com/dustin/go-humanize v1.0.0
 	github.com/google/go-cmp v0.5.5
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.0.12
 	github.com/patrickmn/go-cache v2.1.0+incompatible
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -41,10 +41,16 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
 	nodes := createNUMANodeList(zones)
 	qos := v1qos.GetPodQOS(pod)
 
+	// Node() != nil already verified in Filter(), which is the only public entry point
+	logNumaNodes("container handler NUMA resources", nodeInfo.Node().Name, nodes)
+
 	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
 	// and normal containers in the one scope
 	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
-		if !resourcesAvailableInAnyNUMANodes(nodes, container.Resources.Requests, qos, nodeInfo) {
+		logKey := fmt.Sprintf("%s/%s/%s", pod.Namespace, pod.Name, container.Name)
+		klog.V(6).InfoS("target resources", resourceListToLoggable(logKey, container.Resources.Requests)...)
+
+		if !resourcesAvailableInAnyNUMANodes(logKey, nodes, container.Resources.Requests, qos, nodeInfo) {
 			// definitely we can't align container, so we can't align a pod
 			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align container: %s", container.Name))
 		}
@@ -54,11 +60,14 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
 
 // resourcesAvailableInAnyNUMANodes checks for sufficient resource, this function
 // requires NUMANodeList with properly populated NUMANode, NUMAID should be in range 0-63
-func resourcesAvailableInAnyNUMANodes(numaNodes NUMANodeList, resources v1.ResourceList, qos v1.PodQOSClass, nodeInfo *framework.NodeInfo) bool {
+func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, resources v1.ResourceList, qos v1.PodQOSClass, nodeInfo *framework.NodeInfo) bool {
 	bitmask := bm.NewEmptyBitMask()
 	// set all bits, each bit is a NUMA node, if resources couldn't be aligned
 	// on the NUMA node, bit should be unset
 	bitmask.Fill()
+
+	// Node() != nil already verified in Filter(), which is the only public entry point
+	nodeName := nodeInfo.Node().Name
 
 	for resource, quantity := range resources {
 		// for each requested resource, calculate which NUMA slots are good fits, and then AND with the aggregated bitmask, IOW unset appropriate bit if we can't align resources, or set it
@@ -79,13 +88,17 @@ func resourcesAvailableInAnyNUMANodes(numaNodes NUMANodeList, resources v1.Resou
 			}
 
 			resourceBitmask.Add(numaNode.NUMAID)
+			klog.V(6).InfoS("feasible", "logKey", logKey, "node", nodeName, "NUMA", numaNode.NUMAID, "resource", resource)
 		}
 		bitmask.And(resourceBitmask)
 		if bitmask.IsEmpty() {
+			klog.V(5).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
 			return false
 		}
 	}
-	return !bitmask.IsEmpty()
+	ret := !bitmask.IsEmpty()
+	klog.V(5).InfoS("final verdict", "logKey", logKey, "node", nodeName, "suitable", ret)
+	return ret
 }
 
 func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, numaQuantity resource.Quantity) bool {
@@ -116,7 +129,14 @@ func singleNUMAPodLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nod
 
 	resources := util.GetPodEffectiveRequest(pod)
 
-	if !resourcesAvailableInAnyNUMANodes(createNUMANodeList(zones), resources, v1qos.GetPodQOS(pod), nodeInfo) {
+	logKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+	nodes := createNUMANodeList(zones)
+
+	// Node() != nil already verified in Filter(), which is the only public entry point
+	logNumaNodes("pod handler NUMA resources", nodeInfo.Node().Name, nodes)
+	klog.V(6).InfoS("target resources", resourceListToLoggable(logKey, resources)...)
+
+	if !resourcesAvailableInAnyNUMANodes(logKey, createNUMANodeList(zones), resources, v1qos.GetPodQOS(pod), nodeInfo) {
 		// definitely we can't align container, so we can't align a pod
 		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align pod: %s", pod.Name))
 	}
--- a/pkg/noderesourcetopology/pluginhelpers.go
+++ b/pkg/noderesourcetopology/pluginhelpers.go
@@ -19,6 +19,11 @@ package noderesourcetopology
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/dustin/go-humanize"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -75,6 +80,7 @@ func createNUMANodeList(zones topologyv1alpha1.ZoneList) NUMANodeList {
 				continue
 			}
 			resources := extractResources(zone)
+			klog.V(6).InfoS("extracted NUMA resources", resourceListToLoggable(zone.Name, resources)...)
 			nodes = append(nodes, NUMANode{NUMAID: numaID, Resources: resources})
 		}
 	}
@@ -139,10 +145,45 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
 func extractResources(zone topologyv1alpha1.Zone) v1.ResourceList {
 	res := make(v1.ResourceList)
 	for _, resInfo := range zone.Resources {
-		klog.V(5).InfoS("Extract resources for zone", "resName", resInfo.Name, "resAvailable", resInfo.Available)
 		res[v1.ResourceName(resInfo.Name)] = resInfo.Available
 	}
 	return res
+}
+
+func logNumaNodes(desc, nodeName string, nodes NUMANodeList) {
+	for _, numaNode := range nodes {
+		numaLogKey := fmt.Sprintf("%s/node-%d", nodeName, numaNode.NUMAID)
+		klog.V(6).InfoS(desc, resourceListToLoggable(numaLogKey, numaNode.Resources)...)
+	}
+}
+
+func resourceListToLoggable(logKey string, resources v1.ResourceList) []interface{} {
+	items := []interface{}{"logKey", logKey}
+
+	resNames := []string{}
+	for resName := range resources {
+		resNames = append(resNames, string(resName))
+	}
+	sort.Strings(resNames)
+
+	for _, resName := range resNames {
+		qty := resources[v1.ResourceName(resName)]
+		items = append(items, resName)
+		resVal, _ := qty.AsInt64()
+		if needsHumanization(resName) {
+			items = append(items, humanize.IBytes(uint64(resVal)))
+		} else {
+			items = append(items, strconv.FormatInt(resVal, 10))
+		}
+	}
+	return items
+}
+
+func needsHumanization(resName string) bool {
+	// memory-related resources may be expressed in KiB/Bytes, which makes
+	// for long numbers, harder to read and compare. To make it easier for
+	// the reader, we express them in a more compact form using go-humanize.
+	return resName == string(v1.ResourceMemory) || strings.HasPrefix(resName, v1.ResourceHugePagesPrefix)
 }
 
 func newPolicyHandlerMap() PolicyHandlerMap {
--- a/pkg/noderesourcetopology/pluginhelpers_test.go
+++ b/pkg/noderesourcetopology/pluginhelpers_test.go
@@ -1,0 +1,157 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package noderesourcetopology
+
+import (
+	"bytes"
+	"fmt"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestResourceListToLoggable(t *testing.T) {
+	tests := []struct {
+		name      string
+		logKey    string
+		resources corev1.ResourceList
+		expected  string
+	}{
+		{
+			name:      "empty",
+			logKey:    "",
+			resources: corev1.ResourceList{},
+			expected:  ` logKey=""`,
+		},
+		{
+			name:      "only logKey",
+			logKey:    "TEST1",
+			resources: corev1.ResourceList{},
+			expected:  ` logKey="TEST1"`,
+		},
+		{
+			name:   "only CPUs",
+			logKey: "TEST1",
+			resources: corev1.ResourceList{
+				corev1.ResourceCPU: resource.MustParse("16"),
+			},
+			expected: ` logKey="TEST1" cpu="16"`,
+		},
+		{
+			name:   "only Memory",
+			logKey: "TEST2",
+			resources: corev1.ResourceList{
+				corev1.ResourceMemory: resource.MustParse("16Gi"),
+			},
+			expected: ` logKey="TEST2" memory="16 GiB"`,
+		},
+		{
+			name:   "CPUs and Memory, no logKey",
+			logKey: "",
+			resources: corev1.ResourceList{
+				corev1.ResourceCPU:    resource.MustParse("24"),
+				corev1.ResourceMemory: resource.MustParse("16Gi"),
+			},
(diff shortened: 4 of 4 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/scheduler-plugins
 go 1.17
 
 require (
+	github.com/dustin/go-humanize v1.0.0
```
fromanirh, line 6: not this was already pulled in, because of the indirect dep to etcd:
```
$ go mod why github.com/dustin/go-humanize
warning: ignoring symlink /home/fromani/go/src/github.com/kubernetes-sigs/scheduler-plugins/manifests/install/charts/as-a-second-scheduler/templates/crds
# github.com/dustin/go-humanize
sigs.k8s.io/scheduler-plugins/cmd/controller/app
k8s.io/apiserver/pkg/server
k8s.io/apiserver/pkg/registry/generic
k8s.io/apiserver/pkg/storage/storagebackend/factory
k8s.io/apiserver/pkg/storage/storagebackend/factory.test
k8s.io/apiserver/pkg/storage/etcd3/testserver
go.etcd.io/etcd/server/v3/embed
go.etcd.io/etcd/server/v3/etcdserver
github.com/dustin/go-humanize
```

open to suggestions though.

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -139,12 +144,41 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
...
+		// avoid even the trivial range over nodes
+		return
+	}
+	for _, numaNode := range nodes {
+		numaIdent := fmt.Sprintf("node-%d", numaNode.NUMAID)
+		klog.V(6).InfoS(desc, resourceListToLoggable(numaIdent, numaNode.Resources)...)
+	}
+}
+
+func resourceListToLoggable(ident string, resources v1.ResourceList) []interface{} {
```
Tal-or, line 163: Can we have some unit-tests here?
It can also be useful getting some sense of the format this function return.

pkg/noderesourcetopology/filter.go
```diff
@@ -41,10 +41,16 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
 	nodes := createNUMANodeList(zones)
 	qos := v1qos.GetPodQOS(pod)
 
+	logNumaNodes("cnt handler NUMA resources", nodes)
```
Tal-or, line 44: This should be logged out to the user, so I would suggest not spare few letter and user ```container``` and not ```cnt```

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -139,12 +144,41 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
...
+		} else {
+			// fmt.Sprintf is an overkill, let's use something lighter
+			items = append(items, strconv.FormatInt(resVal, 10))
+		}
+	}
+	return items
+}
+
+func needsHumanization(resName string) bool {
+	return resName == "memory" || strings.HasPrefix(resName, v1.ResourceHugePagesPrefix)
```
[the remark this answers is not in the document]
fromanirh: yep, totally

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: yes, totally. I'll need to add examples and some tests coverage.

pkg/noderesourcetopology/filter.go
fromanirh: good point again. I used `cnt` because it is exactly long as `pod`, but a bit more verbosity helps here.

fromanirh:
I'm happy enough with this PR, so I'm lifting the WIP. The PR is now reviewable.

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -139,12 +145,56 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
...
+		idx++
+	}
+	sort.Strings(resNames)
+
+	itemIdx := 2
+	for _, resName := range resNames {
+		qty := resources[v1.ResourceName(resName)]
+		items[itemIdx] = resName
+		resVal, _ := qty.AsInt64()
+		if needsHumanization(resName) {
```
swatisehgal, line 183: Maybe capture the reason why we are checking for memory and hugepages and handling differently here as a comment here to explain that we expect that the CPUs are already in human readable form as they are expected to be integral.

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -139,12 +145,56 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
...
+		// avoid even the trivial range over nodes
+		return
+	}
+	for _, numaNode := range nodes {
+		numaIdent := fmt.Sprintf("node-%d", numaNode.NUMAID)
+		klog.V(6).InfoS(desc, resourceListToLoggable(numaIdent, numaNode.Resources)...)
+	}
+}
+
+func resourceListToLoggable(ident string, resources v1.ResourceList) []interface{} {
```
swatisehgal, line 164: Why the use of a slice of interface here? `map[string]string` seemed like an obvious choice to me here.

swatisehgal (commented):
I appreciate the effort to improve the scheduler plugin logging. Thanks for all the work.

This is definitely a huge improvement in the filter logging but I couldn't help but notice that we are extracting the resourcelist and storing it in a slice of interface.
I assumed that it would be better to represent the information we are deriving in a more appropriate data structure maybe as a map or a structure? We could also potentially associate a String() method to represent the info captured with instance directly while logging and DeepEqual for comparison.

WDYT?

pkg/noderesourcetopology/filter.go
```diff
@@ -41,10 +41,16 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
...
 	qos := v1qos.GetPodQOS(pod)
 
+	logNumaNodes("container handler NUMA resources", nodes)
+
 	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
 	// and normal containers in the one scope
 	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
-		if resMatchNUMANodes(nodes, container.Resources.Requests, qos, nodeInfo) {
+		// avoid fmt.Sprintf because the format is very simple and to squeeze a bit of speed
+		ident := pod.Namespace + "/" + pod.Name + "/" + container.Name
```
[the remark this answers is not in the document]
fromanirh: `ident` is meant to be the identifier of the object which is driving this flow. It is meant to enable correlate all the log lines pertaining to the same filtering. IOW, grepping the logs using the container name and/or the pod.name, we are supposed to get all the log lines emitted by the filtering.
Is more like a key, here's why si built as concatenation of namespace + pod name + container name (and similarly in the rest of the code).
I'd love to make this concept easier to understand but I'm not sure how. Perhaps renaming `ident` to `logKey` can help?

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: sure thing

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: Because doing like this enables us to write like
```
klog.V(6).InfoS("target resources", resourceListToLoggable(ident, resources)...)
```
which is both terse yet readable in the context (IMVHO) and cheap, because this way the function `resourceListToLoggable` is called (or that's the intention :) ) only when the logs need to be actually emitted, not unconditionally. Returning a map would mean we need to do some extra processing before to feed the data into `InfoS`

fromanirh:
> I appreciate the effort to improve the scheduler plugin logging. Thanks for all the work.
> 
> This is definitely a huge improvement in the filter logging but I couldn't help but notice that we are extracting the resourcelist and storing it in a slice of interface. I assumed that it would be better to represent the information we are deriving in a more appropriate data structure maybe as a map or a structure? We could also potentially associate a String() method to represent the info captured with instance directly while logging and DeepEqual for comparison.
> 
> WDYT?

Thanks Swati for the review! I think in this case we can use a more specialized representation tailored for the needs of structured logging. But by all means let's deep dive into this! I suggest us to keep this conversation nearer to the code, let's continue in the inline comments.

pkg/noderesourcetopology/filter.go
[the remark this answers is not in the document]
swatisehgal: Thanks for the explanation! `LogKey` is definitely a better term here. Thanks for making the updates.

pkg/noderesourcetopology/pluginhelpers.go
swatisehgal: Ack, this certainly helps my understanding. It looks like a trade off to me on whether we want to have a more structured data and process it before logging or have more flexibility at the time of logging and then tackle the extra processing in `kvListFormat()` when we perform the comparison. Either way, I don't feel too strongly about it and don't consider it a blocker, just want us sure that we are a conscious decision.

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -139,12 +145,59 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
...
 	return res
 }
 
+func logNumaNodes(desc string, nodes NUMANodeList) {
+	if !klog.V(6).Enabled() {
+		// avoid even the trivial range over nodes
+		return
+	}
+	for _, numaNode := range nodes {
+		numaIdent := fmt.Sprintf("node-%d", numaNode.NUMAID)
```
swatisehgal, line 159: NIT: might be better to keep things consistent and use `numaLogKey `as the variable name here.

fromanirh: right, fixed. Thanks!

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: In this case, it is indeed a conscious (albeit of course debatable) decision driven by the interface of `klog.InfoS`.
I'm looking for a tradeoff between readability in the logs and readability in the source code, leaning to the former when needed.
ATM I'm not aware of a clear use case for processing the return value of `resourceListToLoggable`, but if this emerges we can surely re-evaluate.

swatisehgal:
Thanks for making the updates.
/lgtm

/assign @Huang-Wei 
for approval

fromanirh:
rebased on top of master branch

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -139,12 +145,59 @@ func makePodByResourceListWithManyContainers(resources *v1.ResourceList, contain
...
+	}
+}
+
+func resourceListToLoggable(logKey string, resources v1.ResourceList) []interface{} {
+	// count pairs + logKey pair
+	items := make([]interface{}, (len(resources)+1)*2)
+	items[0] = "logKey"
+	items[1] = logKey
+
+	idx := 0
```
Huang-Wei, line 170: how about just get rid of the index thing:

```go
func resourceListToLoggable(logKey string, resources v1.ResourceList) []interface{} {
	// count pairs + logKey pair
	items := make([]interface{}, 0)
	items = append(items, "logKey")
	items = append(items, logKey)

	resNames := make([]string, 0)
	for resName := range resources {
		resNames = append(resNames, string(resName))
	}
	sort.Strings(resNames)

	for _, resName := range resNames {
		qty := resources[v1.ResourceName(resName)]
		items = append(items, resName)
		resVal, _ := qty.AsInt64()
		if needsHumanization(resName) {
			items = append(items, humanize.IBytes(uint64(resVal)))
		} else {
			// fmt.Sprintf is an overkill, let's use something lighter
			items = append(items, strconv.FormatInt(resVal, 10))
		}
	}
	return items
}
```

pkg/noderesourcetopology/filter.go
```diff
@@ -41,10 +41,17 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
...
 
+	// Node() != nil already verified in Filter(), which is the only public entry point
+	logNumaNodes("container handler NUMA resources", nodeInfo.Node().Name, nodes)
+
 	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
 	// and normal containers in the one scope
 	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
-		if !resourcesAvailableInAnyNUMANodes(nodes, container.Resources.Requests, qos, nodeInfo) {
+		// avoid fmt.Sprintf because the format is very simple and to squeeze a bit of speed
+		logKey := pod.Namespace + "/" + pod.Name + "/" + container.Name
```
Huang-Wei, line 51: nit: use `fmt.Sprintf`

pkg/noderesourcetopology/filter.go
```diff
@@ -79,13 +89,17 @@ func resourcesAvailableInAnyNUMANodes(numaNodes NUMANodeList, resources v1.Resou
 			}
 
 			resourceBitmask.Add(numaNode.NUMAID)
+			klog.V(6).InfoS("feasible", "logKey", logKey, "node", nodeName, "NUMA", numaNode.NUMAID, "resource", resource)
 		}
 		bitmask.And(resourceBitmask)
 		if bitmask.IsEmpty() {
+			klog.V(4).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
```
Huang-Wei, line 96: make it V(6)?

pkg/noderesourcetopology/filter.go
```diff
@@ -79,13 +89,17 @@ func resourcesAvailableInAnyNUMANodes(numaNodes NUMANodeList, resources v1.Resou
...
 			return false
 		}
 	}
-	return !bitmask.IsEmpty()
+	ret := !bitmask.IsEmpty()
+	klog.V(4).InfoS("final verdict", "logKey", logKey, "node", nodeName, "suitable", ret)
```
Huang-Wei, line 101: same, make it V(6)?

pkg/noderesourcetopology/filter.go
```diff
@@ -116,7 +130,15 @@ func singleNUMAPodLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nod
 
 	resources := util.GetPodEffectiveRequest(pod)
 
-	if !resourcesAvailableInAnyNUMANodes(createNUMANodeList(zones), resources, v1qos.GetPodQOS(pod), nodeInfo) {
+	// avoid fmt.Sprintf because the format is very simple and to squeeze a bit of speed
+	logKey := pod.Namespace + "/" + pod.Name
```
Huang-Wei, line 134: nit: use `fmt.Sprintf()`

pkg/noderesourcetopology/pluginhelpers_test.go
```diff
@@ -0,0 +1,157 @@
...
+		name      string
+		logKey    string
+		resources corev1.ResourceList
+		expected  string
+	}{
+		{
+			name:      "empty",
+			logKey:    "",
+			resources: corev1.ResourceList{},
+			expected:  " logKey=\"\"",
```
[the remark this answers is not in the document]
fromanirh: way nicer indeed! fixed, thanks!

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: this looks nicer indeed. I was doing some premature optimization towards speed, but I don't have any data to back this choice (hence premature), so let's start readable and let's optimize later where/if there is a clear need. Thanks!

pkg/noderesourcetopology/filter.go
fromanirh: Here I'd like to have progressively more information, hence I picked V(4).
I now realized I'm not sure this matches how the verbosity levels in `klog` are intended to work, I'll doublecheck what the docs suggest and what other code in k8s does

pkg/noderesourcetopology/filter.go
fromanirh: same, checking

fromanirh:
/hold
because pending unanswered comments about `V(4)` vs `V(6)`

pkg/noderesourcetopology/filter.go
Huang-Wei: The angle I suggested V(6) (or V(5)) is based on:

1. Most prod cluster set the verbose level to 3 or 4.
2. I totally understand you want to log this information; however, Filter would run against basically all nodes, so it will generate a high volume of logs in a large cluster. With that being said, maybe either make them both at verbose level 5; or log at level 4 on only one case whichever yields less volume of log entries.

fromanirh: The reason I'm inclined to keep this V(4) is that we expect to see this line, and the one right below, at most once per filter invovation, so it seems to me a good compromise. Would this address this concern or would it still be too spammy?

pkg/noderesourcetopology/filter.go
fromanirh: discussing in the above comment

pkg/noderesourcetopology/filter.go
Huang-Wei: > The reason I'm inclined to keep this V(4) is that we expect to see this line, and the one right below, at most once per filter invocation

Yes, per filter invocation, but against basically all nodes, which looks a bit spammy. How about V(5)?

fromanirh: I'm fine with `V(5)`. Thanks!

Huang-Wei:
/lgtm
/approve

Thanks @fromanirh !

fromanirh:
> /lgtm /approve
> 
> Thanks @fromanirh !

thank you for the review!

fromanirh:
/hold cancel
we reached agreement about the `V` level.

PR #359, opened 2022-03-23
noderesourcetopology: revamp the score logging

Tal-or:
This PR is inspired by https://github.com/kubernetes-sigs/scheduler-plugins/pull/354 (but not depend on it)

The goal is to expose more information about how the scoring descion  are made by the plugin.
It also try to make the complex data structure of the NRT object more easier to read in the log.
Finally, we change all logs messages to begin with small letter, to keep consistent convention with the filter logs.

Signed-off-by: Talor Itzhak <titzhak@redhat.com>

#### What type of PR is this?
/kind cleanup

Improve the logging in the noderesourcetopology score plugin,  aiming to better clarify the decision points in the code, and make the logs more readable.

N/A

N/A

### DIFF
--- a/pkg/noderesourcetopology/pluginhelpers.go
+++ b/pkg/noderesourcetopology/pluginhelpers.go
@@ -18,6 +18,7 @@ package noderesourcetopology
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"sort"
 	"strconv"
@@ -191,4 +192,18 @@ func newPolicyHandlerMap() PolicyHandlerMap {
 		topologyv1alpha1.SingleNUMANodePodLevel:       newPodScopedHandler(),
 		topologyv1alpha1.SingleNUMANodeContainerLevel: newContainerScopedHandler(),
 	}
+}
+
+func logNRT(desc string, nrtObj *topologyv1alpha1.NodeResourceTopology) {
+	if !klog.V(6).Enabled() {
+		// avoid the expensive marshal operation
+		return
+	}
+
+	ntrJson, err := json.MarshalIndent(nrtObj, "", " ")
+	if err != nil {
+		klog.V(6).ErrorS(err, "failed to marshal noderesourcetopology object")
+		return
+	}
+	klog.V(6).Info(desc, "noderesourcetopology", string(ntrJson))
 }
--- a/pkg/noderesourcetopology/score.go
+++ b/pkg/noderesourcetopology/score.go
@@ -56,7 +56,7 @@ func (rw resourceToWeightMap) weight(r v1.ResourceName) int64 {
 }
 
 func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState, pod *v1.Pod, nodeName string) (int64, *framework.Status) {
-	klog.V(5).InfoS("Scoring node", "nodeName", nodeName)
+	klog.V(6).InfoS("scoring node", "nodeName", nodeName)
 	// if it's a non-guaranteed pod, every node is considered to be a good fit
 	if v1qos.GetPodQOS(pod) != v1.PodQOSGuaranteed {
 		return framework.MaxNodeScore, nil
@@ -65,17 +65,18 @@ func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState,
 	nodeTopology := findNodeTopology(nodeName, tm.lister)
 
 	if nodeTopology == nil {
+		klog.V(5).InfoS("noderesourcetopology was not found for node", "node", nodeName)
 		return 0, nil
 	}
 
-	klog.V(5).InfoS("NodeTopology found", "nodeTopology", nodeTopology)
+	logNRT("noderesourcetopology found", nodeTopology)
 	for _, policyName := range nodeTopology.TopologyPolicies {
 		if handler, ok := tm.policyHandlers[topologyv1alpha1.TopologyManagerPolicy(policyName)]; ok {
 			// calculates the fraction of requested to capacity per each numa-node.
 			// return the numa-node with the minimal score as the node's total score
 			return handler.score(pod, nodeTopology.Zones, tm.scorerFn, tm.resourceToWeightMap)
 		} else {
-			klog.V(5).InfoS("Policy handler not found", "policy", policyName)
+			klog.V(4).InfoS("policy handler not found", "policy", policyName)
 		}
 	}
 	return 0, nil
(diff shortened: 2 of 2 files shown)


pkg/noderesourcetopology/score.go
```diff
@@ -65,17 +66,20 @@ func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState,
 	nodeTopology := findNodeTopology(nodeName, tm.lister)
 
 	if nodeTopology == nil {
+		klog.V(4).InfoS("noderesourcetopology was not found for node", "node", nodeName)
 		return 0, nil
 	}
 
-	klog.V(5).InfoS("NodeTopology found", "nodeTopology", nodeTopology)
+	if err := logNRT("noderesourcetopology found", nodeTopology); err != nil {
```
fromanirh, line 73: I think we should not fail because of error here. We probably should log the logging error (!!) and carry on

pkg/noderesourcetopology/score.go
```diff
@@ -135,7 +140,25 @@ func containerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList, scorerFn
...
+		// avoid the expensive marshal operation
+		return nil
+	}
+
+	ntrJson, err := json.MarshalIndent(nrtObj, "", " ")
+	if err != nil {
+		return err
 	}
-	return int64(stat.Mean(contScore, nil)), nil
+	klog.V(6).Info(desc, " ", "noderesourcetopology=", string(ntrJson))
```
fromanirh, line 162: nit: we can just write:
```json
klog.V(6).Info(desc, " noderesourcetopology=", string(ntrJson))``
```
I'm wondering if we should do MarshalIndent or we can just do Marshal. Logging in JSON format (and not just dumping the golang struct) is a good improvement. The compact JSON form, however, can be good enough because it's pretty fair to assume `jq` or any other JSON pretty printer is avaialble for the person checking the logs.
OTOH multi-line logs are unwieldly.

fromanirh (commented):
I think this is a nice improvement.

pkg/noderesourcetopology/score.go
Tal-or: > klog.V(6).Info(desc, " noderesourcetopology=", string(ntrJson))``

Definitely possible. I didn't do that, because I wanted the reviewers/readers to know that the space is explicit and not a typo.

> I'm wondering if we should do MarshalIndent or we can just do Marshal.

Just Marshling without indentation doesn't (literally!) look like a significant improvement over dumping the golang struct.

fromanirh: fair point about doing this way to better inform the reader. I have mixed feelings about Marshal vs MarshalIndent, I'll have a deeper look later. I'm not really against unsing MarshalIndent though

pkg/noderesourcetopology/score.go
```diff
@@ -135,7 +140,25 @@ func containerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList, scorerFn
...
+		identifier := pod.Namespace + "/" + pod.Name + "/" + container.Name
 		contScore[i] = float64(scoreForEachNUMANode(container.Resources.Requests, allocatablePerNUMA, scorerFn, resourceToWeightMap))
+		klog.V(6).InfoS("container scope scoring", "container", identifier, "score", contScore[i])
+	}
+	finalScore := int64(stat.Mean(contScore, nil))
+	klog.V(4).InfoS("container scope scoring final node score", "finalScore", finalScore)
+	return finalScore, nil
+}
+
+func logNRT(desc string, nrtObj *topologyv1alpha1.NodeResourceTopology) error {
```
swatisehgal, line 152: I think it would be better to move this helper function to pluginhelper.go in order to keep things consistent with the implementation in https://github.com/kubernetes-sigs/scheduler-plugins/pull/354.

swatisehgal (commented):
Thanks for the PR. I just left a comment with a minor suggestion. Other than that looks good.

pkg/noderesourcetopology/score.go
Tal-or: Done. Thank you Swati for the review!
I presume this change will create a conflict once #354 get merged. I'll keep an eye on it.

Tal-or:
@Huang-Wei Could you PTAL when you have some time?

pkg/noderesourcetopology/score.go
```diff
@@ -65,17 +65,20 @@ func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState,
 	nodeTopology := findNodeTopology(nodeName, tm.lister)
 
 	if nodeTopology == nil {
+		klog.V(4).InfoS("noderesourcetopology was not found for node", "node", nodeName)
```
Huang-Wei, line 68: maybe V(5).

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -192,3 +193,17 @@ func newPolicyHandlerMap() PolicyHandlerMap {
...
+	if !klog.V(6).Enabled() {
+		// avoid the expensive marshal operation
+		return nil
+	}
+
+	ntrJson, err := json.MarshalIndent(nrtObj, "", " ")
+	if err != nil {
+		return err
+	}
+	klog.V(6).Info(desc, " ", "noderesourcetopology=", string(ntrJson))
```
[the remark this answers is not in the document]
Tal-or: Done.

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -192,3 +193,17 @@ func newPolicyHandlerMap() PolicyHandlerMap {
...
+func logNRT(desc string, nrtObj *topologyv1alpha1.NodeResourceTopology) error {
+	if !klog.V(6).Enabled() {
+		// avoid the expensive marshal operation
+		return nil
+	}
+
+	ntrJson, err := json.MarshalIndent(nrtObj, "", " ")
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Tal-or: Sure. Fixed

pkg/noderesourcetopology/score.go
Tal-or: I thought that since it's unusual/unhealthy behavior it should be at a higher (or low depending on how you look at this) verbosity level, but ok I changed it.

fromanirh:
LGTM

Huang-Wei:
/lgtm
/approve

Thanks @Tal-or .

Issue #327, opened 2022-02-03
Migrate scheduler plugins to use controller-runtime


helayoty:
This is to track the effort to migrate all scheduler plugins to use `controller-runtime` aka `Kubebuilder` instead of the custom framework.

The main motivation is to enhance the ability of users to use and/or extend the project.

- [x] ~~Create new APIs~~ Migrate APIs to use `controller-runtime` (#340 )
- [ ] Migrate Capacity Scheduling
- [ ] Migrate Coscheduling
- [ ] Migrate Node Resources
- [ ] Migrate Node Resource Topology
- [ ] Migrate Preemption Toleration
- [ ] Migrate Trimaran 
- [ ] Migrate samples
  - [ ] Cross Node Preemption
  - [ ] Pod State
  - [ ] Quality of Service
- [ ] Update Documentaion 
- [ ] Remove old implementation

Huang-Wei:
@helayoty thanks for driving this. What do you mean by "Create new APIs"?

helayoty:
> @helayoty thanks for driving this. What do you mean by "Create new APIs"?

I'd like to have a separate PR first with all APIs created and all generated code using `kubebuilder` to make sure we agree on them. So I can go and use them in the plugins, wdyt?

Huang-Wei:
No matter the codegen tool is kubebuilder or vanilla codegen, isn't the generated content _almost_ the same? I don't get the point to create APIs - even they're just alias-like APIs.

helayoty:
> No matter the codegen tool is kubebuilder or vanilla codegen, isn't the generated content _almost_ the same? I don't get the point to create APIs - even they're just alias-like APIs.

It should be the exact same, just wanted to have the base point separate

Huang-Wei:
> It should be the exact same, just wanted to have the base point separate

Then the first PR can be merged directly to `master` - replace the raw codegen tags with kubebuilder ones, and update the codegen scripts. The expected result is that the current plugins function the same.

After that, you can start migrating the plugin impls.

Huang-Wei:
BTW: I'm going to cut a 0.22.X release and bump the master to vendor k8s 1.23 soon (within one week).

helayoty:
> > It should be the exact same, just wanted to have the base point separate
> 
> Then the first PR can be merged directly to `master` - replace the raw codegen tags with kubebuilder ones, and update the codegen scripts. The expected result is that the current plugins function the same.
> 
[...]

PR #360, opened 2022-03-28
bugfix about update occupied in podGroup status

NoicFank:
#### What type of PR is this?
/kind bug

currently, The updated information about occupied is edited locally rather than updated into apiserver. See the test `TestFillGroupStatusOccupied` for details.

Fixes #
NONE

```release-note
NONE
```

### DIFF
--- a/pkg/controller/podgroup_test.go
+++ b/pkg/controller/podgroup_test.go
@@ -3,6 +3,8 @@ package controller
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strings"
 	"testing"
 	"time"
 
@@ -25,7 +27,7 @@ func Test_Run(t *testing.T) {
 	cases := []struct {
 		name               string
 		pgName             string
-		minMember          uint32
+		minMember          int32
 		podNames           []string
 		podNextPhase       v1.PodPhase
 		podPhase           v1.PodPhase
@@ -125,7 +127,7 @@ func Test_Run(t *testing.T) {
 			podNames:          []string{"pod91", "pod92"},
 			podPhase:          v1.PodPending,
 			previousPhase:     v1alpha1.PodGroupPending,
-			desiredGroupPhase: v1alpha1.PodGroupPreScheduling,
+			desiredGroupPhase: v1alpha1.PodGroupPending,
 		},
 		{
 			name:              "Group status convert from running to pending",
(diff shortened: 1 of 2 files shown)


pkg/controller/podgroup_test.go
```diff
@@ -183,12 +183,98 @@ func Test_Run(t *testing.T) {
...
+			pgInformerFactory.Start(ctx.Done())
+			informerFactory.Start(ctx.Done())
+
+			go ctrl.Run(1, ctx.Done())
+			err := wait.Poll(200*time.Millisecond, 1*time.Second, func() (done bool, err error) {
+				pg, err := pgClient.SchedulingV1alpha1().PodGroups("default").Get(ctx, c.pgName, metav1.GetOptions{})
+				if err != nil {
+					return false, err
+				}
+				if pg.Status.OccupiedBy != c.desiredGroupOccupied {
```
denkensk, line 257: We can change `desiredGroupOccupied` to `[]string` and compare them by `cmp.Diff`

pkg/controller/podgroup_test.go
```diff
@@ -183,12 +183,98 @@ func Test_Run(t *testing.T) {
...
+				kubeClient = fake.NewSimpleClientset(ps[0], ps[1])
+			}
+			pg := makePG(c.pgName, c.minMember, c.groupPhase, nil)
+			pgClient := pgfake.NewSimpleClientset(pg)
+
+			informerFactory := informers.NewSharedInformerFactory(kubeClient, controller.NoResyncPeriodFunc())
+			pgInformerFactory := schedinformer.NewSharedInformerFactory(pgClient, controller.NoResyncPeriodFunc())
+			podInformer := informerFactory.Core().V1().Pods()
+			pgInformer := pgInformerFactory.Scheduling().V1alpha1().PodGroups()
+			ctrl := NewPodGroupController(kubeClient, pgInformer, podInformer, pgClient)
```
denkensk, line 246: This part of the code has a lot of duplication with another test.
It's better to define a separate function for `setup env for test`.

pkg/controller/podgroup_test.go
```diff
@@ -183,12 +183,98 @@ func Test_Run(t *testing.T) {
...
+	}
+}
+
+func makePods(podNames []string, pgName string, phase v1.PodPhase, reference []metav1.OwnerReference) []*v1.Pod {
 	pds := make([]*v1.Pod, 0)
 	for _, name := range podNames {
 		pod := st.MakePod().Namespace("default").Name(name).Obj()
 		pod.Labels = map[string]string{v1alpha1.PodGroupLabel: pgName}
 		pod.Status.Phase = phase
+		if reference != nil {
```
denkensk, line 275:
```suggestion
		if reference != nil && len(reference) != 0 {
```

pkg/controller/podgroup_test.go
```diff
@@ -183,12 +183,98 @@ func Test_Run(t *testing.T) {
...
+		pgName               string
+		minMember            int32
+		podNames             []string
+		podPhase             v1.PodPhase
+		podOwnerReference    []metav1.OwnerReference
+		groupPhase           v1alpha1.PodGroupPhase
+		desiredGroupOccupied string
+	}{
+		{
+			name:      "fill the desired Occupied",
```
denkensk, line 199: fill the Occupied of PodGroup with a single ownerReference?

pkg/controller/podgroup_test.go
```diff
@@ -183,12 +183,98 @@ func Test_Run(t *testing.T) {
...
+			podOwnerReference: []metav1.OwnerReference{
+				{
+					Name: "new-occupied",
+				},
+			},
+			groupPhase:           v1alpha1.PodGroupPending,
+			desiredGroupOccupied: "default/new-occupied",
+		},
+		{
+			name:      "fill the desired Occupied, with multi ownerReference",
```
denkensk, line 213: fill the Occupied of PodGroup with multi ownerReferences ?

denkensk (commented):
Thanks @NoicFank 
Some comments for you.

pkg/controller/podgroup_test.go
NoicFank: the type of `pg.Status.OccupiedBy` is `string`, thus `cmp.Diff` works? as we change `desiredGroupOccupied ` to `[]string `.

NoicFank:
@denkensk thanks for reviews. its updated, PTAL.

pkg/controller/podgroup_test.go
Huang-Wei: > It's better to define a separate function for setup env for test.

We can leave it to the kubebuilder migration. (@helayoty is working on using kubebuilder and envtest to migrate existing controller-related code).

pkg/controller/podgroup_test.go
denkensk: O.. I forget the OccupiedBy is string....

pkg/controller/podgroup_test.go
denkensk: > We can leave it to the kubebuilder migration. 
Sure

denkensk:
/lgtm
/approve

Thanks. @NoicFank

pkg/controller/podgroup_test.go
helayoty: tagging the kubebuilder issue here so I remember to handle that #327

PR #367, opened 2022-04-15
Grant access to configmaps

wangchen615:
Trimaran plugins may use configmaps to authenticate Prometheus queries in OpenShift, for example, using [routerCAConfigMap](https://github.com/paypal/load-watcher/blob/329422106426c9fa4f797950ebc9d83617e4b191/pkg/watcher/internal/metricsprovider/prometheus.go#L103)

So the configmap access is added.

Fixes #


Huang-Wei:
@wangchen615 [sched-plugins::extension-apiserver-authentication-reader](https://github.com/kubernetes-sigs/scheduler-plugins/blob/1c90b0d319f33c7a33d9456ba412be0b6ffb61bc/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml#L118) is supposed to associate the [scheduler SA](https://github.com/kubernetes-sigs/scheduler-plugins/blob/1c90b0d319f33c7a33d9456ba412be0b6ffb61bc/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml#L126) to a standard Kubernetes offering's [extension-apiserver-authentication-reader](https://github.com/kubernetes-sigs/scheduler-plugins/blob/1c90b0d319f33c7a33d9456ba412be0b6ffb61bc/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml#L123) role - which should have another set of RBAC reading privileges like accessing ConfigMap. Is that something specific in OpenShift like `extension-apiserver-authentication-reader` doesn't exist?

Huang-Wei:
Close due to no response for https://github.com/kubernetes-sigs/scheduler-plugins/pull/367#issuecomment-1100571940.

PR #370, opened 2022-04-26
Return framework.UnschedulableAndUnresolvable in coscheduling#PreFilter

### LINKED ISSUE #300: coscheduling: prefilter should return `UnschedulableAndUnresolvable` status when pod group is denied
https://github.com/kubernetes-sigs/scheduler-plugins/blob/455cb8a6536e8bf8b1d856812851c545255295b4/pkg/coscheduling/coscheduling.go#L145

When pod group was recently denied (`pgMgr.lastDenied`), pre-filter should return `UnschedulableAndUnresolvable` code (instead of `Unschedulable`) in order to prevent preemption. 

Similar when number of pods < minAvailable then pre-filter should also return `UnschedulableAndUnresolvable`. In both cases preemption does not make sense.

Huang-Wei:
#### What type of PR is this?

/kind cleanup

Return framework.UnschedulableAndUnresolvable in coscheduling#PreFilter, to avoid unnecessary preemption attempts as that doesn't help.

Fixes #300


Huang-Wei:
cc @denkensk @mateuszlitwin

denkensk:
/lgtm
/approve

Thanks. @Huang-Wei

PR #371, opened 2022-05-01
noderesourcetopology: fix filter plugin resource alignment when scope=container

### LINKED ISSUE #366: noderesourcetopology: wrong accounting of the filter plugin when scope=container
**Problem Description:**

When **scope=container** and the scheduled pod has **multiple containers**, the noderesourcetopology filter plugin does the accounting in a wrong way.
The plugin treats each container inside the pod individually, and check if it's fit into the (NUMA) node without taking into consideration the resources which already allocated by the previous container's requests.

**Example:**

 A guaranteed pod:
```yaml
Pod:
  Scheduler-Name: topo-aware-scheduler
  Container A:
    CPUs: 2
    Memory: 100M
  Container B:
    CPUs: 2
    Memory: 100M
```
A node candidate for scheduling: 
```yaml
Node:
 NUMA Zone 1:
  Available:
   CPUs: 2
   Memory: 100M
 NUMA Zone 2: 
  Available:
   CPUs: 1
   Memory: 50M
```
The filter plugin iterate over the containers in the pod:
It finds that ```NUMA Zone 1``` is a good fit for ```Container A```.
Next it looks at ```Container B```'s resource request and finds that it fits for ```NUMA Zone 1``` as well and then mark the Node as a possible candidate. 
This is wrong because the available resources under ```NUMA Zone 1``` are **already taken** by ```Container A``` and cannot suffice ```Container B```'s resource request too.

Tal-or:
/kind bug

The noderesourcetopology filter plugin is used to handle resource accounting in the wrong way when topology manager scope = container.

This PR intended to solve it by reimplementing the filtering logic in such a way that it will take the previous container's requested resources into account.  

Fixes #366


pkg/noderesourcetopology/filter.go
```diff
@@ -32,6 +32,8 @@ import (
 	"sigs.k8s.io/scheduler-plugins/pkg/util"
 )
 
+const highestNumaId = 63
```
fromanirh, line 35: nit: capitalize `Id` to `ID`. Should probably be `highestNUMAID`

pkg/noderesourcetopology/filter.go
```diff
@@ -44,23 +46,46 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
 	// Node() != nil already verified in Filter(), which is the only public entry point
 	logNumaNodes("container handler NUMA resources", nodeInfo.Node().Name, nodes)
 
-	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
-	// and normal containers in the one scope
-	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
+	// the init containers are running SERIALLY and BEFORE the normal containers
```
fromanirh, line 49: add a reference (link) to https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#understanding-init-containers

pkg/noderesourcetopology/filter.go
```diff
@@ -44,23 +46,46 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
...
+		_, match := resourcesAvailableInAnyNUMANodes(logKey, nodes, initContainer.Resources.Requests, qos, nodeInfo)
+		if !match {
+			// we can't align init container, so definitely we can't align a pod
+			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align init container: %s", initContainer.Name))
+		}
+	}
+
+	// this flow is changing the values of zones for its internal accounting.
+	// hence, create a pristine copy for this purpose and keep the original intact
+	mutableZones := zones.DeepCopy()
```
fromanirh, line 64: I was wondering that maybe extending `NUMANode` to track the currently available resources can lead to nicer code, and most imporant can save us to call `createNUMANodeList` for each container
```
type NUMANode struct {
	NUMAID    int
	Resources v1.ResourceList
        Available v1.ResourceList
}
```

pkg/noderesourcetopology/filter.go
```diff
@@ -89,16 +114,26 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 
 			resourceBitmask.Add(numaNode.NUMAID)
 			klog.V(6).InfoS("feasible", "logKey", logKey, "node", nodeName, "NUMA", numaNode.NUMAID, "resource", resource)
+
+			// according to TopologyManager, the preferred NUMA affinity, is the narrowest one.
+			// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/policy.go#L155-L157
```
fromanirh, line 119: let's use a link to a tag, so the link won't get stale if the code is refactored: https://github.com/kubernetes/kubernetes/blob/v1.24.0-rc.1/pkg/kubelet/cm/topologymanager/policy.go#L155

pkg/noderesourcetopology/filter_test.go
```diff
@@ -141,6 +141,30 @@ func TestNodeResourceTopology(t *testing.T) {
 				},
 			},
 		},
+		&topologyv1alpha1.NodeResourceTopology{
```
fromanirh, line 144: let's elaborate in the commit message the cases we want to cover. `scope=container` is a more complex scenario wrt to `scope=pod` (the latter is basically like the former with len(containers)=1) so we can make sure we have good coverage

pkg/noderesourcetopology/filter_test.go
```diff
@@ -598,10 +790,40 @@ func makeMultiContainerPodWithResourceList(name string, resourcesList []v1.Resou
...
 	return res
 }
+
+// makeResList intention is to decrease the amount of repetitive code and improve test writing speed
+// it gets a variable amount of strings, where each represent a quantity of different resource in the following order
+// [0] = v1.ResourceCPU
+// [1] = v1.ResourceMemory
+// [2] =  hugepages2Mi
+// [3] = nicResourceName
+func makeResList(quan ...string) v1.ResourceList {
```
fromanirh, line 815: the intent here is good, but this helper now makes more hidden (not more explicit) which resources has which quantity. In the end it seems to be saving this
```
resourceName = resource.MustParse(q)
```
which doesn't look like a big improvement.
Additionally, having the tests very explicit and all the context laid down explicitely, having at the cost of some repetition, is IMO a very good thing in the test flow.

test/integration/noderesourcetopology_test.go
```diff
@@ -555,6 +555,111 @@ func TestTopologyMatchPlugin(t *testing.T) {
 			},
 			expectedNodes: []string{"fake-node-1", "fake-node-2"},
 		},
+		{
+			name: "SingleNUMANodePodLevel: Filtering out nodes that cannot fit resources in case of Guaranteed pod with multi containers",
```
fromanirh, line 563: same for the unit test: let's enumerate the flows and let's check which of them we want to cover

test/integration/noderesourcetopology_test.go
```diff
@@ -52,6 +52,10 @@ const (
 	memory = string(v1.ResourceMemory)
 )
 
+const (
```
fromanirh, line 55: the commit message says `negative unit-tests` but we are changing the integration tests - let's make it consistent

fromanirh (commented):
initial review. The main flow looks good. I need to spend more time reviewing the finer details of the TM behaviour and the test cases

pkg/noderesourcetopology/filter.go
```diff
@@ -89,16 +114,26 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
...
 		if bitmask.IsEmpty() {
 			klog.V(5).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
-			return false
+			return numaID, false
 		}
 	}
-	ret := !bitmask.IsEmpty()
+
+	// at least one NUMA node is available
+	ret := bitmask.Count() > 0
```
Huang-Wei, line 134: And difference with `bitmask.IsEmpty()`? If no, seems `bitmask.IsEmpty()` is more efficient.

pkg/noderesourcetopology/filter.go
```diff
@@ -44,23 +46,46 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
...
-	// We count here in the way TopologyManager is doing it, IOW we put InitContainers
-	// and normal containers in the one scope
-	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
+	// the init containers are running SERIALLY and BEFORE the normal containers
+	// therefore, we don't need to accumulate their resources together
+	for _, initContainer := range pod.Spec.InitContainers {
+		logKey := fmt.Sprintf("%s/%s/%s", pod.Namespace, pod.Name, initContainer.Name)
+		klog.V(6).InfoS("target resources", resourceListToLoggable(logKey, initContainer.Resources.Requests)...)
+
+		_, match := resourcesAvailableInAnyNUMANodes(logKey, nodes, initContainer.Resources.Requests, qos, nodeInfo)
```
Huang-Wei, line 55: Are you going to implement an optimal [container placement strategy](https://github.com/kubernetes-sigs/scheduler-plugins/issues/366#issuecomment-1098617120) in this PR?

pkg/noderesourcetopology/filter_test.go
Tal-or: I was trying to have more "matrix-like" tests to ease the test writing process, so all the developer needs to do is change some numbers, and avoid copy-paste big chunks of code which sometimes might lead to mistakes.

fromanirh: the intent is good but I don't think this specific implementation is on target

pkg/noderesourcetopology/filter.go
Tal-or: Thank you for the quick review.
The thing is that even if we'll implement an optimal placement strategy as you suggested, it is eventually for Kubelet to decide which container is about to be placed on which NUMA. 

Kubelet doesn't have this optimal placement strategy, so in the case above it will fail to schedule the pod, regardless of what the filter plugin thinks or do.

Hence, IMHO we should imitate Kubelet's behavior as much as possible in order to predict the future placements nearly as possible to reality.

pkg/noderesourcetopology/filter.go
Tal-or: I see the point indeed. But its might be a bit confusing since the ```NUMANode.Resources``` value is already taking the ```Available``` resources from the ```Zone```: https://github.com/kubernetes-sigs/scheduler-plugins/blob/7a6dcdc99b1ee9a324823eaf98718cfd9e98e805/pkg/noderesourcetopology/pluginhelpers.go#L149

I think it will be better subtracting the resources from the ```NUMANode``` object directly instead from the ```Zone``` and then recreating the ```NUMANode```

fromanirh: good point. let's postpone this work as further optimization.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -309,6 +333,15 @@ func TestNodeResourceTopology(t *testing.T) {
 			node:       nodes[3],
 			wantStatus: framework.NewStatus(framework.Unschedulable, "cannot align pod: "),
 		},
+		{
+			name: "Guaranteed QoS Topology Container Scope, pod fit",
```
swatisehgal, line 337: This is corresponding to a test case where a pod _doesn't_ fit. We should update the description as follows:
```suggestion
			name: "Guaranteed QoS Topology Container Scope, pod doesn't fit",
```

pkg/noderesourcetopology/filter_test.go
```diff
@@ -553,6 +586,165 @@ func TestNodeResourceTopologyMultiContainerPodScope(t *testing.T) {
...
+	tests := []struct {
+		name       string
+		pod        *v1.Pod
+		node       *v1.Node
+		nrts       []*topologyv1alpha1.NodeResourceTopology
+		avail      []resourceDescriptor
+		wantStatus *framework.Status
+	}{
+		{
+			name: "gu pod, overallocation",
```
swatisehgal, line 633: We should make this test case name more descriptive to maybe something like below:
```suggestion
			name: "gu pod, resources, cannot fit pod (with 3 containers) as not enough resources available on the node",
```

swatisehgal (commented):
Great work on the PR Talor! Left a few comments but overall looks like we are heading in the right direction.

swatisehgal:
Fixing the typo in the title.

/retitle nodresourcetopology: fix filter plugin resource alignment when scope=container

pkg/noderesourcetopology/filter.go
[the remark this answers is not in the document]
fromanirh: +1 (but let's do on a separate PR). This is also a publicly acknowledged limitation, so it's a safe change

pkg/noderesourcetopology/filter_test.go
Tal-or: yes. this is a mistake, thanks!

pkg/noderesourcetopology/filter_test.go
Tal-or: Revert it. Done

pkg/noderesourcetopology/filter.go
Tal-or: The change wasn't big so I applied it already.

pkg/noderesourcetopology/filter.go
Tal-or: Done.

pkg/noderesourcetopology/filter.go
Tal-or: Done.

pkg/noderesourcetopology/filter.go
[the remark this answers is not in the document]
Tal-or: Maybe we should change this number to 8 and just refer to the documentation?

Tal-or:
/hold
I think more tests should be added before moving forward.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -586,6 +586,31 @@ func TestNodeResourceTopologyMultiContainerPodScope(t *testing.T) {
 	}
 }
 
+const (
```
fromanirh, line 589: reading the code down below it seems we save little, coming from
```
state: "Pending"
```
to
```
state: pending
```

pkg/noderesourcetopology/filter_test.go
```diff
@@ -586,6 +586,31 @@ func TestNodeResourceTopologyMultiContainerPodScope(t *testing.T) {
...
+	running = "Running"
+	pending = "Pending"
+)
+
+// should be filled by the user
+type testUserEntry struct {
+	// description contains the test type and tier as described in TEST.md
+	// and a short description of the test itself
+	description  string
+	initCntCount int
```
fromanirh, line 599: can we infer safely the count from `len(initCntReq)`? 
IOW, when do we have `initCntCount != len(initCntReq)`?
same question for app containers

pkg/noderesourcetopology/filter_test.go
```diff
@@ -621,159 +646,232 @@ func TestNodeResourceTopologyMultiContainerContainerScope(t *testing.T) {
...
+			initCntCount: 3,
+			initCntReq:   []map[string]string{{cpu: "20", memory: "50G"}, {cpu: "40", memory: "50G"}, {cpu: "3", memory: "50G"}},
+			cntCount:     2,
+			cntReq:       []map[string]string{{cpu: "20", memory: "40G"}, {cpu: "2", memory: "6G"}},
+			state:        pending,
+			statusErr:    "cannot align init container: cnt-2",
 		},
 	}
 
+	tests := parseTestUserEntry(tue)
```
fromanirh, line 873: this turned out quite nice!

pkg/noderesourcetopology/filter_test.go
Tal-or: > can we infer safely the count from len(initCntReq)?

We do, but think of it from an eye of a user, in case you want to know how many init containers/containers this specific test is going to run. 

If you don't have this number in front of your eyes, you'll need to count how many resource objects you have in each of ```initCntReq``` or ```cntReq``` struct, which is quite confusing and inconvenient.

pkg/noderesourcetopology/filter_test.go
Tal-or: It provides 2 small (but nice, consider the effort) benefits:
1. In case of future name changes - change once
2. You are getting self-completion on modern IDEs, which make it faster and help you avoid typoes.

pkg/noderesourcetopology/filter_test.go
Tal-or: Thanks!

pkg/noderesourcetopology/filter_test.go
fromanirh: I don't have strong feelings here so let's keep it

pkg/noderesourcetopology/filter_test.go
fromanirh: OTOH we are setting the same info twice, which is also annoying. I think the inconvenience of counting can be greatly mitigated with better code formatting, making obvious at glance how many container the test is using without having it to tell it explicitely (again)

pkg/noderesourcetopology/filter_test.go
fromanirh: If we states we care about are only 2 (and this seems the case) I wonder if we can abstract even further and just use a simple `running bool` and infer the pending state automatically

pkg/noderesourcetopology/filter_test.go
Tal-or: Done

Tal-or:
/hold cancel
I'm happy with the PR now. removing hold

pkg/noderesourcetopology/filter.go
```diff
@@ -181,3 +207,22 @@ func resourceFoundOnNode(resName v1.ResourceName, wantQuantity resource.Quantity
 	}
 	return false
 }
+
+func allocateFromNUMA(nodes NUMANodeList, nodeID int, container v1.Container) {
```
lianghao208, line 211: Does `numaID` better than `nodeID`?

Tal-or: Thank you for the review.
I wanted to keep consistency with the first argument which is ```nodes``` but I presume ```numaID``` is clearer indeed.

pkg/noderesourcetopology/filter.go
```diff
@@ -181,3 +207,22 @@ func resourceFoundOnNode(resName v1.ResourceName, wantQuantity resource.Quantity
 	}
 	return false
 }
+
+func allocateFromNUMA(nodes NUMANodeList, numaID int, container v1.Container) {
+	var nRes v1.ResourceList
+	// find the correct NUMA ID's resources
```
fromanirh, line 213: we have
```
type NUMANodeList []NUMANode
```
can we fix (perhaps in a later PR?) createNUMANodeList to guarantee that the `slice index == numaID`, so we can always safely write
```
nRes = nodes[numaID].Resources
```

pkg/noderesourcetopology/filter.go
```diff
@@ -181,3 +207,22 @@ func resourceFoundOnNode(resName v1.ResourceName, wantQuantity resource.Quantity
...
+		if nodes[i].NUMAID == numaID {
+			nRes = nodes[i].Resources
+			break
+		}
+	}
+
+	cReq := container.Resources.Requests
+	for resName, quan := range cReq {
+		nodeResQuan := nRes[resName]
+		// subtract the resource from the available ones under the NUMA node
```
fromanirh, line 224: this is pretty obvious from the line below, isn't it? :)

test/integration/noderesourcetopology_test.go
```diff
@@ -305,7 +305,7 @@ func TestTopologyMatchPlugin(t *testing.T) {
 		{
 			name: "Scheduling of a burstable pod requesting only memory",
 			pods: []*v1.Pod{
-				withContainer(st.MakePod().Namespace(ns).Name("topology-aware-scheduler-pod").Req(map[v1.ResourceName]string{v1.ResourceMemory: "5Gi"}).Obj(), pause),
+				st.MakePod().Namespace(ns).Name("topology-aware-scheduler-pod").Req(map[v1.ResourceName]string{v1.ResourceMemory: "5Gi"}).Obj(),
```
fromanirh, line 312: why is `withContainer` no longer needed?

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,78 @@
...
+| 30        | N > 1           | SUM(R) > Node && R < NUMA        | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 32        | N > 1           | SUM(R) > Node && R < NUMA        | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
+| 33        | N > 1           | SUM(R) > Node && R < NUMA        | N > 1      | NUMA < R < Node                  | Pending |
+| 34        | N > 1           | NUMA < R < Node                  | NVM        | NVM                              | Pending |
+
+### Test Tiers
+
+#### Tier1
+Contains the most fundamental and critical tests.
+Init containers with multi/single-containers placement against multiple NUMAs.
```
fromanirh, line 69: Init and app containers or only init containers? I guess the former, right?

pkg/noderesourcetopology/filter_test.go
```diff
@@ -586,6 +586,23 @@ func TestNodeResourceTopologyMultiContainerPodScope(t *testing.T) {
...
 }
 
+// should be filled by the user
+type testUserEntry struct {
+	// description contains the test type and tier as described in TEST.md
+	// and a short description of the test itself
+	description string
+	initCntReq  []map[string]string
+	cntReq      []map[string]string
+	statusErr   string
```
fromanirh, line 596: I think it should be obvious and explicit the nodes and the NRTs being used here

fromanirh (commented):
partial review. The foundations are good, and the work describing the tests is amazing. IMHO, The test implementations themselves could be improved further for clarity and to make sure (and obvious) there are no regressions wrt the existing testsuite.
I'll review again shortly.

test/integration/noderesourcetopology_test.go
Tal-or: ```withContainer``` is misleading because ```Req``` is already creating a container with given resources.
It is also redundant because ```withContainer```  was setting the image and container name, but it does that with the exact values that ```Req``` already provided.

pkg/noderesourcetopology/TESTS.md
Tal-or: yes I'll mention that explicitly

pkg/noderesourcetopology/filter_test.go
Tal-or: We're always using the same nodes and NRTs object, so why would we need them as variables?

fromanirh: because making that implicit makes harder to understand at glance what the test is supposed to do. One has to lookup the implicit context instead of having it clearly spelled out.

test/integration/noderesourcetopology_test.go
fromanirh: thanks, this fully make sense.

pkg/noderesourcetopology/filter.go
Tal-or: Yes. That can be a good idea, but let's do that on later PR as you mentioned.

Tal-or:
@Huang-Wei I would like to add more integration tests, but I think it would be better to split that into a separate PR since this one is big enough already. 
Please let me know what you think

Huang-Wei:
> @Huang-Wei I would like to add more integration tests, but I think it would be better to split that into a separate PR since this one is big enough already.

I'm fine with that as long as you're going to raise the integration test shortly. I will wait for that one and cut a release.

pkg/noderesourcetopology/filter.go
```diff
@@ -44,23 +46,41 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
...
-			// definitely we can't align container, so we can't align a pod
+		numaID, match := resourcesAvailableInAnyNUMANodes(logKey, nodes, container.Resources.Requests, qos, nodeInfo)
+		if !match {
+			// we can't align container, so definitely we can't align a pod
 			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align container: %s", container.Name))
 		}
+
+		// subtract the resources requested by the container from the given NUMA.
+		// this is necessary, so we won't allocate the same resources for the upcoming containers
+		allocateFromNUMA(nodes, numaID, container)
```
Huang-Wei, line 75: maybe rename to `subtractFromNUMA`?

pkg/noderesourcetopology/filter.go
```diff
@@ -181,3 +207,21 @@ func resourceFoundOnNode(resName v1.ResourceName, wantQuantity resource.Quantity
...
+			break
+		}
+	}
+
+	cReq := container.Resources.Requests
+	for resName, quan := range cReq {
+		nodeResQuan := nRes[resName]
+		nodeResQuan.Sub(quan)
+		nRes[resName] = nodeResQuan
+	}
```
Huang-Wei, line 226: This part can be moved inside the above foo loop:

```go
// Find the correct NUMA ID's resources and subtract from `nodes`.
func subtractFromNUMA(nodes NUMANodeList, numaID int, container v1.Container) {
	for i := 0; i < len(nodes); i++ {
		if nodes[i].NUMAID != numaID {
			continue
		}

		nRes := nodes[i].Resources
		for resName, quan := range container.Resources.Requests {
			nodeResQuan := nRes[resName]
			nodeResQuan.Sub(quan)
			nRes[resName] = nodeResQuan
		}
	}
}
```

pkg/noderesourcetopology/filter.go
Tal-or: Sure.

pkg/noderesourcetopology/filter.go
Tal-or: I don't have a strong preference here, so okay.

Tal-or:
> > @Huang-Wei I would like to add more integration tests, but I think it would be better to split that into a separate PR since this one is big enough already.
> 
> I'm fine with that as long as you're going to raise the integration test shortly. I will wait for that one and cut a release.

PR is ready: https://github.com/kubernetes-sigs/scheduler-plugins/pull/377

fromanirh (commented):
suggested reordering:
```
(HEAD) nrt: refactor and add extra unit-tests
nrt: add unit-tests
nrt: document tests
nrt: fix filter plugin resource alignment when scope=container
```
and move the integration test to the other existing PR, possibly merging commits

pkg/noderesourcetopology/filter_test.go
```diff
@@ -621,159 +639,155 @@ func TestNodeResourceTopologyMultiContainerContainerScope(t *testing.T) {
...
+			statusErr:   "cannot align init container: cnt-1",
+		},
+		{
+			description: "[7][tier1] init container with memory over allocation, multi-containers with good allocation - not fit",
+			initCntReq:  []map[string]string{{cpu: "4", memory: "70G"}},
+			cntReq:      []map[string]string{{cpu: "1", memory: "4G"}, {cpu: "1", memory: "4G"}},
+			statusErr:   "cannot align init container: cnt-1",
+		},
+		{
+			description: "[7][tier1] init container with memory over allocation, multi-containers with good allocation - not fit",
```
fromanirh, line 703: this looks the same as the previous one. is it intentional?

pkg/noderesourcetopology/filter_test.go
```diff
@@ -621,159 +639,155 @@ func TestNodeResourceTopologyMultiContainerContainerScope(t *testing.T) {
...
+			cntReq:      []map[string]string{{cpu: "35", memory: "40G"}},
+			statusErr:   "cannot align container: cnt-1",
+		},
+		{
+			description: "[28][tier1] multi init container with good allocation, multi-containers with good allocation - fit",
+			initCntReq:  []map[string]string{{cpu: "30", memory: "10G"}, {cpu: "30", memory: "10G"}},
+			cntReq:      []map[string]string{{cpu: "20", memory: "40G"}, {cpu: "20", memory: "40G"}},
+		},
+		{
+			description: "[29][tier1] multi init container with over cpu allocation in total, multi-containers with good allocation - fit",
```
fromanirh, line 766: could you please clarify what "in total" means here?

pkg/noderesourcetopology/filter_test.go
```diff
@@ -621,159 +639,155 @@ func TestNodeResourceTopologyMultiContainerContainerScope(t *testing.T) {
...
-				},
-			),
-			node: nodes[0],
-			nrts: []*topologyv1alpha1.NodeResourceTopology{
-				nodeTopologies[0],
-			},
-			avail:      []resourceDescriptor{},
-			wantStatus: nil,
+			description: "[4][tier2] multi-containers with good allocation, spread across NUMAs - fit",
+			cntReq:      []map[string]string{{cpu: "20", memory: "40G"}, {cpu: "20", memory: "40G"}},
```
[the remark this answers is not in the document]
fromanirh: full example:
```
		{
			description: "[29][tier1] multi init container with over cpu allocation in total, multi-containers with good allocation - fit",
			initCntReq: []map[string]string{
				{cpu: "20", memory: "10G"},
				{cpu: "10", memory: "10G"},
				{cpu: "30", memory: "10G"}},
			cntReq: []map[string]string{
				{cpu: "2", memory: "6G"},
				{cpu: "20", memory: "40G"},
				{cpu: "20", memory: "40G"},
			},
		},
```

pkg/noderesourcetopology/filter_test.go
```diff
@@ -424,8 +424,8 @@ func TestNodeResourceTopologyMultiContainerPodScope(t *testing.T) {
 	}{
 		{
 			name: "gu pod fits only on a numa node",
-			pod: makeMultiContainerPodWithResourceList("testpod",
-				[]v1.ResourceList{
+			pod: makePod("testpod",
+				withMultiContainers([]v1.ResourceList{
```
swatisehgal, line 428: This commit includes refactoring and addition of test cases, I would split this commit into two to separate those out for better clarity.

swatisehgal (commented):
Still have to review the last two commits in depth (which I will get to soon) but wanted to share my review comments so far.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -309,6 +333,15 @@ func TestNodeResourceTopology(t *testing.T) {
...
 		},
+		{
+			name: "Guaranteed QoS Topology Container Scope, pod doesn't fit",
+			pod: makePodByResourceListWithManyContainers(&v1.ResourceList{
+				v1.ResourceCPU:             *resource.NewQuantity(3, resource.DecimalSI),
+				v1.ResourceMemory:          resource.MustParse("1Gi"),
+				notExistingNICResourceName: *resource.NewQuantity(0, resource.DecimalSI)}, 3),
+			node:       nodes[4],
+			wantStatus: framework.NewStatus(framework.Unschedulable, "cannot align container: "),
+		},
```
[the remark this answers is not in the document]
fromanirh: That's a good point. We need to make sure we have a testcase which exposes https://github.com/kubernetes-sigs/scheduler-plugins/issues/366 but doesn't overlap with the `NodeResourceFit`.

pkg/noderesourcetopology/filter_test.go
Tal-or: Copy-paste mistake probably.

pkg/noderesourcetopology/filter_test.go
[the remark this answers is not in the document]
Tal-or: > Also, I would add another test case here that demonstrates a positive flow for a scope=container and emphasize that we end up with a pod scheduled successfully on the node even thought all the containers are not allocated from the same NUMA nodes unlike scope=pod where all the containers within the pod HAVE to be allocated resource from the same NUMA node. 

We already have multiple tests that apply to this description:
Test id: ```[4][tier2]```
Test id: ```[11][tier1]```
Test id: ```[17][tier1]```

pkg/noderesourcetopology/filter_test.go
```diff
@@ -553,6 +586,221 @@ func TestNodeResourceTopologyMultiContainerPodScope(t *testing.T) {
 	}
 }
 
+func TestNodeResourceTopologyMultiContainerContainerScope(t *testing.T) {
+	nodeTopologies := []*topologyv1alpha1.NodeResourceTopology{
```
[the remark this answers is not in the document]
Tal-or: 1. It is fine to have _some_ code duplication, especially in test code - https://evgenii.com/blog/code-duplication-in-unit-tests/ 

2. If we would like to change some of the values in the NRT object, it will involve changing the values for the second method as well. Are we sure we want to have that?

pkg/noderesourcetopology/filter_test.go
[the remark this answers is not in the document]
Tal-or: This change was pretty exhausting :) 
We would probably need the same reformat for the integration tests and this is another reason why I preferred all these tests to be available from some public package.
Change once - Change everywhere

pkg/noderesourcetopology/filter_test.go
Tal-or: This is why we have the ```TESTS.md``` for, but yes I'll rephrase it

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,77 @@
...
+
+Since this specific configuration has so many edge cases and permutations, we decided to summarize all of them in a table below. 
+This will help us track the coverage and made sure the plugins are fully tested.
+
+#### Test types:
+- Each test type in the table below can be tested with a different type of resources. For example a test that check for overallocation, can be tested with CPUs overallocation or Memory overallocation or both.
+
+### Table guidelines:
+- All pods are ```Guaranteed``` QoS pods, IOW all containers are asking for an equal ```request``` and ```limit``` values.
+- The table describes the coverage of the ```TopologyManagerPolicy=single-numa-node``` and ```TopologyMangerPolicy=container``` configuration. ```TopologyMangerScope=pod``` is basically the same case as ```TopologyMangerPolicy=container``` but with a single container.
```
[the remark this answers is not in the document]
Tal-or: Thanks! Done!

pkg/noderesourcetopology/filter_test.go
Tal-or: Sure. I'll work on it

pkg/noderesourcetopology/filter.go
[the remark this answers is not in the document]
Tal-or: Done!

test/integration/noderesourcetopology_test.go
```diff
@@ -129,7 +129,7 @@ func TestTopologyMatchPlugin(t *testing.T) {
 
 	// Create a Node.
 	resList := map[v1.ResourceName]string{
-		v1.ResourceCPU:    "4",
+		v1.ResourceCPU:    "16",
```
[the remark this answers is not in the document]
Tal-or: Done!

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,78 @@
...
+| 10        | 1               | R < NUMA                         | N > 1      | R < NUMA                         | Running |
+| 11        | 1               | R < NUMA                         | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 12        | 1               | R < NUMA                         | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
+| 13        | 1               | R < NUMA                         | N > 1      | NUMA < R < Node                  | Pending |
+| 14        | N > 1           | R < NUMA                         | 1          | R < NUMA                         | Running |
+| 15        | N > 1           | R < NUMA                         | 1          | NUMA < R < Node                  | Pending |
+| 16        | N > 1           | R < NUMA                         | N > 1      | R < NUMA                         | Running |
+| 17        | N > 1           | R < NUMA                         | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 18        | N > 1           | R < NUMA                         | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
```
swatisehgal, line 49: Same as my comment above.

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,78 @@
...
+| 19        | N > 1           | R < NUMA                         | N > 1      | NUMA < R < Node                  | Pending |
+| 20        | N > 1           | NUMA < SUM(R) < Node && R < NUMA | 1          | R < NUMA                         | Running |
+| 21        | N > 1           | NUMA < SUM(R) < Node && R < NUMA | 1          | NUMA < R < Node                  | Pending |
+| 22        | N > 1           | NUMA < SUM(R) < Node && R < NUMA | N > 1      | R < NUMA                         | Pending |
+| 23        | N > 1           | NUMA < SUM(R) < Node && R < NUMA | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 24        | N > 1           | NUMA < SUM(R) < Node && R < NUMA | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
```
swatisehgal, line 55: Same as above.

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,78 @@
...
+| 25        | N > 1           | NUMA < SUM(R) < Node && R < NUMA | N > 1      | NUMA < R < Node                  | Pending |
+| 26        | N > 1           | SUM(R) > Node && R < NUMA        | 1          | R < NUMA                         | Running |
+| 27        | N > 1           | SUM(R) > Node && R < NUMA        | 1          | NUMA < R < Node                  | Pending |
+| 28        | N > 1           | SUM(R) > Node && R < NUMA        | N > 1      | R < NUMA                         | Running |
+| 29        | N > 1           | SUM(R) > Node && R < NUMA        | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 30        | N > 1           | SUM(R) > Node && R < NUMA        | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
```
swatisehgal, line 61: Same as above.

swatisehgal (commented):
Was thinking about the commit ordering. I noticed a couple of test cases were added initially and then refactoring was done followed by additional unit tests. Why not refactor first in preparation of adding unit tests and then add them in subsequent commits?
Also, seems a bit of back and forth going on because of the current commit breakdown. As an example `makeMultiInitContainerAndContainerPodWithResourceList` was added [879967b](https://github.com/kubernetes-sigs/scheduler-plugins/pull/371/commits/879967bf489537986b1ab3c364ad9769cce24040) and then the implementation was changed to `withMultiContainers`.
I would propose a commit breakdown something like below
```
nrt: fix filter plugin resource alignment when scope=container 
nrt: document tests 
nrt: refactor unit-tests
nrt: add unit-tests 
(HEAD) nrt: add extra unit-tests 
```
The last two commits (`nrt: add unit-tests` and `nrt: add extra unit-tests`) could also be squashed and mapped with the documented tests in the table unless there is a strong reason for splitting them into two separate commits.

fromanirh (commented):
We are converging, thanks for the updates.
In this PR we have a remarkable quantity of unit tests, which is a very good first step and a huge help for the confidence of the correctness of this change (the algorithm, per se, looks good pretty much since the begining).
We're not there yet because we need to review and likely retarget the unit tests proper. The tests per se looks good, but there is a possible significant overlap with NodeResourceFit which we should address. In some corner cases we may end up adding some overlap, because a little redundancy here is (IMO) better than a little gap. But these case should be indeed exceptionally and clearly marked and documented (acknowledging the overlap and explaining why it helps).
As result of the tests retargeting, we may want to add some extra cases. still TBD depending on the answers of the commits.

From my side, the missing step is to clarify and fix this overlap. Adding missing test cases can be done appending commits to this PR once we reach agreement, or be deferred to a later PR.

pkg/noderesourcetopology/filter_test.go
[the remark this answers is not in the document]
Tal-or: As agreed, I removed this test since it's already covered in the ```TestNodeResourceTopologyMultiContainerContainerScope``` test function.

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,77 @@
...
+| 4         | 0               | N/A                              | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 5         | 0               | N/A                              | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
+| 6         | 0               | N/A                              | N > 1      | NUMA < R < Node                  | Pending |
+| 7         | 1               | NUMA < R < Node                  | NVM        | NVM                              | Pending |
+| 8         | 1               | R < NUMA                         | 1          | R < NUMA                         | Running |
+| 9         | 1               | R < NUMA                         | 1          | NUMA < R < Node                  | Pending |
+| 10        | 1               | R < NUMA                         | N > 1      | R < NUMA                         | Running |
+| 11        | 1               | R < NUMA                         | N > 1      | NUMA < SUM(R) < Node && R < NUMA | Running |
+| 12        | 1               | R < NUMA                         | N > 1      | SUM(R) > Node && R < NUMA        | Pending |
```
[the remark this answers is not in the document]
Tal-or: Sure. I removed these tests from the unit tests (but kept them in the ```TESTS.md``` since it's still a valid and possible case).
I also mentioned your note in the commit that this should be covered by the ```NodeResourceFit```.

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,78 @@
...
+This will help us track the coverage and made sure the plugins are fully tested.
+
+#### Test types:
+- Each test type in the table below can be tested with a different type of resources. For example a test that check for overallocation, can be tested with CPUs overallocation or Memory overallocation or both.
+
+### Table guidelines:
+- All pods are ```Guaranteed``` QoS pods, IOW all containers are asking for an equal ```request``` and ```limit``` values.
+- The table describes the coverage of the ```TopologyManagerPolicy=single-numa-node``` and ```TopologyMangerScope=container``` configuration. `TopologyMangerScope=container` is similar to `TopologyMangerScope=pod` but the key difference is the resource alignment scope which is that of a container (i.e. all resources must be allocated to the container from the same NUMA node) as opposed to a pod scope ( where resources of all containers within the pod must be aligned and hence allocated from the same NUMA node). For further reading please refer to  https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#topology-manager-scopes
```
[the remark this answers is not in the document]
Tal-or: Thanks. Fixed

pkg/noderesourcetopology/TESTS.md
```diff
@@ -0,0 +1,78 @@
...
+The most complicated and common case for these plugins is when ```TopologyManagerPolicy=single-numa-node``` and ```TopologyMangerPolicy=container```.
+
+Since this specific configuration has so many edge cases and permutations, we decided to summarize all of them in a table below. 
+This will help us track the coverage and made sure the plugins are fully tested.
+
+#### Test types:
+- Each test type in the table below can be tested with a different type of resources. For example a test that check for overallocation, can be tested with CPUs overallocation or Memory overallocation or both.
+
+### Table guidelines:
+- All pods are ```Guaranteed``` QoS pods, IOW all containers are asking for an equal ```request``` and ```limit``` values.
```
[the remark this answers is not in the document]
Tal-or: Done.

pkg/noderesourcetopology/TESTS.md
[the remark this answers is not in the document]
fromanirh: It would be nice to either split the table or add a column or add another table to mark which case applies to which test kind (unit, integration, e2e)

fromanirh (commented):
thanks for the changes, from my PoV we're almost there. I'd just like a little extra clarifications on `TESTS.md` and LGTM

Huang-Wei:
@Tal-or Sorry for not responding for a while. Pretty occupied recently. The consensus we had, IIUC, was:

- best-efforts scheduling behavior when NRT comes with scope=container
- move integration test to a separate PR

Other than these 2, if you have any other blockers that need my input, please let me know. Otherwise, I will leave the final tests/documentions discussion to you, Swati and Francesco :)

Tal-or:
> thanks for the changes, from my PoV we're almost there. I'd just like a little extra clarifications on `TESTS.md` and LGTM

There is no clear cut line here wrt whether a test is defined as a unit, integration, or e2e test.
Basically, each test listed in the table can fit into any category, it's just a question of time and capacity.

fromanirh (commented):
LGTM

Huang-Wei:
@Tal-or please rebase. I will approve later.

Tal-or:
> @Tal-or please rebase. I will approve later.

@Huang-Wei Sorry it took a while. I had some personal businesses to deal with this week.

PR #372, opened 2022-05-03
noderesourcetopology: fix handling of non-NUMA-affine resources

### LINKED ISSUE #355: using k8s extended resources makes the pod unschedulable at noderesourcetopology filtering
The logic in the nrt filter.go resMatchNUMANodes function considers kubernetes extended resources as if they would all have some numa relevance. Since some of the extended resources may not be part of the createNUMANodeList results at all in case they are not registered device plugin resources, the end result is that any (not-best-effort) pod which uses such extended resources, will be considered as unschedulable by nrt.

It would make sense that if the check for numa resources shows that the extended resource isn't numa-related (map returns not-ok), then that resource would be considered to be ok for any numa.

Kubernetes extended resource creation with curl is described here: https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/ and the extended resource can then be used in the pod spec resource limits and requests.

A unit test which fails on the issue can be found here: https://github.com/uniemimu/scheduler-plugins/commit/42cb4264dfc8aa9710385fb76223aefd868176c1

fromanirh:
#### What type of PR is this?
/kind bug

Review and reorganize how we handle non-NUMA-affine resources, aka resource which don't declare NUMA affinity, like old device plugins and extended resources.

Fixes #355 

please do not squash commits! I'm using unit tests written by another community member, and we need to preserve their authorship.

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -87,6 +87,7 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 
 		// for each requested resource, calculate which NUMA slots are good fits, and then AND with the aggregated bitmask, IOW unset appropriate bit if we can't align resources, or set it
 		// obvious, bits which are not in the NUMA id's range would be unset
+		hasNUMAAffinity := false
 		resourceBitmask := bm.NewEmptyBitMask()
 		for _, numaNode := range numaNodes {
 			numaQuantity, ok := numaNode.Resources[resource]
@@ -94,6 +95,7 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 				continue
 			}
 
+			hasNUMAAffinity = true
 			if !isNUMANodeSuitable(qos, resource, quantity, numaQuantity) {
 				continue
 			}
@@ -101,6 +103,12 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			resourceBitmask.Add(numaNode.NUMAID)
 			klog.V(6).InfoS("feasible", "logKey", logKey, "node", nodeName, "NUMA", numaNode.NUMAID, "resource", resource)
 		}
+
+		if !hasNUMAAffinity && !v1helper.IsNativeResource(resource) {
+			klog.V(6).InfoS("resource available at node level (no NUMA affinity)", "logKey", logKey, "node", nodeName, "resource", resource)
+			continue
+		}
+
 		bitmask.And(resourceBitmask)
 		if bitmask.IsEmpty() {
 			klog.V(5).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
--- a/pkg/noderesourcetopology/filter_test.go
+++ b/pkg/noderesourcetopology/filter_test.go
@@ -35,6 +35,7 @@ import (
 const (
 	cpu                        = string(v1.ResourceCPU)
 	memory                     = string(v1.ResourceMemory)
+	extended                   = "namespace/extended"
 	hugepages2Mi               = "hugepages-2Mi"
 	nicResourceName            = "vendor/nic1"
 	notExistingNICResourceName = "vendor/notexistingnic"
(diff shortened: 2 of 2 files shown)


pkg/noderesourcetopology/filter.go
```diff
@@ -68,28 +68,49 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 
 	// Node() != nil already verified in Filter(), which is the only public entry point
 	nodeName := nodeInfo.Node().Name
+	nodeResources := util.ResourceList(nodeInfo.Allocatable)
 
 	for resource, quantity := range resources {
+		if quantity.IsZero() {
+			// why bother? everything's fine from the perspective of this resource
```
fromanirh, line 75: this shouldn't happen, let's `klog.V(4)` something here

pkg/noderesourcetopology/filter.go
```diff
@@ -68,28 +68,49 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
...
+			continue
+		}
+
+		// do we have this resource at node level, and if so, do we have enough to met the requirement?
+		isNodeResource := false
+		if nodeQuantity, ok := nodeResources[resource]; ok && nodeQuantity.Cmp(quantity) >= 0 {
```
fromanirh, line 81: this also deserves a `klog.V(6)`

fromanirh:
there are many small cleanups embedded here. I think is better to carefully review them in isolation, so I spawned off https://github.com/kubernetes-sigs/scheduler-plugins/pull/373 from this PR.

pkg/noderesourcetopology/filter.go
```diff
@@ -68,28 +68,51 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 
 	// Node() != nil already verified in Filter(), which is the only public entry point
 	nodeName := nodeInfo.Node().Name
+	nodeResources := util.ResourceList(nodeInfo.Allocatable)
 
 	for resource, quantity := range resources {
+		if quantity.IsZero() {
```
Tal-or, line 74: How a resource request can be zero in the first place?
I don't see why this check is necessary.

pkg/noderesourcetopology/filter.go
```diff
@@ -68,28 +68,51 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
...
+			// why bother? everything's fine from the perspective of this resource
+			klog.V(4).InfoS("ignoring zero-qty resource request", "logKey", logKey, "node", nodeName, "resource", resource)
+			continue
+		}
+
+		// do we have this resource at node level, and if so, do we have enough to met the requirement?
+		isNodeResource := false
+		if nodeQuantity, ok := nodeResources[resource]; ok && nodeQuantity.Cmp(quantity) >= 0 {
```
Tal-or, line 82: If ```!ok``` i.e. ```isNodeResource == false``` we should bailout, because if we can't find it at the node level, then a fortiori we can't find it at the NUMA level.

pkg/noderesourcetopology/filter.go
fromanirh: this check was present in the original code and just moved around. We can probably remove it entirely if we deem it unnecessary.

pkg/noderesourcetopology/filter.go
fromanirh: we can have resources only at node level though, like extended resources (https://github.com/kubernetes-sigs/scheduler-plugins/issues/355) or device plugins which don't report NUMA affinity

Tal-or: I know that it might not be at NUMA level, only at the node level.
But What I'm saying is, if you don't even have it at the **node** level, you can call ```continue``` here.

fromanirh: right, I misread initially. That's true indeed, we can simplify the flow here.

fromanirh:
depends on https://github.com/kubernetes-sigs/scheduler-plugins/pull/373
please let's *not* squash commits in this PR because I cherry-picked the testcase added to #355 and we should keep proper author attribution (cc @Huang-Wei )

pkg/noderesourcetopology/filter.go
fromanirh: still TBD. I'm not completely sure we want to remove this, event though it should not happen

fromanirh:
still on hold: need to check the test coverage. Everything else should be OK now.

fromanirh:
/hold cancel
I'm now happy enough with the test coverage

PR #373, opened 2022-05-04
noderesourcetopology: quicker and simpler fail logic for filter when dealing with node-level resources

fromanirh:
#### What type of PR is this?

/kind cleanup

During the implementation of #372 (which wants to fix issue 355) quite a few of possible filter flow flow simplifications emerged. 
The objective is to make the filtering logic fail quicker while also being more linear (and simpler) when dealing with node-level resources. The implementation of #372 will build on top of this simplification.
I am splitting them out here to make it easier to review them in isolation and to help ensure their correctness.

Fixes None, but related to 355

See #372 for more context

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -76,6 +76,8 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			continue
 		}
 
+		isNodeResource := resourceFoundOnNode(resource, quantity, nodeInfo)
+
 		// for each requested resource, calculate which NUMA slots are good fits, and then AND with the aggregated bitmask, IOW unset appropriate bit if we can't align resources, or set it
 		// obvious, bits which are not in the NUMA id's range would be unset
 		resourceBitmask := bm.NewEmptyBitMask()
@@ -84,7 +86,7 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			// if the requested resource can't be found on the NUMA node, we still need to check
 			// if the resource can be found at the node itself, because there are resources which are not NUMA aligned
 			// or not supported by the topology exporter - if resource was not found at both checks - skip (don't set it as available NUMA node).
-			if !ok && !resourceFoundOnNode(resource, quantity, nodeInfo) {
+			if !ok && !isNodeResource {
 				continue
 			}
 
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -70,6 +70,12 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 	nodeName := nodeInfo.Node().Name
 
 	for resource, quantity := range resources {
+		if quantity.IsZero() {
+			// why bother? everything's fine from the perspective of this resource
+			klog.V(4).InfoS("ignoring zero-qty resource request", "logKey", logKey, "node", nodeName, "resource", resource)
+			continue
+		}
+
 		// for each requested resource, calculate which NUMA slots are good fits, and then AND with the aggregated bitmask, IOW unset appropriate bit if we can't align resources, or set it
 		// obvious, bits which are not in the NUMA id's range would be unset
 		resourceBitmask := bm.NewEmptyBitMask()
@@ -78,8 +84,7 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			// if the requested resource can't be found on the NUMA node, we still need to check
 			// if the resource can be found at the node itself, because there are resources which are not NUMA aligned
 			// or not supported by the topology exporter - if resource was not found at both checks - skip (don't set it as available NUMA node).
-			// if the un-found resource has 0 quantity probably this numa node can be considered.
-			if !ok && !resourceFoundOnNode(resource, quantity, nodeInfo) && !quantity.IsZero() {
+			if !ok && !resourceFoundOnNode(resource, quantity, nodeInfo) {
 				continue
 			}
 
@@ -116,11 +121,7 @@ func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, 
 			return true
 		}
 	}
-	// 3. set numa node as possible node if zero quantity for non existing resource was requested
-	if quantity.IsZero() {
-		return true
-	}
-	// 4. otherwise check amount of resources
+	// 3. otherwise check amount of resources
 	return numaQuantity.Cmp(quantity) >= 0
 }
 
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -19,11 +19,11 @@ package noderesourcetopology
 import (
 	"context"
 	"fmt"
-	"strings"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/klog/v2"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
 	bm "k8s.io/kubernetes/pkg/kubelet/cm/topologymanager/bitmask"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
@@ -108,7 +108,7 @@ func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, 
 		if resource == v1.ResourceMemory {
 			return true
 		}
-		if strings.HasPrefix(string(resource), v1.ResourceHugePagesPrefix) {
+		if v1helper.IsHugePageResourceName(resource) {
 			return true
 		}
 		// 2. set numa node as possible node if resource is CPU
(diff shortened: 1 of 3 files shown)


fromanirh:
@Tal-or @swatisehgal

pkg/noderesourcetopology/filter.go
```diff
@@ -77,20 +77,19 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			continue
 		}
 
-		isNodeResource := false
-		if nodeQuantity, ok := nodeResources[resource]; ok && nodeQuantity.Cmp(quantity) >= 0 {
-			isNodeResource = true
+		if !isNodeSuitable(nodeResources, resource, quantity) {
```
fromanirh, line 80: this seeminly correct changes breaks the unit tests (locally). Investigating.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -222,14 +222,6 @@ func TestNodeResourceTopology(t *testing.T) {
...
 			wantStatus: nil,
 		},
-		{
-			name: "Burstable QoS, pod doesn't fit",
-			pod: makePodByResourceList(&v1.ResourceList{
-				v1.ResourceCPU:  *resource.NewQuantity(14, resource.DecimalSI),
-				nicResourceName: *resource.NewQuantity(3, resource.DecimalSI)}),
-			node:       nodes[1],
-			wantStatus: nil, // number of cpu is exceeded, but in case of burstable QoS for cpu resources we rely on fit.go
-		},
```
swatisehgal, line 232: Seems a bit strange to me in this test that the pod doesn't fit and yet the return nil as the status. Is that why we are removing it from here?
Also, (in this and a few other tests) for creation of burstable pod we seem to be calling [makePodByResourceList](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/pluginhelpers.go#L91-L104) which creates a pod such that request =limit.

pkg/noderesourcetopology/filter.go
```diff
@@ -81,8 +81,9 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			// some resources may not expose NUMA affinity (device plugins, extended resources), but all resources
 			// must be reported at node level; thus, if they are not present at node level, we can safely assume
 			// we don't have the resource at all.
-			klog.V(5).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
-			return false
+			klog.V(5).InfoS("short circuit", "logKey", logKey, "node", nodeName, "resource", resource)
+			bitmask.Clear()
+			break
```
swatisehgal, line 86: I would prefer to stick to return (the way we currently do) rather than break. Seems more "go" way of doing things to me :)

pkg/noderesourcetopology/filter.go
```diff
@@ -98,8 +99,9 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 		}
 		bitmask.And(resourceBitmask)
 		if bitmask.IsEmpty() {
-			klog.V(5).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
-			return false
+			klog.V(5).InfoS("short circuit", "logKey", logKey, "node", nodeName, "resource", resource)
+			bitmask.Clear()
```
swatisehgal, line 103: Bitmask is already empty here, why do we need to clear it?

pkg/noderesourcetopology/filter.go
```diff
@@ -77,20 +77,20 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
...
 		}
 
-		isNodeResource := false
-		if nodeQuantity, ok := nodeResources[resource]; ok && nodeQuantity.Cmp(quantity) >= 0 {
-			isNodeResource = true
+		if !isNodeSuitable(nodeResources, resource, quantity) {
+			// some resources may not expose NUMA affinity (device plugins, extended resources), but all resources
+			// must be reported at node level; thus, if they are not present at node level, we can safely assume
+			// we don't have the resource at all.
+			klog.V(5).InfoS("early verdict", "logKey", logKey, "node", nodeName, "resource", resource, "suitable", "false")
```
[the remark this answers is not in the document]
fromanirh: thanks, adopted a slight variation of your suggestion.

pkg/noderesourcetopology/filter_test.go
fromanirh: I'm removing this test because the plugin was implicitely depending on `noderesourcesfit` to perform this check. I think the plugin should be self-sufficient and check all the conditions it is meant to enforce. So the NRT filter should actually fail this test.

pkg/noderesourcetopology/filter.go
fromanirh: ok, so let's drop this commit.

pkg/noderesourcetopology/filter.go
fromanirh: sure, but I'm dropping this commit anyway

pkg/noderesourcetopology/filter_test.go
swatisehgal: Thanks for the explanation, makes sense to not have assumptions that the plugin is configured to run alongside  `noderesourcesfit`.

swatisehgal:
Thanks for your work on this PR and making the code more optimized.
/lgtm

/assign @Huang-Wei 
for review/approval

fromanirh:
/hold
Thanks @swatisehgal for the review! but your point about removing the test deserves a bit more thought from myself. Let me hold this change while I re-evaluate the best course of action

pkg/noderesourcetopology/filter.go
```diff
@@ -70,6 +70,12 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 	nodeName := nodeInfo.Node().Name
 
 	for resource, quantity := range resources {
+		if quantity.IsZero() {
```
Tal-or, line 74: I still cannot figure out how a user can ask for zero resources, is that even possible when you're filling in the pod spec?

pkg/noderesourcetopology/filter.go
```diff
@@ -77,20 +77,20 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 			continue
 		}
 
-		isNodeResource := false
```
Tal-or: We just inline this 3 lines of code in previous commit. Why we're moving it back into a function again?
if that's the case, let's remove the commit with the inline work

Tal-or (commented):
Great work! Make the code much clearer.

I left few minor comments

pkg/noderesourcetopology/filter.go
fromanirh: there are two questions here:
1. why function -> inline -> another function? to illustrate the transformative step and to make each step as obvious as possible, providing a rationale for each step. I expect this PR to ber squashed past review anyway, as it is pretty much the standard in this project
2. should this code be inlined or in a function? I actually don't have strong feelings here. I just need we should try very hard to be consistent between the node suitable and the numa zone suitable check: if one is a function, the other should must likely be as we well, bar glaring obvious reasons. This is for consistency. So, `isNUMANodeSuitable` is a function, then I'm making as final state `isNodeSuitable` a function as well. But indeed, it's debatable.

pkg/noderesourcetopology/filter.go
```diff
@@ -68,22 +68,28 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
...
+	nodeResources := util.ResourceList(nodeInfo.Allocatable)
 
 	for resource, quantity := range resources {
+		if quantity.IsZero() {
+			// why bother? everything's fine from the perspective of this resource
+			klog.V(4).InfoS("ignoring zero-qty resource request", "logKey", logKey, "node", nodeName, "resource", resource)
+			continue
+		}
+
+		if !isNodeSuitable(nodeResources, resource, quantity) {
```
Huang-Wei, line 80: Sounds like we're duplicating the regular check which is already covered in the in-tree [NodeResourcesFit](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/noderesources/fit.go) plugin? I'd suggest to define a list of resources (via pluginArgs or by default) so NRT plugin will check those resources only?

fromanirh: Indeed @Huang-Wei , this is something I'm investigating and I'll clarify ASAP. Thanks for highlighting this!

pkg/noderesourcetopology/filter.go
fromanirh: not for cpu and memory, but kube 1.24.0 accepts pod requesting 0 devices.

pkg/noderesourcetopology/filter.go
fromanirh: thinking more about it, I can do a bit better and avoid the ping pong here. Thanks for raising.

fromanirh:
/hold cancel
just forgot to remove it, the reason to have it is long gone

pkg/noderesourcetopology/filter.go
Huang-Wei: SGTM.

Huang-Wei:
/approve

/hold for doc update :) (feel free to cancel it when doc is updated)
@fromanirh could you update the title and(or) description to reveal the core idea of this cleanup. Basically, IIUC, this PR adds an optimal logic to fail quick when a requested resource is not present at Node level.

fromanirh:
> /approve
> 
> /hold for doc update :) (feel free to cancel it when doc is updated) @fromanirh could you update the title and(or) description to reveal the core idea of this cleanup. Basically, IIUC, this PR adds an optimal logic to fail quick when a requested resource is not present at Node level.

sure thing, done!

fromanirh:
thanks everyone for the reviews!

PR #385, opened 2022-06-05
Minor fix for coscheduling plugin's PostBind logic

Huang-Wei:
#### What type of PR is this?

/kind cleanup

- lock is not needed as both Get() and Patch() are lock-free
- remove unneeded assignment for local PodGroup object

### DIFF
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -216,9 +216,8 @@ func (pgMgr *PodGroupManager) Permit(ctx context.Context, pod *corev1.Pod) Statu
 }
 
 // PostBind updates a PodGroup's status.
+// TODO: move this logic to PodGroup's controller.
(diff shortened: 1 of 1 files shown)


denkensk (commented):
If `pg.status.phase` is changed, how about adding a flag to represent it, without using `deepcopy` here?

Huang-Wei:
> If `pg.status.phase` is changed, how about adding a flag to represent it, without using `deepcopy` here?

Actually, we should remove PostBind as scheduler plugin should not (1) either update the in-memory object, esp. the shared one, and (2) patch the PodGroup object in scheduler plugin.

denkensk:
Yes. Actually `status` isn't used in the internal of scheduler.  Maintaining the status of pod groups in the scheduler isn't suitable. The "reconcile" in the controller can do this uniformly.

/lgtm

Huang-Wei:
Yup, we need to document some best practices for scheduler plugin's development. For example, "do NOT update core API or CR objects during the scheduling plugin; update them separately in a standalone controller".

PR #388, opened 2022-06-17
Nrt fix ext resources

### LINKED ISSUE #355: using k8s extended resources makes the pod unschedulable at noderesourcetopology filtering

fromanirh:
#### What type of PR is this?

/kind bug

Fix handling of resources w/o NUMA affinity. Some resources legitimate lack NUMA affinity. Extended resources
are a great example: https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/

Fixes #355

please do not squash this PR: I'm including unit tests contributed by another community member, and we should preserver their authorship.

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -87,6 +87,7 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 
 		// for each requested resource, calculate which NUMA slots are good fits, and then AND with the aggregated bitmask, IOW unset appropriate bit if we can't align resources, or set it
 		// obvious, bits which are not in the NUMA id's range would be unset
+		hasNUMAAffinity := false
 		resourceBitmask := bm.NewEmptyBitMask()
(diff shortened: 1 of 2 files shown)


fromanirh:
replaces: #372

fromanirh:
sorry, fumbled out with rebases!

PR #392, opened 2022-06-22
Add vendor Directory

### LINKED ISSUE #233: include /vendor directory in the src code
Followup of https://github.com/kubernetes-sigs/scheduler-plugins/pull/222#issuecomment-902788190:

> From Mike: I would be +1 to start committing the vendor directory here. The original argument against it was that this was meant to be more of a toolbox of plugins, but now that this repo is being published as its own scheduler I think committing an up-to-date vendor directory is important (eg, for development, offline builds, and now trying to generate CRDs from vendored api packages).

There is always a debate on whether or not to commit the /vendor directory to a project's source code repo. Some projects (like k/k) do but some (like kind) don't.

As time goes, more and more questions regarding how to configure a compatible dev env or perform local build surface, so I'm planning to include the /vendor directory in the latest master branch, but still leave /vendor absent in other branches.

Please let me know your thoughts.

seanmalloy:
#### What type of PR is this?
/kind cleanup

Adds all Go project dependencies in the `vendor` directory. Prior to this change the `vendor` directory was in the `.gitignore` file and was not committed to this repository.

This will make it easier for developers to have a compatible development environment and reduce the number of local build problems/questions.

Fixes #233 

The vendor directory was added using the below steps with Go 1.17.11.

```
go mod vendor
go mod tidy
make build
```

The update-vendor dependency was removed from all Makefile targets because the vendor directory no longer needs to be updated on the fly at build time.


seanmalloy:
Running `docker build` is updating the `vendor` directory. Need to sort this out before merging.

/hold

seanmalloy:
> Running `docker build` is updating the `vendor` directory. Need to sort this out before merging.
> 
> /hold

I removed the `update-vendor` dependency from all Makefile targets, so that the `vendor` directory does not get updated at build time(i.e. docker build) any more.

/hold cancel

seanmalloy:
> Thanks @seanmalloy. Would you mind split this PR into two commits: one with changes in `vendor/` only, and the other with real changes. This would make review easier.

@Huang-Wei  I have split the changes into two separate commits.
* 9ccdd2b283d7f37b8854a82f18bb846625ab7e2f - add vendor directory and update `.gitignore`
* 38b9b384fb1782b9a702e976269f2695beaafcdf - Makefile changes

PR #395, opened 2022-06-28
Update From Alpine 3.12 to 3.16

seanmalloy:
#### What type of PR is this?
/kind cleanup

Alpine Linux 3.12 went end of life on May 1st, 2022. Therefore update to
use the latest stable release(3.16).

ref: https://alpinelinux.org/releases/

N/A

I verified that the container images build by running the command `make local-image` locally.


Huang-Wei:
/lgtm
/approve

Thanks @seanmalloy .

PR #399, opened 2022-07-05
add sysched design KEP

mvle:
#### What type of PR is this?

/kind feature
KEP for the design of SySched scoring plugin. Allows the scheduling of pods based on pods' system call profiles. Has the potential of using scheduling to enhance overall cluster security.

Fixes #

Some description of the plugin can be found in the mailing list:
https://groups.google.com/g/kubernetes-sig-scheduling/c/fXTaIf4dbh8

We presented the proposal to the community on April 7, 2022 but it looks like the recording has not been uploaded.


kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+### System call Profiles
+
+**How the Scheduler gets access to pods' system call profiles**
+
+We assume that the system call profiles of pods have explicitly been made available to the scheduler through a pod's security context. The seccomp profile directory on the host is mounted into the scheduler's pod. Then it reads and parses the seccomp profiles and obtains the the system call set for that pod. If the Security Profile Operator (SPO) is used (see below), the scheduler can instead use API server to access the seccomp profile CRD to get the system call profile. With the SPO, there's no need to mount the host's seccomp profile directory into the scheduler pod.
```
Huang-Wei, line 119: > The seccomp profile directory on the host is mounted into the scheduler's pod.

Can you elaborate? How the scheduler (usually deployed in the control plane) access the seccomp profiles specified in a to-schedule pod? The seccomp profile settings are specific to a particular node which is not determined at the scheduling time yet.

mvle: The assumption is that we would have seccomp profiles for all the to-be-deployed pods out of band. This can be either using the SPO or just the developers/users generating the profiles for their pods and use cases. The seccomp profiles are then collected and mounted as a volume in the scheduler's pod. This works but not very flexible as using the CRD from the SPO. We will make changes to better clarify in the KEP. We might make this option a minor alternative to using the SPO as that's a better way of doing things.

Huang-Wei: Reading from the SPO CRD makes sense to me. It's just I don't think reading and mounting the regular seccomp profile(s) is an option.

mvle: Agreed. We've updated the KEP to reflect only the SPO option and have removed the mounting option.

mvle:
@Huang-Wei Curious to see if there are any other concerns or comments? I've updated the KEP based on your previous comment in case you haven't seen.

Huang-Wei:
@mvle will review later this week (was on vacation last week). Meanwhile, I'd like a reviewer who is competent on k8s security to chime in.

mvle:
@Huang-Wei can you recommend k8 security reviewers or how else we can proceed?

The proposed plugin is pretty self-contained. The security benefits are really dependent on many factors including cluster size, workload characteristics and other topological/performance requirements. 

More than happy to discuss further over a different medium if so desired.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+</p>
+
+Figure 1  illustrates how our scoring algorithm works and how pods are placed. The figure shows the placement of six pods (C1, C2, C3, C4, C5,and  C6)  in  two  nodes  (Node 1  and  Node 2).  The  numbers inside a rectangular box for a pod indicate the system calls the pod use, i.e., C1 {1, 2, 3, 5}, C2 {4, 7, 8}, C3 {1, 3, 5, 9}, C4 {2, 5, 6, 8}, C5 {2, 4, 5, 7}, and C6 {2, 7, 8}. Initially, to schedule pod C1, the ExS  scores for both Node 1 and Node 2 are the same (i.e., 0), since there are no pods running on the nodes (Figure 1(a)). Thus, our algorithm randomly picks one of the two nodes, Node 1 in this case, and updates its system call usage list using the system calls of C1 (Figure 1(b)). The ExS  scores for scheduling pods C2 are 7 and 0 for Node 1 and Node 2, respectively (Figure 1(b)). In this case, the scheduling algorithm picks Node 2 since Node 2 has a lower ExS  score. Figure 1(c) shows the updated system call lists for Node 2. This process repeats for all incoming pods. After all pods are placed, the cluster-wide ExS score in this example is 12, i.e., ExS1 is 5 and ExS2 is 7, where 1 and 2 are node numbers.
```
Huang-Wei, line 56: Can you be more specific about the meaning of the numbers (in C1 {1, 2, 3, 5})? Like it's a particular type of system call?

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+We assume the pod's configuration contains seccomp security context provided by the user or automatically configured using the Security Profile Operator (more on this below). We make use of the seccomp profile to retrieve what system calls are used by the pods. seccomp is also used to filter the system calls for each pod to enforce system call restriction.
+
+Creating a seccomp profile that is inaccurate can result in pod execution failure. Creating an accurate seccomp filter can be challenging but this is an existing difficulty of using seccomp. The Security Profile Operator can be used to ease and automate this process.
+
+### Risks and Mitigations
+
+Pods with the same workload might get scheduled on the same set of nodes, potentially impacting performance. Can remedy this by applying existing tainting or anti-affinity policies to spread those pods out.
+  
+
```
Huang-Wei, line 82: remove this empty line.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+## Design Details: Overall architecture
+
+Our plugin utilizes system call profiles of pods obtained through their security contexts in the form of seccomp profiles. Our plugin computes a score for each feasible node by comparing the system call profile of the pod with the system call profiles of the existing pods in the node. Our plugin keeps track of the placements of pods in the entire cluster. To compute the scores, the plugin uses a metric called Extraneous System Call (ExS). The ExS metic captures the system calls that a pod does not need, but is present in the node.  Our plugin then returns the computed normalized ExS scores to be combined with other scores in Kubernetes for ranking nodes. If a pod has no seccomp profile associated with it, then our plugin effectively is a noop. The following figure shows the integration of our plugin with the Kubernetes scheduler.
```
Huang-Wei, line 87: metic -> metric

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+</p>
+
+### Scheduling Plugin
+
+We developed a new plugin to implement the scoring extension API point in the Kubernetes scheduler while leaving other filters and scoring plugins intact. We leverage existing filtering and scoring operations of the Kubernetes scheduler to handle other aspects of pod placement such as spreading the pods and ensuring resource availability.
+
+ There are two main components of our plugin: 1) a lightweight in memory state store that holds the up-to-date node to pod mapping and 2) mechanisms for retrieving the system call sets used by pods and for computing the ExS scores. The state store is implemented as a thread that listens to events of a pod’s lifecycle, e.g., creation, stop, destruction, run, etc. Using these event notifications, this thread tracks where pods are currently running in a cluster. When a pod stops running on a node, our plugin also removes it from its internal mapping. While the entire state of pod placement in a cluster can be retrieved dynamically by querying the API server, this can be an expensive task to do per pod scheduling event, hence we choose to maintain this state internally in our scheduler.
+
```
Huang-Wei, line 114: We need a stanza to describe the detailed Scoring algorithm here. In one of the above pictures, it shows Node 1/2 were given different scores, we should detail how that score gets calculated.

And also explain the logic that if a pod doesn't carry syscall-related attributes, this plugin does nothing, etc, etc.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+<img src="images/arch.png" style="width:70%">
+</p>
+
+<p align="center">
+<b>Figure 3: Overall architecture</b>
+</p>
+
+### Scheduling Metrics - Extraneous System Call (ExS)
+
+At the heart of SySched is a single-valued score used to quantify the extraneous system call exposure of a pod on a node. Informally, ExS captures the additional system calls accessible on a node that the pod itself does not use. This score reflects the potential “danger” a pod must face when running on the target node as vulnerabilities in these extraneous system calls from neighboring pods can jeopardize its own security. In the following figure, we can obtain the ExS for pod C1, C2, and C3 for a node with a kernel of nine system calls.
```
Huang-Wei, line 99: Can you be more specific about how this metric is obtained? Using a concrete example like a pod's seccomp spec, or dependent CRD spec would help.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+</p>
+
+### Scheduling Plugin
+
+We developed a new plugin to implement the scoring extension API point in the Kubernetes scheduler while leaving other filters and scoring plugins intact. We leverage existing filtering and scoring operations of the Kubernetes scheduler to handle other aspects of pod placement such as spreading the pods and ensuring resource availability.
+
+ There are two main components of our plugin: 1) a lightweight in memory state store that holds the up-to-date node to pod mapping and 2) mechanisms for retrieving the system call sets used by pods and for computing the ExS scores. The state store is implemented as a thread that listens to events of a pod’s lifecycle, e.g., creation, stop, destruction, run, etc. Using these event notifications, this thread tracks where pods are currently running in a cluster. When a pod stops running on a node, our plugin also removes it from its internal mapping. While the entire state of pod placement in a cluster can be retrieved dynamically by querying the API server, this can be an expensive task to do per pod scheduling event, hence we choose to maintain this state internally in our scheduler.
```
Huang-Wei, line 113: These two stanzas do not reveal essential info to illustrate how a pod is scored. Some of the info like how it relates with filter and other score plugins are assumed to be known by readers, so we usually don't need to mention it.

I guess the second stanza here is trying to illustrate that there is a cache defined to syn with pod events, so that we keep an up-to-date state about what types of syscalls are present on what nodes. If it's the case, we can simplify it a bit.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,161 @@
...
+
+### System call Profiles
```
Huang-Wei, line 115: As mentioned above, it'd be very helpful to have concrete examples here.

Huang-Wei:
> @Huang-Wei can you recommend k8 security reviewers or how else we can proceed?

If you're acquainted with someone in https://github.com/kubernetes-sigs/security-profiles-operator, getting thoughts from them would be super helpful.

mvle:
@Huang-Wei Thanks for your comments! We've updated the KEP to incorporate your suggestions. Please take another look.

mvle:
> If you're acquainted with someone in https://github.com/kubernetes-sigs/security-profiles-operator, getting thoughts from them would be super helpful.

We are not acquainted but will seek their input on this KEP.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,217 @@
...
+#### **How the Scheduler gets access to pods' system call profiles**
+
+We assume that the system call profiles of pods have explicitly been made available to the scheduler via the SPO. We rely on the SPO to generate and bind seccomp profile CRDs to the pods. The scheduler uses the API server to access a pod's seccomp profile CRD and parses the seccomp profile CRD to obtain the list of system calls for a pod. 
+
+An example of an SPO's seccomp profile CRD:
+
+```
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1beta1
+kind: SeccompProfile
```
saschagrunert, line 118: We have multiple components of the seccomp profile to be taken into account depending on how the scoring calculation works in detail:

- A default action of `SCMP_ACT_ALLOW` may require us to collect / assume all available syscalls on a node
- The architecture has to match
- Additional arguments may not be allowed (it would be non-trivial to evaluate those)

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,217 @@
...
+    - fchdir
+    - sendfile
+    - getdents64
+    - setsockopt
+    - recvfrom
+    - sigaltstack
+    - fstatfs
+    - umask
+    - getegid
+    - read
```
saschagrunert, line 139: I'm wondering if we wanna create groups of syscalls depending on their security impact. For example the `RuntimeDefault` profile already tries to provide strong defaults.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,217 @@
...
+```
+
+An example of pod's security context specifying the associated seccomp profile:
+
+```
+...
+    spec:
+      securityContext:
+        seccompProfile:
+          type: Localhost
```
saschagrunert, line 150: We could also take `Unconfined` into account and consider those as pods/containers with the highest security risk.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,217 @@
...
+
+While pod-centric security schemes such as seccomp are necessary, they alone are insufficient. We introduce a practical way to curtail the impact of these types of threats by leveraging a new security-aware pod scheduling scheme that we've developed for Kubernetes to co-locate pods based on their system call usage. So even if such attacks were successful, we show that it would impact fewer victim pods (up to 40% less) on fewer nodes (up to 1/3 less) in the cluster than the existing scheduler.
+ 
+### Goals
+
+1. Develop a score plugin to favor nodes, for a given incoming pod, that have low ExS scores, i.e., nodes with pods that have similar system call profiles.
```
saschagrunert, line 20: Depending on how often syscalls are being called, we may end up in a performance issue: Assuming we have workloads doing a lot of IO intense operations and we group them on a single node because of their read/write syscalls, then we may stress the node on that. The same would apply to networking intense applications and so on. I think we could outline that the plugin favors security over resource distribution, which may or may not be fine. :shrug:

saschagrunert (commented):
Great idea, I'm thrilled to see this (as a PoV) in action! I left a few comments and thoughts on the KEP, they're more informal than blocking.

kep/399-sysched-scoring/README.md
mvle: These are great points. 

Right now we are just considering `SCMP_ACT_ALLOW` but should definitely also handle other action types when they are present. Syscall arguments are not something we are considering using as of yet. Ultimately, we just care about what syscalls a pod is allowed to use. 

We will update these points in caveats/notes/limitations.

kep/399-sysched-scoring/README.md
mvle: Grouping syscalls based on their security impact is definitely something interesting. We thought about that aspect when we mentioned the use of weights to denote the "criticality" of certain syscalls. We can define criticality in a few ways but one way is to base on historical CVE data associated with syscalls or how syscalls are used in known exploits. 

Introducing this groupings/weights through the scheduling plugin is one way or encode it somehow in the SPO CRD would also be useful.

kep/399-sysched-scoring/README.md
mvle: Taking into account the type and kind of seccomp profile being used in the scheduling besides just purely the syscalls could be really interesting. Right now we treat pods w/o seccomp profiles as no-ops. But on considering your point, we could instead, treat pods w/o seccomp profiles or `unconfined` filters as having access to all syscalls, hence being riskiest to all other pods. 

I think this aspect relates to your previous point on considering different components of a seccomp profile.

kep/399-sysched-scoring/README.md
mvle: Yes, this would be a concern if the plugin is used as the sole decider of placement. However, we don't anticipate it being used as such since if there are performance critical workloads or other scheduling metric that is more important, you would want to use tagging or other scheduling plugins in conjunction with our security plugin and provide the appropriate weights to ensure some form of QoS is maintained. The workloads w/o specific restrictions can then benefit from our approach. 

We'll try to expand the limitations section to make this consideration more clear.

mvle:
@saschagrunert Thank you for taking the time to review our KEP!

@Huang-Wei We've updated the KEP to incorporate your comments and also @saschagrunert (mainly to clarify what and how seccomp is used). Please take another look and let us know your thoughts.

mvle:
@Huang-Wei Can you please let us know if you have further comments or concerns?

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+Initially, to schedule pod $P_1$, the ExS  scores for both Node 1 and Node 2 are the same (i.e., 0), since there are no pods running on the nodes (Figure 3(a)). Thus, we randomly pick one of the two nodes, Node 1 in this case, and updates Node 1's system call usage list using $P_1$'s system call profile (Figure 3(b)). When pod $P_2$ arrives to be scheduled, our plugin computes the ExS scores 7 and 0 for $P_2$ w.r.t. Node 1 and Node 2, respectively (Figure 3(b)). In this case, the plugins ranked Node 2 over Node 1 since Node 2 has a lower ExS  score (i.e., higher normalized score). Figure 3(c) shows the updated system call lists for Node 2. This process repeats for all incoming pods. After all pods are placed, the cluster-wide ExS score in this example is 12, i.e., $ExS_1$ is 5 and $ExS_2$ is 7, where 1 and 2 are node numbers.
```
Huang-Wei, line 103: Is score `7` a hypothetical number, or it's precisely calculated?

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+<p align="center">
+<img src="images/example_complete.png" style="width:50%">
+</p>
+
+<p align = "center">
+<b>Figure 3. Scheduling algorithm and pod placement example</b>
+</p>
+
+Figure 3 also highlights the opportunities not only for reducing the exposure to extraneous system calls (i.e., ExS score) but also the node’s attack surface. If $P_3$ had been scheduled on Node 2 instead of Node 1, then Node 2 would have a total of 7 system calls "open" $(1,3,4,5,7,8,9)$, instead of only 3 system calls under the SySched approach.
```
Huang-Wei, line 113: > instead of only 3 system calls under the SySched approach.

Is that `5` instead `3`? (1,2,3,5,9)

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### **How the Scheduler gets access to pods' system call profiles**
+
+We assume that the system call profiles of pods have explicitly been made available to the scheduler via the SPO. We rely on the SPO to generate and bind seccomp profile CRDs to the pods. The scheduler uses the API server to access a pod's seccomp profile CRD and parses the seccomp profile CRD to obtain the list of system calls for a pod. 
+
+An example of an SPO's seccomp profile CRD:
+
+```
```
Huang-Wei, line 125:
```
```yaml
```
to prettify the rendering.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Scheduling Plugin
+
+There are two main components of our plugin: 1) a lightweight in-memory cache that is kept synchronize with pod events in order to keep an up-to-date node to pod mapping in the cluster, and 2) mechanisms for retrieving the system call sets used by pods for computing the ExS scores. The in-memory cache reduces the cost of querying the API server for each pod scheduling event. We obtain the system call profiles for pods from their seccomp profiles. Our plugin leverages the Kubernetes sigs community-developed [Security Profiles Operator](https://github.com/kubernetes-sigs/security-profiles-operator) (SPO) that creates Custom Resource Definitions (CRDs) for seccomp profiles. If a pod has no seccomp profile associated with it, i.e., `unconfined`, it will have access to all system calls on a host (denoting highest risk), hence, our plugin will associate all system calls available on a particular host to the pod when calculating ExS.
```
Huang-Wei, line 117: for (1), scheduler framework provides a `handle` to obtain pods on a given node, so maybe not necessary to build yourself. But I will leave the review to code impl.

Huang-Wei (commented):
The latest version is pretty neat! Thanks for the diligent work!

Only a few nits.

kep/399-sysched-scoring/README.md
```diff
@@ -0,0 +1,229 @@
+﻿# KEP: SySched  - System call-based scheduling
+
+## Summary
+
+We propose the use of pod placement as a way to improve the security of nodes in a cluster. Specifically, we propose a new scheduler scoring plugin (SySched) that enables the ranking of feasible nodes based on the relative risks of pods' system call usage. Key to this risk calculation is the Extraneous System Call (ExS) metric &mdash; a metric defined by us &mdash; which measures the amount of excess system call a pod is exposed to on a given node. The benefits of this approach are the following:
```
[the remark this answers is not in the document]
mvle: Fixed.

kep/399-sysched-scoring/README.md
mvle: The numbers are precisely calculated from the hypothetical example. However, there was a mistake in cluster-wide ExS score that is now fixed. We also added some text and equations to better define node-wide and cluster-wide ExS scores and to clarify the explanation.

kep/399-sysched-scoring/README.md
mvle: Fixed

kep/399-sysched-scoring/README.md
mvle: Good to know and we'll investigate this.

mvle:
@Huang-Wei Thanks for the comments. Updated the KEP. Please check it out again.

Huang-Wei:
@mvle thanks for the work again.

/lgtm
/approve

/label tide/merge-method-squash

PR #412, opened 2022-08-09
Trimaran: add low-risk overcommitment scheduler plugin

### LINKED ISSUE #357: Overcommitment aware scheduler plugin KEP
Though containers are allowed to specify resource limit values beyond the requested (guaranteed) values, most schedulers are only concerned with the requested values. This may result in a cluster where some (or all) nodes are overcommitted, thus leading to potential CPU congestion and performance degradation and/or memory OOM situations and pod evictions. A scheduler which takes into consideration (1) the resource limit values of pods (limit-aware) and (2) the actual load (utilization) on the nodes (load-aware) could provide a low risk environment for pods and alleviate issues with overcommitment, while allowing pods to use their limits.

This proposal describes a scheduler (score) plugin which evaluates the performance risk of overcommitment and selects the node with lowest risk. It achieves this goal by combining two risk factors: limit risk and load risk. The limit risk is based on requests and limits values. And, the load risk is based on observed load. Thus, the proposed scheduler is risk-aware as well as load-aware. The outcome is that burstable and best effort pods are placed on nodes where the chance of being impacted by overcommitment is minimized, while providing them a chance to burst up to their full limits.

[Presentation](https://docs.google.com/presentation/d/1u4SJxN_tRq8qT-G5mkh5omhXmgrFIbV865zBdMVCFps/edit?usp=sharing)

[Documentation](https://docs.google.com/document/d/15XPacVfK0EapUtTUX6JcKt014YEOrxQMflV2M8PN2PY/edit?usp=sharing)

atantawi:
#### What type of PR is this?
/kind feature
/kind api-change

add an overcommitment-aware scheduler plugin to the trimaran family

Fixes #357 

```release-note
Trimaran: implementation for low-risk overcommitment scheduler plugin.
```


atantawi:
\release-note ```Added LowRiskOverCommitment scheduler plugin```

pkg/trimaran/resourcestats.go
```diff
@@ -136,3 +136,124 @@ func GetResourceRequested(pod *v1.Pod) *framework.Resource {
...
+	nodeRequest := &framework.Resource{}
+	nodeLimit := &framework.Resource{}
+	unlimitedCpu := false
+	unlimitedMem := false
+	nodeRequestMinusPod := &framework.Resource{}
+	nodeLimitMinusPod := &framework.Resource{}
+	// set capacities
+	nodeCapacity := &framework.Resource{}
+	allocatableResources := node.Status.Allocatable
+	amCpu := allocatableResources[v1.ResourceCPU]
```
Huang-Wei, line 192: Other than cpu and memory, other core resources can also be overcommitted, like ephemeral storage.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+		return score, framework.NewStatus(framework.Error, fmt.Sprintf("getting node %q from Snapshot: %v", nodeName, err))
+	}
+	// get node metrics
+	metrics, _ := pl.collector.GetNodeMetrics(nodeName)
+	if metrics == nil {
+		klog.InfoS("Failed to get metrics for node; using minimum score", "nodeName", nodeName)
+		return score, nil
+	}
+	// calculate score
+	totalScore := pl.computeRank(metrics, nodeInfo.Node(), pod) * float64(framework.MaxNodeScore)
```
Huang-Wei, line 119: Why don't you pass in NodeInfo, and then use nodeInfo.Pods to obtain the pods?

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+		capacity = nodeCapacity.Memory
+	} else {
+		// invalid resource
+		klog.V(6).InfoS("Unexpected resource", "resourceName", resourceName)
+		return 0
+	}
+
+	// (1) riskLimit : calculate overcommit potential load
+	excess := limit - request
+	if excess < 0 {
```
Huang-Wei, line 204: This cannot happen, right? If it does, there is a bug in the code.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+	maxLoad := capacity
+	if limit < capacity {
+		maxLoad = limit
+	}
+	allowed := maxLoad - request
+	if allowed < 0 {
+		allowed = 0
+	}
+	if excess > 0 && allowed < excess {
+		riskLimit = 1 - float64(allowed)/float64(excess)
```
Huang-Wei, line 216: Is that the higher riskLimit is, the more risk it has? If so, if a pod request 0.001 cpu, and limits a whole node's allocatable cpu, the riskLimit is ~0 in this case?

```
	// node's capacity: 1
	// pod's req/limits: 0.01/1
	// an existing pod with req/limits: 0.01/0.01
	// limit = 1.01 (1 + 0.01)
	// request = 0.01
	// excess = 1
	// maxLoad = 1
	// allowed = 0.99
```

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	klog.V(6).InfoS("RiskLimit", "node", klog.KObj(node), "resource", resourceName, "riskLimit", riskLimit)
+
+	// (2) riskLoad : calculate measured overcommitment
+	zeroRequest := &framework.Resource{}
+	stats, ok := trimaran.CreateResourceStats(metrics, node, zeroRequest, resourceName, resourceType)
+	if ok {
+
+		// fit a beta distribnution to the measured load stats
```
Huang-Wei, line 225:
```suggestion
		// fit a beta distribution to the measured load stats
```

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+	if excess > 0 && allowed < excess {
+		riskLimit = 1 - float64(allowed)/float64(excess)
+	}
+	klog.V(6).InfoS("RiskLimit", "node", klog.KObj(node), "resource", resourceName, "riskLimit", riskLimit)
+
+	// (2) riskLoad : calculate measured overcommitment
+	zeroRequest := &framework.Resource{}
+	stats, ok := trimaran.CreateResourceStats(metrics, node, zeroRequest, resourceName, resourceType)
+	if ok {
+
```
Huang-Wei, line 224: remove the empty line

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+		allowed = 0
+	}
+	if excess > 0 && allowed < excess {
+		riskLimit = 1 - float64(allowed)/float64(excess)
+	}
+	klog.V(6).InfoS("RiskLimit", "node", klog.KObj(node), "resource", resourceName, "riskLimit", riskLimit)
+
+	// (2) riskLoad : calculate measured overcommitment
+	zeroRequest := &framework.Resource{}
+	stats, ok := trimaran.CreateResourceStats(metrics, node, zeroRequest, resourceName, resourceType)
```
Huang-Wei, line 222: Why pass in a zeroRequest? Does this mean the score of this portion won't take the incoming pod's limits into consideration?

apis/config/v1beta3/types.go
```diff
@@ -120,6 +120,21 @@ type LoadVariationRiskBalancingArgs struct {
...
+// LowRiskOverCommitmentArgs holds arguments used to configure LowRiskOverCommitment plugin.
+type LowRiskOverCommitmentArgs struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Common parameters for trimaran plugins
+	TrimaranSpec `json:",inline"`
+	// Smoothing window size for usage data in the load watcher
+	SmoothingWindowSize *int64 `json:"smoothingWindowSize,omitempty"`
+	// Weight of risk due to limit
+	RiskLimitWeight *float64 `json:"riskLimitWeight,omitempty"`
```
Huang-Wei, line 135: so IIUC this weight is to configure the ratio of riskLimits, and (1 - RiskLimitWeight) is the ratio of riskLoad? If so, I'd suggest to make their semantics explicit like:

```yaml
- strategies:
  - foo
    - foo-specific-arg: 
    - weight: 0.5
  - bar
    - bar-specific-arg: 
    - weight: 0.5
```

pkg/trimaran/lowriskovercommitment/beta.go
```diff
@@ -0,0 +1,264 @@
...
+
+// lgamma : log gamma function
+func lgamma(x float64) float64 {
+	y, _ := math.Lgamma(x)
+	return y
+}
+
+// ComputeProbability : The probability that the resource utilization is less than or equal to a given threshold value
+func ComputeProbability(mu, sigma, threshold float64) (float64, *BetaDistribution) {
+
```
Huang-Wei, line 246: nit: move the first empty line (applies elsewhere)

pkg/trimaran/lowriskovercommitment/lowriskovercommitment_test.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+func TestLowRiskOverCommitment_New(t *testing.T) {
+	watcherResponse := watcher.WatcherMetrics{}
+	server := httptest.NewServer(http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
+		bytes, err := json.Marshal(watcherResponse)
+		assert.Nil(t, err)
+		resp.Write(bytes)
+	}))
+	defer server.Close()
```
Huang-Wei, line 78: Along with latest refactoring, we don't need this anymore. We're switching to `envtest` to launch the integration test. You can refer to other integration tests to refactor it.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+	if ok {
+
+		// fit a beta distribnution to the measured load stats
+		mu, sigma := trimaran.GetMuSigma(stats)
+		// adjust standard deviation due to data smoothing
+		sigma *= math.Pow(float64(pl.args.SmoothingWindowSize), 0.5)
+		// limit the standard deviation close to the allowed maximum for the beta distribution
+		sigma = math.Min(sigma, math.Sqrt(GetMaxVariance(mu)*MaxVarianceAllowance))
+
+		// calculate area under beta probability curve beyond total allocated, as overuse risk measure
```
Huang-Wei, line 232: TBH I may not be the best guy to review the correctness/efficiency of this portion of code. Maybe @wangchen615 can give a hand on this "beta probability algorithm"?

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+		}
+
+		riskLoad = 1 - allocProb
+		klog.V(6).InfoS("RiskLoad", "node", klog.KObj(node), "resource", resourceName,
+			"allocThreshold", allocThreshold, "allocProb", allocProb, "riskLoad", riskLoad)
+	}
+
+	// combine two components of risk into a total risk as a weighted sum
+	totalRisk = (pl.args.RiskLimitWeight*riskLimit + (1-pl.args.RiskLimitWeight)*riskLoad)
```
Huang-Wei, line 255:
```suggestion
	totalRisk = pl.args.RiskLimitWeight*riskLimit + (1-pl.args.RiskLimitWeight)*riskLoad
```

apis/config/v1beta3/types.go
Huang-Wei: Typically, cpu can be overcommitted more tolerantly than memory/storage, so can we add args to distinguish the importance/weight of individual resources?

apis/config/v1beta3/defaults.go
```diff
@@ -60,6 +60,15 @@ var (
 	// DefaultSafeVarianceSensitivity is one
 	DefaultSafeVarianceSensitivity = 1.0
 
+	// Defaults for LowRiskOverCommitment plugin
+
+	// The moothing window size for usage data in the load watcher.
```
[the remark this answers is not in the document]
atantawi: done

apis/config/v1beta3/types.go
```diff
@@ -120,6 +120,21 @@ type LoadVariationRiskBalancingArgs struct {
...
+// +k8s:defaulter-gen=true
+
+// LowRiskOverCommitmentArgs holds arguments used to configure LowRiskOverCommitment plugin.
+type LowRiskOverCommitmentArgs struct {
+	metav1.TypeMeta `json:",inline"`
+
+	// Common parameters for trimaran plugins
+	TrimaranSpec `json:",inline"`
+	// Smoothing window size for usage data in the load watcher
+	SmoothingWindowSize *int64 `json:"smoothingWindowSize,omitempty"`
```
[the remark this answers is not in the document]
atantawi: The units here are number of windows. Clarified in comments.

pkg/trimaran/handler.go
```diff
@@ -68,6 +68,12 @@ func New() *PodAssignEventHandler {
 	return &p
 }
 
+// Returns a new instance of PodAssignEventHandler, without deleting old pods from cache
```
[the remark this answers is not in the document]
atantawi: Removed the need for caching for this plugin. Now using the SnapshotSharedLister.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+					utilruntime.HandleError(fmt.Errorf("unable to convert object %T to *v1.Pod", obj))
+					return false
+				default:
+					utilruntime.HandleError(fmt.Errorf("unable to handle object: %T", obj))
+					return false
+				}
+			},
+			Handler: podAssignEventHandler,
+		},
+	)
```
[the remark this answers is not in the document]
atantawi: Thanks, done

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+func isAssigned(pod *v1.Pod) bool {
+	return len(pod.Spec.NodeName) != 0
+}
+
+// computeRank : rank function for the LowRiskOverCommitment
+func (pl *LowRiskOverCommitment) computeRank(metrics []watcher.Metric, node *v1.Node, pod *v1.Pod) float64 {
+	// get all pods on node
+	nodeName := node.GetName()
+	podsOnNode := make([]*v1.Pod, 0)
+	pl.eventHandler.RLock()
```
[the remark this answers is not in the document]
atantawi: Even though the overhead is not that significant, I decided to take your second suggestion of using the SnapshotSharedLister. Much cleaner.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,258 @@
...
+
+	klog.V(6).InfoS("Node rank", "nodeName", nodeName, "riskCPU", riskCPU, "riskMemory", riskMemory, "rank", rank)
+
+	return rank
+}
+
+// computeRisk : calculate the risk of scheduling on node for a given resource
+func (pl *LowRiskOverCommitment) computeRisk(metrics []watcher.Metric, resourceName v1.ResourceName,
+	resourceType string, node *v1.Node, nodeRequestsAndLimits *trimaran.NodeRequestsAndLimits) float64 {
+
```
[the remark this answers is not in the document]
atantawi: done

pkg/trimaran/resourcestats.go
```diff
@@ -136,3 +136,124 @@ func GetResourceRequested(pod *v1.Pod) *framework.Resource {
...
+	// NodeRequestMinusPod is the NodeRequest without the requests of the pending pod
+	NodeRequestMinusPod *framework.Resource
+	// NodeLimitMinusPod is the NodeLimit without the limits of the pending pod
+	NodeLimitMinusPod *framework.Resource
+	// Nodecapacity is the capacity (allocatable) of node
+	Nodecapacity *framework.Resource
+}
+
+// GetNodeRequestsAndLimits : total requested and limits of resources on a given node plus a pod
+func GetNodeRequestsAndLimits(podsOnNode []*v1.Pod, node *v1.Node, pod *v1.Pod) *NodeRequestsAndLimits {
```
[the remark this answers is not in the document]
atantawi: I cleaned up the functions as much as I understand. When a new utility function becomes available I'll incorporate it. The use of accumulated node requests and limits in this plugin is simply to assess the gap between the two. So, the calculation does not have to be very precise, as it is not used for allocation.

pkg/trimaran/resourcestats.go
atantawi: Yes, it would be nice to add support for additional resources such as GPUs and ephemeral storage. We'll hold to that until the load watcher supports exporting usage metrics for these resources.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: Thanks, done.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: Typically, it should not happen, unless some containers do not specify requested values and a default minimum is used instead. The sum of such minima across containers may become larger than the sum of limits. So, to be on the safe side, I handle that case.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: Yes, the risk is almost zero in this case. And, that makes sense. The reasoning is that if the pod is placed on that node it should be able to burst from its allocated 0.01 to the full node capacity without any risk.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: done

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: done

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: This part is about utilization. Since the pod hasn't started yet, we don't know how much utilization it will have. Currently, zero is assumed. We could use an estimate, but that would be a little risky.

apis/config/v1beta3/types.go
atantawi: Thanks, done.

pkg/trimaran/lowriskovercommitment/beta.go
atantawi: done

pkg/trimaran/lowriskovercommitment/lowriskovercommitment_test.go
atantawi: This is for the unit test. I believe you mean `envtest` for integration test?

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: @wangchen615 is familiar with this plugin

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: done

atantawi:
/release-note-edit ```release-note Added LowRiskOverCommitment scheduler plugin```

atantawi:
Pinging @Huang-Wei for a second review. I addressed all your comments/suggestions in your first review as described in comments above. Thanks

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,242 @@
...
+		// calculate risk
+		riskLoad = 1 - allocProb
+		klog.V(6).InfoS("RiskLoad", "node", klog.KObj(node), "resource", resourceName,
+			"allocThreshold", allocThreshold, "allocProb", allocProb, "riskLoad", riskLoad)
+	}
+
+	// combine two components of risk into a total risk as a weighted sum
+	w := pl.riskLimitWeightsMap[string(resourceName)]
+	totalRisk = w*riskLimit + (1-w)*riskLoad
```
Huang-Wei, line 239: isn't `w` used to weight scores among different resources? and here it's weighting riskLimit vs. riskLoad?

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
Huang-Wei: Hold on here... why the pod can burst to the full node capacity is at a low risk?

If a pod is specifying a req/lim request as 0.001/1, it implies it's somehow cheating the system's quota system, and if we pack pods to this kind to the same node, they may co-burst at the same time and that yields OutOfMemory or(and) CPU throttling.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,242 @@
...
+	}
+	// create map of resource risk limit weights
+	m := make(map[string]float64)
+	m[v1.ResourceCPU.String()] = 0.5
+	m[v1.ResourceMemory.String()] = 0.5
+	for _, rs := range args.RiskLimitWeights {
+		if rs.Weight >= 0 && rs.Weight <= 100 {
+			m[rs.Name] = float64(rs.Weight) / 100
+		}
+	}
```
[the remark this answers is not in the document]
atantawi: OK, thanks, I will use a different construct than `ResourceSpec`.

pkg/trimaran/resourcestats.go
```diff
@@ -136,3 +154,150 @@ func GetResourceRequested(pod *v1.Pod) *framework.Resource {
...
+	// add up limit of all containers
+	result = &framework.Resource{}
+	for _, container := range pod.Spec.Containers {
+		result.Add(container.Resources.Limits)
+		if container.Resources.Limits.Cpu().IsZero() {
+			unlimitedCPU = true
+		}
+		if container.Resources.Limits.Memory().IsZero() {
+			unlimitedMemory = true
+		}
```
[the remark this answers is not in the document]
atantawi: Let me clarify.

1. The idea of using `unlimitedCPU` is to capture the fact that if a container specifies CPU requests but no CPU limits, it can potentially surge its usage to 100% node utilization. Of course this depends on what the other containers on the node are doing. The Kubelet will allow it to happen. Right? So, we would rather schedule a new pod on a different node, with limited CPU, than this unlimited node. This way, there would be less risk for the new pod if it needs to burst beyond its requested value, and, at the same time, let this unlimited container burst as well! If scheduled to the same node, then both pods will share the excess capacity (beyond requests) proportional to their requested values. However, if scheduled on different nodes, there is no competition.
2. If an init container does not specify a limit, then again it could burst and use up the full node, if allowed. So, the pod with this init container, as a whole, could use 100% of the node, even only during initialization, which may last a significant amount of time, depending on the appllication. This situation poses a risk to other pods being scheduled to this node. I think this is similar to the SpringBoot scenario?

pkg/trimaran/resourcestats.go
```diff
@@ -136,3 +154,150 @@ func GetResourceRequested(pod *v1.Pod) *framework.Resource {
...
+	// take max_limit(sum_pod, any_init_container)
+	for _, container := range pod.Spec.InitContainers {
+		for rName, rQuantity := range container.Resources.Limits {
+			switch rName {
+			case v1.ResourceCPU:
+				if rQuantity.MilliValue() == 0 {
+					unlimitedCPU = true
+				} else {
+					setMax(&result.MilliCPU, rQuantity.MilliValue())
+				}
```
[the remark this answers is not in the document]
atantawi: But, as described above, couldn't this container burst its usage beyond its requested value?

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: Yes, that's the meaning of the weight `w`. It's about weighting two types of risks: `riskLimit` and `riskLoad`.  The `totalRisk` is a weighted sum. There is a `totalRisk` per resource, CPU and memory. The worst `totalRisk` among the resources is taken to be the risk of scheduling a pod on this node. Makes sense?

The idea of this plugin is **not** to restrict pods from going beyond their requests. To the contrary, it's about scheduling pods on nodes so that they can burst and **reach their limits**, without bumping into each other and creating OOM and throttling. It should work whether the aggregate in the cluster is under or over-committed.

Huang-Wei: > To the contrary, it's about scheduling pods on nodes so that they can burst and reach their limits, without bumping into each other and creating OOM and throttling.

In this way, a tenant can hack around this design by specifying low-priority and best-efforts Pods (0 limits) to exploit  cluster's resources as those pods will be spread across the cluster. I understand it's useful for applications like spring-boot, but it doesn't sound fair to other workloads that specify limits - those Pods will be more likely to be bin-packed right?

pkg/trimaran/resourcestats.go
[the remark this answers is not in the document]
Huang-Wei: I think the principle for best-efforts Pods is: try your best to tune the limits. What we can try to achieve is that based on the semantics of requests/limits, try to pack more workloads in the cluster while with less disruption/eviction.

In other words, it's impractical to promise users their apps can achieve optimal scheduling behavior by specifying 0 limits. As I mentioned in another comment, giving Pods with 0 limits a "preference" is hazard as users can make use of it.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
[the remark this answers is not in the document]
Huang-Wei: Yes please. BTW: the `name` field is used to represent resource type, while here it represents the "strategy" (`limits` and `load`).

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
Huang-Wei: > Yes, that's the meaning of the weight w

Understood now. Commented in another comment about pluginArgs.

atantawi:
@Huang-Wei Thanks for the clarifications and suggestions. I agree in principle. Let me think a little bit about the issues that you raised.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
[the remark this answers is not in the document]
atantawi: Done. Used a simple map.

apis/config/v1beta3/types.go
```diff
@@ -120,6 +120,21 @@ type LoadVariationRiskBalancingArgs struct {
 	SafeVarianceSensitivity *float64 `json:"safeVarianceSensitivity,omitempty"`
 }
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:defaulter-gen=true
+
+// LowRiskOverCommitmentArgs holds arguments used to configure LowRiskOverCommitment plugin.
+type LowRiskOverCommitmentArgs struct {
```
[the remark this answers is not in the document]
atantawi: Done.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,242 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
atantawi: OK

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,242 @@
...
+	collector, err := trimaran.NewCollector(&args.TrimaranSpec)
+	if err != nil {
+		return nil, err
+	}
+	// create map of resource risk limit weights
+	m := make(map[string]float64)
+	m[v1.ResourceCPU.String()] = 0.5
+	m[v1.ResourceMemory.String()] = 0.5
+	for _, rs := range args.RiskLimitWeights {
+		if rs.Weight >= 0 && rs.Weight <= 100 {
```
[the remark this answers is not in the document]
atantawi: Done.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,242 @@
...
+func (pl *LowRiskOverCommitment) NormalizeScore(context.Context, *framework.CycleState, *v1.Pod, framework.NodeScoreList) *framework.Status {
+	return nil
+}
+
+// computeRank : rank function for the LowRiskOverCommitment
+func (pl *LowRiskOverCommitment) computeRank(metrics []watcher.Metric, nodeInfo *framework.NodeInfo, pod *v1.Pod) float64 {
+	// get all pods on node
+	node := nodeInfo.Node()
+	nodeName := node.GetName()
+	podsOnNode := make([]*v1.Pod, 0)
```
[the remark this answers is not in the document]
atantawi: OK

pkg/trimaran/resourcestats.go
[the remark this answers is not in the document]
atantawi: Yes, agreed.

pkg/trimaran/resourcestats.go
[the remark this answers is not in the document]
atantawi: OK

pkg/trimaran/resourcestats.go
```diff
@@ -107,24 +115,34 @@ func GetResourceData(metrics []watcher.Metric, resourceType string) (avg float64
...
+// GetResourceRequested : calculate the resource requests of a pod (CPU and Memory); zero requests replaced with a minimum default
 func GetResourceRequested(pod *v1.Pod) *framework.Resource {
-	// add up demand of all containers
+	// add up requests of all containers
 	result := &framework.Resource{}
 	for _, container := range pod.Spec.Containers {
 		result.Add(container.Resources.Requests)
+		if container.Resources.Requests.Cpu().IsZero() {
+			result.MilliCPU += MinimumMilliCPURequest
+		}
```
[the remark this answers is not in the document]
atantawi: Yes

pkg/trimaran/resourcestats.go
```diff
@@ -26,6 +26,15 @@ import (
 )
 
 const (
+	// MinimumMilliCPURequest : minimum CPU request, if unspecified (or zero)
+	// (Minimum values are not used to check if a pod with zero requests container(s) fits.
+	// For the LoadVariationRiskBalancing plugin, they are used to assess load due to such a pod.
+	// For the LowRiskOverCommitment plugin, they are used to assess the range between requests
+	// and limits, which impacts the overcommitment risks.
+	// Higher minimum values increase the limit risk component and decrease the load risk component.)
+	MinimumMilliCPURequest int64 = 10 // 0.01 core
```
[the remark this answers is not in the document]
atantawi: Removed it.

pkg/trimaran/resourcestats.go
```diff
@@ -136,3 +154,150 @@ func GetResourceRequested(pod *v1.Pod) *framework.Resource {
...
+			// adjust limits if unlimited
+			if nodeUnlimitedCPU {
+				setMax(&nodeLimitMinusPod.MilliCPU, capCpu)
+			}
+			if nodeUnlimitedMemory {
+				setMax(&nodeLimitMinusPod.Memory, capMem)
+			}
+		}
+		// get requests and limits for pod
+		requested := GetResourceRequested(p)
```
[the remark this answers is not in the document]
atantawi: Thanks for the suggestion. Done.

atantawi:
In a nutshell, the behavior of this plugin is now as follows.

For the sake of this discussion, I will use the following terms. Resource usage above requests, up to limits, is referred to as **overload**. If the sum of pod limits in a node is above the allocatable, the node is **over-comitted**. Unspecified limits are assumed equal to requests. Unspecified requests are assumed zero. Also, as a matter of design principle, the plugin does not micro-manage usage and over-commitment at the container and pod levels, rather at a macro level, at the node level. Further, best effort pods (zero requests and limits) are not addressed by this plugin. As such, the plugin does not accommodate the bursting of best effort pods (hence, no opportunity for users to _game_ the plugin). The plugin is targeted for guaranteed and bursting pods. (Even though guaranteed pods are not affected, their placement on nodes affects bursting pods.)

So, starting from an idle cluster, this plugin will schedule pods on nodes such all nodes remain under-committed, favoring least over-loaded nodes. Then, as some nodes become over-committed, the plugin will favor nodes that are least over-loaded and least over-committed. This combined measure is controlled by a weight parameter (per resource), defaulting to a half. And, as all nodes become over-committed, least over-loaded nodes are favored.

Of course, the above description is rather coarse and not very precise, but it summarizes the behavior. Hope it helps.

atantawi:
Avoiding copyright issues, I replaced functions `Complete()` and `RegularizedIncomplete()` in `beta.go` to directly use package `gonum.org/v1/gonum,` which is already in required list in `go.mod`. The original,  removed code was based on code in a book.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,308 @@
...
+		excess = 0
+	}
+	maxLoad := capacity
+	if limit < capacity {
+		maxLoad = limit
+	}
+	allowed := maxLoad - request
+	if allowed < 0 {
+		allowed = 0
+	}
```
Huang-Wei, line 236: suppose request/limit/capacity as 1/1/10, then allowed=0, it doesn't look right to me.

Huang-Wei (commented):
still has some questions on the rational of calculating riskLimit. comments below

apis/config/v1/defaults.go
```diff
@@ -61,6 +61,19 @@ var (
...
 
+	// Defaults for LowRiskOverCommitment plugin
+
+	// The default number of windows over which usage data metrics are smoothed.
+	// DefaultSmoothingWindowSize is 5 (used by Prometheus)
+	DefaultSmoothingWindowSize int64 = 5
+	// The default weight of risk due to limit for a resource
+	DefaultRiskLimitWeight float64 = 0.5
+	// Resources fractional weight of risk due to limits specification [0,1]
+	DefaultRiskLimitWeights = map[v1.ResourceName]float64{
```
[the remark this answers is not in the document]
atantawi: Let me clarify. The risk of a given resource `r `(CPU or memory) on node `n` is evaluated as a weighted sum of two risks: (1) risk based on limits specifications `riskLimit` and (2) risk based on usage `riskLoad`. Let the weight be w, where 0<=w<=1, then:

> totalRisk(n, r) = w(r) * riskLimit(n, r) + (1-w(r)) * riskLoad(n, r)

The default value of w is defined by `DefaultRiskLimitWeight` as 0.5.
The user may specify different values of w, say w(CPU)=0.4 and w(memory)=0.8, as

>       riskLimitWeights:
>         cpu: 0.4
>         memory: 0.8

The totalRisk(n) of node n is taken to be the maximum of totalRisk(n, CPU) and totalRisk(n, memory), and not a weighted sum of the two quantities:

> totalRisk(n) = max (totalRisk(n, CPU), totalRisk(n, memory))

The reason being that the risk of a node is dictated by the worst risk among all resources. If a node has a high risk due to CPU over-commitment and/or overload, then the risk due to memory seems irrelevant. Another node with same CPU risk and less memory risk does not make the other node any more desirable for scheduling. That's the current logic. Please feel free to comment.

atantawi:
@Huang-Wei I'll make the changes you recommended. Please check above comments if I answered your questions. Thanks

atantawi:
@Huang-Wei done with update 3

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
Huang-Wei: I still feel this section is hard to read and can be simplified:

- if limits <= capacity, it seems we can just claim riskLimit = 0, correct? because in this case, allowed === excess
- if limits > capacity, some logic is valid, but what if limits equals requests, like request/limits/capacity as 15/15/10, it seems in this case, riskLimit is also 0?

Huang-Wei:
@atantawi a final comment https://github.com/kubernetes-sigs/scheduler-plugins/pull/412/files#r1119654048

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: @Huang-Wei Agreed. My proposal for simplification is to replace
```
// (1) riskLimit : calculate overcommit potential load
	excess := limit - request
	// This should not happen
	if excess < 0 {
		klog.V(6).InfoS("Negative excess", "node", klog.KObj(node), "resource", resourceName,
			"request", request, "limit", limit)
		excess = 0
	}
	maxLoad := capacity
	if limit < capacity {
		maxLoad = limit
	}
	allowed := maxLoad - request
	if allowed < 0 {
		allowed = 0
	}
	if excess > 0 && allowed < excess {
		riskLimit = 1 - float64(allowed)/float64(excess)
	}
```
with
```
// (1) riskLimit : calculate overcommit potential load
	if limit > capacity {
		riskLimit = float64(limit - capacity)/float64(limit - request)
	}
```

The reasons I believe it should work are:

- `riskLimit` is initialized to zero, so the case of `limit <= capacity` is handled automatically
- conditions `request <= limit` and `request <= capacity` are already ensured by `GetNodeRequestsAndLimits()`
- denominator cannot be zero since `limit > capacity` and `request <= capacity`
- if `limit == request` then it must be that `limit <= capacity`, hence `riskLimit` will remain zero

One interpretation of the simplified equation is: the numerator captures the potential of bursting beyond capacity (overload), the denominator captures the potential of bursting beyond allocation, and the ratio is the chance of overload when bursting (between 0 and 1).

Now, we come to the case "requests/limits/capacity as 15/15/10". This case will not arise since requests (15) cannot be larger than capacity (10). (The pods will not fit on the node.) If that was 8/8/10, then the risk is 0.

I will make the change, unless you have other suggestions. Thanks.

atantawi:
Addressed comment in https://github.com/kubernetes-sigs/scheduler-plugins/pull/412#discussion_r1147941924

atantawi:
@Huang-Wei This PR is waiting on lgtm. All comments addressed already. Thanks

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
```diff
@@ -0,0 +1,290 @@
...
+		capacity = nodeCapacity.Memory
+	} else {
+		// invalid resource
+		klog.V(6).InfoS("Unexpected resource", "resourceName", resourceName)
+		return 0
+	}
+
+	// (1) riskLimit : calculate overcommit potential load
+	if limit > capacity {
+		riskLimit = float64(limit-capacity) / float64(limit-request)
```
Huang-Wei, line 207: Thanks @atantawi. Sorry for late response. The new simplified logic calculating risk sounds good. It's just could you make it UTed? in `TestLowRiskOverCommitment_computeRisk`.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment_test.go
```diff
@@ -0,0 +1,398 @@
...
+		node                  *v1.Node
+		nodeRequestsAndLimits *trimaran.NodeRequestsAndLimits
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		want   float64
+	}{
+		{name: "test-cpu",
```
Huang-Wei, line 324: nit: put it in a new line is more conventional (applies elsewhere)

```suggestion
		{
			name: "test-cpu",
```

pkg/trimaran/lowriskovercommitment/lowriskovercommitment_test.go
```diff
@@ -0,0 +1,398 @@
...
+		resourceName          v1.ResourceName
+		resourceType          string
+		node                  *v1.Node
+		nodeRequestsAndLimits *trimaran.NodeRequestsAndLimits
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		want   float64
```
Huang-Wei, line 322: could you simplify the args building by just plumbing the variants in each sub-test? If i read the code right, the tests can be:

```go
tests := []struct {
		name         string
		resourceName v1.ResourceName
		resourceType string
		want         float64
```

and per previous comment, it'd be desirable to test different kinds of combinations of requests/limits/capacity, so plumb in necessary parameters as well.

test/integration/lowriskovercommitment_test.go
```diff
@@ -0,0 +1,205 @@
...
+		node := st.MakeNode().Name(nodeNames[i]).Label("node", nodeNames[i]).Capacity(capacity).Obj()
+		_, err := cs.CoreV1().Nodes().Create(testCtx.Ctx, node, metav1.CreateOptions{})
+		assert.Nil(t, err)
+	}
+
+	// Pods divided into numExistingPods already scheduled and the rest to be scheduled
+	var pods []*corev1.Pod
+	podNames := []string{"pod-1", "pod-2", "pod-3"}
+	requestCPU := []int64{500, 500, 500}
+	limitCPU := []int64{500, 500, 1000}
```
Huang-Wei, line 159: This test case seems to be too simple to "prove" this plugin functions, even in a happy path.

Usually it's fine to just compose a happy path test in integration test, and plumb complex ones in UTs. So here I'd like to see a test case to **pass and only pass** when this plugin is enabled.

pkg/trimaran/lowriskovercommitment/lowriskovercommitment.go
atantawi: No problem, thanks

pkg/trimaran/lowriskovercommitment/lowriskovercommitment_test.go
atantawi: Done

pkg/trimaran/lowriskovercommitment/lowriskovercommitment_test.go
atantawi: Done

test/integration/lowriskovercommitment_test.go
atantawi: Done

Huang-Wei: We're close, but not there yet. For now, w/ or w/o this plugin's enablement, the pod always land on node-1, I'd like to apply this change:

```diff
diff --git a/test/integration/lowriskovercommitment_test.go b/test/integration/lowriskovercommitment_test.go
index e888291d..c64a73e4 100644
--- a/test/integration/lowriskovercommitment_test.go
+++ b/test/integration/lowriskovercommitment_test.go
@@ -139,7 +139,7 @@ func TestLowRiskOverCommitmentPlugin(t *testing.T) {
 
 	// Cluster with two nodes
 	nodeNames := []string{"node-1", "node-2"}
-	capCPU := []string{"2", "1"}
+	capCPU := []string{"2", "2"}
 	capMemory := []string{"256", "256"}
 	for i := 0; i < len(nodeNames); i++ {
 		t.Logf("Creating Node %q", nodeNames[i])
@@ -156,7 +156,7 @@ func TestLowRiskOverCommitmentPlugin(t *testing.T) {
 	var pods []*corev1.Pod
 	podNames := []string{"pod-1", "pod-2", "pod-3"}
 	requestCPU := []int64{500, 100, 500}
-	limitCPU := []int64{500, 500, 1000}
+	limitCPU := []int64{500, 1200, 1000}
 	var requestMemory int64 = 64
 	var limitMemory int64 = 64
 	scheduledNodes := []string{"node-1", "node-2"}
```

So that:

- w/o this plugin's enablement, by default using LeastAllocated, it'd land on node-2
- w/ this plugin's enablement, it's expected to land on node-1

then we're able to catch potential regression introduced someday in the future.

Huang-Wei (commented):
one final comment. 

@atantawi along with resolving that comment, could you squash the commits into 3? API changes into one, vendor/ as another one, and core logic as the last one.

test/integration/lowriskovercommitment_test.go
atantawi: Perfect.

ok  	sigs.k8s.io/scheduler-plugins/test/integration	559.475s

Huang-Wei:
/lgtm
/approve

Thanks @atantawi !

Issue #419, opened 2022-08-25
[Helm] support customizing pluginConfig


Huang-Wei:
Follow up of https://github.com/kubernetes-sigs/scheduler-plugins/issues/379#issuecomment-1227798395:

> the recent Helm chart refactoring PR https://github.com/kubernetes-sigs/scheduler-plugins/pull/326 doesn't support pluginConfig's customization. And in recent refactoring to leverage v1beta3, this support isn't added either. I will create an issue to track on this.

It'd be nice to add support for customizing pluginConfig in `as-a-second-scheduler/templates/configmap.yaml`.

lianghao208:
For now, I think only coscheduling, noderesources, noderesourcetopology and trimaran have their onw pluginConfig. I'll work on pluginConfig's customization of these 4 plugins. @Huang-Wei

Huang-Wei:
@lianghao208 could you help cherry-pick this to `release-1.23`?

lianghao208:
> @lianghao208 could you help cherry-pick this to `release-1.23`?

@Huang-Wei Sure, working on it.

PR #421, opened 2022-08-27
enhancement: support customizing pluginConfig

lianghao208:
Fixes: #419 
/release-note-none

```
 NONE
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/values.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/values.yaml
@@ -23,3 +23,15 @@ plugins:
   enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch","NodeResourcesAllocatable"]
   disabled: ["PrioritySort"] # only in-tree plugins need to be defined here
 
+# Customize the enabled plugins' config.
+# Refer to the "pluginConfig" section of manifests/<plugin>/scheduler-config.yaml.
+# For example, for Coscheduling plugin, you want to customize the permit waiting timeout to 10 seconds:
+pluginConfig:
+- name: Coscheduling
+  args:
+    permitWaitingTimeSeconds: 10 # default is 60
+# Or, customize the other plugins
+# - name: NodeResourceTopologyMatch
+#   args:
+#     scoringStrategy:
+#       type: MostAllocated # default is LeastAllocated
--- a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
@@ -23,6 +23,9 @@ data:
           {{- range $.Values.plugins.disabled }}
           - name: {{ title . }}
           {{- end }}
+      {{- if $.Values.pluginConfig }}
+      pluginConfig: {{ toYaml $.Values.pluginConfig | nindent 6 }}
+      {{- end }}
 
   {{- /* TODO: wire CRD installation with enabled plugins. */}}
 {{- end }}


manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -23,3 +23,33 @@ plugins:
   enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch","NodeResourcesAllocatable"]
   disabled: ["PrioritySort"] # only in-tree plugins need to be defined here
 
+pluginConfig:
+  plugins:
+    - Coscheduling
+    - NodeResourceTopologyMatch
+    - NodeResourcesAllocatable
+    - LoadVariationRiskBalancing
+    - TargetLoadPacking
```
Huang-Wei, line 32: Is it possible to remove this section? ideally, we should refrain from adding new "formats" of config if unnecessary.

In other words, does it technically work by incorporating the original pluginConfig:

```yaml
pluginConfig:
- name: foo
  args:
- name: bar
  args:
```

Huang-Wei (commented):
Thanks for working on this.

manifests/install/charts/as-a-second-scheduler/values.yaml
lianghao208: That's a good point, I've done some changes, PTAL.

manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -23,3 +23,27 @@ plugins:
...
+    scoringStrategy:
+      type: "LeastAllocated"
+- name: NodeResourcesAllocatable
+  args:
+    mode: Least
+    resources:
+      - name: cpu
+        weight: 1000000
+      - name: memory
+        weight: 1
```
Huang-Wei, line 42: let's remove this part as well.

manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -23,3 +23,27 @@ plugins:
   enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch","NodeResourcesAllocatable"]
   disabled: ["PrioritySort"] # only in-tree plugins need to be defined here
 
+pluginConfig:
+- name: Coscheduling
+  args:
+    permitWaitingTimeSeconds: 10
+    deniedPGExpirationTimeSeconds: 3
```
Huang-Wei, line 30: this field is now removed in v1beta3.

manifests/install/charts/as-a-second-scheduler/values.yaml
```diff
@@ -23,3 +23,27 @@ plugins:
   enabled: ["Coscheduling","CapacityScheduling","NodeResourceTopologyMatch","NodeResourcesAllocatable"]
   disabled: ["PrioritySort"] # only in-tree plugins need to be defined here
 
+pluginConfig:
```
Huang-Wei, line 26: Let's add some comments in this section:

```yaml
# Customize the enabled plugins' config.
# Refer to the "pluginConfig" section of manifests/<plugin>/scheduler-config.yaml.
# For example, for Coscheduling plugin, you want to customize the permit waiting timeout to 10 seconds:
- name: Coscheduling
  args:
    permitWaitingTimeSeconds: 10 # default is 60
# Or, customize the other plugins
# - name: NodeResourceTopologyMatch
#   args:
#     scoringStrategy:
#       type: MostAllocated # default is LeastAllocated
```

lianghao208: done.

lianghao208: @Huang-Wei I've updated this section, do I need to add more configs such as `manifests/<plugin>/scheduler-config-v1beta3.yaml` to support v1beta3?

Huang-Wei: We can leave that as a follow-up.

PR #425, opened 2022-09-09
noderesourcetopology: refine best-effort pod early exit logic

swatisehgal:

In the current implementation for best effort pods, we exit and leave it
the responsibility of scheduling such a pod to other scheduler plugins
which might still result in TopologyAffinityError.

The correct behavior would be to exit return early ONLY if we are 
evaluating best effort pods AND they are not requesting devices 
(non-native resources).

It is important to handle burstable and best effort pods requesting devices
carefully because irrespective of the QoS class of the pod requesting devices,
devices are always exclusively allocated unlike CPU and memory which are
allocated from the shared pool in case of BE/burstable pods.

Check related to whether requested resource can fit on a NUMA node
in case of non-guaranteed pods is already done in `isNUMANodeSuitable`
helper function so we don't need to duplicate that logic.

#### What type of PR is this?
/kind bug

```release-note
noderesourcetopology: refine best-effort pod early exit logic
```
<!--

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -191,7 +191,7 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
 	if nodeInfo.Node() == nil {
 		return framework.NewStatus(framework.Error, "node not found")
 	}
-	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort {
+	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort && !hasNonNativeResource(pod) {
 		return nil
 	}
 
@@ -235,4 +235,24 @@ func subtractFromNUMA(nodes NUMANodeList, numaID int, container v1.Container) {
 			nRes[resName] = nodeResQuan
 		}
 	}
+}
+
+func hasNonNativeResource(pod *v1.Pod) bool {
+	for _, initContainer := range pod.Spec.InitContainers {
+		for resource := range initContainer.Resources.Requests {
+			if !v1helper.IsNativeResource(resource) {
+				return true
+			}
+
+		}
+	}
+	for _, container := range pod.Spec.Containers {
+		for resource := range container.Resources.Requests {
+			if !v1helper.IsNativeResource(resource) {
+				return true
+			}
+
+		}
+	}
+	return false
 }
--- a/pkg/noderesourcetopology/filter_test.go
+++ b/pkg/noderesourcetopology/filter_test.go
@@ -232,6 +232,34 @@ func TestNodeResourceTopology(t *testing.T) {
 			wantStatus: nil,
 		},
 		{
+			name: "Best effort QoS requesting devices, Pod Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[2],
+			wantStatus: nil,
+		},
+		{
+			name: "Best effort QoS requesting devices, Pod Scope Topology policy; pod doesn't fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				nicResourceName: *resource.NewQuantity(20, resource.DecimalSI)}),
+			node:       nodes[2],
+			wantStatus: framework.NewStatus(framework.Unschedulable, "cannot align pod: "),
+		},
+		{
+			name: "Best effort QoS requesting devices, Container Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[1],
+			wantStatus: nil,
+		},
+		{
+			name: "Best effort QoS requesting devices, Container Scope Topology policy; pod doesn't fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				nicResourceName: *resource.NewQuantity(20, resource.DecimalSI)}),
+			node:       nodes[0],
+			wantStatus: framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align container: %s", containerName)),
+		},
+		{
 			name: "Guaranteed QoS, minimal, pod fit",
 			pod: makePodByResourceList(&v1.ResourceList{
 				v1.ResourceCPU:    *resource.NewQuantity(2, resource.DecimalSI),
@@ -291,6 +319,38 @@ func TestNodeResourceTopology(t *testing.T) {
 				nicResourceName: *resource.NewQuantity(11, resource.DecimalSI)}),
 			node:       nodes[1],
 			wantStatus: framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align container: %s", containerName)),
+		},
+		{
+			name: "Burstable QoS requesting CPU (enough on NUMA) and devices, Pod Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				v1.ResourceCPU:  *resource.NewQuantity(2, resource.DecimalSI),
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[2],
+			wantStatus: nil,
+		},
+		{
+			name: "BurstableQoS requesting CPU (not enough on NUMA) and devices, Pod Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				v1.ResourceCPU:  *resource.NewQuantity(20, resource.DecimalSI),
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[2],
+			wantStatus: nil,
+		},
+		{
+			name: "Burstable QoS requesting CPU (enough on NUMA) and devices, Container Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				v1.ResourceCPU:  *resource.NewQuantity(2, resource.DecimalSI),
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[1],
+			wantStatus: nil,
+		},
+		{
+			name: "Burstable QoS requesting CPU (not enough on NUMA) and devices, Container Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				v1.ResourceCPU:  *resource.NewQuantity(4, resource.DecimalSI),
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[1],
+			wantStatus: nil,
 		},
 		{
 			name: "Guaranteed QoS, hugepages, pod doesn't fit",
--- a/pkg/noderesourcetopology/filter_test.go
+++ b/pkg/noderesourcetopology/filter_test.go
@@ -226,6 +226,15 @@ func TestNodeResourceTopology(t *testing.T) {
 			wantStatus: nil,
 		},
 		{
+			name: "Guaranteed QoS, pod with extended resource no devices; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				v1.ResourceCPU:    *resource.NewQuantity(2, resource.DecimalSI),
+				v1.ResourceMemory: resource.MustParse("2Gi"),
+				extended:          resource.MustParse("1")}),
+			node:       nodes[4],
+			wantStatus: nil,
+		},
+		{
 			name:       "Best effort QoS, pod fit",
 			pod:        &v1.Pod{},
 			node:       nodes[0],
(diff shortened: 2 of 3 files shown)


pkg/noderesourcetopology/filter.go
```diff
@@ -150,21 +150,25 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 }
 
 func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, numaQuantity resource.Quantity) bool {
-	// Check for the following:
+	// // Check for the following:
```
fromanirh, line 153: why the double `//`?

fromanirh (commented):
seems fine bar a minor nit. I'd like to have a deeper pass later to be really sure.

pkg/noderesourcetopology/filter.go
swatisehgal: Was a mistake, fixed.

fromanirh (commented):
thanks for fixing the nit :) looks good, let me just have a deeper pass later to doublecheck

pkg/noderesourcetopology/filter_test.go
```diff
@@ -231,6 +231,66 @@ func TestNodeResourceTopology(t *testing.T) {
...
+		{
+			name: "Best effort QoS requesting devices, Pod Scope Topology policy; pod fit",
+			pod: makeBEPodByResourceList(&v1.ResourceList{
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[2],
+			wantStatus: nil,
+		},
+		{
+			name: "Best effort QoS requesting CPU (enough on NUMA) and devices, Pod Scope Topology policy; pod fit",
+			pod: makeBEPodByResourceList(&v1.ResourceList{
```
shajmakh, line 243: A pod that requests one of memory or cpus is counted as of burstable QoS, will be best-effort only if not specifying any of CPUs and memory. reference: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#create-a-pod-that-gets-assigned-a-qos-class-of-burstable 
same note applies for the multiple similar tests below. Maybe also need to adjust the name of makeBEPodByResourceList to not specify BE (as in best effort).

swatisehgal: Good point, will update.

swatisehgal: Now that I am rethinking this, we don't actually need a separate function for creating BE or burstable pods. QoS can be influenced by changing the resources passed to `makePodByResourceList`.

pkg/noderesourcetopology/filter_test.go
```diff
@@ -231,6 +231,66 @@ func TestNodeResourceTopology(t *testing.T) {
...
 		},
+		{
+			name: "Best effort QoS requesting devices, Pod Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[2],
+			wantStatus: nil,
+		},
+		{
+			name: "Best effort QoS requesting CPU (enough on NUMA) and devices, Pod Scope Topology policy; pod fit",
```
fromanirh, line 242: let's make sure to fix the name of the testcases to be consistent with the expected pod QoS

swatisehgal: Fixed. Thanks

pkg/noderesourcetopology/filter_test.go
```diff
@@ -292,6 +320,38 @@ func TestNodeResourceTopology(t *testing.T) {
...
+			wantStatus: nil,
+		},
+		{
+			name: "Burstable QoS requesting CPU (not enough on NUMA) and devices, Container Scope Topology policy; pod fit",
+			pod: makePodByResourceList(&v1.ResourceList{
+				v1.ResourceCPU:  *resource.NewQuantity(4, resource.DecimalSI),
+				nicResourceName: *resource.NewQuantity(5, resource.DecimalSI)}),
+			node:       nodes[1],
+			wantStatus: nil,
+		},
```
fromanirh, line 371: to avoid introducing regressions: do we have tests covering the case on which we have burstable/best-effort pods requesting cpu/memory/both such that requested quantity is not available on a single NUMA zone? these pods should still be admitted (per explanation in the commit message)

fromanirh (commented):
the change looks good. I think we should have tests minimizing the risk of regressions. Perhaps are there already, didn't look at them deeply yet (will do shortly).
Once we have some coverage to prevent regressions, LGTM.

pkg/noderesourcetopology/filter_test.go
swatisehgal: Yeah right, I added more test cases to cover these scenarios.

Tal-or: > do we have tests covering the case on which we have burstable/best-effort pods requesting cpu/memory/both such that requested quantity is not available on a single NUMA zone

Isn't the described state a bit of a paradox?
I mean if it's a burstable/best-effort pod it means it shares its resources with others (pod/services) i.e. we don't care about performance so much.
Hence, why do we care from which NUMA the resources are coming?

Tal-or: > to avoid introducing regressions: do we have tests covering the case on which we have burstable/best-effort pods requesting cpu/memory/both such that requested quantity is not available on a single NUMA zone?

Isn't the described above a bit of a paradox? 
I mean if it's a burstable/best-effort pod it means it shares its resources with other pods/services, i.e. this pod doesn't care about performance so much.
Hence, why the scheduler should care about NUMA alignment in that case?

According to the [docs](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#policy-single-numa-node ) we won't end up with `TopologyAffinityErrror` anyway.
and I quote: 
_In the case of the BestEffort pod, the static CPU Manager policy would send back the default topology hint as there is no CPU request and the Device Manager would send back the hints for each of the requested devices._

swatisehgal: > > to avoid introducing regressions: do we have tests covering the case on which we have burstable/best-effort pods requesting cpu/memory/both such that requested quantity is not available on a single NUMA zone?
> 
> Isn't the described above a bit of a paradox? I mean if it's a burstable/best-effort pod it means it shares its resources with other pods/services, i.e. this pod doesn't care about performance so much. Hence, why the scheduler should care about NUMA alignment in that case?
> 
> According to the [docs](https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#policy-single-numa-node) we won't end up with `TopologyAffinityErrror` anyway. and I quote: _In the case of the BestEffort pod, the static CPU Manager policy would send back the default topology hint as there is no CPU request and the Device Manager would send back the hints for each of the requested devices._

We need to take into consideration that CPU and memory are handled differently from devices in case of non-gu pods. In case of burstable and best effort pods, CPU and memory is not allocated exclusively and therefore such pods obtain resources from the shared pool. Devices however, are always exclusively allocated (as per the device plugin API design) irrespective of pod QoS. As per your quote above, `In the case of the BestEffort pod, ... Device Manager sends back hints for each requested device` because of which we _should_ check if the requested devices are available on any of the NUMA nodes on a candidate node. Without the check proposed in this PR, the pod could potentially get scheduled on a node for it to be later rejected with `TopologyAffinityErrror` by Topology Manager.

fromanirh (commented):
LGTM

pkg/noderesourcetopology/filter_test.go
Tal-or: > We need to take into consideration that CPU and memory are handled differently from devices in case of non-gu pods. In case of burstable and best effort pods, CPU and memory is not allocated exclusively and therefore such pods obtain resources from the shared pool. Devices however, are always exclusively allocated (as per the device plugin API design) irrespective of pod QoS. As per your quote above, In the case of the BestEffort pod, ... Device Manager sends back hints for each requested device because of which we should check if the requested devices are available on any of the NUMA nodes on a candidate node. Without the check proposed in this PR, the pod could potentially get scheduled on a node for it to be later rejected with TopologyAffinityErrror by Topology Manager.

I think we're basically saying the same thing.
I was talking about adding more cases with non-gu pods that are asking for CPU/memory - This won't cause  ```TopologyAffinityErrror``` what so ever. only wrong device alignment

swatisehgal: > Just doublechecking: this is in the narrow, but still relevant, case of a pod requesting _only_ devices, all of which should be available on the same NUMA node. Right?

Yep, I have some test cases [here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/425/commits/149ee7de1ff9ba64efe03d840a736c295e74ba6d#diff-3da058c06e05f2fc9bfa2e6f69abd193781cc354b75c44c4660dc7caaeadf38cR235-R261) corresponding to pods requesting only devices all of which should be allocated from the same NUMA node.

swatisehgal: > I think we're basically saying the same thing. I was talking about adding more cases with non-gu pods that are asking for CPU/memory - This won't cause `TopologyAffinityErrror` what so ever.
Ah I see. Apologies, I misunderstood your comment previously. I already added more test cases for non-gu pods in the second commit. Let me know if you think I am missing any case that should be added. 
>  Please keep me honest here, but with non-gu pod, only wrong device alignment can cause `TopologyAffinityErrror`
Yes, that's right!

pkg/noderesourcetopology/filter.go
```diff
@@ -152,19 +152,23 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, numaQuantity resource.Quantity) bool {
 	// Check for the following:
 	if qos != v1.PodQOSGuaranteed {
-		// 1. set numa node as possible node if resource is memory or Hugepages
+		// 1. set numa node as possible node if resource is a non-native resources but sufficiently available on NUMA node(e.g. device)
```
pendoragon, line 155: I got a bit confused by the code here. CMIIW, the code before this patch says for guaranteed pod, check for every resource type if current numa node is sufficient; for non guaranteed pod, ignore `memory`, `hugepage` and `cpu`, and check for the rest of the resource types if current numa node is sufficient. so for non guaranteed pod, the non-native resources are already checked?

Tal-or: > for non guaranteed pod, ignore memory, hugepage and cpu, and check for the rest of the resource types if current numa node is sufficient. so for non guaranteed pod, the non-native resources are already checked?

(A) With non guaranteed pods the **previous** code devides into 2:
1. `BestEffort` pods - return immediately, IOW mark this node as a suitable one.
2.  `Burstable` pods -  `memory`, `hugepage` and `cpu` (AKA native resources) are not taken into account for `Burstable` pod so we don't need to check them at the NUMA level. And yes, native  resources already checked node-wide in the NodeResourceFit plugin which part of the native scheduler and prior to this plugin: https://github.com/kubernetes/kubernetes/blob/v1.25.2/pkg/scheduler/framework/plugins/noderesources/fit.go#L218 

(B) With non guaranteed pods **this** code has the same logic for both `BestEffort` and `Burstable` pods:
It checks whether a non-native device can be accounted for the pods, since unlike native resources, devices cannot 
  be shared among pods, even if they're not guaranteed ones.
  and then continues with step (A)2

I hope it addresses your question.

pendoragon: Thanks for the explanation! I guess i'm on the same page now.
But according to the previous code, the non-native resources will be checked as well for `BestEffort` and `Burstable` pods I guess🤔? e.g. for `BestEffort` pods, the flow will fall through to L172  which is `return numaQuantity.Cmp(quantity) >= 0`, which means if I request devices in best effort pod, the quantity will be checked at numa level

Tal-or: Not exactly :)
On the previous code,  for `BestEffort` pods we would immediately return here:
https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/filter.go#L194

pendoragon: aah sorry I missed that part xD. so `BestEffort` pod is not a proper example. But then it's the same thing with Burstable pod, i.e. for Burstable pod, if I request devices, the quantity is already checked by previous code? 

So to handle best effort and burstable pods with devices, I think we only have to remove https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/filter.go#L194-L196? Actually if I comment out the changes in function `isNUMANodeSuitable`, the test case will still pass.

Tal-or: hmm, you got a point.
I tend to agree, because if the resource is not one of the native resources it would fall off to the check on:
https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/filter.go#L168

pkg/noderesourcetopology/filter.go
```diff
@@ -191,9 +195,6 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
 	if nodeInfo.Node() == nil {
 		return framework.NewStatus(framework.Error, "node not found")
 	}
-	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort {
-		return nil
-	}
```
Huang-Wei, line 196: My 2 cents: pre-check if the Pod is "best efforts" is still the semantics we're pursuing - so that we do early return. It's just v1qos.GetPodQOS() is not the right function to use which only checks CPU and memory:

```go
var supportedQoSComputeResources = sets.NewString(string(core.ResourceCPU), string(core.ResourceMemory))
```

So WDYT on composing a "best-efforts++" function to filter out the pods this plugin doesn't care?

pkg/noderesourcetopology/filter.go
pendoragon: Yes exactly!

pkg/noderesourcetopology/filter.go
```diff
@@ -152,19 +152,23 @@ func resourcesAvailableInAnyNUMANodes(logKey string, numaNodes NUMANodeList, res
 func isNUMANodeSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, numaQuantity resource.Quantity) bool {
 	// Check for the following:
 	if qos != v1.PodQOSGuaranteed {
-		// 1. set numa node as possible node if resource is memory or Hugepages
+		// 1. set numa node as possible node if resource is a non-native resources but sufficiently available on NUMA node(e.g. device)
+		if !v1helper.IsNativeResource(resource) && numaQuantity.Cmp(quantity) >= 0 {
```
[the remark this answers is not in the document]
swatisehgal: Ack, Thanks.

pkg/noderesourcetopology/filter.go
swatisehgal: Thanks @pendoragon and @Tal-or, what you are saying is totally correct and is a great catch! 

As per @Huang-Wei's suggestion [here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/425#discussion_r985297798), I will add a check and return early in case we have `BestEffort` pods without devices which would rule out a bunch of unwanted cases. Checks for requested quantity for both best effort pods and burstable pods would be handled properly automatically [here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/filter.go#L168).

I will add this change as a separate commit for clarity and squash into the 1st commit once everyone is happy.

Thanks again for your input on this.

pkg/noderesourcetopology/filter.go
swatisehgal: Sure, sounds good. Instead of returning early if the pod belongs to `BestEffort` QoS, we need to return if pod belongs to `BestEffort` AND doesn't request devices (non-native resources). I will add a helper function to determine if the pod is requesting non-native resources.

Huang-Wei:
/approve

Would leave /lgtm to @Tal-or and @fromanirh .

fromanirh (commented):
initial review

pkg/noderesourcetopology/filter.go
```diff
@@ -195,6 +191,9 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
 	if nodeInfo.Node() == nil {
 		return framework.NewStatus(framework.Error, "node not found")
 	}
+	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort && !hasNonNativeResource(pod) {
```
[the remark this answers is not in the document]
swatisehgal: Sure, I will fix that ASAP.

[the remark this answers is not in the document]
swatisehgal: Fixed.

Tal-or (commented):
LGTM

fromanirh (commented):
/lgtm

PR #432, opened 2022-09-27
PR for inclusion of the networkAware plugins

jpedro1992:
**What type of PR is this?**

PR for networkAware plugins approved here https://github.com/kubernetes-sigs/scheduler-plugins/pull/282.

The PR includes two plugins (NetworkOverhead and TopologicalSort).

/kind feature

**What this PR does / why we need it:**

This proposal introduces an end-to-end solution to model/weight a cluster's network latency and topological information, and leverage that to better schedule latency- and bandwidth-sensitive workloads. Further information is available in our merged [KEP](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/260-network-aware-scheduling).

**Special notes for your reviewer:**

After feedback received in the previous PR https://github.com/kubernetes-sigs/scheduler-plugins/pull/348, we have hosted both CRDs and controllers outside of sig-scheduling and included only the networkAware plugins in the new PR.  

We decided to open a new PR since the code changed significantly and most components in the previous PR are now hosted in [diktyo-io](https://github.com/diktyo-io). 

All feedback is welcome. Thanks.

Extra components hosted in diktyo-io:

[AppGroup CRD](https://github.com/diktyo-io/appgroup-api) 
[NetworkTopology CRD](https://github.com/diktyo-io/networktopology-api) 
[AppGroup controller](https://github.com/diktyo-io/appgroup-controller) 
[NetworkTopology controller](https://github.com/diktyo-io/networktopology-controller) 

Cc: @Huang-Wei @wangchen615 @ingvagabund

Thanks in advance. I look forward to your feedback!


pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+	var ok error
+
+	numOK, numNotOK, ok = checkMaxNetworkCostRequirements(scheduledList, dependencyList, nodeInfo, region, zone, costMap, no)
+	if ok != nil {
+		return framework.NewStatus(framework.Error, "pod hostname not found")
+	}
+
+	klog.V(4).Infof("NumNotOk: %v / numOK: %v ", numNotOK, numOK)
+
+	if numNotOK > numOK {
```
Huang-Wei, line 205: shouldn't it be a yes or no logic? why compare the quantifies here?

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+	// check if maxNetworkCost fits
+	for _, podAllocated := range scheduledList { // For each pod already allocated
+		if podAllocated.Hostname != "" { // if hostname not empty...
+			for _, d := range dependencyList { // For each pod dependency
+				if podAllocated.Selector == d.Workload.Selector { // If the pod allocated is an established dependency
+					if podAllocated.Hostname == nodeInfo.Node().Name { // If the Pod hostname is the node being filtered, requirements are checked via extended resources
+						numOK += 1
+					} else { // If Nodes are not the same
+						// Get NodeInfo from pod Hostname
+						podHostname, err := no.handle.SnapshotSharedLister().NodeInfos().Get(podAllocated.Hostname)
```
Huang-Wei, line 431: move this out? as it's likely to be reused in most cases.

Huang-Wei: podHostname -> nodeInfo

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+							} else { // belong to a different zone, check maxNetworkCost
+								cost, costOK := costMap[networkawareutil.CostKey{ // Retrieve the cost from the map (origin: zone, destination: pod zoneHostname)
+									Origin:      zone, // Time Complexity: O(1)
+									Destination: zonePodHostname,
+								}]
+								if costOK {
+									if cost <= d.MaxNetworkCost {
+										numOK += 1
+									} else {
+										numNotOK += 1
```
Huang-Wei, line 455: in this case, there is some constraint violated, instead of comparing the number of the violations with some other number, shouldn't we return to fail immediately?

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+						} else { // belong to a different region
+							cost, costOK := costMap[networkawareutil.CostKey{ // Retrieve the cost from the map (origin: zone, destination: pod zoneHostname)
+								Origin:      region, // Time Complexity: O(1)
+								Destination: regionPodHostname,
+							}]
+							if costOK {
+								if cost <= d.MaxNetworkCost {
+									numOK += 1
+								} else {
+									numNotOK += 1
```
Huang-Wei, line 468: ditto.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+						}
+					}
+				}
+			}
+		}
+	}
+	return cost, nil
+}
+
+func findAppGroupNetworkOverhead(agName string, no *NetworkOverhead) (*agv1alpha1.AppGroup, error) {
```
Huang-Wei, line 540: better passing a simple lister instead the plugin?

```suggestion
func findAppGroupNetworkOverhead(agName string, agLister aglisters.AppGroupLister) (*agv1alpha1.AppGroup, error) {
```

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+	// SameZone : If pods belong to hosts in the same zone, then consider cost as 1
+	SameZone = 1
+)
+
+// NetworkOverhead : Filter and Score nodes based on Pod's AppGroup requirements: MaxNetworkCosts requirements among Pods with dependencies
+type NetworkOverhead struct {
+	handle      framework.Handle
+	podLister   corelisters.PodLister
+	agLister    *aglisters.AppGroupLister
+	ntLister    *ntlisters.NetworkTopologyLister
```
Huang-Wei, line 58: AppGroupLister and NetworkTopologyLister are interfaces, usually we don't apply another `*` to them:

```
agLister    aglisters.AppGroupLister
ntLister    ntlisters.NetworkTopologyLister
```

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+
+func findAppGroupNetworkOverhead(agName string, no *NetworkOverhead) (*agv1alpha1.AppGroup, error) {
+	klog.V(6).Infof("namespaces: %s", no.namespaces)
+	var err error
+	for _, namespace := range no.namespaces {
+		klog.V(6).Infof("ag.lister: %v", no.agLister)
+
+		// AppGroup could not be placed in several namespaces simultaneously
+		lister := no.agLister
+		appGroup, err := (*lister).AppGroups(namespace).Get(agName)
```
Huang-Wei, line 548: If you chance the definition in NetworkOverhead, you don't need to deference the interface object anymore.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+}
+
+func findNetworkTopologyNetworkOverhead(no *NetworkOverhead) (*ntv1alpha1.NetworkTopology, error) {
+	klog.V(6).Infof("namespaces: %s", no.namespaces)
+	var err error
+	for _, namespace := range no.namespaces {
+		klog.V(6).Infof("nt.lister: %v", no.ntLister)
+		// NetworkTopology could not be placed in several namespaces simultaneously
+		lister := no.ntLister
+		networkTopology, err := (*lister).NetworkTopologies(namespace).Get(no.ntName)
```
Huang-Wei, line 567: ditto.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+			continue
+		}
+		if appGroup != nil {
+			return appGroup, nil
+		}
+	}
+	return nil, err
+}
+
+func findNetworkTopologyNetworkOverhead(no *NetworkOverhead) (*ntv1alpha1.NetworkTopology, error) {
```
Huang-Wei, line 560: ditto. pass-in the specific lister.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+		if nodeScore.Score < min {
+			min = nodeScore.Score
+		}
+	}
+
+	return min, max
+}
+
+// sortNetworkTopologyCosts : sort costs if manual weights were selected
+func sortNetworkTopologyCosts(no *NetworkOverhead, networkTopology *ntv1alpha1.NetworkTopology) {
```
Huang-Wei, line 355: replace `no` with `weightsName`?

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+	if no.weightsName != ntv1alpha1.NetworkTopologyNetperfCosts { // Manual weights were selected
+		for _, w := range networkTopology.Spec.Weights {
+			// Sort Costs by TopologyKey, might not be sorted since were manually defined
+			sort.Sort(networkawareutil.ByTopologyKey(w.TopologyList))
+		}
+	}
+	return
```
Huang-Wei, line 362: this is redundant.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+}
+
+// populateCostMap : Populates costMap based on the node being filtered/scored
+func populateCostMap(no *NetworkOverhead, costMap map[networkawareutil.CostKey]int64, networkTopology *ntv1alpha1.NetworkTopology, region string, zone string) {
+	seen := false
+	for _, w := range networkTopology.Spec.Weights { // Check the weights List
+		if w.Name == no.weightsName { // If its the Preferred algorithm
```
Huang-Wei, line 369: if we do `if w.Name != no.weightsName then continue`, we can save indents, that will make the code more readable.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+				costs := networkawareutil.FindOriginCosts(topologyList, zone)
+
+				// Add Zone Costs
+				for _, c := range costs {
+					costMap[networkawareutil.CostKey{ // Add the cost to the map
+						Origin:      zone,
+						Destination: c.Destination}] = c.NetworkCost
+				}
+			}
+			seen = true
```
Huang-Wei, line 408: Isn't it equivalent to an immediate `return` from this entire function?

Huang-Wei (commented):
Thanks @jpedro1992 !

First round of review. (TODO: score and tests)

BTW: would you mind crafting an integration test?

apis/config/v1beta2/types.go
```diff
@@ -146,3 +146,27 @@ type NodeResourceTopologyMatchArgs struct {
 
 // PreemptionTolerationArgs reuses DefaultPluginArgs.
 type PreemptionTolerationArgs schedulerconfigv1beta2.DefaultPreemptionArgs
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type TopologicalSortArgs struct {
```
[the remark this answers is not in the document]
jpedro1992: Ok! I removed it from v1beta2.

apis/config/v1beta3/types.go
```diff
@@ -150,3 +150,27 @@ type NodeResourceTopologyMatchArgs struct {
...
+	metav1.TypeMeta `json:",inline"`
+
+	// Namespaces to be considered by NetworkMinCost plugin
+	Namespaces []string `json:"namespaces,omitempty"`
+
+	// Preferred weights (Default: UserDefined)
+	WeightsName *string `json:"weightsName,omitempty"`
+
+	// The NetworkTopology CRD name
+	NetworkTopologyName *string `json:"networkTopologyName,omitempty"`
```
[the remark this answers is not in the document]
jpedro1992: I would keep it as a string since users can decide which name to give to their weights and networkTopology CRs. We consider a few names as default, but users are free to name them according to their liking.

apis/config/scheme/scheme_test.go
```diff
@@ -18,6 +18,8 @@ package scheme
 
 import (
 	"bytes"
+	"sigs.k8s.io/scheduler-plugins/pkg/networkaware/networkoverhead"
+	"sigs.k8s.io/scheduler-plugins/pkg/networkaware/topologicalsort"
```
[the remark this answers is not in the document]
jpedro1992: Sure. Thanks

cmd/scheduler/main.go
```diff
@@ -52,6 +54,8 @@ func main() {
 		// app.WithPlugin(crossnodepreemption.Name, crossnodepreemption.New),
 		app.WithPlugin(podstate.Name, podstate.New),
 		app.WithPlugin(qos.Name, qos.New),
+		app.WithPlugin(networkoverhead.Name, networkoverhead.New),
+		app.WithPlugin(topologicalsort.Name, topologicalsort.New),
```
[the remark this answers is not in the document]
jpedro1992: The NetworkOverhead plugin can be enabled without the TopologicalSort plugin. We created the TopologicalSort plugin as an additional plugin to sort pods based on their service topology information. Nevertheless, we recommend users to enable both plugins to achieve higher performance, typically higher throughput and lower latency.

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,143 @@
...
+// 2) Otherwise, follow the strategy of the QueueSort Plugin
+func (ts *TopologicalSort) Less(pInfo1, pInfo2 *framework.QueuedPodInfo) bool {
+	p1AppGroup := util.GetPodAppGroupLabel(pInfo1.Pod)
+	p2AppGroup := util.GetPodAppGroupLabel(pInfo2.Pod)
+
+	// If pods do not belong to an AppGroup, follow QoS Sort
+	if len(p1AppGroup) == 0 || len(p2AppGroup) == 0 {
+		s := &qos.Sort{}
+		return s.Less(pInfo1, pInfo2)
+	}
```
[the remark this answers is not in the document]
jpedro1992: Thanks for the suggestion!

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,577 @@
...
+								} else {
+									numNotOK += 1
+								}
+							}
+						}
+					}
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
jpedro1992: Updated. Thanks.

pkg/networkaware/networkoverhead/networkoverhead.go
jpedro1992: Thanks for bringing up this question. In our experiments, we saw that if we filtered out a node based on only 1 unmet dependency, we might get unscheduled pods since the problem becomes harder to solve and available nodes might not be able to accommodate all dependencies. So, we considered quantifies, and we only filter a node if the number of unmet dependencies is higher than the ones that are respected.

For 1 pod dependency, this would fail.

For 10 pod dependencies, filtering a node that respects 9 dependencies makes the scheduling problem hard to solve. That's why our design decision.

pkg/networkaware/networkoverhead/networkoverhead.go
jpedro1992: What do you mean exactly here? Essentially, if costs are manually added, we check if they are alphabetically sorted since we apply binary search to get costs from a specific topology key.

pkg/networkaware/networkoverhead/networkoverhead.go
jpedro1992: Yes. Code updated.

jpedro1992:
> Thanks @jpedro1992 !
> 
> First round of review. (TODO: score and tests)
> 
> BTW: would you mind crafting an integration test?

Thanks for the review @Huang-Wei!

I made most modifications and I left a few comments. 

BTW: I created two integration tests: [topologicalSort](https://github.com/jpedro1992/scheduler-plugins/blob/development/test/integration/topologicalsort_test.go) and [networkOverhead](https://github.com/jpedro1992/scheduler-plugins/blob/development/test/integration/networkoverhead_test.go) but the creation of both CRDs is failing. Any step I might have missed in my implementation? Thanks in advance.

apis/config/scheme/scheme_test.go
```diff
@@ -18,6 +18,7 @@ package scheme
 
 import (
 	"bytes"
+	"sigs.k8s.io/scheduler-plugins/apis/config/v1beta3"
```
Huang-Wei, line 21: nit: move it below with other scheduler-plugins imports.

cmd/scheduler/main.go
```diff
@@ -18,12 +18,14 @@ package main
 
 import (
 	"os"
+	"sigs.k8s.io/scheduler-plugins/pkg/networkaware/topologicalsort"
```
Huang-Wei, line 21: nit: move it below with other scheduler-plugins imports.

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,136 @@
...
+		handle:     handle,
+		agLister:   agLister,
+		namespaces: args.Namespaces,
+	}
+	return pl, nil
+}
+
+// Less is the function used by the activeQ heap algorithm to sort pods.
+// 1) Sort Pods based on their AppGroup and corresponding service topology graph.
+// 2) Otherwise, follow the strategy of the QueueSort Plugin
```
Huang-Wei, line 84:
```suggestion
// 2) Otherwise, follow the strategy of the in-tree QueueSort Plugin
```

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,136 @@
...
+func (ts *TopologicalSort) Less(pInfo1, pInfo2 *framework.QueuedPodInfo) bool {
+	p1AppGroup := networkawareutil.GetPodAppGroupLabel(pInfo1.Pod)
+	p2AppGroup := networkawareutil.GetPodAppGroupLabel(pInfo2.Pod)
+
+	// If pods do not belong to an AppGroup, or being to different AppGroups, follow vanilla QoS Sort
+	if p1AppGroup != p2AppGroup || len(p1AppGroup) == 0 {
+		klog.V(4).Infof("Pods do not belong to the same appGroup: %v and %v, following Qos strategy...", p1AppGroup, p2AppGroup)
```
Huang-Wei, line 91: let's migrate to `InfoS`

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,136 @@
...
+		s := &qos.Sort{}
```
Huang-Wei, line 92: nope, we should turn to the upstream's prioritysort: https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/queuesort/priority_sort.go

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,136 @@
...
+		return s.Less(pInfo1, pInfo2)
+	}
+
+	klog.V(4).InfoS("Pods: %v and %v from appGroup %v", pInfo1.Pod.Name, pInfo2.Pod.Name, p1AppGroup)
```
Huang-Wei, line 96: This functin is on the hot path, and then hence this can be a frequent call, so maybe remove this log or increase the verbose level to V(5).

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,136 @@
...
+
+	// Binary search to find both order index since topology list is ordered by Workload Name
+	orderP1 := networkawareutil.FindPodOrder(appGroup.Status.TopologyOrder, labelsP1[agv1alpha.AppGroupSelectorLabel])
+	orderP2 := networkawareutil.FindPodOrder(appGroup.Status.TopologyOrder, labelsP2[agv1alpha.AppGroupSelectorLabel])
+
+	klog.V(6).InfoS("1) Pod %v order: %v", pInfo1.Pod.Name, orderP1)
+	klog.V(6).InfoS("2) Pod %v order: %v", pInfo2.Pod.Name, orderP2)
+
+	// Lower is better, thus invert result!
+	return !(orderP1 > orderP2)
```
Huang-Wei, line 117:
```suggestion
	return orderP1 <= orderP2
```

pkg/networkaware/topologicalsort/topologicalsort.go
```diff
@@ -0,0 +1,136 @@
...
+		return s.Less(pInfo1, pInfo2)
+	}
+
+	klog.V(4).InfoS("Pods: %v and %v from appGroup %v", pInfo1.Pod.Name, pInfo2.Pod.Name, p1AppGroup)
+	agName := p1AppGroup
+	appGroup, err := ts.findAppGroupTopologicalSort(agName)
+
+	if err != nil {
+		klog.ErrorS(err, "Error while returning AppGroup")
+		s := &qos.Sort{}
```
Huang-Wei, line 102: ditto. fall back to prioritysort

pkg/networkaware/util/util.go
```diff
@@ -0,0 +1,300 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"context"
+	agv1alpha1 "github.com/diktyo-io/appgroup-api/pkg/apis/appgroup/v1alpha1"
```
Huang-Wei, line 21: reorganize the imports?

pkg/networkaware/networkoverhead/networkoverhead.go
Huang-Wei: ok, how about renaming `numOK, numNotOK` to `satisfied, violated`, and adding a comment above `if satisfied > violated` to explain the intent?

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,578 @@
...
+		return framework.NewStatus(framework.Unschedulable,
+			fmt.Sprintf("Node %v does not meet several network requirements from Workload dependencies: OK: %v NotOK: %v", nodeInfo.Node().Name, numOK, numNotOK))
+	}
+	return nil
+}
+
+// Score : evaluate score for a node
+func (no *NetworkOverhead) Score(ctx context.Context, cycleState *framework.CycleState, pod *v1.Pod, nodeName string) (int64, *framework.Status) {
+
+	klog.V(4).InfoS("Calculating score for pod %q on node %q", pod.GetName(), nodeName)
```
Huang-Wei, line 213: Ditto. It's of little value to log for every Score(), maybe remove it or increase the verbose level.

pkg/networkaware/networkoverhead/networkoverhead.go
```diff
@@ -0,0 +1,578 @@
...
+		return score, framework.NewStatus(framework.Error, fmt.Sprintf("getting pods from lister: %v", err))
+	}
+
+	// Return min score if pods is empty...
+	if pods == nil {
+		return score, framework.NewStatus(framework.Success, "No pods yet allocated: minimum score")
+	}
+
+	// Pods already scheduled: Get scheduledList (Deployment name, replicaID, hostname)
+	scheduledList := networkawareutil.GetScheduledList(pods)
```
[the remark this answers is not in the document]
Huang-Wei: But I will leave it to you to do the optimization in this PR, or add a TODO comment here and finish in a follow-up.

pkg/networkaware/networkoverhead/networkoverhead.go
Huang-Wei: I mean the line `return` can be removed as this function doesn't need to return anything.

pkg/networkaware/networkoverhead/networkoverhead_test.go
```diff
@@ -0,0 +1,1546 @@
...
+				handle:      fh,
+				agLister:    agLister,
+				podLister:   podLister,
+				ntLister:    ntLister,
+				namespaces:  []string{"default"},
+				weightsName: "UserDefined",
+				ntName:      "nt-test",
+			}
+
+			// Without sleep sometimes the pods are not created in time
```
Huang-Wei, line 335: To avoid flakiness, we usually add a wait.Poll() call to wait for the pods to be available. This is more optimal than time.Sleep().

pkg/networkaware/networkoverhead/networkoverhead_test.go
```diff
@@ -0,0 +1,1546 @@
...
+				ntLister:    ntLister,
+				namespaces:  []string{"default"},
+				weightsName: "UserDefined",
+				ntName:      "nt-test",
+			}
+
+			state := framework.NewCycleState()
+
+			// Without sleep sometimes the pods are not created in time
+			time.Sleep(1 * time.Second)
```
Huang-Wei, line 753: ditto. we should avoid using time.Sleep() for unexpected flakiness.

pkg/networkaware/networkoverhead/networkoverhead_test.go
```diff
@@ -0,0 +1,1546 @@
...
+			for i := 0; i < b.N; i++ {
+				gotList := make(framework.NodeScoreList, len(nodes))
+				scoreNode := func(i int) {
+					n := nodes[i]
+					score, _ := pl.Score(ctx, state, tt.pod, n.Name)
+					gotList[i] = framework.NodeScore{Name: n.Name, Score: score}
+				}
+				Until(ctx, len(nodes), scoreNode)
+
+				// b.Logf("Score List (Before normalization): %v", gotList)
```
Huang-Wei, line 766: remove it.

pkg/networkaware/networkoverhead/networkoverhead_test.go
```diff
@@ -0,0 +1,1546 @@
...
+				status := pl.NormalizeScore(ctx, state, tt.pod, gotList)
+				assert.True(b, status.IsSuccess())
+
+				// b.Logf("Score List (After normalization): %v", gotList)
```
Huang-Wei, line 770: remove it.

pkg/networkaware/util/util.go
```diff
@@ -0,0 +1,300 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
Huang-Wei, line 2: 2022

test/integration/utils.go
```diff
@@ -19,9 +19,14 @@ package integration
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
```
Huang-Wei, line 22: reorg the imports

test/integration/topologicalsort_test.go
```diff
@@ -0,0 +1,199 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Huang-Wei, line 2: 2022

test/integration/networkoverhead_test.go
```diff
@@ -0,0 +1,271 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Huang-Wei, line 2: 2022

Huang-Wei (commented):
Regarding the integration test failure, I guess you forgot to include the CRDs for AppGroup? It's expected to be located in `manifests/crds`. And this folder's CRDs will be loaded in the init step of an integration test's setup (via `envtest`).

pkg/networkaware/networkoverhead/networkoverhead_test.go
jpedro1992: What do you think of the following change? Is there a cleaner way? 

```go
// Wait for the pods to be scheduled.
if err := wait.Poll(1*time.Second, 20*time.Second, func() (bool, error) {
return true, nil
}); err != nil {
b.Errorf("pods not scheduled yet: %v ", err)
}
```

jpedro1992:
Hi @Huang-Wei,

Thank you for your feedback! Your suggestion to move some of the logic to Pre-Filter makes perfect sense. I will check this first and see if I can incorporate it already in this PR, otherwise, I will add a TODO as suggested. 

Thank you once again for the revision!

Huang-Wei:
> --- FAIL: TestNetworkOverheadPlugin (3.02s)
> networkoverhead_test.go:75: Timed out waiting for CRDs to be ready: timed out waiting for the condition

It seems the integration test doesn't install the CRD properly.

jpedro1992:
> > --- FAIL: TestNetworkOverheadPlugin (3.02s)
> > networkoverhead_test.go:75: Timed out waiting for CRDs to be ready: timed out waiting for the condition
> 
> It seems the integration test doesn't install the CRD properly.

Yes, the integration tests fail at the CRD installation. I have made the following steps: 

1) I added both AppGroup CRD and NetworkTopology CRD in the [verify-crdgen.sh](https://github.com/kubernetes-sigs/scheduler-plugins/pull/432/files#diff-4e3ddc507e7011467ddc0a1168a0eb8355821fe18f934b3a9042fb6ceef37cac) 

2) I added both AppGrup and NetwporkTopology folders in [manifests](https://github.com/kubernetes-sigs/scheduler-plugins/pull/432/files#diff-57d3ffa0dc93dcc558b764f59446a9a6f1787bd395ac7bcd845271d4e897d384). 

Am I missing something? 

If I install both crd.yaml, they both work in a K8s cluster: 

crd_check

jpedro1992:
Hi @Huang-Wei, 

Regarding the PreFilter state suggestion, I made a few adjustments to the NetworkOverhead plugin, and essentially now the longest extension point is the PreFilter State since the filtering and scoring are just retrieving the results needed from the PreFilterState. 

I am sharing benchmark results below based on unit tests.

goos: windows
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i7-12700H
Test: 1 AppGroup with 10 pods already scheduled, execution time to deploy an extra pod:

| PreFilter | Description | Number of Instructions | Execution Time (in ms/op) |
| ----------- | ----------- | ----------- | ----------- |
|  | 10 nodes     | 40490 | 0.031 |
|  | 100 nodes   | 2209   | 0.84   |
|  | 500 nodes   | 285     |  4.35  |
|  | 3000 nodes   | 160   | 7.5     |
|  | 5000 nodes   | 75     | 21.33 |
|  | 10000 nodes   | 16   | 64.6   |

| Filter | Description | Number of Instructions | Execution Time (in ms/op) |
| ----------- | ----------- | ----------- | ----------- |
|  | 10 nodes   | 84877 | 0.012 |
|  | 100 nodes  | 23818 | 0.062 |
|  | 500 nodes   | 3552 | 0.031 |
|  | 3000 nodes   | 872 | 1.35  |
|  | 5000 nodes   | 544 | 2.21  |
|  | 10000 nodes   |255 | 4.15  |


| Score | Description | Number of Instructions | Execution Time (in ms/op) |
| ----------- | ----------- | ----------- | ----------- |
|  | 10 nodes    | 125080 | 0.0087 |
|  | 100 nodes   | 34197  | 0.032  |
|  | 500 nodes   | 15489  | 0.071  |
|  | 3000 nodes  | 4443   | 0.27   |
|  | 5000 nodes  | 2852   | 0.40   |
|  | 10000 nodes | 1714   | 0.63   | 

I can include this version already in this PR, otherwise, I can leave it for a new PR since a few modifications are made to include the PreFilter extention point. 

btw, for the network overhead plugin, we do not see an use for the add/removePod functions for now, so we just return here.  

Any feedback on my previous comment on the failing CRD installation? Thanks in advance!

Huang-Wei:
@jpedro1992 let's put the PreFilter optimization in a follow-up PR - we have a jumbo PR already :)

> btw, for the network overhead plugin, we do not see an use for the add/removePod functions for now, so we just return here.

It's ok. Those methods are for preemption purpose.

> Any feedback on my previous comment on the failing CRD installation?

Regarding the CRD installation issue, for now you seem to duplicate the CRD yamls:

```
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins|pr/432 ⇒  ll manifests/crds
total 16K
-rw-r--r-- 1 weih 8.0K Dec  5 22:45 appgroup.diktyo.k8s.io_appgroups.yaml
-rw-r--r-- 1 weih 6.4K Dec  5 22:45 networktopology.diktyo.k8s.io_networktopologies.yaml
lrwxr-xr-x 1 weih   30 May 25  2022 scheduling.sigs.k8s.io_elasticquotas.yaml -> ../capacityscheduling/crd.yaml
lrwxr-xr-x 1 weih   24 May 25  2022 scheduling.sigs.k8s.io_podgroups.yaml -> ../coscheduling/crd.yaml
lrwxr-xr-x 1 weih   32 May 25  2022 topology.node.k8s.io_noderesourcetopologies.yaml -> ../noderesourcetopology/crd.yaml
```

We should soft link them under the folder `manifests/crds`:

```
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins/manifests/crds|pr/432 ⇒  rm appgroup.diktyo.k8s.io_appgroups.yaml networktopology.diktyo.k8s.io_networktopologies.yaml
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins/manifests/crds|pr/432⚡ ⇒  ln -s ../appgroup/crd.yaml appgroup.diktyo.k8s.io_appgroups.yaml
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins/manifests/crds|pr/432⚡ ⇒  ln -s ../networktopology/crd.yaml networktopology.diktyo.k8s.io_networktopologies.yaml
```

Could you make the ^^ change and let's see if CI can pass?

jpedro1992:
Hi @Huang-Wei, 

> We should soft link them under the folder `manifests/crds`:
> 
> ```
> weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins/manifests/crds|pr/432 ⇒  rm appgroup.diktyo.k8s.io_appgroups.yaml networktopology.diktyo.k8s.io_networktopologies.yaml
> weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins/manifests/crds|pr/432⚡ ⇒  ln -s ../appgroup/crd.yaml appgroup.diktyo.k8s.io_appgroups.yaml
> weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins/manifests/crds|pr/432⚡ ⇒  ln -s ../networktopology/crd.yaml networktopology.diktyo.k8s.io_networktopologies.yaml
> ```
> 

Thanks! I knew I was missing something! Also, I was trying to create these soft links in Windows, which was definitively not a good idea... I solved the issue with the integration tests. However, I still have issues with the CRDs verification because of the _-required, minimum and maximum_ tags. Should I remove these tags, or is there any way of adding these to my API definition so that the test assumes them when creating the CRDs?  

> @jpedro1992 let's put the PreFilter optimization in a follow-up PR - we have a jumbo PR already :)
> 
> > btw, for the network overhead plugin, we do not see an use for the add/removePod functions for now, so we just return here.
> 
> It's ok. Those methods are for preemption purpose.
> 

I kept a simple PreFilter implementation that I worked during the past weeks based on the previous implementation. I leave its optimization for a further PR. This one is indeed quite large! :) 

Thanks for your effort on revising this!

Huang-Wei:
> However, I still have issues with the CRDs verification because of the -required, minimum and maximum tags

I suppose those tags are expected to be absent in the generated crd yaml? as they're absent in the original types.go. Take `numMembers` for example, it's [defined](https://github.com/diktyo-io/appgroup-api/blob/main/pkg/apis/appgroup/v1alpha1/types.go#L52) as:

```go
NumMembers int32 `json:"numMembers,omitempty" protobuf:"bytes,1,opt,name=numMembers"`
```

So I'd suggest to first run `hack/verify-crdgen.sh`, then you will see some local changes, then `git add` all changes **except** for the lines with `api-approved.kubernetes.io` - that is still required in lower version `controller-gen` (I will bump its version later). Then rerun `hack/verify-crdgen.sh`, it should give a green signal.

If you're confused with how it works, here is a [diff](https://gist.github.com/Huang-Wei/057a8927649bbc46e466005c3fe2eac0) you can apply to pass `hack/verify-crdgen.sh`.

Huang-Wei:
> Thanks, @Huang-Wei! I was able to solve all issues with the verify and integration tests! Thanks once again for your revision and comments!

Glad to hear that! Thanks for the continuous work to bring an innovative scheduler plugin!

/lgtm
/approve

PR #434, opened 2022-10-10
fix repeat field description

0xff-dev:
#### What type of PR is this?
/kind bug

fix repeat field description.

Fixes #


Huang-Wei:
/lgtm
/approve

Thanks for the catch!

PR #439, opened 2022-10-21
Add local cache for noderesourcetopology using the reserve plugin [3/3] - enhancements

fromanirh:
#### What type of PR is this?

/kind feature

Finalize the implemention of the noderesourcetopology pessimistic cache allocation, building on top of #437 .
Includes enhancmement needed to let the feature work outside the simplest case, and add introspection facilities to learn and troubleshoot the status.

Fixes N/A

design doc and release note added on #315


fromanirh:
depends on and includes #437

fromanirh:
/release-note-none
set in #315

pkg/noderesourcetopology/cache/overallocating.go
```diff
@@ -74,6 +74,9 @@ func NewOverallocating(lister listerv1alpha1.NodeResourceTopologyLister, indexer
 
 func (ov *Overallocating) GetByNode(nodeName string, pod *corev1.Pod) (*topologyv1alpha1.NodeResourceTopology, bool) {
 	logKey := keyFromPod(pod)
+
+	ov.resyncState(logKey)
```
fromanirh, line 87: the commit message has enough context, but the explanation is poor and should be rephrased.
I spent some more time thinking about how to solve this issue, and I want to give a honest try to the "use an intermediate object" approach before to resort to the solution proposed here.
I think there is room for a clean and minimally invasive fix, which I'll experiment shortly.

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -64,6 +69,14 @@ func initNodeTopologyInformer(tcfg *apiconfig.NodeResourceTopologyMatchArgs, han
 		return nil, err
 	}
 
+	// TODO: do we have a better name to introspect the profile name?
+	if schedName, ok := os.LookupEnv(schedulerNameEnvVar); ok {
```
Tal-or, line 73: This requires an update of both docs and manifests

pkg/noderesourcetopology/cache/overallocating.go
```diff
@@ -49,9 +49,10 @@ type Overallocating struct {
 	nodeTainted      counter
 	lister           listerv1alpha1.NodeResourceTopologyLister
 	indexer          NodeIndexer
+	statusDirectory  string
```
Tal-or, line 52: Pods are replaceable units, what would happen if the plugin pod gets crashed, restarted, etc?
Do we keep a copy of the file somewhere on the host?

Tal-or (commented):
Left a couple of comments.
I need to reiterate the doc design in order to comment on the resync loop issue. 
I'll try to get to it ASAP

pkg/noderesourcetopology/cache/overallocating.go
```diff
@@ -221,13 +240,43 @@ func (ov *Overallocating) FlushNodes(logKey string, nrts ...*topologyv1alpha1.No
...
+		pod, ok := obj.(*corev1.Pod)
+		if !ok {
+			klog.V(3).InfoS("nrtcache: taint: unsupported object %T", obj)
+			return
+		}
+		if pod.Spec.SchedulerName == schedulerName {
+			// nothing to do here
+			return
+		}
+		if pod.Spec.NodeName == "" {
```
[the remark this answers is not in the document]
fromanirh: when the pod was just posted, the nodeName will be empty for a little while (hopefully little) until the scheduler picks it up.

pkg/noderesourcetopology/cache/overallocating.go
```diff
@@ -221,13 +240,43 @@ func (ov *Overallocating) FlushNodes(logKey string, nrts ...*topologyv1alpha1.No
...
+			return
+		}
+
+		cc.MarkNodeTainted(pod.Spec.NodeName, pod)
+		klog.V(6).InfoS("nrtcache: tainted", "logKey", keyFromPod(pod), "node", pod.Spec.NodeName, "podUID", pod.UID)
+	}
+
+	podInformer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
+		AddFunc: taintCache,
+		UpdateFunc: func(oldObj, newObj interface{}) {
```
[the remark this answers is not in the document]
fromanirh: yeah, this is still out of paranoia mostly. I can remove it to be consistent with the other code.

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: I think it's better to just postpone this change. I'll remove the related commit from the next upload

pkg/noderesourcetopology/cache/overallocating.go
fromanirh: we lose this data, and it's fine. It's ephemeral reporting data anyway. But it's also half-baked borderline hack, so let's postpone this.

fromanirh:
uhm, known broken. I'll address the comments on https://github.com/kubernetes-sigs/scheduler-plugins/pull/437 and reshape it a bit before resubmit.

pkg/noderesourcetopology/cache/overallocating.go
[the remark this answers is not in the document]
Tal-or: Keep it for now I guess, better have some redundant than nasty bug

fromanirh:
pending:
- add tests to exercise Resync (https://github.com/kubernetes-sigs/scheduler-plugins/pull/437#pullrequestreview-1180659205)

fromanirh:
/hold
need to add the resync integration tests as discussed during the review of #437

fromanirh:
/hold cancel
retargeted this PR to contains fixes only to avoid to become too large, like we already observed in #315 and #437 (especially the latter). Tests proper spawned off in #455

fromanirh:
this PR is now downsized and ready for review

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: this still needs to be addressed

fromanirh: in general, need to add a commit to document the new option into the README. Will fix in the next upload.

Huang-Wei: hrm, one scheduler binary can contain multiple profiles that the NRT plugin is enabled in. Let me think about it.

pkg/noderesourcetopology/cache/cache.go
```diff
@@ -28,7 +28,9 @@ type Interface interface {
 	// Over-reserved resources are the resources consumed by pods scheduled to that node after the last update
 	// of NRT pertaining to the same node, pessimistically overallocated on ALL the NUMA zones of the node.
 	// The pod argument is used only for logging purposes.
-	GetCachedNRTCopy(nodeName string, pod *corev1.Pod) *topologyv1alpha1.NodeResourceTopology
+	// Returns a boolean to signal the caller if the NRT data is clean. If false, then the node is possibly
+	// tainted - so it should be ignored or handled differently by the caller.
```
Huang-Wei, line 32: After reading through the logic, I realized `tainted` here doesn't mean the k8s `Taint`... so maybe reword it a bit to avoid confusion, or add a comment saying it's not the k8s Taint.

pkg/noderesourcetopology/pluginhelpers.go
Huang-Wei: Not sure if it works: enforce cast `handle` to `framework.Framework` in plugin's initialization, and call its `ProfileName()` to get the profile name and store it in the plugin. (but even if it works, it doesn't quite fit the case that NRT is enabled in multiple profiles unless we put a shared variable to store the profile names)

pkg/noderesourcetopology/cache/cache.go
fromanirh: good point, will rename. Also good point about multiple profiles: in my testing (and in the usecases we are covering) we will have a secondary scheduler with a single profile, this is why I missed this before.
A way to fix is to turn the environment variable value into a comma separate list of profile names. This is pretty ugly and duplicates information, but AFAICT should work.

fromanirh: Addressed the multiple profile comments. I'm not too happy about the UX but this is the best I can think of atm. I'd like to keep iterating here in the next revision cycles.

pkg/noderesourcetopology/pluginhelpers.go
fromanirh: addressed the comment docs and manifests. Thanks for pointing this out @Tal-or !

Huang-Wei: > We can pivot to a (lock-protected) module-level sets.String if you prefer.

It can be lock-free as each plugin's New() is called sequentially in the scheduler framework core.

fromanirh: very cool, thanks. Moving to this approach. UX is much more streamlined.

Huang-Wei:
/lgtm
/approve

Thanks @fromanirh !

fromanirh:
thank you @Huang-Wei for the quick review! I'll continue over the next days (/weeks considering the holidays) to implement the integration tests.

Huang-Wei:
> thank you @Huang-Wei for the quick review! I'll continue over the next days (/weeks considering the holidays) to implement the integration tests.

SG. I will soon (before Christmas) cut a v0.24 release and bump main to use k8s v1.25.

fromanirh:
> > thank you @Huang-Wei for the quick review! I'll continue over the next days (/weeks considering the holidays) to implement the integration tests.
> 
> SG. I will soon (before Christmas) cut a v0.24 release and bump main to use k8s v1.25.

thanks for the heads up! At cost of being redundant, I expect both of these activities to have no impact on integration tests addition (#455).

PR #454, opened 2022-12-08
Add LeastNUMANodes ScoringStrategy for NodeResourceTopologyMatch plugin

PiotrProkop:
#### What type of PR is this?

/kind feature

This is a KEP for introducing new ScoringStrategy `LeastNUMANodes` for `NodeResourceTopologyMatch` plugin that will score nodes based on how many NUMA nodes are required to satisfy resource requirements of given pod.

Fixes #

```release-note
NONE
```

### DIFF
--- a/kep/454-numa-nodes-scoring/README.md
+++ b/kep/454-numa-nodes-scoring/README.md
@@ -1,0 +1,235 @@
+# LeastNUMANodes ScoringStrategy for NodeResourceTopologyMatch plugin
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [ScoringStrategy implementation details](#scoringstrategy-implementation-details)
+    - [Description of calculating required NUMA nodes](#description-of-calculating-required-numa-nodes)
+    - [Description of normalizing score](#description-of-normalizing-score)
+    - [Example](#example)
+- [Use cases](#use-cases)
+- [Known limitations](#known-limitations)
+- [Test plans](#test-plans)
+- [Graduation criteria](#graduation-criteria)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature enablement and rollback](#feature-enablement-and-rollback)
+- [Implementation history](#implementation-history)
+<!-- /toc -->
+
+# Summary
+
+This document describes behaviour of a new ScoringStrategy `LeastNUMANodes` for NodeResourceTopologyMatch plugin that scores nodes based on how many NUMA nodes are required to run a given pod.
+
+# Motivation
+
+Consuming resources from multiple NUMA nodes can cause significant performance degradation
+in latency-critical execution and high-throughput applications.
+Topology Manager assigns resources from least amount of NUMA nodes but the scheduler is unaware
+of different NUMA topologies. The best case scenario would be to schedule pod on the node that
+can satisfy resource requirements from least amount of NUMA nodes to minimize latency.
+
+## Goals
+
+- Make better scheduling decisions that take NUMA topology into consideration.
+
+## Non-Goals
+
+- Change the PodSpec to allow requesting a maximum amount of NUMA nodes required to satisfy resource requirements.
+
+# Proposal
+
+A new ScoringStrategy `LeastNUMANodes` would check how many NUMA nodes are required to run a given pod. It will use [CRD][1]
+to get available resources on the worker nodes and identify which topology policy is enabled, the same as other ScoringStrategies.
+
+For now available ScoringStrategies are running only for `single-numa-node` TopologyManager policy.
+`LeastNUMANodes` strategy can score nodes running all TopologyManager policies.
+
+## ScoringStrategy implementation details
+
+Since topology of the node is stored in the CR, kube-scheduler should be subscribed for updates of appropriate CRD type.
+Same mechanism is used by other ScoringStrategies for `NodeResourceTopologyMatch` plugin.
+
+### Description of calculating required NUMA nodes
+
+It's important to notice the we need to mimic how Topology Manger chooses which bitmask is the most suitable if there are multiple combinations of NUMA nodes that can satisfy the request.
+The narrower bitmask is preffered or if the same number of bits is set then the mask with more lower-numbered bits set wins out([link](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/bitmask/bitmask.go#L143)).
+
+To calculate how many NUMA nodes are required to run given container/pod the plugin will:
+* generate possible combinations starting from narrowest possible bitmask to widest
+* the [Combination][2] function can used to generate combinations since it generate combinations from the lowest values
+* combine resources available in combination by iterating over every Resources present in every NUMA node and adding them together
+* each NUMA nodes combination will be evaluated to see if there are enough resources to satisfy POD requirements
+* it's valid to request only non NUMA resources, so if pod asks only for non NUMA resources, it requires 0 NUMA nodes
+
+
+The Topology Manager can deal with the alignment of resources in a couple of distinct scopes:
+* container
+* pod
+
+Calculating required NUMA nodes for a `pod` scope is straightforward, we just need to calculate it for pod's effective resources.
+
+When it comes to `container` scope we need to calculate `required` NUMA nodes for every container in proper order([same as TopologyManager asks for hints](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/scope_container.go#L52))
+and temporarly substract those resources from available resources in given combination and store only the maximum for given pod:
+```go
+    res := getRequiredNUMANodes(container.Resources.Requests)
+    // take max NUMA we want to score nodes based on the worst scenario
+	if res > maxNuma {
+		maxNuma = res
+	}
+```
+
+### Description of normalizing score
+
+Score will be calculated as follows:
+
+```go
+func normalizeScore(numaNodes int) int64 {
+	numaScore := framework.MaxNodeScore / highestNUMAID
+	return framework.MaxNodeScore - int64(numaNodes)*numaScore
+}
+```
+
+So pod which only uses non NUMA resources will receive `MaxNodeScore` and pod that requires 2 NUMA nodes will receive:
+
+```
+100 - (2*(100/8)) = 100 - (2*(12)) = 100 - 24 = 76
+```
+
+### Example
+
+Suppose we have 2 candidate nodes after going through Filter phase. Each nodes has 2 NUMA nodes but available resources looks as follow:
+
+**Node 1**
+
+| Resource name | Available resource in NUMA node 0 | Available resource in node NUMA node 1 |
+|---------------|-----------------------------------|----------------------------------------|
+| cpu           | 2                                 | 4                                      |
+
+**Node 2**
+
+| Resource name | Available resource in NUMA node 0 | Available resource in NUMA node 1 |
+|---------------|-----------------------------------|-----------------------------------|
+| cpu           | 8                                 | 8                                 |
+
+The pod consists of 2 containers each asking for 3 cpus.
+
+Score calculation for `Node 1` will look as follows:
+- First container:
+    - Generate combinations for 2 NUMA nodes starting from narrowest and lowest to widest and greatest
+    - Generate combinations for bitmask of length 1 `[(10) (01)]`
+    - Check if container can fit first combination (10) -> it can't
+    - Check if container can fit second combination (01) -> it can, return required NUMA nodes == 1 and bitmask (01)
+    - Substract container resources from combined resources:
+
+    **Node 1**
+
+    | Resource name | Available resource in NUMA node 0 | Available resource in NUMA node 1 |
+    |---------------|-----------------------------------|-----------------------------------|
+    | cpu           | 2                                 | 1                               |
+
+- Second container:
+    - Generate combinations for 2 NUMA nodes starting from narrowest and lowest to widest and greatest
+    - Generate combinations for bitmask of length 1 `[(10) (01)]`
+    - Check if container can fit first combination (10) -> it can't
+    - Check if container can fit second combination (01) -> it can't
+    - No bitmask with width of 1 can satisfy resource requirements of container -> increase bitmask width
+    - Generate combinations for bitmask of length 2 `[(11)]`
+    - Combine resources from `node 0` and `node 1` -> available cpus 3
+    - Check if container can fit third combination (11) -> it can, return required NUMA nodes == 2 and bitmask (11)
+    - Substract container resources from combined resources:
+
+    **Node 1**
+
+    | Resource name | Available resource in NUMA node 0 | Available resource in NUMA node 1 |
+    |---------------|-----------------------------------|-----------------------------------|
+    | cpu           | 0                                 | 0                                 |
+
+- Score calculations:
+  - Take max of required NUMA nodes `max([1, 2]) = 2`
+  - Calculate normalize score = 76
+
+Score calculation for `Node 2` will look as follows:
+- First container:
+    - Generate combinations for 2 NUMA nodes starting from narrowest and lowest to widest and greatest
+    - Generate combinations for bitmask of length 1 `[(10) (01)]`
+    - Check if container can fit first combination (10) -> it can
+    - Substract container resources from combined resources:
+
+    **Node 2**
+
+    | Resource name | Available resource in NUMA node 0 | Available resource in NUMA node 1 |
+    |---------------|-----------------------------------|-----------------------------------|
+    | cpu           | 5                                 | 8                                 |
+
+- Second container:
+    - Generate combinations for 2 NUMA nodes starting from narrowest and lowest to widest and greatest
+    - Generate combinations for bitmask of length 1 `[(10) (01)]`
+    - Check if container can fit first combination (10) -> it can
+    - Substract container resources from combined resources:
+
+    **Node 2**
+
+    | Resource name | Available resource in NUMA node 0 | Available resource in NUMA node 1 |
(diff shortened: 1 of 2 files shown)


PiotrProkop:
fyi @fromanirh @swatisehgal

fromanirh:
will be reviewing this before December 16 (worst case)

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Non-Goals
+
+- Change the PodSpec to allow requesting a maximum amount of NUMA nodes required to satisfy resource requirements.
+- This Proposal requires exposing NUMA topology information. `NodeResourceTopology` API will be used to retrieve it.
+  Similiar to `NodeResourceTopology` plugins.
+
+# Proposal
+
+A Score plugin would check how many NUMA nodes are required to run given pod. Proposed Score plugin will use [CRD][1]
```
fromanirh, line 47: I'm not pushing for this (I'm fine both ways) but the first obvious question is if this new logic can be added to the existing  NodeResourceTopologyMatch score plugin, perhaps using an option

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,286 @@
...
+to get available resources on the worker nodes and identify which topology policy is enabled, similliar to `NodeResourceTopology` plugins.
+It will run for all Topology Manager.
```
fromanirh, line 49: it will be very nice if you can highlight the differences with NodeResourceTopology score. This can serve an explanation why a separate plugin is the best approach, or at least we can have a discussion around this area

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Score plugin implementation details
+
+Since topology of the node is stored in the CR, kube-scheduler should be subscribed for updates of appropriate CRD type.
+Same mechanism will be used as in `NodeResourceTopologyMatch` plugin.
+
+### Description of calculating required NUMA nodes
+
+To calculate how many NUMA nodes are required to run given container/pod the plugin will generate combinations of all NUMA nodes for given bitmask and see if there are enough resources in specific combination.
+It's important to notice the we need to mimic how Topology Manger chooses which bitmask is the most suitable if there are multiple combinations of NUMA nodes that can satisfy the request.
```
fromanirh, line 63: so this is another plugin which -out of necessity- has to mimic Topology Manager behavior. We should probably discuss if and how makes sense to have these plugins clearly marked or perhaps merged together if possible/makes sense (see previous comments)

fromanirh (commented):
I like the proposal in general and I don't see any issue with that. I have some questions and possible concerns about the implementation proper if we want/need/should reuse code or infra added for NodeResourceTopologyMatch. Explained as inline questions.
Overall this looks already pretty good to me.

kep/numa-nodes-scoring/README.md
PiotrProkop: I am fine with that, I have written small POC and I have reused many functions from `noderesourcetopology` package. We should just have proper documentation to outline that using this score plugin with `NodeResourceTopologyMatch` filter plugin is kinda pointless 😄

kep/numa-nodes-scoring/README.md
PiotrProkop: 👍 good point

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,286 @@
...
+It will run for all Topology Manager.
+
+## Topology format
+
+The `NodeResourceTopology` CRD API will be used to collect all necessary information regarding worker node topology and available resources.
+
+## Score plugin implementation details
+
+Since topology of the node is stored in the CR, kube-scheduler should be subscribed for updates of appropriate CRD type.
+Same mechanism will be used as in `NodeResourceTopologyMatch` plugin.
```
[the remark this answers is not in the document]
PiotrProkop: For now I think reading objects without using cacheis enough and if we agree that this plugin adds value we can create issue to adjust `nrt cache` for this plugin.

kep/numa-nodes-scoring/README.md
PiotrProkop: +1 for that

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+# Use cases
+
+Numbers of kubernetes worker nodes with different hardware configuration on bare metal with NUMA topology. TopologyManager feature gate enabled on the nodes. In this configuration, the user would like to deploy application and make sure it will land on a node where its resource requirements can be satisfied from the least amount of NUMA nodes to
+minimize latency.
+
+# Known limitations
+
+Kube-scheduler makes an assumption about current resource usage on the worker node, since kube-scheduler knows which pod assigned to node. This assumption makes right after kube-scheduler choose a node. But in case of scheduling with NUMA topology only TopologyManager on the worker node knows exact NUMA node used by pod, this information about NUMA node delivers to kube-scheduler with latency. In this case kube-scheduler will not know actual NUMA topology until topology exporter will send it back. It could be mitigated if kube-scheduler in proposed plugin will add a hint on which NUMA id pod could be assigned, further Topology Manager on the worker node may take it into account.
```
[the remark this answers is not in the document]
PiotrProkop: Rather short term, I know that pushing such non trivial feature to Kubernetes can take some time.

PiotrProkop:
Thanks for review @fromanirh , will adjust the proposal today.
I have also came up with an idea that with upcoming `TopologyManagerPolicyOptions`. 
We can adjust the score further by checking if NUMA nodes allocation is `optimal` in the context of `prefer-closest-allocation` `policy-option`. But it can also be introduced later with seperate `KEP`.

kep/numa-nodes-scoring/README.md
PiotrProkop: After some time I have realized that this new scoring `plugin` or `scoringStrategy` can safely coexist with filter plugin since it only runs for `single-numa-node` policy

fromanirh: I for myself I'm fine with both a separate plugin or exesting the NodeResourceTopologyMatch score. I'll be happy to review in both cases.

PiotrProkop: I am changing the proposal to add new `ScoringStrategy` called `LeastNUMANodes`, I think it would minimize code duplication and allow to reuse some patterns.

fromanirh: nice!

PiotrProkop: Proposal is changed!

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,280 @@
...
+- Make better scheduling decisions that take NUMA topology into consideration.
+
+## Non-Goals
+
+- Change the PodSpec to allow requesting a maximum amount of NUMA nodes required to satisfy resource requirements.
+
+# Proposal
+
+A new ScoringStrategy `LeastNUMANodes` would check how many NUMA nodes are required to run given pod. It will use [CRD][1]
+to get available resources on the worker nodes and identify which topology policy is enabled, the same as other ScoringStrategies.
```
fromanirh, line 45: I think it makes sense to add this scoring strategy. CC @Tal-or @swatisehgal

kep/numa-nodes-scoring/README.md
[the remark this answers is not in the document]
fromanirh: More than that though. If the hypotethical scheduler-sent topology manager hints are prescriptive (TM is obliged to adhere to them or fail) then we are effectively moving the allocation logic in the scheduler. OTOH if the scheduler-sent topology manager hints are hints the TM is allowed to ignore, we will still need to deal with uncertainity at the scheduler side.
So good we mention here this direction, and I really think of (and agree with) this is very long term.

fromanirh (commented):
LGTM. Tagging other reviewers.

Tal-or:
@PiotrProkop Looking like a good idea to me as well.
Since the main idea of this plugin is to work with non `single-numa-node` policies as well, how is it going to cope with the `Filter` plugin? 
Because as of today, the Filter plugin supports `single-numa-node` policies only 
https://github.com/kubernetes-sigs/scheduler-plugins/blob/a64627b6d909b80afdc3a51ec6c0f57d94990293/pkg/noderesourcetopology/pluginhelpers.go#L200-L201

PiotrProkop:
@Tal-or the `Filter` plugin won't be changed. If I understand current codepath correctly if the node is configured with a policy other than `single-numa-node`, the `Filter` plugin will noop without discarding the node so this node will be present in node pool that will be scored by the new `ScoringStrategy`.

fromanirh:
> @Tal-or the `Filter` plugin won't be changed. If I understand current codepath correctly if the node is configured with a policy other than `single-numa-node`, the `Filter` plugin will noop without discarding the node so this node will be present in node pool that will be scored by the new `ScoringStrategy`. If the cluster nodes are configured with different `TopologyManager` policies, the `LeastNUMANodes` strategy will still favor nodes that can offer resources from single `NUMA` node.

we can test this! On top of that, it's totally possible to enable NodeResourceTopologyMatch `Score` only, without `Filter`. Or, well, this should be possible, and if not it's likely a bug.

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,286 @@
...
+* Integration Tests
+    *  Default configuration (this plugin is disabled)
+        * no side effect on basic scheduling flow (and performance)
+        * no side effect no matter the CRD is installed or not
+    *  Enable this plugin
+        * basic workflow of this feature works (the pod lands on the node that requires the least amount of NUMA nodes to run it)
+        * verify the behavior when the CRD is and isn't installed
+* End-to-end tests
+
+Integration and End-to-end would Implementation of it does not constitute a difficulty, but requires appropriate multi-numa hardware for comprehensive testing of this feature. Comprehensive E2E testing of this would be done in order to graduate this feature from Alpha to Beta.
```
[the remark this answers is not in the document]
swatisehgal: There is ongoing discussion in the community in this area: https://github.com/kubernetes/test-infra/issues/28211. The `NodeResourceTopologyMatch` plugin could potentially benefit from availability of multi-NUMA systems in the test infrastructure.

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	// we will generate combination of numa nodes from len = 1 to the number of numa nodes present on the machine
+	for i := 1; i <= len(numaNodes); i++ {
+		// generate combinations of len i
+		numaNodesCombination := combin.Combinations(len(numaNodes), i)
+		// iterate over combinations for given i
+		for _, combination := range numaNodesCombination {
+			// accumulate resources for given combination
+			combinationResources := combineResources(numaNodes, combination)
```
swatisehgal, line 78: Can you please clarify what `combineResources` function does here? Does it return the total amount of resources  present on the evaluated NUMA Node combination?

swatisehgal (commented):
Thanks for your work on this!

I totally see the benefit of LeastNUMANodes scoring strategy in the `NodeResourceTopologyMatch` scheduler plugin. The proposal looks good overall. Left a few minor comments/questions.

PiotrProkop:
@swatisehgal thanks for the review! especially for typos (English is not my first language :p ). Updating the proposal now.

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,280 @@
...
+- [Test plans](#test-plans)
+- [Graduation criteria](#graduation-criteria)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature enablement and rollback](#feature-enablement-and-rollback)
+- [Implementation history](#implementation-history)
+<!-- /toc -->
+
+# Summary
+
+This document describes behaviour of the new ScoringStrategy `LeastNUMANodes` for NodeResourceTopologyMatch plugin that scores nodes based on how many NUMA nodes are required to run given pod.
```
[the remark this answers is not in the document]
PiotrProkop: 👍

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,280 @@
...
+
+- Make better scheduling decisions that take NUMA topology into consideration.
+
+## Non-Goals
+
+- Change the PodSpec to allow requesting a maximum amount of NUMA nodes required to satisfy resource requirements.
+
+# Proposal
+
+A new ScoringStrategy `LeastNUMANodes` would check how many NUMA nodes are required to run given pod. It will use [CRD][1]
```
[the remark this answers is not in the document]
PiotrProkop: 👍

kep/numa-nodes-scoring/README.md
PiotrProkop: Yes, it returns the sum of all Resources in given NUMA Node combination. I'll add it to proposal.

PiotrProkop: Added.

fromanirh:
/lgtm
I'd be happy to have this scoring strategy added to NodeResourceTopologyMatch
I suggest to defer further discussions about topology manager hints to the future. Is a topic which gathers a lot of interest, but it's also outside the scope of this change.

fromanirh:
/assign @Huang-Wei 
for approval review

kep/numa-nodes-scoring/kep.yaml
```diff
@@ -0,0 +1,12 @@
+title: Scoring nodes based on number of NUMA nodes required to run the pod.
+kep-number: placeholder
```
Huang-Wei, line 2:
```suggestion
kep-number: 454
```

kep/numa-nodes-scoring/kep.yaml
```diff
@@ -0,0 +1,12 @@
+title: Scoring nodes based on number of NUMA nodes required to run the pod.
```
Huang-Wei, line 1: Can you also rename the folder to `454-numa-nodes-scoring`?

Huang-Wei: Can the title be a bit concise like "NUMA aware scoring"?

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,297 @@
...
+    *  Enable this plugin
+        * basic workflow of this feature works (the pod lands on the node that requires the least amount of NUMA nodes to run it)
+        * verify the behavior when the CRD is and isn't installed
+* End-to-end tests
+
+Integration and End-to-end would Implementation of it does not constitute a difficulty, but requires appropriate multi-numa hardware for comprehensive testing of this feature. Comprehensive E2E testing of this would be done in order to graduate this feature from Alpha to Beta.
+
+# Graduation criteria
+
+* Alpha (v0.24)
```
Huang-Wei, line 276:
```suggestion
* Alpha (v0.25)
```

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,297 @@
...
+Same mechanism is used by other ScoringStrategies for `NodeResourceTopologyMatch` plugin.
+
+### Description of calculating required NUMA nodes
+
+To calculate how many NUMA nodes are required to run given container/pod the plugin will generate combinations of all NUMA nodes for given bitmask and see if there are enough resources in specific combination.
+It's important to notice the we need to mimic how Topology Manger chooses which bitmask is the most suitable if there are multiple combinations of NUMA nodes that can satisfy the request.
+The bitmask with less bits set or if the same number of bits is set then the mask with more lower-numbered bits set wins out([link](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/bitmask/bitmask.go#L143)).
+The [Combination][2] function can used for that since it generate combinations from the lowest values.
+
+```go
```
Huang-Wei, line 62: I'd avoid using source code to explain the core idea as source code would evolve quickly as time goes. Would you mind either abstracting the logic in a few bullets or embedding pseudo code only?

I think it may read better to put up an example from this section to "Description of normalizing score". For example, suppose we have 2 candidate nodes after going through Filter phase, and illustrate how the NUMA capacity in each node is like, then explain in each section how the logic is implemented.

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,297 @@
...
+* container
+* pod
+
+Calculating required NUMA nodes for a `pod` scope is straightforward, we just need to calculate it for pod's effective resources.
+
+When it comes to `container` scope we need to calculate `required` NUMA nodes for every container in proper order and
+temporarly substract those resources from available resources. Return value would be the maximum amount of NUMA nodes required to run
+container in given pod.
+
+```go
```
Huang-Wei, line 155: Ditto. I'd prefer using concise stanzas or pseudo code to explain the core logic.

kep/numa-nodes-scoring/README.md
```diff
@@ -0,0 +1,297 @@
...
+    return nil
+}
+```
+
+
+### Description of normalizing score
+
+Score will be calculated as follows:
+
+```go
```
Huang-Wei, line 232: This kind of short code snippet makes sense in a design doc.

PiotrProkop:
@Huang-Wei thanks for the review. I tried to apply all you suggestions 😄

kep/numa-nodes-scoring/README.md
PiotrProkop: I have added `Example` section.

kep/numa-nodes-scoring/README.md
PiotrProkop: 👍

kep/numa-nodes-scoring/kep.yaml
PiotrProkop: :+1:

kep/numa-nodes-scoring/kep.yaml
PiotrProkop: :+1:

Huang-Wei:
Thanks @PiotrProkop !

/lgtm
/approve

PR #455, opened 2022-12-13
integration tests for the NodeResourceTopology cache and resync loop

fromanirh:
#### What type of PR is this?

/kind cleanup

Add integration tests deferred from #437 

Fixes # N/A

Split again #439 trying to simplify the review process which proven difficult so far. I see the integration tests requiring more work, and #439 is collecting a fair amount of important changes anyway. Trying to prevent #439 to grow too much in size.

### DIFF
--- a/test/integration/noderesourcetopology_cache_test.go
+++ b/test/integration/noderesourcetopology_cache_test.go
@@ -21,12 +21,15 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
@@ -37,6 +40,7 @@ import (
 
 	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
 	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
+	"github.com/k8stopologyawareschedwg/podfingerprint"
 
 	schedconfig "sigs.k8s.io/scheduler-plugins/apis/config"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology"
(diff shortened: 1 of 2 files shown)


fromanirh:
Regardless of the bugs in the current implementation, which I'm aware of and will fix (obv.) I'm seriously considering adding a new lane just for these tests

fromanirh:
demo of the [existing e2e test run](https://github.com/openshift-kni/numaresources-operator/tree/main/test/e2e/serial) which we are looking to reimplement here as integration test, at least partially reimplement: https://gist.github.com/fromanirh/d43c06e49edcbdd73a1f774428c2c5bb

Makefile
```diff
@@ -119,6 +119,10 @@ install-envtest:
 integration-test: install-envtest
 	$(INTEGTESTENVVAR) hack/integration-test.sh
 
+.PHONY: integration-test-cache
+integration-test-cache: install-envtest
+	$(INTEGTESTENVVAR) hack/integration-test-cache.sh
```
Huang-Wei, line 124: Hrm, why make it a separate test?

hack/integration-test-cache.sh
```diff
@@ -0,0 +1,45 @@
...
+  kube::log::status "Configuring envtest"
+  source "${TEMP_DIR}/setup-envtest"
+
+  kube::log::status "Running integration test cases"
+  # TODO: make args customizable.
+  go test -timeout=40m -mod=vendor -v \
+	  test/integration/noderesourcetopology_cache_test.go \
+	  test/integration/nrtutils.go \
+	  test/integration/utils.go \
+	  test/integration/main_test.go
```
Huang-Wei, line 40: This is already covered in:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/a64627b6d909b80afdc3a51ec6c0f57d94990293/hack/integration-test.sh#L36

Huang-Wei: And `go test` runs against packages instead of files. You have to use `-run TestXYZ` to target the specific tests if they're in the same package.

fromanirh: I should probably have looked harder at go docs instead of reinventing my wheel :) I'll just drop these in my next upload.

Makefile
fromanirh: it was the quickest (admittedly, did poor research before to take this route) to focus on the new tests and cut development time. Thanks for your suggestions below!

fromanirh:
done! there's another commit I'll probably drop soon enough. The PR is still very much WIP and incomplete (need to add more tests)

fromanirh:
My original plan was to add more tests to this PR, but in hindsight it's probably better to build incrementally and have many smaller PRs. So I'm removing the WIP label.

fromanirh:
it seems we need https://github.com/kubernetes-sigs/scheduler-plugins/pull/469 before to move forward

swatisehgal (approved):
/lgtm

PR #458, opened 2022-12-14
update the k8s patch version to v1.24.9

Huang-Wei:
#### What type of PR is this?

/kind cleanup

Bump the k8s dependency: from v1.24.3 to v1.24.9. After this PR and some other cleanup PRs, I will cut an official scheduler-plugins v0.24.9 release.

Fixes #


yibozhuang (approved):
/lgtm

PR #460, opened 2022-12-15
topologysort integrationtest update

### LINKED ISSUE #457: possible flake: topologicalsort_test
It seems this integration test is flaking. Observed with different unrelated PRs, demonstrated with #456, see https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=scheduler-plugins&pr=456 for details.

Filing this issue for tracking purposes.

jpedro1992:
#### What type of PR is this?

Discuss and solve issue #457 about the flakiness of the topologysort integration test.

The integration test has been updated. It seems that when adding 3 pods to the scheduling queue, the correct order does not occur every time, thus the flakiness. When adding only 2 pods, the order seems to occur always correctly. A complex test has been added where 11 pods are added to the scheduling queue. So, the test has been extended to see if the issues are solved.   

Fixes #457

### DIFF
--- a/test/integration/topologicalsort_test.go
+++ b/test/integration/topologicalsort_test.go
@@ -96,7 +96,8 @@ func TestTopologicalSortPlugin(t *testing.T) {
 		scheduler.WithFrameworkOutOfTreeRegistry(fwkruntime.Registry{topologicalsort.Name: topologicalsort.New}),
 	)
 	syncInformerFactory(testCtx)
-	go testCtx.Scheduler.Run(testCtx.Ctx)
+	// Do not start the scheduler.
+	// go testCtx.Scheduler.Run(testCtx.Ctx)
 	t.Log("Init scheduler success")
 	defer cleanupTest(t, testCtx)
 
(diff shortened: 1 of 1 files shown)


Huang-Wei:
The root cause seems to be in:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/07d6327976a4b60662a4b5a677f15dea1f343b57/test/integration/topologicalsort_test.go#L209

`NextPod()` is a blocking operation, hence needs to be pretty cautious when using in integration test. When you call `NextPod()` manually, it means you wan to take the control of the scheduler's running - pod popping, flushing unschedulable pod etc. So you should not start scheduler.run(), which will make the test pass:

```diff
diff --git a/test/integration/topologicalsort_test.go b/test/integration/topologicalsort_test.go
index 9b3fe51..0c21872 100644
--- a/test/integration/topologicalsort_test.go
+++ b/test/integration/topologicalsort_test.go
@@ -96,7 +96,8 @@ func TestTopologicalSortPlugin(t *testing.T) {
 		scheduler.WithFrameworkOutOfTreeRegistry(fwkruntime.Registry{topologicalsort.Name: topologicalsort.New}),
 	)
 	syncInformerFactory(testCtx)
-	go testCtx.Scheduler.Run(testCtx.Ctx)
+	// Do not start the scheduler.
+	// go testCtx.Scheduler.Run(testCtx.Ctx)
 	t.Log("Init scheduler success")
 	defer cleanupTest(t, testCtx)

```

jpedro1992:
> The root cause seems to be in:
> 
> https://github.com/kubernetes-sigs/scheduler-plugins/blob/07d6327976a4b60662a4b5a677f15dea1f343b57/test/integration/topologicalsort_test.go#L209
> 
> `NextPod()` is a blocking operation, hence needs to be pretty cautious when using in integration test. When you call `NextPod()` manually, it means you wan to take the control of the scheduler's running - pod popping, flushing unschedulable pod etc. So you should not start scheduler.run(), which will make the test pass:
> 
> ```diff
> diff --git a/test/integration/topologicalsort_test.go b/test/integration/topologicalsort_test.go
> index 9b3fe51..0c21872 100644
> --- a/test/integration/topologicalsort_test.go
> +++ b/test/integration/topologicalsort_test.go
> @@ -96,7 +96,8 @@ func TestTopologicalSortPlugin(t *testing.T) {
>  		scheduler.WithFrameworkOutOfTreeRegistry(fwkruntime.Registry{topologicalsort.Name: topologicalsort.New}),
>  	)
>  	syncInformerFactory(testCtx)
> -	go testCtx.Scheduler.Run(testCtx.Ctx)
> +	// Do not start the scheduler.
> +	// go testCtx.Scheduler.Run(testCtx.Ctx)
>  	t.Log("Init scheduler success")
>  	defer cleanupTest(t, testCtx)
> ```

Thanks @Huang-Wei! I missed it... The QoS plugin indeed skips the start of the scheduler.

Huang-Wei:
/lgtm
/approve

Thanks @jpedro1992 .

PR #471, opened 2022-12-30
update rbac to allow all verbs for podgroups/status

KunWuLuan:
When I install scheduler-plugins by `helm install`, I face the following problem:
image

It seems the rabc in chart doesn't requirement the rules for podgroup/status, so I fix it.
FYI:https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources
image

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
@@ -64,7 +64,7 @@ rules:
   verbs: ["get", "list", "watch"]
 # resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
-  resources: ["podgroups", "elasticquotas"]
+  resources: ["podgroups", "elasticquotas", "podgroups/status", "elasticquotas/status"]
   verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
 ---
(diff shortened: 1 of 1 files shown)


manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -64,7 +64,7 @@ rules:
   verbs: ["get", "list", "watch"]
 # resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.sigs.k8s.io"]
-  resources: ["podgroups", "elasticquotas"]
```
Huang-Wei, line 67: Could you please also add `elasticquotas/status`?

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -96,7 +96,7 @@ rules:
...
```
Huang-Wei, line 99: same here.

Huang-Wei (commented):
Thanks for the catch!

manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
KunWuLuan: Ok, done. : )

PR #474, opened 2023-01-05
NUMA aware scoring

PiotrProkop:
#### What type of PR is this?

/kind feature
Introduce new ScoringStrategy `LeastNUMANodes` for NodeResourceTopologyMatch.
This is the implementation of [NUMA aware scoring KEP](https://github.com/kubernetes-sigs/scheduler-plugins/pull/454)

```release-note
Introduce new ScoringStrategy LeastNUMANodes for NodeResourceTopologyMatch plugin that will score nodes based on how many NUMA nodes are required to satisfy resource requirements of given pod.
```


PiotrProkop:
@Huang-Wei @fromanirh @swatisehgal please review. There are a lot of lines added due to upgrade of 2 packages(I can split this PR into 2 -> one implementation and second upgrading packages and vendor directory) for now I have just split it into separate commits.

fromanirh (commented):
/ok-to-test

go.mod
```diff
@@ -6,14 +6,14 @@ require (
...
-	github.com/google/go-cmp v0.5.5
+	github.com/google/go-cmp v0.5.7
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.0.12
 	github.com/k8stopologyawareschedwg/podfingerprint v0.1.1
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/paypal/load-watcher v0.2.2
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.0
-	gonum.org/v1/gonum v0.6.2
+	gonum.org/v1/gonum v0.12.0
```
ffromani, line 16: could you please add a couple line in the commit message to explain why we need the dep bump? (not obvious from the next commits)

go.mod
```diff
@@ -6,14 +6,14 @@ require (
 	github.com/diktyo-io/appgroup-api v0.0.9-alpha
 	github.com/diktyo-io/networktopology-api v0.0.8-alpha
 	github.com/dustin/go-humanize v1.0.0
-	github.com/google/go-cmp v0.5.5
+	github.com/google/go-cmp v0.5.7
```
ffromani, line 9: perhaps rephrase the commit message as "upgrade go-cmp and gonum packages"?

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/diktyo-io/networktopology-api v0.0.8-alpha
 	github.com/dustin/go-humanize v1.0.0
 	github.com/google/go-cmp v0.5.7
-	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.0.12
+	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.0.13
```
ffromani, line 10: please mention in the commit message that you want to consume the new policies (https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/23) )

pkg/noderesourcetopology/filter.go
```diff
@@ -208,8 +208,8 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
 
 	klog.V(5).InfoS("Found NodeResourceTopology", "nodeTopology", klog.KObj(nodeTopology))
 	for _, policyName := range nodeTopology.TopologyPolicies {
-		if handler, ok := tm.policyHandlers[topologyv1alpha1.TopologyManagerPolicy(policyName)]; ok {
-			if status := handler.filter(pod, nodeTopology.Zones, nodeInfo); status != nil {
+		if handler, ok := tm.filterHandlers[topologyv1alpha1.TopologyManagerPolicy(policyName)]; ok {
```
ffromani, line 211: could you please add few lines in the commit message to explain the rationale for this split?
this is important to document why we choose this direction

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,192 @@
...
+func leastNUMAContainerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList) (int64, *framework.Status) {
+	nodes := createNUMANodeList(zones)
+	qos := v1qos.GetPodQOS(pod)
+
+	maxNuma := 0
+	// the order how TopologyManager asks for hint is important so doing it in the same order
+	// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/scope_container.go#L52
+	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
+		identifier := fmt.Sprintf("%s/%s/%s", pod.Namespace, pod.Name, container.Name)
+		numaNodes, res := numaNodesRequired(identifier, qos, nodes, container.Resources.Requests)
```
ffromani, line 27: `res` -> `numaNodesCount` ? (or in general a bit more explicit identifier)

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,192 @@
...
+	"k8s.io/klog/v2"
+	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
+	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
+	"k8s.io/kubernetes/pkg/kubelet/cm/topologymanager/bitmask"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+	"sigs.k8s.io/scheduler-plugins/pkg/util"
+)
+
+func leastNUMAContainerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList) (int64, *framework.Status) {
+	nodes := createNUMANodeList(zones)
```
ffromani, line 19: in general (not related to this PR) we will need to ensure this function returns a fully separate object without shallow copies hidden inside. I'll file a separate PR later.

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,192 @@
...
+			}
+
+			nRes := numaNodes[node].Resources
+			if available, ok := nRes[resName]; ok {
+				switch quantity.Cmp(available) {
+				case 0: // the same
+					// basically zero container resources
+					quantity.Sub(available)
+					// zero NUMA quantity
+					nRes[resName] = resource.MustParse("0")
```
ffromani, line 84: nit: I'd move this `MustParse` up and reuse `zeroQty` as needed

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,192 @@
...
+				continue
+			}
+			resources[resource] = quantity
+		}
+	}
+
+	return resources
+}
+
+func isCombinationSuitable(qos v1.PodQOSClass, resource v1.ResourceName, quantity, numaQuantity resource.Quantity) bool {
```
ffromani, line 175: this seems a copy-paste of `isNUMANodeSuitable`. I'm not against copy-paste per se ("a little duplication is better than a little dependency") but maybe we can evaluate other options? like rename/refactor the former and reuse it here.  Perhaps rename `isNUMANodeSuitable` to `isResourceSetSuitable` can fit and be clear enough in both call sites.

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,192 @@
...
+
+	return normalizeScore(res), nil
+}
+
+func normalizeScore(numaNodes int) int64 {
+	numaScore := framework.MaxNodeScore / highestNUMAID
+	return framework.MaxNodeScore - int64(numaNodes)*numaScore
+}
+
+func subtractFromNUMAs(numaNodes NUMANodeList, nodes []int, resources v1.ResourceList) {
```
ffromani, line 69: any chance to reuse/refactor `subtractFromNUMA` to fit here?

ffromani (commented):
initial review

go.mod
PiotrProkop: Commit reworded

go.mod
PiotrProkop: :+1:

go.mod
PiotrProkop: done

pkg/noderesourcetopology/filter.go
PiotrProkop: 👍

pkg/noderesourcetopology/least_numa.go
PiotrProkop: Good catch.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: will do, I have started this implemetation as separate plugin and moved this inside `TopologyMatch` later, so there can be more places which can reuse/refeactor functions from this package.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: Good idea.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: I have remove old `subtractFromNUMA` function and make `filter` plugin to use this new `subtractFromNUMAs` function.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: Done.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: done

pkg/noderesourcetopology/least_numa.go
PiotrProkop: done

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,173 @@
+package noderesourcetopology
```
ffromani, line 1: let's add the copyright boilerplate. Feel free to copy it from any other source file, but please make sure copyright year is 2023

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,173 @@
...
+
+import (
+	"fmt"
+
+	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	"gonum.org/v1/gonum/stat/combin"
```
ffromani, line 7: `combin` should sit in its own group:
```
stdlib

3rd party packages

kube packages
```
(perhaps inverting 3rd party and kube)

pkg/noderesourcetopology/least_numa_test.go
```diff
@@ -0,0 +1,271 @@
+package noderesourcetopology
```
ffromani, line 1: same, let's add the copyright boilerplate

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,173 @@
...
+	return normalizeScore(res), nil
+}
+
+func normalizeScore(numaNodes int) int64 {
+	numaScore := framework.MaxNodeScore / highestNUMAID
+	return framework.MaxNodeScore - int64(numaNodes)*numaScore
+}
+
+func subtractFromNUMAs(numaNodes NUMANodeList, nodes []int, resources v1.ResourceList) {
+	zeroQty := resource.MustParse("0")
```
ffromani, line 69: perhaps even more radical, like we are doing in `cache/store.go`:
```
var zeroQty resource.Quantity

func init() {
	zeroQty = resource.MustParse("0")
}
```

pkg/noderesourcetopology/least_numa.go
ffromani: ok, this means I need to spend a bit more time checking the two functions, then.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: 👍

pkg/noderesourcetopology/least_numa.go
PiotrProkop: 👍

pkg/noderesourcetopology/least_numa_test.go
PiotrProkop: :+1:

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,195 @@
...
+		if numaNodesCount < 1 {
+			errMsg := fmt.Sprintf("cannot calculate NUMA score for container: %s", container.Name)
+			klog.ErrorS(fmt.Errorf(errMsg), "failed to calculate numaNodesRequired for: %s", identifier)
+			return 0, framework.NewStatus(framework.Error, errMsg)
+		}
+
+		if numaNodesCount > maxNuma {
+			maxNuma = numaNodesCount
+		}
+
```
swatisehgal, line 59: Just wondering if we need to handle the case of a very large pods that requires resources from more than 8 numa nodes assuming there are nodes with that capability in the cluster. Maybe we should mention here as being out of scope because  the maximum number of numa nodes supported by Topology manager is 8.

pkg/noderesourcetopology/least_numa_test.go
```diff
@@ -0,0 +1,287 @@
...
+			expected: NUMANodeList{
+				{
+					NUMAID: 0,
+					Resources: v1.ResourceList{
+						v1.ResourceCPU:    *resource.NewQuantity(0, resource.DecimalSI),
+						v1.ResourceMemory: resource.MustParse("8Gi"),
+					},
+				},
+				{
+					NUMAID: 0,
```
swatisehgal, line 94: This should be NUMAID: 1 right?

swatisehgal (commented):
Looking good overall, left a couple of comments.

I believe we would require integration tests as well, are there plans of adding them to this PR or in a separate PR?

PiotrProkop:
@swatisehgal Thanks for the review! I plan to add integration tests in separate PR, this one became very large 😸

pkg/noderesourcetopology/least_numa.go
PiotrProkop: I think the name of the variable is confusing. The intend of  this check was to see which container in a POD requires the most NUMA nodes and return it :) I'll change the variable name to `maxNUMANodesCount`

pkg/noderesourcetopology/least_numa_test.go
PiotrProkop: 👍

pkg/noderesourcetopology/least_numa.go
PiotrProkop: renamed

pkg/noderesourcetopology/least_numa_test.go
PiotrProkop: fixed

PiotrProkop:
A lot of :
```
"Load watcher client failed" err="Get \"http://127.0.0.1:38061/watcher\": dial tcp 127.0.0.1:38061: connect: connection refused"
"Unable to update metrics" err="Get \"http://127.0.0.1:38061/watcher\": dial tcp 127.0.0.1:38061: connect: connection refused"
```
seems unrelated to this PR, especially I am planning to add integration tests in new PR

PiotrProkop:
> /retest

Hmm, bot didn't picked it up. Trying once again.
/retest

pkg/noderesourcetopology/least_numa.go
ffromani: thanks, I'm still forming thoughts

ffromani: after some thinking, I'm more confident on the proposed approach. I think is good enough for now. I'm still not completely happy with how the code looks, but this alone is not anymore sufficient to hold this PR. I'll have a final pass and if everything else looks good (as IIRC it should) I'll LGTM, and we can iterate over improvements on future PR.

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,195 @@
...
+func leastNUMAContainerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList) (int64, *framework.Status) {
+	nodes := createNUMANodeList(zones)
+	qos := v1qos.GetPodQOS(pod)
+
+	maxNUMANodesCount := 0
+	// the order how TopologyManager asks for hint is important so doing it in the same order
+	// https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/topologymanager/scope_container.go#L52
+	for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
+		identifier := fmt.Sprintf("%s/%s/%s", pod.Namespace, pod.Name, container.Name)
+		numaNodes, numaNodesCount := numaNodesRequired(identifier, qos, nodes, container.Resources.Requests)
```
ffromani, line 49: can `numaNodesCount` ever be different from `len(numaNodes.GetBits())` ?
I guess only on error. If so it could be worth to return `bitmask.Bitmask, error` WDYT?

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,195 @@
...
+		if numaNodesCount < 1 {
+			errMsg := fmt.Sprintf("cannot calculate NUMA score for container: %s", container.Name)
```
ffromani, line 51: if `numaNodesRequired` returns error, we can build this error inside the function and just consume it here (as `err.String()` or something like it)

ffromani (commented):
I''m happy enough with `subtractFromNUMAs`. I think the following patch can improve it even more, but we can defer to other improvement PRs in this area (your call @PiotrProkop ).
`leastNUMAContainerScopeScore` needs from further review from my side, on it. Everything else LGTM
```
diff --git a/pkg/noderesourcetopology/filter.go b/pkg/noderesourcetopology/filter.go
index 7f3b376..c100b15 100644
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -80,7 +80,7 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
                }
                // subtract the resources requested by the container from the given NUMA.
                // this is necessary, so we won't allocate the same resources for the upcoming containers
-               subtractFromNUMAs(nodes, []int{idx}, container.Resources.Requests)
+               subtractFromNUMAs(container.Resources.Requests, nodes, idx)
        }
        return nil
 }
diff --git a/pkg/noderesourcetopology/least_numa.go b/pkg/noderesourcetopology/least_numa.go
index be6de0c..bbea82b 100644
--- a/pkg/noderesourcetopology/least_numa.go
+++ b/pkg/noderesourcetopology/least_numa.go
@@ -59,7 +59,7 @@ func leastNUMAContainerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList)
 
                // subtract the resources requested by the container from the given NUMA.
                // this is necessary, so we won't allocate the same resources for the upcoming containers
-               subtractFromNUMAs(nodes, numaNodes.GetBits(), container.Resources.Requests)
+               subtractFromNUMAs(container.Resources.Requests, nodes, numaNodes.GetBits()...)
        }
 
        return normalizeScore(maxNUMANodesCount), nil
@@ -88,7 +88,7 @@ func normalizeScore(numaNodes int) int64 {
        return framework.MaxNodeScore - int64(numaNodes)*numaScore
 }
 
-func subtractFromNUMAs(numaNodes NUMANodeList, nodes []int, resources v1.ResourceList) {
+func subtractFromNUMAs(resources v1.ResourceList, numaNodes NUMANodeList, nodes ...int) {
        for resName, quantity := range resources {
                for _, node := range nodes {
                        // quantity is zero no need to iterate through another NUMA node, go to another resource
diff --git a/pkg/noderesourcetopology/least_numa_test.go b/pkg/noderesourcetopology/least_numa_test.go
index 3b11903..a1c96e6 100644
--- a/pkg/noderesourcetopology/least_numa_test.go
+++ b/pkg/noderesourcetopology/least_numa_test.go
@@ -103,7 +103,7 @@ func TestSubstractNUMA(t *testing.T) {
 
        for _, tcase := range tcases {
                t.Run(tcase.description, func(t *testing.T) {
-                       subtractFromNUMAs(tcase.numaNodes, tcase.nodes, tcase.resources)
+                       subtractFromNUMAs(tcase.resources, tcase.numaNodes, tcase.nodes...)
                        for i, node := range tcase.numaNodes {
                                for resName, quantity := range node.Resources {
                                        if !tcase.expected[i].Resources[resName].Equal(quantity) {
```

pkg/noderesourcetopology/least_numa.go
PiotrProkop: Totally makes sense 😄 my first attempt on this function was just to return `numaNodesCount` and after further refinements and adding `numaNodes` as return value I just didn't think of using it. Will change it now.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: ok, I'll make `subtractFromNodes` a function of `NUMANodeList` in this PR. Let me just implement this today 😄

pkg/noderesourcetopology/least_numa.go
PiotrProkop: I'll change `numaNodesRequired` to return `(bitmask.Bitmask, error)` , so we can just do traditional Golangish:
```
if err != nil{
}
```

pkg/noderesourcetopology/least_numa.go
PiotrProkop: changed.

pkg/noderesourcetopology/least_numa.go
PiotrProkop: changed

pkg/noderesourcetopology/least_numa.go
PiotrProkop: changed

ffromani: I'm not convinced this is the best way forward, but I'll surely evaluate. Thanks for addressing the comments quickly!

PiotrProkop: @ffromani sorry for the confusion i haven't noticed your git diff for the function and I assumed the `proposed` approach is the struct func 🤦  I like the variadic function approach better, it gets rid of passing `[]int{idx}` as param. Let me change this PR quickly.

PiotrProkop: I have changed the implementation to apply you patch 😄

ffromani: no problem!

ffromani (commented):
thansk @PiotrProkop !
I'm (very) happy with this PR. I'm spending some time to make sure I completely understand the `leastNUMAContainerScopeScore` flow and there are not bugs which should surface during the review ;)
Once I complete this, LGTM is on its way

PiotrProkop:
@ffromani thanks for the update before weekend 😄  !

ffromani:
/lgtm
I think I get it and looks OK.
The only pending concern could be computing all the combinations, which can be costly.
Costly, but not worse than other things we do already (thinking of pod fingerprinting computation as easy example), and same as for those other things, no clearly better alternative exist yet.
I think we can keep on eye on performances (perhaps adding metrics on future PRs) and iterating as we go.

ffromani:
/assign @Huang-Wei 
for final review

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,157 @@
...
+
+func leastNUMAPodScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList) (int64, *framework.Status) {
+	nodes := createNUMANodeList(zones)
+	qos := v1qos.GetPodQOS(pod)
+
+	identifier := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+
+	resources := util.GetPodEffectiveRequest(pod)
+
+	numaNodes, err := numaNodesRequired(identifier, qos, nodes, resources)
```
Huang-Wei, line 74: in reality, how possible `numaNodesRequired()` would return an error? and is this error _critical_?

I asked this is b/c in the scheduler framework, a score plugin's [failure](https://github.com/kubernetes/kubernetes/blob/9787f46bbf05349ca53cd8b04cd2f7a76771646a/pkg/scheduler/framework/runtime/framework.go#L939-L957) would cause [the whole scheduling](https://github.com/kubernetes/kubernetes/blob/9787f46bbf05349ca53cd8b04cd2f7a76771646a/pkg/scheduler/schedule_one.go#L371-L374) to entirely fail. So it's pretty rare that a Score plugin to return critical error.

Huang-Wei (commented):
Thanks for the PR @PiotrProkop ! Some comments below.

Also, we need an integration test and UT covering the new ScoringStrategyType.

apis/config/types.go
```diff
@@ -128,6 +128,8 @@ const (
 	BalancedAllocation ScoringStrategyType = "BalancedAllocation"
 	// LeastAllocated strategy favors node with the most amount of available resource
 	LeastAllocated ScoringStrategyType = "LeastAllocated"
+	// LeastNUMANodes strategy favors nodes which requires least amount of NUMA nodes to satisfy resource requests for given pod
+	LeastNUMANodes ScoringStrategyType = "LeastNUMANodes"
```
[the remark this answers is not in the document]
PiotrProkop: 👍

pkg/noderesourcetopology/filter.go
```diff
@@ -73,9 +73,14 @@ func singleNUMAContainerLevelHandler(pod *v1.Pod, zones topologyv1alpha1.ZoneLis
 			return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("cannot align container: %s", container.Name))
 		}
 
+		// nodes list doesn't have to be sorted be ID key, map numaID to index in nodes list
+		idx := numaIDxToID(nodes, numaID)
```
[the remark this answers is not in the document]
PiotrProkop: No. I can introduce new `func` to subtract from nodes in scoring plugin and do not touch how it was calculated in `filter` plugin. We can move to new func in the future.

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,157 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
+	"k8s.io/kubernetes/pkg/kubelet/cm/topologymanager/bitmask"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+
+	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	"gonum.org/v1/gonum/stat/combin"
```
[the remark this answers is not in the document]
PiotrProkop: I have sorted imports in way how `filter.go` and `score.go` does it. Example from `filter.go`:
```go
import (
	"context"
	"fmt"

	v1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/resource"
	"k8s.io/klog/v2"
	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
	bm "k8s.io/kubernetes/pkg/kubelet/cm/topologymanager/bitmask"

	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
)
```
So the new imports should be like:
```go
import (
	"fmt"

	v1 "k8s.io/api/core/v1"
	"k8s.io/apimachinery/pkg/api/resource"
	"k8s.io/klog/v2"
	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
	"k8s.io/kubernetes/pkg/kubelet/cm/topologymanager/bitmask"

	topologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
	"gonum.org/v1/gonum/stat/combin"

)
```
?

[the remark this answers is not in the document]
ffromani: It's very possible `filter.go` and `score.go` are not great examples and can use some cleanup. Material for a future, different PR of course (I can work on that)

pkg/noderesourcetopology/plugin.go
```diff
@@ -43,31 +43,49 @@ type NUMANode struct {
...
-		filter: singleNUMAContainerLevelHandler,
-		score:  containerScopeScore,
-	}
-}
+type filterFn func(pod *v1.Pod, zones topologyv1alpha1.ZoneList, nodeInfo *framework.NodeInfo) *framework.Status
+type scoringFn func(*v1.Pod, topologyv1alpha1.ZoneList) (int64, *framework.Status)
 
-type PolicyHandlerMap map[topologyv1alpha1.TopologyManagerPolicy]tmScopeHandler
+type PolicyFilterMap map[topologyv1alpha1.TopologyManagerPolicy]filterFn
+type PolicyScoringMap map[topologyv1alpha1.TopologyManagerPolicy]scoringFn
```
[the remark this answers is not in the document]
PiotrProkop: 👍

pkg/noderesourcetopology/least_numa.go
```diff
@@ -0,0 +1,157 @@
...
+	"sigs.k8s.io/scheduler-plugins/pkg/util"
+)
+
+var zeroQty resource.Quantity
+
+func init() {
+	zeroQty = resource.MustParse("0")
+}
```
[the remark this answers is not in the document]
ffromani: nice, TIL

pkg/noderesourcetopology/least_numa.go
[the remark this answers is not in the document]
PiotrProkop: ok, will sort them like my example above ^^

pkg/noderesourcetopology/least_numa.go
PiotrProkop: The only case when `numaNodesRequired` returns error is when the container/pod requires more resources than available on the worker node, so it's rather impossible since score plugin should  run after `NodeResourcesFit` plugin. Instead of an error I can print warning message and return `minScore`(bitmask with all 8 bits set) here.

pkg/noderesourcetopology/least_numa.go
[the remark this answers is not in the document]
PiotrProkop: will do!

pkg/noderesourcetopology/least_numa.go
PiotrProkop: So I have changed this function signature to return only `bitmask.Bitmask`. The problem I have is that right now I have chosen to return `nil` when we can't fit resources onto the worker node and just check for `nil` to return quickly(in container scope it's necessary not to attempt subtractFromNUMAs):
```go
		if numaNodes == nil {
			return framework.MinNodeScore, nil
		}
```
I am not fully satisfied with this approach but I can't think of any simpler solution either 😞 
I was playing around with filling bitmask with all bits up to `highestNUMAiD` but it is very clumsy:
```go
allNodesBitmask, _ := bitmask.NewBitMask(0, 1, 2, 3, 4, 5, 6, 7, 8)
```

PiotrProkop:
> Thanks for the PR @PiotrProkop ! Some comments below.
> 
> Also, we need an integration test and UT covering the new ScoringStrategyType.

I was planning to add integration tests as separate PR(this one is quite big already) but If we feel like it's necessary to merge it I am happy to add them this week. 

This PR is improving unit test coverage for `noderesourcetopology` from:
```sh
~/Repos/pprokop/scheduler-plugins/pkg/noderesourcetopology (master) $ go test -cover .
ok      sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology  0.021s  coverage: 69.5% of statements
```
to
```sh
~/Repos/pprokop/scheduler-plugins/pkg/noderesourcetopology (least-numa-scoring-strategy %) $ go test -cover .
ok      sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology  0.038s  coverage: 73.1% of statements
```

PiotrProkop:
@Huang-Wei @ffromani I have added integration tests as a new commit in this PR.

ffromani:
> @Huang-Wei @ffromani I have added integration tests as a new commit in this PR.

thanks, I'll review them shortly

test/integration/noderesourcetopology_test.go
```diff
@@ -622,6 +628,319 @@ func TestTopologyMatchPlugin(t *testing.T) {
...
+							noderesourcetopology.MakeTopologyResInfo(cpu, "4", "4"),
+							noderesourcetopology.MakeTopologyResInfo(memory, "8Gi", "8Gi"),
+						}).
+					Zone(
+						topologyv1alpha1.ResourceInfoList{
+							noderesourcetopology.MakeTopologyResInfo(cpu, "4", "4"),
+							noderesourcetopology.MakeTopologyResInfo(memory, "8Gi", "3Gi"),
+						}).Obj(),
+			},
+			expectedNodes: []string{"fake-node-1", "fake-node-2"},
```
ffromani, line 837: The slice is sorted from the less desirable (leftmost) to most desirable (rightmost) right?

ffromani (commented):
just a question inside. If the answer is positive, then LGTM

test/integration/noderesourcetopology_test.go
PiotrProkop: What do you mean by desirable? Both nodes will have the same score(requires same amount of NUMA nodes to satisfy resource requests).

PiotrProkop: There are couple of those:
* `Scheduling Guaranteed pod with LeastNUMANodes strategy scheduler, different allocation pod scope`
* `Scheduling Guaranteed multi container pod with LeastNUMANodes strategy scheduler and pod scope`
* `Scheduling Guaranteed pod with multiple containers with LeastNUMANodes strategy scheduler, different resource allocation, pod scope`
* `Scheduling Guaranteed pod with multiple containers with LeastNUMANodes strategy scheduler, different resource allocation, container scope`

Only one test is checking the case where a pod can't fit on a node:
* `Scheduling Guaranteed multiple containers pod with LeastNUMANodes strategy scheduler, can't fit one node, pod scope`

ffromani: Thanks! what I would like to have covered is a case on which the score is actually different from 2+ compute nodes without interference from NodeResourceTopoologyMatch `filter` . Unless I'm mistaken, 
in `Scheduling Guaranteed pod with LeastNUMANodes strategy scheduler, different allocation pod scope`
the pod here can't fit on either NUMA zones on fake-node-1 (requests too many CPUs), so the filter is ruling it out, so the score would only score fake-node-2
same for `Scheduling Guaranteed multi container pod with LeastNUMANodes strategy scheduler and pod scope` and few others.
So existing integration tests looks good, I would just like to see 1+ tests on which the score is the deciding factor without being constrained by `filter` into a obvious decision.

I reckon that creating such condition can be nontrivial, exactly because we should NOT disable `filter` if possible

ffromani (commented):
/lgtm
considering our discussion about integration tests, we can add more tests in the future. The existing coverage seems good enough start.

test/integration/noderesourcetopology_test.go
PiotrProkop: I think `filter` plugins only works for `SingleNUMANode` policies and is noop for other policies, so in this example it's won't interfere.

PiotrProkop: Yeah, after passing `-v=5` to tests we can see:
```sh
=== RUN   TestTopologyMatchPlugin/Scheduling_Guaranteed_pod_with_multiple_containers_with_LeastNUMANodes_strategy_scheduler,_different_resource_allocation,_pod_scope
    noderesourcetopology_test.go:1173: Start-topology-match-test Scheduling Guaranteed pod with multiple containers with LeastNUMANodes strategy scheduler, different resource allocation, pod scope
    noderesourcetopology_test.go:1183: Creating Pod "topology-aware-scheduler-pod"
I0124 13:17:34.911563 2433617 eventhandlers.go:116] "Add event for unscheduled pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:34.911619 2433617 scheduling_queue.go:957] "About to try and schedule pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:34.911636 2433617 schedule_one.go:85] "Attempting to schedule pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:34.911749 2433617 passthrough.go:38] "Lister for nodeResTopoPlugin" lister=&{indexer:0xc000536108}
I0124 13:17:34.911771 2433617 filter.go:209] "Found NodeResourceTopology" nodeTopology="fake-node-1"
I0124 13:17:34.911779 2433617 passthrough.go:38] "Lister for nodeResTopoPlugin" lister=&{indexer:0xc000536108}
I0124 13:17:34.911806 2433617 filter.go:217] "Policy handler not found" policy="BestEffortPodLevel"
I0124 13:17:34.911812 2433617 filter.go:209] "Found NodeResourceTopology" nodeTopology="fake-node-2"
I0124 13:17:34.911827 2433617 filter.go:217] "Policy handler not found" policy="BestEffortPodLevel"
I0124 13:17:34.911918 2433617 passthrough.go:38] "Lister for nodeResTopoPlugin" lister=&{indexer:0xc000536108}
I0124 13:17:34.911929 2433617 passthrough.go:38] "Lister for nodeResTopoPlugin" lister=&{indexer:0xc000536108}
I0124 13:17:34.912130 2433617 default_binder.go:52] "Attempting to bind pod to node" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod" node="fake-node-2"
I0124 13:17:34.915339 2433617 cache.go:402] "Finished binding for pod, can be expired" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:34.915400 2433617 schedule_one.go:266] "Successfully bound pod to node" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod" node="fake-node-2" evaluatedNodes=2 feasibleNodes=2
I0124 13:17:34.915834 2433617 eventhandlers.go:159] "Delete event for unscheduled pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:34.915847 2433617 eventhandlers.go:184] "Add event for scheduled pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
    noderesourcetopology_test.go:1200: Pod topology-aware-scheduler-pod scheduled
    noderesourcetopology_test.go:1207: Pod "topology-aware-scheduler-pod" is on a nodes as expected.
    noderesourcetopology_test.go:1234: Case Scheduling Guaranteed pod with multiple containers with LeastNUMANodes strategy scheduler, different resource allocation, pod scope finished
I0124 13:17:35.923940 2433617 eventhandlers.go:204] "Update event for scheduled pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:35.925712 2433617 eventhandlers.go:229] "Delete event for scheduled pod" pod="integration-test-8149a00c-fb72-4a0d-b849-53d6277354b3/topology-aware-scheduler-pod"
I0124 13:17:35.938264 2433617 nrtutils.go:136] init scheduler success
```
and line `I0124 13:17:34.911827 2433617 filter.go:217] "Policy handler not found" policy="BestEffortPodLevel"` means the `filter` plugin is not being triggered for this pod

ffromani: ok, I was at fault: I was misremembering what `expectedNodes` means in this context, so the above tests are indeed good and cover the cases I was hinting. It seems we're all good.

pkg/noderesourcetopology/filter.go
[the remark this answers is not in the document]
Huang-Wei: Thanks. Feel free to add the filter changes in follow-up PR.

pkg/noderesourcetopology/plugin.go
```diff
@@ -96,19 +126,27 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
...
 	for _, resource := range tcfg.ScoringStrategy.Resources {
 		resToWeightMap[v1.ResourceName(resource.Name)] = resource.Weight
 	}
 
+	var scoringHandlers scoreHandlersMap
+
+	if tcfg.ScoringStrategy.Type == apiconfig.LeastNUMANodes {
+		scoringHandlers = leastNUMAscoreHandlers()
+	} else {
+		strategy, err := getScoringStrategyFunction(tcfg.ScoringStrategy.Type)
```
Huang-Wei, line 139: (continue the comment from https://github.com/kubernetes-sigs/scheduler-plugins/pull/474#discussion_r1083109506)

I _thought_  scheduler-plugin has the logic in place for plugin arguments, but seems not. Fine with current logic.

A general follow-up can be done to catch all illegal plugin arguments like https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/apis/config/validation/validation_pluginargs.go.

pkg/noderesourcetopology/least_numa.go
Huang-Wei: > so it's rather impossible since score plugin should run after NodeResourcesFit plugin

This is very true. Maybe add a comment here:

```go
if numaNodes == nil {
	// Should not reach here.
	return framework.MinNodeScore, nil
}
```

Huang-Wei:
> I was planning to add integration tests as separate PR(this one is quite big already) but If we feel like it's necessary to merge it I am happy to add them this week.

Aha, I thought you _missed_ that. (next time you can say "integration test will come in a follow-up PR", then I know you have that in your mind)

Huang-Wei:
/approve

Thanks @PiotrProkop !

PR #477, opened 2023-01-16
Clean up manifests for all-in-one installation

tenzen-y:
#### What type of PR is this?

/kind bug
/kind cleanup
/kind documentation

I cleaned up manifests for all-in-one installation. Also, I fix the installation guide.

1. To avoid the following errors, I added ClusterRole to allow all verbs for  "podgroups/status", "elasticquotas/status".

```shell
E0115 18:05:06.743727       1 podgroup.go:178] "Error syncing pod group" err="podgroups.scheduling.sigs.k8s.io \"tfjob-simple\" is forbidden: User \"system:serviceaccount:scheduler-plugins:scheduler-plugins-controller\" cannot patch resource \"podgroups/status\" in API group \"scheduling.sigs.k8s.io\" in the namespace \"kubeflow\"" podGroup="kubeflow/tfjob-simple"
```

2. I replaced the [deprecated](https://github.com/kubernetes/api/blob/59fcd23597fd090dba6b7e903eb0a8c9e8efb0a6/core/v1/generated.proto#L3600-L3604) serviceAccount field with serviceAccountName field.
3. Fix the image version for the installation guide.

Fixes #

```release-note

```

### DIFF
--- a/manifests/install/all-in-one.yaml
+++ b/manifests/install/all-in-one.yaml
@@ -77,7 +77,7 @@ spec:
       labels:
         app: scheduler-plugins-controller
     spec:
-      serviceAccount: scheduler-plugins-controller
+      serviceAccountName: scheduler-plugins-controller
       containers:
       - name: scheduler-plugins-controller
         image: registry.k8s.io/scheduler-plugins/controller:v0.24.9
--- a/manifests/install/all-in-one.yaml
+++ b/manifests/install/all-in-one.yaml
@@ -6,7 +6,7 @@ metadata:
   name: system:kube-scheduler:plugins
 rules:
 - apiGroups: ["scheduling.sigs.k8s.io"]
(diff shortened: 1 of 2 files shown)


manifests/install/all-in-one.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   name: system:kube-scheduler:plugins
 rules:
 - apiGroups: ["scheduling.sigs.k8s.io"]
-  resources: ["podgroups", "elasticquotas"]
+  resources: ["podgroups", "elasticquotas", "podgroups/status", "elasticquotas/status"]
```
Huang-Wei, line 9: only the helm chart (#471) was updated 😑

Huang-Wei (commented):
Thanks @tenzen-y !

Issue #445, opened 2022-11-04
hack/update-codegen.sh is not useful


zffocussss:
Hi team:
  as I want to add a new out-of-tree plugin,after preparing the data structure, issuing hack/update-codegen.sh,no deepcopy is generated.can you help me?

JingbenShi668:
> Hi team: as I want to add a new out-of-tree plugin,after preparing the data structure, issuing hack/update-codegen.sh,no deepcopy is generated.can you help me?

You need to notice the storage location of your cloned repo. For example, if you attempt to generate code for your out-of-tree plugin in Linux environment, the cloned repo location should be like $GOPATH/src/sigs.k8s.io/scheduler-plugins. 

In scheduler-plugins/hack/update-codegen.sh file, the prefix of the directory positions in bash command in the red location may be custom. 
<img width="748" alt="image" src="https://user-images.githubusercontent.com/39040502/200188227-2faeedf7-c6a4-42ab-ab19-cd33f1b0fded.png">

I just use the default repo name and it is scheduler-plugins. I can successfully  generate code for my custom plugin. But it seems that some generated code is wrong in the scheduler-plugins/pkg/apis/config/v1beta1/zz_generated.defaults.go file. 
In other words, it doesn't generate setDefaultValue function.

Huang-Wei:
@zffocussss did you place your repo under GOPATH? and as @JingbenShi668 mentioned, you may need to tweak some literals in that script if you're running at a different folder.

zffocussss:
> 

thanks,I will try as your advice

zffocussss:
> 

@Huang-Wei no,I just put my repo in other place,without relation to go env.

zwpaper:
after https://github.com/kubernetes-sigs/scheduler-plugins/pull/481 get merged, scheduler plugins can now work outside of GOPATH.

this issue should be ok to close, @Huang-Wei @zffocussss

Huang-Wei:
/close

Workaround for this issue: leverage #481, or place the repo under GOPATH.

PR #481, opened 2023-01-24
update 'update-codegen.sh' to make it works outside GOPATH

zwpaper:

#### What type of PR is this?

/kind bug

https://github.com/kubernetes-sigs/scheduler-plugins/commit/b5ea0f1f188e37f770868c7a74a9ae91107dd4e1#diff-149dfe7bb29d1191dceae3a52915e750e64b7f87257a5fb309c29d3056e2a95d This commit previously deleted the `generate-group.sh`, but then can not gen clientset, informers and listers. 

this PR:
1. added the `generate-group.sh` back to generate clientset, informers, listers for scheduler-plugins
2. bump controller gen to `v0.11.1` to match https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/hack/verify-crdgen.sh#L30
3. add `output-base` for codegen to make it works outside GOPATH

Fixes #

ref: https://github.com/kubernetes-sigs/kwok/pull/226

```release-note

```

### DIFF
--- a/config/crd/bases/scheduling.sigs.k8s.io_elasticquotas.yaml
+++ b/config/crd/bases/scheduling.sigs.k8s.io_elasticquotas.yaml
@@ -1,11 +1,10 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/52 # edited manually
-    controller-gen.kubebuilder.io/version: v0.6.2
+    api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/52
+    controller-gen.kubebuilder.io/version: v0.11.1
   creationTimestamp: null
   name: elasticquotas.scheduling.sigs.k8s.io
 spec:
@@ -79,9 +78,5 @@ spec:
         type: object
     served: true
     storage: true
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
+    subresources:
+      status: {}
--- a/config/crd/bases/scheduling.sigs.k8s.io_podgroups.yaml
+++ b/config/crd/bases/scheduling.sigs.k8s.io_podgroups.yaml
@@ -1,11 +1,10 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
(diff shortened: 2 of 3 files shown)


Huang-Wei:
@zwpaper thanks for the PR, but I cannot merge as-is now.

There is a history for https://github.com/kubernetes-sigs/scheduler-plugins/commit/b5ea0f1f188e37f770868c7a74a9ae91107dd4e1#diff-149dfe7bb29d1191dceae3a52915e750e64b7f87257a5fb309c29d3056e2a95d. Given it's pretty common for scheduler plugins to manipulate CR objects, and more important, **_different_** CR objects. So it makes a lot of sense to not use generated clients (the traditional k/k way); instead, leverage controller-runtime and follow kubebuilder's annotation syntac to use general (dynamic) client to manipulate the CR objects. However, that migration (from static clients to controller-runtime's client) is half-way done due to lack of volunteers. We need to finish it up sometime so we don't need static clients anymore.

zwpaper:
Hi @Huang-Wei, got it! as for the volunteers part, where can I dig deeper into the controller-runtime topic? I would like to contribute to it, I used the operator from time to time, and maybe there is something I could help with.

Huang-Wei:
> Hi @Huang-Wei, got it! as for the volunteers part, where can I dig deeper into the controller-runtime topic? I would like to contribute to it, I used the operator from time to time, and maybe there is something I could help with.

I need to take a look at the codebase first.. to see where we were left at. And then come up with a plan to work on the migration. Stay tuned.

zwpaper:
Hi @Huang-Wei, I have updated the PR, and deleted the generated clientset related parts.

it is still necessary to do this fix because we still need `update-internal-group.sh` to generate deepcopy,defaults,conversion, and it should work outside GOPATH after this PR get merged.

also, I have updated the `controller-gen` to `0.11.1` to match https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/hack/verify-crdgen.sh?rgh-link-date=2023-01-24T09%3A50%3A10Z#L30

zwpaper:
as for the controller runtime part, I will dig into the code base before your plan is ready😁

zwpaper:
happened to find this PR may fix https://github.com/kubernetes-sigs/scheduler-plugins/issues/445 as it failed due to not putting `scheduler-plugins` under GOPATH

Huang-Wei:
> I have updated the PR, and deleted the generated clientset related parts.

Thanks. The current state looks good.

/ok-to-test

PR #486, opened 2023-01-29
Fix list items numbering

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

#### What type of PR is this?

/kind documentation

When reading the document through the code, it turns out that the order list in the document is wrong, and this pr fixes the problem

Fixes #

```release-note
NONE
```

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -193,7 +193,7 @@ Now, we're able to verify how the coscheduling plugin works.
     $ kubectl apply -f podgroup.yaml
     ```
 
-1. Create a deployment labelled `pod-group.scheduling.sigs.k8s.io: pg1` to associated with PodGroup
+2. Create a deployment labelled `pod-group.scheduling.sigs.k8s.io: pg1` to associated with PodGroup
    `pg1` created in the previous step.
 
     ```yaml
@@ -230,7 +230,7 @@ Now, we're able to verify how the coscheduling plugin works.
     >       schedulerName: scheduler-plugins-scheduler
     > ```
 
-1. As PodGroup `pg1` requires at least 3 pods to be scheduled all-together, and there are only 2 Pods
+3. As PodGroup `pg1` requires at least 3 pods to be scheduled all-together, and there are only 2 Pods
    so far, so it's expected to observer they are pending:
 
     All nginx pods are expected to be `Pending` as they cannot be co-scheduled altogether.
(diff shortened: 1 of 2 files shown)


PiotrProkop:
@Fish-pro the syntax is correct Markdown allows to use only `1.` in ordered lists and it will be automatically converted into proper indexes, check reference here:
https://www.markdownguide.org/basic-syntax/#ordered-lists

Fish-pro:
> @Fish-pro the syntax is correct Markdown allows to use only `1.` in ordered lists and it will be automatically converted into proper indexes, check reference here: https://www.markdownguide.org/basic-syntax/#ordered-lists

Thanka for tips，I understand that either way is OK, whether it would be better to follow the order？

PiotrProkop:
> > @Fish-pro the syntax is correct Markdown allows to use only `1.` in ordered lists and it will be automatically converted into proper indexes, check reference here: https://www.markdownguide.org/basic-syntax/#ordered-lists
> 
> Thanka for tips，I understand that either way is OK, whether it would be better to follow the order？

I think it brings unnecessary burden 😄 for me using `1` everywhere is simpler and consistent. But this is my opinion.

Fish-pro:
> > > @Fish-pro the syntax is correct Markdown allows to use only `1.` in ordered lists and it will be automatically converted into proper indexes, check reference here: https://www.markdownguide.org/basic-syntax/#ordered-lists
> > 
> > 
> > Thanka for tips，I understand that either way is OK, whether it would be better to follow the order？
> 
> I think it brings unnecessary burden 😄 for me using `1` everywhere is simpler and consistent. But this is my opinion.

@PiotrProkop Thanks, I will close this pr

Fish-pro:
> > > @Fish-pro the syntax is correct Markdown allows to use only `1.` in ordered lists and it will be automatically converted into proper indexes, check reference here: https://www.markdownguide.org/basic-syntax/#ordered-lists
> > 
> > 
> > Thanka for tips，I understand that either way is OK, whether it would be better to follow the order？
> 
> I think it brings unnecessary burden 😄 for me using `1` everywhere is simpler and consistent. But this is my opinion.

@PiotrProkop Thanks, I will close this pr

Huang-Wei:
I'm with @PiotrProkop on using constant `1` for bullet listing :)

PR #488, opened 2023-01-30
noderesourcetopology: overhaul Topology Manager configuration management

ffromani:
#### What type of PR is this?

/kind cleanup

The NodeResourceTopology API is deprecating the `TopologyPolicies` field, which is the way the agents (NFD) exported the relevant Topology Manager options (see: https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/issues/24 https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/25).
The new recommended option is to expose the configuration using Attributes.
Add relevant code to generalize the configuration management while keep backward compatibility with the current generation of updaters

Fixes N/A

N/A

```release-note
the NodeTopologyMatch plugins now depend on v1alpha2 version of the noderesourcetopology API.
```


ffromani:
/hold cancel
@swatisehgal: do we want to wait for NRT v1alpha2 and put this PR on hold up until that or do we want to go with this fake-zone approach? (I have a _small_ preference towards the fake-zone approach because it becomes kinda like a namespace, even if arguably we're [almost] abusing the zone concept)
@PiotrProkop this is the general direction we are pursuing after a conversation with NFD folks. Could be either top-level attributes (we will need v1alpha2 for that, see the last updates on https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/25) or this fake-zone approach, but we agreed attributes are better fit than labels/annotations.

PiotrProkop:
@ffromani thanks for the info. I'll review this tomorrow.

swatisehgal:
> /hold cancel @swatisehgal: do we want to wait for NRT v1alpha2 and put this PR on hold up until that or do we want to go with this fake-zone approach? (I have a _small_ preference towards the fake-zone approach because it becomes kinda like a namespace, even if arguably we're [almost] abusing the zone concept) 

I would prefer to get `v1alpha2` merged first primarily for the reasons that you have pointed out: it seems like an abuse of the zone concepts. In addition to that, it seems like a workaround when we have already agreed on a clear way forward. If we were to go ahead with this change, we would have to cleanup once we would transition to v1alpha2. I would rather us do it the right way in one shot.

We can easily unblock the transition to v1alpha2. I noticed that the PR is WIP and that is why never reviewed. I will take a look and then we can update this PR to consume v1alpha2 version of the API. 

Sounds reasonable?

ffromani:
/hold
let's wait for `v1alpha2` then!

Huang-Wei:
@swatisehgal @ffromani Just a heads-up, once you proceed to bump the version of NRT, please include a release-note to explain its potential behavior change in brief. 🙏

ffromani:
> @swatisehgal @ffromani Just a heads-up, once you proceed to bump the version of NRT, please include a release-note to explain its potential behavior change in brief. pray

ACK, I'll need to revisit the `release-note` field

PiotrProkop:
> > /hold cancel @swatisehgal: do we want to wait for NRT v1alpha2 and put this PR on hold up until that or do we want to go with this fake-zone approach? (I have a _small_ preference towards the fake-zone approach because it becomes kinda like a namespace, even if arguably we're [almost] abusing the zone concept)
> 
> I would prefer to get `v1alpha2` merged first primarily for the reasons that you have pointed out: it seems like an abuse of the zone concepts. In addition to that, it seems like a workaround when we have already agreed on a clear way forward. If we were to go ahead with this change, we would have to cleanup once we transition to v1alpha2. I would rather us do it the right way in one shot.
> 
> We can easily unblock the transition to v1alpha2. I noticed that the PR is WIP and that is why never reviewed. I will take a look and then we can update this PR to consume v1alpha2 version of the API.
> 
> Sounds reasonable?

I agree with @swatisehgal that we should wait for `v1alpha2` not to abuse `fake` zone concept(such hacks have tendencies to linger for a long time).

pkg/noderesourcetopology/config.go
```diff
@@ -0,0 +1,124 @@
...
+func updateTopologyManagerConfigFromAttributes(conf *TopologyManagerConfig, attrs topologyv1alpha2.AttributeList) {
+	for _, attr := range attrs {
+		if attr.Name == AttributeScope && IsValidScope(attr.Value) {
+			conf.Scope = attr.Value
+			continue
+		}
+		if attr.Name == AttributePolicy && IsValidPolicy(attr.Value) {
+			conf.Policy = attr.Value
+			continue
+		}
```
PiotrProkop, line 91: Can we add `TODO` to add `TopologyManagerPolicies` when `scheduler-plugins` supports k8s 1.26. Basically when this [PR](https://github.com/kubernetes-sigs/scheduler-plugins/pull/489) is merged

PiotrProkop: Yeah, my bad :) comment updated

ffromani: sure thing, it was planned already (https://github.com/kubernetes-sigs/scheduler-plugins/pull/488/commits/b131bdada31d1be1547beab439a4b1b5476cad17#diff-3532db6795688cc2309165cf54c8ad8d256cb477ca3f9a5b0f778f85aabf1b49R132) but forgot to add TODOs here

manifests/noderesourcetopology/crd.yaml
```diff
@@ -3,7 +3,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870 # edited manually
```
PiotrProkop, line 6: I think this annotation has to stay to be able to create CRD in `k8s.io` domain

pkg/noderesourcetopology/config.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return false
+}
+
+type TopologyManagerConfig struct {
+	Scope  string
+	Policy string
+}
+
+func (conf TopologyManagerConfig) IsComplete() bool {
```
PiotrProkop, line 63: I have a question, it is valid not specify both `scope` and `policy` in Kubelet config, and their will be defaulted to `container` for scope and `none` for policy. Do we assume that NFD(or whichever component will advertise those config option) will make sure to make explicit conversion from "" to default value or we should just assume here that those values weren't set and we can just change "" to  default values here.

manifests/noderesourcetopology/crd.yaml
ffromani: I agree. But it should be added automatically (not manually) if possible, hence: https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/29

pkg/noderesourcetopology/config.go
ffromani: that's a good point. I don't have an answer yet for NFD, and I'm revisiting this function and how we use it.

manifests/noderesourcetopology/crd.yaml
PiotrProkop: Ok, I can see you got this covered 😄

ffromani: hopefully! :)

pkg/noderesourcetopology/config.go
ffromani: These details should be discussed and documented, because they are a key part of the producer-consumer contract. I started a change (in this PR) to add this documentation in `pkg/noderesourcetopology/README.md`, let's continue here. If we find we need more space and discussion, we can move later to another forum (e.g. gh issue, google doc...)

PiotrProkop: Yeah, I just feel producer should not do a lot of preprocessing so we can just emit what's configured in kubelet and not bother about complicated logic. IMHO setting defaults for those values should happen in consumer(in this case NodeResourceTopologyMatch plugin).

ffromani: > Yeah, I just feel producer should not do a lot of preprocessing so we can just emit what's configured in kubelet and not bother about complicated logic. IMHO setting defaults for those values should happen in consumer(in this case NodeResourceTopologyMatch plugin).

yes, I'm converging towards this point. What I think now (subject to change/further thinking):

- producers (NFD topology updater) should try their hardest to provide the values, but it the kubelet doesn't provide sufficient introspection mechanism, they should NOT guess or worse hardcode values. I think we're already in agreement here.
- consumer (scheduler plugin) CAN have defaults, because out of necessity we duplicate TM logic and constants anyway. But I'm still reluctant to hardcode defaults here, let's see.

PiotrProkop: >     * consumer (scheduler plugin) CAN have defaults, because out of necessity we duplicate TM logic and constants anyway. But I'm still reluctant to hardcode defaults here, let's see.

No biggie for us when it comes with empty `policy` NodeResourceTopologyMatch  is noop for `none` policy. But we should probably assume that if `scope` is empty it means it is `container` scope(I bet many Kubernetes configurations uses TopologyManager but do not set `scope` and just rely on default here)

ffromani: I'm not fully sold yet, but hardcoding default policy=none and default scope=container and using them as fallback if nothing is given from the producer (NFD) is something we can surely try out and evaluate.

ffromani: Added in the last upload as promised

ffromani:
still WIP: even though we expect no major changes left for NRT API v1alpha2, let's keep this PR on hold until we have a new tag for the NRT API. I'll finish the PR once we have that.

ffromani:
the PR is now reviewable again

ffromani:
/hold cancel
we have v1apha2 API: https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/releases/tag/v0.1.0

pkg/noderesourcetopology/score.go
```diff
@@ -152,3 +148,29 @@ func containerScopeScore(pod *v1.Pod, zones topologyv1alpha1.ZoneList, scorerFn
...
+			return leastNUMAPodScopeScore
+		}
+		if conf.Scope == ScopeContainer {
+			return leastNUMAContainerScopeScore
+		}
+		return nil // cannot happen
+	}
+	if conf.Policy != PolicySingleNUMANode {
+		return nil
+	}
```
PiotrProkop, line 164: I think you can just call `getScoringStrategyFunction` func here and pass it down. By doing it you can remove `scoreStrategyFunc` field from `TopologyMatch` struct, but it would mean that either this function will return additional `error` or we should add some validation in `TopologyMatch` constructor, WDYT?

PiotrProkop:
And can you update compatibility Matrix? 😄

ffromani:
> And can you update compatibility Matrix? smile

absolutely, thanks for reminding me!

pkg/noderesourcetopology/config.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+}
+
+func updateTopologyManagerConfigFromTopologyPolicies(conf *TopologyManagerConfig, nodeName string, topologyPolicies []string) {
+	if len(topologyPolicies) == 0 {
+		klog.V(3).InfoS("Cannot determine policy", "node", nodeName)
+		return
+	}
+	if len(topologyPolicies) > 1 {
+		klog.V(4).InfoS("Ignoring extra policies", "node", nodeName, "policies", len(topologyPolicies)-1)
```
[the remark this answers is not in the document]
ffromani: thanks, fixed!

pkg/noderesourcetopology/score.go
ffromani: ok, unfortunately I have to take it back, this approach looks nicer and carrying `scoreStrategyFunc` is unfortunate, but it allows us to do the validation in the` TopologyMatch` constructor. Erroring out later in `Score` is too late and can only lead to silent failures - no way to abort, we can only set the score. Let me at least document this rationale with a code comment.

PiotrProkop: Another thing we can try is to implement suggestion from this comment https://github.com/kubernetes-sigs/scheduler-plugins/pull/474#discussion_r1086235462, I think this can vastly simplify things in the future :) or I can just make an issue to add args validation for `TopologyMatch` plugin.

ffromani:
/hold
let us wait until NFD master branch is updated to work with v1alpha2

PiotrProkop (commented):
/lgtm 
@swatisehgal for proper LGTM 🚀

pkg/noderesourcetopology/score.go
PiotrProkop: 👍

pkg/noderesourcetopology/README.md
```diff
@@ -155,6 +155,27 @@ Should the cluster need to have different settings (e.g. topology manager) or NU
...
+
+Additionally to properly partition logically your cluster like explained above, to work correctly the Topology-aware scheduler needs to know key node-specific configuration settings for the topology manager
+running on the worker nodes.
+This data is expected to be provided as top-level `Attributes` of the NodeResourceTopology objects:
+
+NodeResourceTopology producers should add top-level `Attributes` in the following format
+- For `Name` and `Value` of attributes, words should be `snakeCase`
+- The `Name` of each attribute should be **the same of the corresponding kubelet configuration option**.
+  - example: `--topology-manager-scope` becomes `topologyManagerScope`
+  - example: `topologyManagerPolicy` becomes `topologyMaanagerPolicy`
```
PiotrProkop, line 170: and I found the typo
```suggestion
  - example: `topologyManagerPolicy` becomes `topologyManagerPolicy`
```

ffromani: good catch, thanks! fixed

ffromani:
depends on https://github.com/kubernetes-sigs/node-feature-discovery/pull/1054 (just completing the cross-link reference)

ffromani:
/hold cancel
both NFD and RTE have been updated

ffromani:
@PiotrProkop @swatisehgal 
did minor cleanups/enhancements since last LGTM, could you PTAL again when you have some time?

PiotrProkop:
LGTM

pkg/noderesourcetopology/config.go
```diff
@@ -0,0 +1,124 @@
...
+	return TopologyManagerConfig{
+		Scope:  ScopeContainer,
+		Policy: PolicyNone,
+	}
+}
+
+func topologyManagerConfigFromNodeResourceTopology(nodeTopology *topologyv1alpha2.NodeResourceTopology) TopologyManagerConfig {
+	conf := makeTopologyManagerConfigDefaults()
+	// backward compatibility (v1alpha2 and previous)
+	updateTopologyManagerConfigFromTopologyPolicies(&conf, nodeTopology.Name, nodeTopology.TopologyPolicies)
```
swatisehgal, line 75: IIUC, we need both these ways of obtaining topology policies as the exporter still might be using the `topologyPolicies` field for exposing information and we want to continue supporting that as  a precautionary measure. We are intentionally taking the route of soft deprecation and not fully deprecating the use of `topologyPolicies` in the v1alpha2 version of NRT API . Let's emphasize this and the fact that in the next iteration (when we move to v1beta1 version) we would be deprecating the use of `topologyPolicies` field.

test/integration/noderesourcetopology_cache_test.go
```diff
@@ -531,10 +531,17 @@ func TestTopologyCachePluginWithUpdates(t *testing.T) {
...
 				Annotations(map[string]string{
-					podfingerprint.Annotation: mkPFP("fake-node-cache-1", tt.podDescs[0].pod),
+					podfingerprint.Annotation: pfpSign,
+				}).
+				Attributes(topologyv1alpha2.AttributeList{
+					{
+						Name:  podfingerprint.Attribute,
+						Value: pfpSign,
+					},
 				}).
```
swatisehgal, line 545: I am really happy about the use of attributes for capturing podfingerprint. It is much nicer than the use of annotations we did previously!

pkg/noderesourcetopology/README.md
```diff
@@ -155,6 +155,28 @@ Should the cluster need to have different settings (e.g. topology manager) or NU
 using [affinity](https://kubernetes.io/docs/user-guide/node-selection/#node-affinity-beta-feature) or also
 [taints](https://kubernetes.io/docs/user-guide/node-selection/#taints-and-toleations-beta-feature).
 
+#### Topology Manager configuration
+
+***Target audience: developers and operators of topology updaters (NodeResourceTopology producers)***
+
+Additionally to properly partition logically your cluster like explained above, to work correctly the Topology-aware scheduler needs to know key node-specific configuration settings for the topology manager
+running on the worker nodes.
```
swatisehgal, line 163: Just a small rephrase: 
```suggestion
In addition to logically partitioning a cluster like explained above, Topology-aware scheduler needs to know key node-specific configuration settings like Topology manager policy and scope.
```

pkg/noderesourcetopology/README.md
```diff
@@ -155,6 +155,28 @@ Should the cluster need to have different settings (e.g. topology manager) or NU
...
+  - example: `topologyManagerPolicy` becomes `topologyManagerPolicy`
+- The `Value` of each attribute should be **one of the value of the corresponding kubelet configuration option, VERBATIM**.
+  - example: `single-numa-node` becomes `single-numa-node`
+- Should `topologyManagerOptions` be exposed:
+  - they should be expanded in key-value pairs, using the `String()` representation
+  - each key-value pair should be preceded by the `topologyManagerOption` prefix
+  - every other provision described above applies
+  - example: the `prefer-closest-numa-nodes` option becomes `topologyManagerOptionPreferClosestNumaNodes`, accepting exactly one of either `true` and `false`.
+  - **RATIONALE**: this representation wants to guarantee all the Attribute Names are unique (no aliasing). It must be noted this is a stricter requirement with respect to the Attribute representation
+    in NRT objects, and this requirement could be lifted in the future (a upgrade path will be provided).
```
swatisehgal, line 179: NIT:
``` suggestion
    in NRT objects, and this requirement could be lifted in the future (an upgrade path will be provided).
```

swatisehgal (commented):
Thanks for your work on this PR!

I left some minor comments and pointed out some NITs but overall the PR looks good to me.

pkg/noderesourcetopology/objects.go
```diff
@@ -0,0 +1,127 @@
...
+				Requests: r,
+				Limits:   r,
+			},
+		})
+	}
+
+	return pod
+}
+
+func makePodWithReqByResourceList(resources *corev1.ResourceList) *corev1.Pod {
```
Tal-or, line 60: We can use functional options here instead https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis 
But probably in separate PR

pkg/noderesourcetopology/objects.go
```diff
@@ -0,0 +1,127 @@
...
+					Resources: corev1.ResourceRequirements{
+						Requests: *resources,
+					},
+				},
+			},
+		},
+	}
+}
+
+func makePodWithReqAndLimitByResourceList(resourcesReq, resourcesLim *corev1.ResourceList) *corev1.Pod {
```
Tal-or, line 74: dito

Tal-or (commented):
Thank you for the PR.
LGTM

pkg/noderesourcetopology/config.go
```diff
@@ -0,0 +1,124 @@
...
+	ScopePod       = "pod"
+)
+
+// TODO: can we reuse constants from k/k/pkg/kubelet/cm/topology_manager ?
+const (
+	PolicyNone           = "none"
+	PolicyBestEffort     = "best-effort"
+	PolicyRestricted     = "restricted"
+	PolicySingleNUMANode = "single-numa-node"
+)
```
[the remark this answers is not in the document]
ffromani: good idea, done!

pkg/noderesourcetopology/config.go
ffromani: Right. Added a code comment *AND* a warning

pkg/noderesourcetopology/README.md
ffromani: thanks, fixed

pkg/noderesourcetopology/README.md
ffromani: thanks, fixed

pkg/noderesourcetopology/objects.go
ffromani: I tend to agree, but let's indeed defer to a separate PR.

swatisehgal (approved):
/lgtm

/assign @Huang-Wei 
for approval

Huang-Wei:
/approve

Thanks @ffromani and @swatisehgal @Tal-or @PiotrProkop for the thorough review!

PR #491, opened 2023-02-01
Remove unnecessary functions and optimize judgments

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

#### What type of PR is this?

/kind cleanup

Fixes #

```release-note
NONE
```


pkg/networkaware/util/util.go
```diff
@@ -63,12 +63,11 @@ func GetNodeRegion(node *v1.Node) string {
...
 
-	zone, _ := labels[v1.LabelTopologyRegion]
-	if zone == "" {
-		return ""
+	if region, ok := labels[v1.LabelTopologyRegion]; ok {
+		return region
 	}
 
-	return zone
+	return ""
```
PiotrProkop, line 70: you can just:
```go
return labels[v1.LabelTopologyRegion]
```
if `v1.LabelTopologyRegion` is not in map it will return default value for string which is `""`

pkg/networkaware/util/util.go
```diff
@@ -78,12 +77,11 @@ func GetNodeZone(node *v1.Node) string {
 		return ""
 	}
 
-	region, _ := labels[v1.LabelTopologyZone]
-	if region == "" {
-		return ""
+	if zone, ok := labels[v1.LabelTopologyZone]; ok {
```
PiotrProkop, line 80: ditto

pkg/networkaware/util/util.go
Huang-Wei: +1

Fish-pro: Thanks, it has been modified

pkg/networkaware/util/util.go
Fish-pro: Thanks, it has been modified

PR #504, opened 2023-02-07
Add new NRT cache type DiscardReservedNodes

PiotrProkop:
#### What type of PR is this?
/kind feature

Add new NRT cache type DiscardReservedNodes, which discards nodes from being scheduled if there are any pods reserved by NRT plugin for it.

DiscardReserved is intended to solve similiar problem as Overreserve Cache,
which is to minimize amount of incorrect scheduling decisions based on stale NRT data.
Unfortunately Overreserve cache only works for single-numa-node Topology Manager policy.
Dis tries to minimize amount of Admission Errors and non-optimal placement
when NodeResourceTopologyMatch plugin is used to schedule PODs requesting resources from multiple NUMA domains.
There are scenarios where using DiscardReserved won't mitigate drawbacks of using Passthrough cache.
NRT update is expected once PostBind triggers, but there's no guarantee about when this will happen.
In cases like:
- NFD(or any other component that advertises NRT) can be nonfunctional
- network can be slow
- Pod being scheduled after PostBind trigger and before NRT update
in those cases DiscardReserved cache will act same as Passthrough cache.

Fixes #

```release-note
Add new DiscardReservedNodes cache type for NodeResourceTopologyPlugin.
```

### DIFF
--- a/apis/config/v1/types.go
+++ b/apis/config/v1/types.go
@@ -149,6 +149,10 @@ type NodeResourceTopologyMatchArgs struct {
 	ScoringStrategy *ScoringStrategy `json:"scoringStrategy,omitempty"`
 	// If > 0, enables the caching facilities of the reserve plugin - which must be enabled
 	CacheResyncPeriodSeconds *int64 `json:"cacheResyncPeriodSeconds,omitempty"`
+	// if set to true, exclude node from scheduling if there are any reserved pods for given node
+	// this option takes precedence over CacheResyncPeriodSeconds
+	// if DiscardReservedNodes is enabled, CacheResyncPeriodSeconds option is noop
+	DiscardReservedNodes bool `json:"discardReservedNodes,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/apis/config/v1beta2/types.go
+++ b/apis/config/v1beta2/types.go
@@ -148,6 +148,10 @@ type NodeResourceTopologyMatchArgs struct {
 	// implicitely enables the caching. If zero, disables the caching entirely.
 	// If the cache is enabled, the Reserve plugin must be enabled.
 	CacheResyncPeriodSeconds *int64 `json:"cacheResyncPeriodSeconds,omitempty"`
+	// if set to true, exclude node from scheduling if there are any reserved pods for given node
+	// this option takes precedence over CacheResyncPeriodSeconds
+	// if DiscardReservedNodes is enabled, CacheResyncPeriodSeconds option is noop
+	DiscardReservedNodes bool `json:"discardReservedNodes,omitempty"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/noderesourcetopology/cache/cache.go
+++ b/pkg/noderesourcetopology/cache/cache.go
@@ -56,4 +56,10 @@ type Interface interface {
 
 	// UnreserveNodeResources decrement from the node assumed resources the resources required by the given pod.
 	UnreserveNodeResources(nodeName string, pod *corev1.Pod)
+
+	// PostBind is called after a pod is successfully bound. These plugins are
+	// informational. A common application of this extension point is for cleaning
+	// up. If a plugin needs to clean-up its state after a pod is scheduled and
+	// bound, PostBind is the extension point that it should register.
+	PostBind(nodeName string, pod *corev1.Pod)
 }
--- a/pkg/noderesourcetopology/cache/discardreserved.go
+++ b/pkg/noderesourcetopology/cache/discardreserved.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+
+	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	listerv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/listers/topology/v1alpha2"
+)
+
+// DiscardReserved is intended to solve similiar problem as Overreserve Cache,
+// which is to minimize amount of incorrect scheduling decisions based on stale NRT data.
+// Unfortunately Overreserve cache only works for single-numa-node Topology Manager policy.
+// Dis tries to minimize amount of Admission Errors and non-optimal placement
+// when NodeResourceTopologyMatch plugin is used to schedule PODs requesting resources from multiple NUMA domains.
+// There are scenarios where using DiscardReserved won't mitigate drawbacks of using Passthrough cache.
+// NRT update is expected once PostBind triggers, but there's no guarantee about when this will happen.
+// In cases like:
+// - NFD(or any other component that advertises NRT) can be nonfunctional
+// - network can be slow
+// - Pod being scheduled after PostBind trigger and before NRT update
+// in those cases DiscardReserved cache will act same as Passthrough cache
+type DiscardReserved struct {
+	rMutex         sync.RWMutex
+	reservationMap map[string]map[types.UID]bool // Key is NodeName, value is Pod UID : reserved status
+	lister         listerv1alpha2.NodeResourceTopologyLister
+}
+
+func NewDiscardReserved(lister listerv1alpha2.NodeResourceTopologyLister) Interface {
+	return &DiscardReserved{
+		lister:         lister,
+		reservationMap: make(map[string]map[types.UID]bool),
+	}
+}
+
+func (pt *DiscardReserved) GetCachedNRTCopy(nodeName string, _ *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
+	pt.rMutex.RLock()
+	defer pt.rMutex.RUnlock()
+	if t, ok := pt.reservationMap[nodeName]; ok {
+		if len(t) > 0 {
+			return nil, false
+		}
+	}
+
+	nrt, err := pt.lister.Get(nodeName)
+	if err != nil {
+		return nil, false
+	}
+	return nrt, true
+}
+
+func (pt *DiscardReserved) NodeMaybeOverReserved(nodeName string, pod *corev1.Pod) {}
+func (pt *DiscardReserved) NodeHasForeignPods(nodeName string, pod *corev1.Pod)    {}
+
+func (pt *DiscardReserved) ReserveNodeResources(nodeName string, pod *corev1.Pod) {
+	klog.V(5).InfoS("nrtcache NRT Reserve", "logID", klog.KObj(pod), "UID", pod.GetUID(), "node", nodeName)
+	pt.rMutex.Lock()
+	defer pt.rMutex.Unlock()
+
+	if pt.reservationMap[nodeName] == nil {
+		pt.reservationMap[nodeName] = make(map[types.UID]bool)
+	}
+	pt.reservationMap[nodeName][pod.GetUID()] = true
+}
+
+func (pt *DiscardReserved) UnreserveNodeResources(nodeName string, pod *corev1.Pod) {
+	klog.V(5).InfoS("nrtcache NRT Unreserve", "logID", klog.KObj(pod), "UID", pod.GetUID(), "node", nodeName)
+
+	pt.removeReservationForNode(nodeName, pod)
+}
+
+// PostBind is invoked to cleanup reservationMap
+func (pt *DiscardReserved) PostBind(nodeName string, pod *corev1.Pod) {
+	klog.V(5).InfoS("nrtcache NRT PostBind", "logID", klog.KObj(pod), "UID", pod.GetUID(), "node", nodeName)
+
+	pt.removeReservationForNode(nodeName, pod)
+}
+
+func (pt *DiscardReserved) removeReservationForNode(nodeName string, pod *corev1.Pod) {
+	pt.rMutex.Lock()
+	defer pt.rMutex.Unlock()
+
+	delete(pt.reservationMap[nodeName], pod.GetUID())
+}
--- a/pkg/noderesourcetopology/cache/discardreserved_test.go
+++ b/pkg/noderesourcetopology/cache/discardreserved_test.go
@@ -1,0 +1,172 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cache
+
+import (
+	"reflect"
+	"testing"
+
+	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	faketopologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned/fake"
+	topologyinformers "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/informers/externalversions"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestDiscardReservedNodesGetNRTCopy(t *testing.T) {
+	fakeClient := faketopologyv1alpha2.NewSimpleClientset()
+	fakeInformer := topologyinformers.NewSharedInformerFactory(fakeClient, 0).Topology().V1alpha2().NodeResourceTopologies()
+
+	nrtCache := NewDiscardReserved(fakeInformer.Lister())
+	var nrtObj *topologyv1alpha2.NodeResourceTopology
+	nrtObj, _ = nrtCache.GetCachedNRTCopy("node1", &corev1.Pod{})
+	if nrtObj != nil {
+		t.Fatalf("non-empty object from empty cache")
+	}
+
+	nodeTopologies := []*topologyv1alpha2.NodeResourceTopology{
+		{
+			ObjectMeta:       metav1.ObjectMeta{Name: "node1"},
+			TopologyPolicies: []string{string(topologyv1alpha2.SingleNUMANodeContainerLevel)},
+			Zones: topologyv1alpha2.ZoneList{
+				{
+					Name: "node-0",
+					Type: "Node",
+					Resources: topologyv1alpha2.ResourceInfoList{
+						MakeTopologyResInfo(cpu, "20", "4"),
+						MakeTopologyResInfo(memory, "8Gi", "8Gi"),
+						MakeTopologyResInfo(nicResourceName, "30", "10"),
+					},
+				},
+				{
+					Name: "node-1",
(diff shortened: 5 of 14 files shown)


PiotrProkop:
@ffromani @swatisehgal please review.

ffromani:
/ok-to-test
I had offline chat with @PiotrProkop and I'm very in favor of this general direction and open to this kind of additions. However I need to deep dive into the design and the code to see how we can clearly fit this code in the current framework, and unfortunately I won't have the time for that in the next few days.

ffromani:
@PiotrProkop just FYI additionally, I would finish https://github.com/kubernetes-sigs/scheduler-plugins/pull/488 before to tackle this one

PiotrProkop:
> @PiotrProkop just FYI additionally, I would finish #488 before to tackle this one

ok, let me know when #488 will be ready to review.

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -70,27 +70,27 @@ func NewOverReserve(lister listerv1alpha1.NodeResourceTopologyLister, indexer No
...
 	defer ov.lock.Unlock()
 	if ov.nodesWithForeignPods.IsSet(nodeName) {
-		return nil, false
+		return nil, fmt.Errorf("foreign pods scheduled on node: %s", nodeName)
 	}
 
 	nrt := ov.nrts.GetNRTCopyByNodeName(nodeName)
 	if nrt == nil {
-		return nil, true
+		return nil, fmt.Errorf("cannot get NodeResourceTopology for node: %s", nodeName)
```
ffromani, line 82: This is a trap in the existing code and would have deserved a comment. For historical reasons, NOT having NRT info is a legit case which is meant to be handled as passthrough (no filter). We can revisit this behavior of course, but in another PR :)

ffromani (commented):
still need to find time to properly review, but something caught my eye already

ffromani:
@PiotrProkop could you please rebase? I'd be happy to review the week starting on March 6 2023

PiotrProkop:
> @PiotrProkop could you please rebase? I'd be happy to review the week starting on March 6 2023

I am on it!

PiotrProkop:
> > @PiotrProkop could you please rebase? I'd be happy to review the week starting on March 6 2023
> 
> I am on it!

@ffromani  Should be good to review now!

ffromani:
> > > @PiotrProkop could you please rebase? I'd be happy to review the week starting on March 6 2023
> > 
> > 
> > I am on it!
> 
> @ffromani Should be good to review now!

Thanks! I'll start my review later today (March 6)

ffromani:
Sorry for the delay. Though week. Reviewing this PR is still on the top of my TODO.

pkg/noderesourcetopology/cache/cache.go
```diff
@@ -30,7 +30,7 @@ type Interface interface {
 	// The pod argument is used only for logging purposes.
 	// Returns a boolean to signal the caller if the NRT data is clean. If false, then the node has foreign
 	// Pods detected - so it should be ignored or handled differently by the caller.
-	GetCachedNRTCopy(nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool)
+	GetCachedNRTCopy(nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, error)
```
ffromani, line 33: this is probably a good idea in general. Ideally, this would be its own commit.

pkg/noderesourcetopology/cache/cache.go
```diff
@@ -56,4 +56,10 @@ type Interface interface {
 
 	// UnreserveNodeResources decrement from the node assumed resources the resources required by the given pod.
 	UnreserveNodeResources(nodeName string, pod *corev1.Pod)
+
+	// PostBind is called after a pod is successfully bound. These plugins are
+	// informational. A common application of this extension point is for cleaning
+	// up. If a plugin needs to clean-up its state after a pod is scheduled and
+	// bound, PostBind is the extension point that it should register.
+	PostBind(nodeName string, pod *corev1.Pod)
```
ffromani, line 64: interesting. I'll review if this can helps the other cache types.

pkg/noderesourcetopology/cache/overreserve.go
ffromani: I suggest to turn the error into a log and `return nil, nil` here

pkg/noderesourcetopology/cache/passthrough.go
```diff
@@ -34,17 +36,17 @@ func NewPassthrough(lister listerv1alpha2.NodeResourceTopologyLister) Interface
...
 }
 
-func (pt Passthrough) GetCachedNRTCopy(nodeName string, _ *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
+func (pt Passthrough) GetCachedNRTCopy(nodeName string, _ *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, error) {
 	klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", pt.lister)
 	nrt, err := pt.lister.Get(nodeName)
 	if err != nil {
-		klog.V(5).ErrorS(err, "Cannot get NodeTopologies from NodeResourceTopologyLister")
-		return nil, true
+		return nil, fmt.Errorf("cannot get NodeTopologies from NodeResourceTopologyLister")
```
ffromani, line 43: this is another trap, the original code didn't fail. We need to check what happens on the call site with this change, which looks indeed more correct.

pkg/noderesourcetopology/filter.go
```diff
@@ -198,12 +198,10 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
...
 	nodeName := nodeInfo.Node().Name
-	nodeTopology, ok := tm.nrtCache.GetCachedNRTCopy(nodeName, pod)
-	if !ok {
-		return framework.NewStatus(framework.Unschedulable, fmt.Sprintf("invalid node topology data for node %s", nodeName))
-	}
-	if nodeTopology == nil {
-		return nil
+	nodeTopology, err := tm.nrtCache.GetCachedNRTCopy(nodeName, pod)
+	if err != nil {
+		klog.Error(err)
```
ffromani, line 203: maybe `klog.ErrorS(err, ...)` and keep the original log message?

pkg/noderesourcetopology/filter.go
```diff
@@ -198,12 +198,10 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
...
+		return framework.NewStatus(framework.Unschedulable, err.Error())
```
ffromani, line 204: why the other check was removed?

pkg/noderesourcetopology/score.go
```diff
@@ -63,15 +63,10 @@ func (tm *TopologyMatch) Score(ctx context.Context, state *framework.CycleState,
...
-	if !ok {
+	nodeTopology, err := tm.nrtCache.GetCachedNRTCopy(nodeName, pod)
+	if err != nil {
 		klog.V(4).InfoS("noderesourcetopology is not valid for node", "node", nodeName)
-		return 0, nil
-	}
-	if nodeTopology == nil {
-		klog.V(5).InfoS("noderesourcetopology was not found for node", "node", nodeName)
-		return 0, nil
+		return framework.MinNodeScore, nil
```
ffromani, line 69: again not sure (literally, need to check) that removing the `if nodeTopology == nil` guard has no ill effects in the flow.

apis/config/v1/types.go
```diff
@@ -149,6 +149,8 @@ type NodeResourceTopologyMatchArgs struct {
 	ScoringStrategy *ScoringStrategy `json:"scoringStrategy,omitempty"`
 	// If > 0, enables the caching facilities of the reserve plugin - which must be enabled
 	CacheResyncPeriodSeconds *int64 `json:"cacheResyncPeriodSeconds,omitempty"`
+	// if set to true, exclude node from scheduling if there are any reserved pods for given node
+	DiscardReservedNodes bool
```
ffromani, line 153: missing JSON annotations. Please make sure the field is fully optional. Do you need to distinguish between "unspecified" and "zero"?

ffromani (commented):
initial partial review

pkg/noderesourcetopology/cache/passthrough.go
PiotrProkop: this is what it looks like in lister:
```
	obj, exists, err := s.indexer.GetByKey(name)
	if err != nil {
		return nil, err
	}
	if !exists {
		return nil, errors.NewNotFound(v1alpha2.Resource("noderesourcetopology"), name)
	}
```
So if NRT doesn't exist we'll get Not Found error.

pkg/noderesourcetopology/filter.go
PiotrProkop: In my first implementation I removed possibility of returning `nil,nil` from GetCachedNRTCopy, with next patchset I am going to revert it so will add this check again:
```go
if nodeTopology == nil {
	return nil
}
```

pkg/noderesourcetopology/filter.go
PiotrProkop: :+1:

pkg/noderesourcetopology/score.go
PiotrProkop: added this check again

apis/config/v1/types.go
PiotrProkop: I don't need to use `*bool` I can rely on default value of boolean which is `false` 😄

PiotrProkop:
> initial partial review

thanks for the review. I have applied changes from this review.

apis/config/v1/types.go
ffromani: fair enough!

ffromani:
For the record, overall from what I see so far I'm still very much convinced this change is a good direction for the NRT scheduler plugin and that the only issues could be related to properly integrate in the codebase - not about the design, which seems to fit nicely in the current framework and goals. There are already few points that could be generalized to other cache impls to make the codebase better.

pkg/noderesourcetopology/cache/discardreserved.go
```diff
@@ -0,0 +1,98 @@
...
+	//readLock
+	pt.rMutex.RLock()
+	defer pt.rMutex.RUnlock()
+	if t, ok := pt.reservationMap[nodeName]; ok {
+		if len(t) > 0 {
+			return nil, fmt.Errorf("node %s is already reserved by another pod", nodeName)
+		}
+	}
+
+	klog.V(5).InfoS("Lister for nodeResTopoPlugin", "lister", pt.lister)
```
ffromani, line 53: minor: logging the lister is rarely useful in my experience. Perhaps you may want to log some properties of the lister, not the full object?

pkg/noderesourcetopology/cache/discardreserved.go
```diff
@@ -0,0 +1,98 @@
...
+func (pt *DiscardReserved) ReserveNodeResources(nodeName string, pod *corev1.Pod) {
+	klog.V(5).InfoS("nrtcache NRT Reserve", "logID", klog.KObj(pod), "node", nodeName)
+	// write lock
+	pt.rMutex.Lock()
+	defer pt.rMutex.Unlock()
+
+	if pt.reservationMap[nodeName] == nil {
+		pt.reservationMap[nodeName] = make(map[string]bool)
+	}
+	podName := klog.KObj(pod).String()
```
ffromani, line 73: does this guarantee the name is stable and namespace prefixed?

pkg/noderesourcetopology/cache/discardreserved.go
```diff
@@ -0,0 +1,98 @@
...
+	pt.reservationMap[nodeName][podName] = true
```
ffromani, line 74: a design decision we inherited by the podresources api is to use the namespace/name pair to identify the pods.
Changing the podresources API would have been too hard, so this decision leaked all across the stack.
But pod.UID is a stronger identifier and in some case we would have preferred to use it.
So, is the namespace/name pair good enough for you? would using pod.UID lead to more robust/simpler code?

pkg/noderesourcetopology/cache/discardreserved.go
```diff
@@ -0,0 +1,98 @@
...
+
+	klog.V(5).InfoS("nrtcache NRT", "logID", klog.KObj(pod), "node", nodeName, "reservationMap", pt.reservationMap[nodeName])
```
ffromani, line 76: this easily gets big (thus poorly readable) when the nodes get populated, though. It's fine to log this only at high verbosiness level, but I'm wondering if we can log somehing more useful

pkg/noderesourcetopology/cache/discardreserved.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (pt *DiscardReserved) UnreserveNodeResources(nodeName string, pod *corev1.Pod) {
+	klog.V(5).InfoS("nrtcache NRT Unreserve", "logID", klog.KObj(pod), "node", nodeName)
+	pt.removeReservationForNode(nodeName, pod)
+}
+
+// PostBind is invoked to cleanup reservationMap
+func (pt *DiscardReserved) PostBind(nodeName string, pod *corev1.Pod) {
```
ffromani, line 85: this is interesting. I think I get it but could you please elaborate (gh comment is fine) why you want to remove the reservation on postbind?

ffromani (commented):
almost done, need to finish looking at the tests and at the core changes

pkg/noderesourcetopology/cache/passthrough.go
ffromani: Uhm, I still tend to believe we DO NOT want to error out here, I recall some past conversations in this area. I'll do dome archeology.

pkg/noderesourcetopology/score.go
ffromani: thanks. Could you please add back the log as well?

pkg/noderesourcetopology/cache/discardreserved.go
PiotrProkop: You're right! I need to work on logging part of this PR.

pkg/noderesourcetopology/cache/discardreserved.go
PiotrProkop: Haven't though about it, probably should use pod.UID.

pkg/noderesourcetopology/cache/discardreserved.go
PiotrProkop: pod.UID FTW!

ffromani (commented):
Besides a couple items in the initial `GetCachedNRTCopy` PR, the prep work LGTM. I'll finish the review of the core changes later.

pkg/noderesourcetopology/score.go
PiotrProkop: :+1:

pkg/noderesourcetopology/cache/passthrough.go
ffromani: let's hold on this and the other item regarding GetCachedNRTCopy. I'll do some research and come up with the suggestion. I'm confident postponing these changes should be little hassle.

PiotrProkop:
> Besides a couple items in the initial `GetCachedNRTCopy` PR, the prep work LGTM. I'll finish the review of the core changes later.

Thanks a lot! I'll update PR till EOW.

pkg/noderesourcetopology/cache/discardreserved.go
PiotrProkop: After checking I used this function cause it prefixes pod with namespace but I like using `UID` better, so I am changing implementation to use it right now!

PiotrProkop:
> > Besides a couple items in the initial `GetCachedNRTCopy` PR, the prep work LGTM. I'll finish the review of the core changes later.
> 
> Thanks a lot! I'll update PR till EOW.

PR updated. Decided to go with `pod.UID` as `reservationMap` key. I have also dropped the commit that changed the signature of `GetCachedNRTCopy`.

apis/config/v1beta2/types.go
```diff
@@ -148,6 +148,8 @@ type NodeResourceTopologyMatchArgs struct {
 	// implicitely enables the caching. If zero, disables the caching entirely.
 	// If the cache is enabled, the Reserve plugin must be enabled.
 	CacheResyncPeriodSeconds *int64 `json:"cacheResyncPeriodSeconds,omitempty"`
+	// if set to true, exclude node from scheduling if there are any reserved pods for given node
+	DiscardReservedNodes bool `json:"discardReservedNodes,omitempty"`
```
ffromani, line 152: my only remaining concern is this effectively makes the code ignore `CacheResyncPeriodSeconds` (if given).
I'd start documenting this. We can think about a better UX later on.

pkg/noderesourcetopology/cache/discardreserved.go
```diff
@@ -0,0 +1,91 @@
...
+type DiscardReserved struct {
+	rMutex         sync.RWMutex
+	reservationMap map[string]map[types.UID]bool // Key is NodeName, value is Pod UID : reserved status
+	lister         listerv1alpha2.NodeResourceTopologyLister
+}
+
+func NewDiscardReserved(lister listerv1alpha2.NodeResourceTopologyLister) Interface {
+	return &DiscardReserved{
+		lister:         lister,
+		reservationMap: make(map[string]map[types.UID]bool),
```
ffromani, line 39: minor: I wonder if we should use a set here (k8s.io/apimachinery/pkg/util/sets)

pkg/noderesourcetopology/cache/discardreserved.go
ffromani: note really looks very similar (possibly the same) reason we introduced the resync mechanism (using podfingerprint). You can't really be sure about timings here. The NRT update can come anytime. They will come in order, sure, but there's no guarantee about timing and about the sequence of events considering the scheduler event and the NRT update events.

ffromani (commented):
thanks for the updates. Codewise LGTM with a note about the config, please check.
I'm OK adding this code to `noderesourcetopology` codebase, it's self-contained and I'm happy to assist your use case, but looks to me you folks are hitting very similar issues which led to the design of the `overreserve` cache, so maybe worth a deeper dive here?

pkg/noderesourcetopology/cache/discardreserved.go
ffromani: Unless I'm missing something, you are totally correct that a NRT update is expected once `PostBind` triggers, but there's no guarantee about WHEN this will happen. NFD can be crashlooping, network can be slow, we can have any network glitch which prevents the NRT update to be propagated [temporary glitch]. But most notably there's no guarantee with respect the occurrence of new pods to be scheduled.

In yet another different way (but still related): what ensures that a NRT update is received before a new scheduling attempt is performed against this node when a new pod comes (new pod scheduling is of course completely async).

In the case of the overreserve, we have the podfingerprint machinery and we intentionally are using a pessimistic approach.

PiotrProkop: You're right but `overreserve` doesn't cache anything if a POD requests resources from multiple NUMA domains(thanks to pesimistic approach and subtracting from every NUMA domain). And it will fetch data from NRT anyway (and temporary glitches also affects this 😄 ). The idea behind Locking nodes is to minimize Admission Errors not to eliminate them.

ffromani:
(copypasting here to ensure visibility)

We covered all the facets I can think of. I'll do a last pass for due diligence to integration tests, but we are almost there. We need two things more, both should be quick and easy:

-  please document as comment in the config that enabling DiscardReservedNodes takes precedence over CacheResyncPeriodSeconds. A code comment is fine.
-  please describe (for historical record) the usecase and the design/conceptual differences between this cache impl and overreserve in the PR cover letter and/or in `cache/discardreserved.go` (preferably both, just copypaste the same block of text). 

I think we already discussed the rationale for this cache impl, it would be fine to just summarize our past conversation and copy/paste it as comment so we have the record for the future in a easy to find place.

PiotrProkop:
> (copypasting here to ensure visibility)
> 
> We covered all the facets I can think of. I'll do a last pass for due diligence to integration tests, but we are almost there. We need two things more, both should be quick and easy:
> 
>     * please document as comment in the config that enabling DiscardReservedNodes takes precedence over CacheResyncPeriodSeconds. A code comment is fine.
> 
>     * please describe (for historical record) the usecase and the design/conceptual differences between this cache impl and overreserve in the PR cover letter and/or in `cache/discardreserved.go` (preferably both, just copypaste the same block of text).
> 
> 
> I think we already discussed the rationale for this cache impl, it would be fine to just summarize our past conversation and copy/paste it as comment so we have the record for the future in a easy to find place.

Thanks for the review!

I have added comments in the code and updated PR description to explain rationale.

pkg/noderesourcetopology/cache/discardreserved_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	nrtCache.ReserveNodeResources("node1", &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "pod",
+			Namespace: "test",
+			UID:       "some-uid",
+		},
+	})
+	nodePods, ok := nrtCache.reservationMap["node1"]
```
ffromani, line 112: let's improve the tests later on
1. adding (more) integration tests
2. avoiding to access private fields/methods

ffromani (commented):
/approve
/lgtm

self-contained change. Useful to explore new directions which may benefit `overreserve in the future, so win-win

PR #505, opened 2023-02-07
cleanup: remove ioutil for new go version

kinzhi:
What type of PR is this?
/kind cleanup

What this PR does / why we need it:
The io/ioutil package has been deprated in go 1.16: https://go.dev/doc/go1.16#ioutil
Mybe we should remove it.

Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:
Does this PR introduce a user-facing change?


Huang-Wei:
Thanks for the PR, but we don't modify the content in `vendor/`. You can raise PR at https://github.com/fsnotify/fsnotify, and vendor here using an updated version.

/close

PR #509, opened 2023-02-12
CapacityScheduling: fix usedOverMinWith/usedOverMaxWith/usedOverMin func and add unit test

### LINKED ISSUE #508: [CapacityScheduling] If ElasticQuota doesn't have Max / Min, usedOverMaxWith / usedOverMinWith return panic
Hello,

Thanks for maintaining this amazing repository.
I found a bug related to ElasticQuota.

When ElasticQuota's Max or Min isn't set up, usedOverMaxWith or usedOverMinWith returns panic.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/capacityscheduling/elasticquota.go#L94-L96

```
func (e *ElasticQuotaInfo) usedOverMaxWith(podRequest *framework.Resource) bool {
	return cmp2(podRequest, e.Used, e.Max)
}
```

I guess this is unintended behavior.
Can I fix this bug?

Gekko0114:
#### What type of PR is this?
/kind bug

When ElasticQuota's Max or Min isn't set up, usedOverMaxWith or usedOverMinWith returns panic.
So I fixed this bug and added unit tests.

Fixes #508 

Thank you very much for creating and maintaining this amazing repository.
This is the first time I created PR for this repo.
If I do wrong things, please let me know.

```release-note
NONE
```

### DIFF
--- a/pkg/capacityscheduling/elasticquota.go
+++ b/pkg/capacityscheduling/elasticquota.go
@@ -17,6 +17,8 @@ limitations under the License.
 package capacityscheduling
 
 import (
+	"reflect"
+
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
@@ -88,15 +90,30 @@ func (e *ElasticQuotaInfo) unreserveResource(request framework.Resource) {
 }
 
 func (e *ElasticQuotaInfo) usedOverMinWith(podRequest *framework.Resource) bool {
-	return cmp2(podRequest, e.Used, e.Min)
+	eqMin := reflect.ValueOf(e).Elem().FieldByName("Min")
+	if eqMin.IsValid() && !eqMin.IsZero() {
+		return cmp2(podRequest, e.Used, e.Min)
+	}
+	// "ElasticQuotaInfo doesn't have Min" means used values exceeded min(0)
+	return true
 }
 
 func (e *ElasticQuotaInfo) usedOverMaxWith(podRequest *framework.Resource) bool {
-	return cmp2(podRequest, e.Used, e.Max)
+	eqMax := reflect.ValueOf(e).Elem().FieldByName("Max")
+	if eqMax.IsValid() && !eqMax.IsZero() {
+		return cmp2(podRequest, e.Used, e.Max)
+	}
+	// "ElasticQuotaInfo doesn't have Max" means there are no limitations(infinite)
+	return false
 }
 
 func (e *ElasticQuotaInfo) usedOverMin() bool {
-	return cmp(e.Used, e.Min)
+	eqMin := reflect.ValueOf(e).Elem().FieldByName("Min")
+	if eqMin.IsValid() && !eqMin.IsZero() {
+		return cmp(e.Used, e.Min)
+	}
+	// "ElasticQuotaInfo doesn't have Min" means used values exceeded min(0)
+	return true
 }
 
 func (e *ElasticQuotaInfo) clone() *ElasticQuotaInfo {
--- a/pkg/capacityscheduling/elasticquota.go
+++ b/pkg/capacityscheduling/elasticquota.go
@@ -17,8 +17,6 @@ limitations under the License.
 package capacityscheduling
 
 import (
-	"reflect"
-
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


pkg/capacityscheduling/elasticquota.go
```diff
@@ -88,15 +90,30 @@ func (e *ElasticQuotaInfo) unreserveResource(request framework.Resource) {
 }
 
 func (e *ElasticQuotaInfo) usedOverMinWith(podRequest *framework.Resource) bool {
-	return cmp2(podRequest, e.Used, e.Min)
+	eqMin := reflect.ValueOf(e).Elem().FieldByName("Min")
+	if eqMin.IsValid() && !eqMin.IsZero() {
+		return cmp2(podRequest, e.Used, e.Min)
+	}
+	// "ElasticQuotaInfo doesn't have Min" means used values exceeded min(0)
+	return true
```
PiotrProkop, line 98: Will checking if `e.Min == nil` works too, without using `reflect`? something like:
```go
if e.Min == nil {
    return true
}
return cmp2(podRequest, e.Used, e.Min)
```

Gekko0114: Thanks for your suggestion. Fixed it.

pkg/capacityscheduling/elasticquota.go
```diff
@@ -88,14 +88,26 @@ func (e *ElasticQuotaInfo) unreserveResource(request framework.Resource) {
...
+	// "ElasticQuotaInfo doesn't have Max" means there are no limitations(infinite)
+	if e.Max == nil {
+		return false
+	}
 	return cmp2(podRequest, e.Used, e.Max)
 }
 
 func (e *ElasticQuotaInfo) usedOverMin() bool {
+	// "ElasticQuotaInfo doesn't have Min" means used values exceeded min(0)
+	if e.Min == nil {
```
PiotrProkop, line 108: I am not very familliar with this plugin but is it possible that `e.Used` will also be `nil`?

pkg/capacityscheduling/elasticquota.go
```diff
@@ -88,14 +88,26 @@ func (e *ElasticQuotaInfo) unreserveResource(request framework.Resource) {
 }
 
 func (e *ElasticQuotaInfo) usedOverMinWith(podRequest *framework.Resource) bool {
+	// "ElasticQuotaInfo doesn't have Min" means used values exceeded min(0)
+	if e.Min == nil {
```
Huang-Wei, line 92: if Min/Max/Used can be nil, maybe we should use a concrete type instead of pointer in ElasticQuotaInfo, to get rid of nil check burden.

@denkensk was there any particular reason making them pointers in the beginning?

denkensk: It's been so long that I can't remember exactly why. 

Maybe We will often call methods in this package https://github.com/kubernetes/kubernetes/blob/5d9d06aa4749ef9ad73f94fef0ef2e56719f9190/pkg/scheduler/framework/types.go#L432; And need to pass a pointer of the struct as an argument to function argument. So I store these as pointers for convenience.

Gekko0114: Thanks for your comments!

> if Min/Max/Used can be nil, maybe we should use a concrete type instead of pointer in ElasticQuotaInfo, to get rid of nil check burden.

@Huang-Wei 
Do you mean like this?
```
type ElasticQuotaInfo struct {
	Namespace string
	pods      sets.String
	Min       framework.Resource
	Max       framework.Resource
	Used      framework.Resource
}

```

In my opinion, Min / Max / Used should be used as pointer.
Firstly, as @denkensk said, it seems kubernetes's framework package recommends us to use pointer.
The return value of NewResource is pointer as well.
https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/types.go#L444-L448

Secondly, if we change Min/Max/Used as concrete type, we would have to fix many parts in capacityScheduling.
This bug is so small, so we don't need to change many parts, just only for this reason.
This problem should be fixed in each useOver functions.

From these reasons, I think Min / Max / Used should be used as pointer.
But you two are so familiar with scheduler-plugin, so any advice would be appreciated.
Thanks,

Huang-Wei: > This bug is so small, so we don't need to change many parts, just only for this reason.

Makes sense to me.

Huang-Wei:
/lgtm
/approve

Thanks @Gekko0114 !

PR #512, opened 2023-02-13
Introduce pluginargs validation for NodeResourceTopologyMatch plugin

### LINKED ISSUE #507: noderesourcetopology plugin: implement proper plugin args validation logic
TL;DR: address https://github.com/kubernetes-sigs/scheduler-plugins/pull/474#discussion_r1086235462

This plugin should hook in the existing plugin argument validation logic, removing code from the plugin application code itself.

PiotrProkop:
#### What type of PR is this?
/kind feature

Introduce pluginargs validation for NodeResourceTopologyMatch plugin.

Fixes #507 

```release-note
NONE
```

### DIFF
--- a/apis/config/validation/validation_plugingargs_test.go
+++ b/apis/config/validation/validation_plugingargs_test.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validation
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"sigs.k8s.io/scheduler-plugins/apis/config"
+)
+
+func TestValidateNodeResourceTopologyMatchArgs(t *testing.T) {
+	testCases := []struct {
+		args        *config.NodeResourceTopologyMatchArgs
+		expectedErr error
+		description string
+	}{
+		{
+			description: "correct config",
+			args: &config.NodeResourceTopologyMatchArgs{
+				ScoringStrategy: config.ScoringStrategy{
+					Type: config.MostAllocated,
+				},
+			},
+		},
+		{
+			description: "incorrect config, wrong ScoringStrategy type",
+			args: &config.NodeResourceTopologyMatchArgs{
+				ScoringStrategy: config.ScoringStrategy{
+					Type: "not existent",
+				},
+			},
+			expectedErr: fmt.Errorf("scoringStrategy.type: Invalid value:"),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			err := ValidateNodeResourceTopologyMatchArgs(nil, testCase.args)
+			if testCase.expectedErr != nil {
+				if err == nil {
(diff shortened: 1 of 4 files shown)


PiotrProkop:
@ffromani please take a look.

pkg/noderesourcetopology/plugin.go
```diff
@@ -120,6 +121,10 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 		return nil, fmt.Errorf("want args to be of type NodeResourceTopologyMatchArgs, got %T", args)
 	}
 
+	if err := validation.ValidateNodeResourceTopologyMatchArgs(nil, tcfg); err != nil {
```
ffromani, line 124: passing `nil` here is suspiscious, is this correct? if so worth 1-line comment to remind/explain why

ffromani (commented):
thanks looks nice, just one question inside.
Is this PR meant to set the foundation and expand later/in different PRs or do you plan to move/add more validations? I'm fine with both approaches.

pkg/noderesourcetopology/plugin.go
PiotrProkop: This is on purpose, I have based my validation work on how in-tree plugins do it 😄 example https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/plugins/defaultpreemption/default_preemption.go#L71

ffromani: nice, TIL. I'll have another, deeper, last pass later as my final review.

PiotrProkop:
My plan was to set foundation and add validation for `NodeResourceTopologyMatch` plugin in this PR. Validation for other plugins can/will be made in different PRs.

pkg/noderesourcetopology/plugin.go
```diff
@@ -120,6 +121,10 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
...
+		return nil, err
+	}
+
```
PiotrProkop, line 127: @ffromani  I can also refactor `getScoringStrategyFunction` not to return `error` but I think it's not worth it cause this PR https://github.com/kubernetes-sigs/scheduler-plugins/pull/488 is changing the way how we want to handle handler initialization.

ffromani:
> My plan was to set foundation and add validation for `NodeResourceTopologyMatch` plugin in this PR. Validation for other plugins can/will be made in different PRs.

Fine, and probably the best approach thinking about it.

pkg/noderesourcetopology/plugin.go
ffromani: Agreed, let's defer to a future PR.

pkg/noderesourcetopology/plugin.go
ffromani: Note to self: passing `nil` here is handled by `k/k/apimachinery/pkg/util/validation/field` Path objects and means "root" of the path to explore. A bit unintuitive maybe, but totally supported and idiomatic indeed.

apis/config/validation/validation_plugingargs_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			err := ValidateNodeResourceTopologyMatchArgs(nil, testCase.args)
+			if testCase.expectedErr != nil {
+				if err == nil {
+					t.Errorf("expected err to equal %v not nil", testCase.expectedErr)
+				}
+
+				if !strings.Contains(err.Error(), testCase.expectedErr.Error()) {
```
ffromani, line 60: minor: it would be nice to see if we can use `error.Is` but this is likely require some rework. Let's maybe add as (mental) note for future improvements.

ffromani (commented):
/lgtm
thanks @PiotrProkop

ffromani:
/assign @Huang-Wei 
for approval review

Huang-Wei:
(Sorry for the delay.)

/approve

PR #513, opened 2023-02-13
build: bump go to 1.20

zwpaper:

I met some build errors when I build images locally in arm64, it can be fixed by upgrading go.

also, go 1.20 has released and k/k used go 1.20 to build, https://github.com/kubernetes/kubernetes/blob/master/.go-version,

It may be time to bump go to 1.20

Fixes #

```release-note
NONE
```

### DIFF
--- a/build/controller/Dockerfile
+++ b/build/controller/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ARG ARCH
-FROM golang:1.17.11
+FROM golang:1.20.0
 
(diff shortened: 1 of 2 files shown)


build/controller/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ARG ARCH
-FROM golang:1.17.11
```
Huang-Wei, line 15: Can you bump it to 1.19? as we should make it consistent with:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/b86321ab439e564cd0ff53f7b90393c587c75475/go.mod#L3

PR #514, opened 2023-02-13
build: add script to build images

zwpaper:

scheduler plugins already have `release-image.arm64v8` and `build-controller/scheduler.arm64v8` but does not have a `local-image.arm64v8`, this PR add one so that arm64 mac/linux users can use it directly

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,6 @@ COMMONENVVAR=GOOS=$(shell uname -s | tr A-Z a-z)
 BUILDENVVAR=CGO_ENABLED=0
 INTEGTESTENVVAR=SCHED_PLUGINS_TEST_VERBOSE=1
 
-LOCAL_REGISTRY=localhost:5000/scheduler-plugins
-LOCAL_IMAGE=kube-scheduler:latest
-LOCAL_CONTROLLER_IMAGE=controller:latest
-
 # RELEASE_REGISTRY is the container registry to push
 # into. The default is to push to the staging
 # registry, not production(k8s.gcr.io).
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -77,6 +77,11 @@ local-image: clean
 	docker build -f ./build/scheduler/Dockerfile --build-arg ARCH="amd64" --build-arg RELEASE_VERSION="$(RELEASE_VERSION)" -t $(LOCAL_REGISTRY)/$(LOCAL_IMAGE) .
 	docker build -f ./build/controller/Dockerfile --build-arg ARCH="amd64" -t $(LOCAL_REGISTRY)/$(LOCAL_CONTROLLER_IMAGE) .
 
+.PHONY: local-image.arm64v8
+local-image.arm64v8: clean
+	docker build -f ./build/scheduler/Dockerfile --build-arg ARCH="arm64v8" --build-arg RELEASE_VERSION="$(RELEASE_VERSION)" -t $(LOCAL_REGISTRY)/$(LOCAL_IMAGE) .
```
Huang-Wei, line 82: For local build, it may make more sense to detect the local GOARCH directly, and then use the single local-image to build it.

Could you incorporate logic of detecting GOARCH to the local-image Makefile entry? `go env GOARCH` should be helpful.

zwpaper:
@Huang-Wei updated PR, there is too much magic in Makefile, I follow the other commands to create a `hack/build-images.sh` building the local image depends on `GOARCH`, and also used in release image commands.

PTAL

Huang-Wei (commented):
Thanks @zwpaper for the diligent work! Really appreciate it.

/approve
/lgtm

Issue #420, opened 2022-08-26
co-scheduler does not get installed in custom namespace


asm582:
When user tries to install helm in custom namespace other than `default` the installation fails with below error:

```
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: Namespace "scheduler-plugins" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "scheduler-plugins": current value is "default"
```

zwpaper:
fixed by PR specified above

/close

PR #519, opened 2023-02-18
chart: use release.namespace to specify target ns

zwpaper:
currently, we let users set the scheduler and controller namespace respectively in `values.yaml`, and only create the controller's namespace in templates.

but helm can manager namespace by `--create-namespace` and `--namespace` args, 
we should leverage helm's namespace management ability.

this PR:
1. drop the controller namespace in templates
2. let helm to manager the target namespace by using `.Release.Namespace`
3. use the same namespace for the controller and secondary scheduler

Fixes # https://github.com/kubernetes-sigs/scheduler-plugins/issues/420

```release-note
Support custom namespace installation in Helm Chart
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: scheduler-config
-  namespace: {{ .Values.scheduler.namespace }}
+  namespace: {{ .Release.Namespace }}
 data:
   scheduler-config.yaml: |
     apiVersion: kubescheduler.config.k8s.io/v1beta3
@@ -29,4 +29,3 @@ data:
 
   {{- /* TODO: wire CRD installation with enabled plugins. */}}
 {{- end }}
-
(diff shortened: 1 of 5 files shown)


Huang-Wei:
I verified in a kind 1.25 env:

```
⇒  helm install scheduler-plugins manifests/install/charts/as-a-second-scheduler/ --create-namespace --namespace foo
⇒  k get po -n foo
NAME                                            READY   STATUS    RESTARTS   AGE
scheduler-plugins-controller-58987f447f-jshdl   1/1     Running   0          39s
scheduler-plugins-scheduler-5756bc7f87-lzdj9    1/1     Running   0          39s
⇒  k get crd
NAME                                              CREATED AT
appgroups.appgroup.diktyo.k8s.io                  2023-03-03T01:21:25Z
elasticquotas.scheduling.sigs.k8s.io              2023-03-03T01:21:25Z
networktopologies.networktopology.diktyo.k8s.io   2023-03-03T01:21:25Z
noderesourcetopologies.topology.node.k8s.io       2023-03-03T01:21:25Z
podgroups.scheduling.sigs.k8s.io                  2023-03-03T01:21:25Z
```

But I need to submit a PR to correct the CRD's API group (migrate to x-k8s.io)

czybjtu:
Do we need to add some note in `https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/manifests/install/charts/as-a-second-scheduler/README.md#installing-the-chart` ?

zwpaper:
Hi @czybjtu, yes, we need the doc to be updated aligning with the new installation mechanism, it is appreciated that you found it, could you like to raise a PR for the update?

PR #520, opened 2023-02-19
CapacityScheduling: Fix default values for Max/Min in ElasticQuota

### LINKED ISSUE #424: [CapacityScheduling] ElasticQuota default value for "max" is zero
Hello everyone,

When creating an `ElasticQuota` resource, if the "max" field is not provided then its default value is zero. Is this the intended behavior? Quoting [KEP 9 - Capacity Scheduling](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/9-capacity-scheduling): "_Max is infinite by default, min is 0 by default_", so I was expecting the plugin to threat the max field as "infinite" when the custom resource does not specify it.

Having the max field with "infinite" as the default value would be helpful in those cases where someone needs to define just the minimum resources that need to be guaranteed to a namespace, without specifying any upper limit constraint. In such a case, I would like to create the `ElasticQuota` resource as follows, without specifying any value for the max field:

```yaml
kind: ElasticQuota
metadata:
  name: quota1
  namespace: quota1
spec:
   min:
    cpu: 4
```

https://github.com/kubernetes-sigs/scheduler-plugins/issues/424#tasklist-block-5282e773-1998-444a-8604-324ddd2a1d32

Gekko0114:
#### What type of PR is this?
/kind bug

This PR sets default values for the Max/Min of CPU/Memory/EphemeralStorage of ElasticQuota. The default value of Max is the maximum number of int64, which is 9223372036854775807. The default value of Min is 0. Default value is set when creating elasticquotainfo.

Additionally, if a Pod has other resource fields but Max/Min doesn't have it, the cmp2 function will use the default value of min/max.

This fix is needed to ensure that the ElasticQuota is functioning properly and to avoid unexpected behavior when dealing with resource limits.

Fixes #424 

Please let me know if I have made any mistakes or if there are any improvements that could be made.
Your comments and feedback are appreciated.

```release-note
NONE
```

### DIFF
--- a/apis/scheduling/v1alpha1/types.go
+++ b/apis/scheduling/v1alpha1/types.go
@@ -37,7 +37,8 @@ type ElasticQuota struct {
 
 	// ElasticQuotaSpec defines the Min and Max for Quota.
 	// +optional
-	Spec ElasticQuotaSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
+	// +kubebuilder:default={max:{cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807},min:{cpu:0,memory:0,ephemeral-storage:0}}
+	Spec ElasticQuotaSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
 
 	// ElasticQuotaStatus defines the observed use.
 	// +optional
@@ -49,13 +50,13 @@ type ElasticQuotaSpec struct {
 	// Min is the set of desired guaranteed limits for each named resource.
 	// +kubebuilder:default={cpu:0,memory:0,ephemeral-storage:0}
 	// +optional
-	Min v1.ResourceList `json:"min" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
+	Min v1.ResourceList `json:"min,omitempty" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
 
 	// Max is the set of desired max limits for each named resource. The usage of max is based on the resource configurations of
 	// successfully scheduled pods.
 	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807}
 	// +optional
-	Max v1.ResourceList `json:"max" protobuf:"bytes,2,rep,name=max, casttype=ResourceList,castkey=ResourceName"`
+	Max v1.ResourceList `json:"max,omitempty" protobuf:"bytes,2,rep,name=max, casttype=ResourceList,castkey=ResourceName"`
 }
 
 // ElasticQuotaStatus defines the observed use.
--- a/apis/scheduling/v1alpha1/types.go
+++ b/apis/scheduling/v1alpha1/types.go
@@ -37,7 +37,6 @@ type ElasticQuota struct {
 
 	// ElasticQuotaSpec defines the Min and Max for Quota.
 	// +optional
-	// +kubebuilder:default={max:{cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807},min:{cpu:0,memory:0,ephemeral-storage:0}}
 	Spec ElasticQuotaSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
 
 	// ElasticQuotaStatus defines the observed use.
@@ -48,13 +47,11 @@ type ElasticQuota struct {
 // ElasticQuotaSpec defines the Min and Max for Quota.
 type ElasticQuotaSpec struct {
 	// Min is the set of desired guaranteed limits for each named resource.
-	// +kubebuilder:default={cpu:0,memory:0,ephemeral-storage:0}
 	// +optional
 	Min v1.ResourceList `json:"min,omitempty" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
 
 	// Max is the set of desired max limits for each named resource. The usage of max is based on the resource configurations of
 	// successfully scheduled pods.
-	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807}
 	// +optional
 	Max v1.ResourceList `json:"max,omitempty" protobuf:"bytes,2,rep,name=max, casttype=ResourceList,castkey=ResourceName"`
 }
--- a/apis/scheduling/v1alpha1/types.go
+++ b/apis/scheduling/v1alpha1/types.go
@@ -47,13 +47,13 @@ type ElasticQuota struct {
 // ElasticQuotaSpec defines the Min and Max for Quota.
 type ElasticQuotaSpec struct {
 	// Min is the set of desired guaranteed limits for each named resource.
-	// +kubebuilder:default={cpu:0,memory:0}
+	// +kubebuilder:default={cpu:0,memory:0,ephemeral-storage:0}
 	// +optional
 	Min v1.ResourceList `json:"min" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
 
 	// Max is the set of desired max limits for each named resource. The usage of max is based on the resource configurations of
 	// successfully scheduled pods.
-	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807}
+	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807}
 	// +optional
 	Max v1.ResourceList `json:"max" protobuf:"bytes,2,rep,name=max, casttype=ResourceList,castkey=ResourceName"`
 }
--- a/apis/scheduling/v1alpha1/types.go
+++ b/apis/scheduling/v1alpha1/types.go
@@ -37,7 +37,7 @@ type ElasticQuota struct {
 
 	// ElasticQuotaSpec defines the Min and Max for Quota.
 	// +optional
-	Spec ElasticQuotaSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+	Spec ElasticQuotaSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
 
 	// ElasticQuotaStatus defines the observed use.
 	// +optional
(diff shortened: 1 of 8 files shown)


pkg/capacityscheduling/elasticquota.go
```diff
@@ -17,12 +17,17 @@ limitations under the License.
...
 import (
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"sigs.k8s.io/scheduler-plugins/pkg/util"
 )
 
+const (
+	UpperBoundOfMax = 9223372036854775807
```
PiotrProkop, line 27: you can use `math.MaxInt64` here https://pkg.go.dev/math#pkg-constants

apis/scheduling/v1alpha1/types.go
```diff
@@ -47,13 +47,15 @@ type ElasticQuota struct {
...
 type ElasticQuotaSpec struct {
 	// Min is the set of desired guaranteed limits for each named resource.
+	// +kubebuilder:default={cpu:0,memory:0}
 	// +optional
-	Min v1.ResourceList `json:"min,omitempty" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
+	Min v1.ResourceList `json:"min" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
 
 	// Max is the set of desired max limits for each named resource. The usage of max is based on the resource configurations of
 	// successfully scheduled pods.
+	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807}
```
PiotrProkop, line 56: what about other resources? Like `gpu`, hugepages, ephemeral-storage or devices advertised by device plugins?

Gekko0114: Thank you for your comment.

In my opinion, I don't think it's necessary to cover all resources. One reason is that since new resources may be added in the future, it would be difficult to accommodate them all. For example, for the GPU plugin, there are nvidia.com/gpu and amd.com/gpu, and more may be added in the future(https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/). It's not necessary to research and cover all of them.

Another reason is that there are limits to the fields that the Resource type can hold. Max/min is converted to ElasticQuotaInfo as a type Resource(https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/framework/types.go#L432-L441). The fields that the Resource type explicitly has are only MilliCPU, Memory, EphemeralStorage, and AllowedPodNumber (AllowedPodNumber is not related to ElasticQuota), so I don't think it's necessary to declare fields other than cpu, memory, and ephemeral-storage. And other fields are handled as ScalarResources(https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/capacityscheduling/elasticquota.go#L186), so I don't use specific name other than CPU, Memory and EphemeralStorage.

However, while writing this reply, I thought that it may be okay to add ephemeral-storage, so I will include it.

Please let me know if there is anything wrong with my response.

pkg/capacityscheduling/elasticquota.go
Gekko0114: Thanks. Fixed it.

apis/scheduling/v1alpha1/types.go
```diff
@@ -37,7 +37,7 @@ type ElasticQuota struct {
 
 	// ElasticQuotaSpec defines the Min and Max for Quota.
 	// +optional
-	Spec ElasticQuotaSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
+	Spec ElasticQuotaSpec `json:"spec" protobuf:"bytes,2,opt,name=spec"`
```
Huang-Wei, line 40: remove `omitempty` in a Spec field doesn't seem conventional. With the new CRD, how it looks if the user creates a bare ElastiQuota without specifying `Spec`?

apis/scheduling/v1alpha1/types.go
```diff
@@ -47,13 +47,15 @@ type ElasticQuota struct {
 // ElasticQuotaSpec defines the Min and Max for Quota.
 type ElasticQuotaSpec struct {
 	// Min is the set of desired guaranteed limits for each named resource.
+	// +kubebuilder:default={cpu:0,memory:0,ephemeral-storage:0}
 	// +optional
```
Huang-Wei, line 51: Let's remove `+optional` as well to keep consistency (given `omitempty` was deleted).

apis/scheduling/v1alpha1/types.go
```diff
@@ -47,13 +47,15 @@ type ElasticQuota struct {
...
-	Min v1.ResourceList `json:"min,omitempty" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
+	Min v1.ResourceList `json:"min" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
 
 	// Max is the set of desired max limits for each named resource. The usage of max is based on the resource configurations of
 	// successfully scheduled pods.
+	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807}
 	// +optional
```
Huang-Wei, line 57: Ditto.

apis/scheduling/v1alpha1/types.go
```diff
@@ -47,13 +47,15 @@ type ElasticQuota struct {
...
 type ElasticQuotaSpec struct {
 	// Min is the set of desired guaranteed limits for each named resource.
+	// +kubebuilder:default={cpu:0,memory:0,ephemeral-storage:0}
 	// +optional
-	Min v1.ResourceList `json:"min,omitempty" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
+	Min v1.ResourceList `json:"min" protobuf:"bytes,1,rep,name=min, casttype=ResourceList,castkey=ResourceName"`
 
 	// Max is the set of desired max limits for each named resource. The usage of max is based on the resource configurations of
 	// successfully scheduled pods.
+	// +kubebuilder:default={cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807}
```
Huang-Wei, line 56: this looks a bit strange to me. Is there a better way to represent this?

Is it worth introducing `map[ResourceName]*resource.Quantity`?

Huang-Wei:
It'd nice to have @denkensk also chimed in to share some thoughts :)

apis/scheduling/v1alpha1/types.go
Gekko0114: Thank you for your feedback.
I noticed that if I remove the optional tag, I am unable to apply the following Yaml, which is not what I intended.
Also I realized that I don't need to remove the `omitempty` tag. Therefore, I have decided to keep both `+optional` and `omitempty` tags.
```
kind: ElasticQuota
metadata:
  name: test
  namespace: test
spec:
  min:
    cpu: 20
    memory: 40Gi
    nvidia.com/gpu: 2
```

apis/scheduling/v1alpha1/types.go
Gekko0114: Thank you for your feedback.
I realized that I don't need to remove the omitempty tag. Therefore, I have decided to keep both `+optional` and `omitempty` tags.

> With the new CRD, how it looks if the user creates a bare ElastiQuota without specifying Spec?

If the user creates a bare ElasticQuota without specifying `Spec` like following Yaml, `max` and `min`are not created as default, which is not what I intended. Therefore, I have set up a default value for `Spec` as well.

Here is an example Yaml without specifying `Spec`.

```
kind: ElasticQuota
metadata:
  name: test
  namespace: test
```

apis/scheduling/v1alpha1/types.go
Gekko0114: I understand from your comment that you suggested creating a variable and assigning it to `kubebuilder`'s marker. I tried it, but I couldn't assign a variable to `kubebuilder`'s marker. I also read `kubebuilder`'s documentation, but I couldn't find a good way to do it.
If there is a better way, please let me know. Also, please let me know if I misunderstood your comment.

This is what I did.
First, I created a variable
`MaxBound := "{cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807}"` and used it in `kubebuilder`'s marker. However it doesn't work.

denkensk: It's also a little strange to me.  Does it look like this if I describe the CR?
```yaml
  max:
    cpu: 9223372036854775807
    ephemeral-storage: 9223372036854775807
    memory: 9223372036854775807
  min:
    cpu: 10
    ephemeral-storage: 0
    memory: 0
```

I prefer not to set the default value for min and max. We can set the default value in the scheduler when we init the cache but do not need to put it in CR.

Gekko0114: Thank you for your comments. It is very helpful for me.

If you apply the following resource, yes you can get CR as you wrote.
```
  min:
    cpu: 10
    ephemeral-storage: 0
    memory: 0
```

I see. Then I will handle it in the cache and try not to change default CR.

apis/scheduling/v1alpha1/types.go
```diff
@@ -37,6 +37,7 @@ type ElasticQuota struct {
 
 	// ElasticQuotaSpec defines the Min and Max for Quota.
 	// +optional
+	// +kubebuilder:default={max:{cpu:9223372036854775807,memory:9223372036854775807,ephemeral-storage:9223372036854775807},min:{cpu:0,memory:0,ephemeral-storage:0}}
 	Spec ElasticQuotaSpec `json:"spec,omitempty" protobuf:"bytes,2,opt,name=spec"`
```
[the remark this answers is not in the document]
Gekko0114: If you set ElasticQuotaSpec but not set ElasticQuota, we can't handle the case like "EQ with Max but without Min".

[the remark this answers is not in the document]
Gekko0114: As you wrote other comment, it may be strange. So I will handle it within cache or somewhere, not the default of CR

Gekko0114:
Hi @denkensk, @Huang-Wei 

Thank you very much for your comments. I have updated the PR as per your feedback.
Instead of setting the default value in the CR itself, I have set it when creating the cache (ElasticQuotaInfo).

Could you take a look when you have time?

Thanks,

Issue #440, opened 2022-10-29
Pod group label has invalid key


sftim:
https://github.com/kubernetes-sigs/scheduler-plugins/blob/66dabd41eb42dd6e96e1762c89cf96b4eff05bdd/pkg/coscheduling/README.md mentions a label:
`pod-group.scheduling.sigs.k8s.io`

This label doesn't follow the official format, which should be a DNS domain name, then `/`, then a key string.

For example: `scheduling.kubernetes.io/pod-group`.

This issue is to report that design bug and request that we fix it. We can also support the legacy label indefinitely as a deprecated substitute.

sftim:
Once we fix this, we should file an issue about documenting the new valid label in https://kubernetes.io/docs/reference/labels-annotations-taints/

/kind bug

denkensk:
Thanks. 
/assign

cmssczy:
hi @denkensk , I want to work on it if you haven't started. 
Do you think `scheduling.sigs.k8s.io/pod-group` is a good choice?

denkensk:
>  I want to work on it if you haven't started.

OK Thanks @cmssczy 

> Do you think scheduling.sigs.k8s.io/pod-group is a good choice?

I think it's better to both support `scheduling.sigs.k8s.io/pod-group` and `pod-group.scheduling.sigs.k8s.io` for the compatibility with existing users.

Huang-Wei:
This is completed.

PR #521, opened 2023-02-21
feat: update PodGroup label key

czybjtu:
#### What type of PR is this?
/kind feature

Fixes #https://github.com/kubernetes-sigs/scheduler-plugins/issues/440

```release-note
using new pod label key `scheduling.x-k8s.io/pod-group` to define a PodGroup. 
```

### DIFF
--- a/pkg/controllers/podgroup_test.go
+++ b/pkg/controllers/podgroup_test.go
@@ -281,7 +281,6 @@ func setUp(ctx context.Context,
 	objs := []runtime.Object{pg}
 	if len(podNames) != 0 {
 		ps := makePods(podNames, pgName, podPhase, podOwnerReference)
-		// s.AddKnownTypes(clientgoscheme.SchemeGroupVersion, ps)
 		objs = append(objs, ps[0], ps[1])
 	}
 	client := fake.NewFakeClient(objs...)
--- a/pkg/coscheduling/README.md
+++ b/pkg/coscheduling/README.md
@@ -16,7 +16,7 @@ This folder holds the coscheduling plugin implementations based on [Coscheduling
 
 ### PodGroup
 
-We use a special label named `pod-group.scheduling.x-k8s.io` to define a PodGroup. Pods that set this label and use the same value belong to the same PodGroup.
+We use a special label named `scheduling.x-k8s.io/pod-group` to define a PodGroup. Pods that set this label and use the same value belong to the same PodGroup.
 
 ```
 # PodGroup CRD spec
@@ -28,9 +28,9 @@ spec:
   scheduleTimeoutSeconds: 10
   minMember: 3
 ---
-# Add a label `pod-group.scheduling.x-k8s.io` to mark the pod belongs to a group
+# Add a label `scheduling.x-k8s.io/pod-group` to mark the pod belongs to a group
 labels:
-  pod-group.scheduling.x-k8s.io: nginx
+  scheduling.x-k8s.io/pod-group: nginx
 ```
 
 We will calculate the sum of the Running pods and the Waiting pods (assumed but not bind) in scheduler, if the sum is greater than or equal to the minMember, the Waiting pods
@@ -108,7 +108,7 @@ spec:
       name: nginx
       labels:
         app: nginx
-        pod-group.scheduling.x-k8s.io: nginx
+        scheduling.x-k8s.io/pod-group: nginx
     spec:
       containers:
       - name: nginx
--- a/pkg/util/podgroup.go
+++ b/pkg/util/podgroup.go
@@ -47,12 +47,12 @@ func CreateMergePatch(original, new interface{}) ([]byte, error) {
 	return patch, nil
 }
 
-// GetPodGroupLabel get pod group from pod annotations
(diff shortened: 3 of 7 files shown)


czybjtu:
@denkensk PTAL

pkg/util/podgroup.go
```diff
@@ -47,9 +47,12 @@ func CreateMergePatch(original, new interface{}) ([]byte, error) {
 	return patch, nil
 }
 
-// GetPodGroupLabel get pod group from pod annotations
+// GetPodGroupLabel get pod group name from pod annotations
```
zwpaper, line 50:
```suggestion
// GetPodGroupLabel get pod group name from pod labels
```

pkg/controllers/podgroup_controller.go
```diff
@@ -99,7 +99,20 @@ func (r *PodGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		log.Error(err, "List pods for group failed")
 		return ctrl.Result{}, err
 	}
-	pods := podList.Items
+	pods := podList.DeepCopy().Items
+	// use deprecated label key if cannot list pod by new label.
```
zwpaper, line 103:
```suggestion
	// also list pods by the deprecated key for compatibility
```

pkg/controllers/podgroup_test.go
```diff
@@ -310,6 +326,20 @@ func makePods(podNames []string, pgName string, phase v1.PodPhase, reference []m
 	return pds
 }
 
+func makePodsWithDeprecatedLabel(podNames []string, pgName string, phase v1.PodPhase, reference []metav1.OwnerReference) []*v1.Pod {
```
zwpaper, line 329: how about adding an args for `makePods` instead of a new function?

zwpaper (changes requested):
@denkensk I have 2 more questions,
1. Should we handle the case that a Pod has both the new and deprecated key?
2. Should we add docs in README noticing the key changes?

denkensk:
> Should we handle the case that a Pod has both the new and deprecated key?

If so, the new key first

> Should we add docs in README noticing the key changes? 

Yes. It's essential and necessary.

czybjtu:
> > Should we handle the case that a Pod has both the new and deprecated key?
> 
> If so, the new key first

Having implement this in function `GetPodGroupLabel`.

> > Should we add docs in README noticing the key changes?
> 
> Yes. It's essential and necessary.

Added a notice in `pkg/coscheduling/README.md`. Please let me know if it is not suitable

pkg/controllers/podgroup_controller.go
```diff
@@ -99,7 +99,20 @@ func (r *PodGroupReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		log.Error(err, "List pods for group failed")
 		return ctrl.Result{}, err
 	}
-	pods := podList.Items
+	pods := podList.DeepCopy().Items
+	// also list pods by the deprecated key for compatibility.
```
Huang-Wei, line 103: we don't need to actually, given the version is alpha.

Also I'm planning to standardize the group name in this release. So it's a breaking change anyway.

pkg/coscheduling/README.md
```diff
@@ -15,8 +15,9 @@ This folder holds the coscheduling plugin implementations based on [Coscheduling
 ## Tutorial
 
 ### PodGroup
+>❗We have changed this label from `pod-group.scheduling.sigs.k8s.io` to `scheduling.sigs.k8s.io/pod-group` in new version.
```
Huang-Wei, line 18: Once #526 gets merged, update it to `scheduling.x-k8s.io/pod-group`

pkg/controllers/podgroup_controller.go
czybjtu: @denkensk WDYT?

pkg/controllers/podgroup_test.go
```diff
@@ -281,7 +281,6 @@ func setUp(ctx context.Context,
 	objs := []runtime.Object{pg}
 	if len(podNames) != 0 {
 		ps := makePods(podNames, pgName, podPhase, podOwnerReference)
-		// s.AddKnownTypes(clientgoscheme.SchemeGroupVersion, ps)
```
Huang-Wei, line 284: remove this.

Huang-Wei (commented):
LGTM. Just a nit.

pkg/controllers/podgroup_test.go
czybjtu: This comment had been removed.

Huang-Wei: It's still there :)

czybjtu: Am I misunderstanding you? This line had been remove in this pr already. 
Did you mean remove this line or uncomment this line?

Huang-Wei: Ah my bad... somehow I always perceived this line was newly introduced...

Huang-Wei:
/lgtm
/approve

Thanks @czybjtu !

PR #526, opened 2023-02-28
⚠️ BREAKING CHANGE: switch API group to x-k8s.io

Huang-Wei:
#### What type of PR is this?

/kind api-change

#### What this PR does / why we need it:

It's requested to document all usages of "k8s.io" and "kubernetes.io" in k/website. However, scheduler-plugins is not supposed to be part of upstream.

To comply with https://kubernetes.io/docs/reference/labels-annotations-taints/, we should stop using "k8s.io" and "kubernetes.io" in all places of this repo, including:

- API group
- Labels
- Annotations

#### Which issue(s) this PR fixes:

Fixes #525

#### Does this PR introduce a user-facing change?

```release-note
ACTION REQUIRED:
- The CRDs `PodGroup` and `ElasticQuota` are now using a new API group: `scheduling.x-k8s.io`. When you upgrade scheduler-plugins to version v0.25.x, run `kubectl replace -f <crd yamls>` to register the new CRDs.
- The Pods work with `PodGroup` feature need to migrate to use new label: `scheduling.x-k8s.io/pod-group`.
```

PR #528, opened 2023-03-01
switch API group to x-k8s.io for network-aware CRDs and plugins

jpedro1992:
#### What type of PR is this?

/kind api-change

To comply with https://kubernetes.io/docs/reference/labels-annotations-taints/, we should stop using "k8s.io" and "kubernetes.io" in all places of this repo, including:

Fixes #526

```release-note
ACTION REQUIRED:
- The CRDs `AppGroup` and `NetworkTopology` are now using a new API group: `diktyo.x-k8s.io`. When you upgrade scheduler-plugins to version v0.25.x, run `kubectl replace -f <crd yamls>` to register the new CRDs.
- Pods need to migrate to use new label: `appgroup.diktyo.x-k8s.io/v1alpha1`.
```


Huang-Wei (commented):
/lgtm
/approve

Thanks @jpedro1992 !

Huang-Wei:
@jpedro1992 could you run `go mod tidy` and then `go mod vendor` and check-in the go modules bits as well? That would fix the CI failure.

Huang-Wei:
@jpedro1992 found a nit: could you also update the `+groupName` field in `doc.go` of the two projects? For now, they still point to k8s.io.

PR #530, opened 2023-03-03
Fix an issue arm64 image is not built properly

Huang-Wei:
#### What type of PR is this?

/kind regression

Fixes #

```release-note
NONE
```


Huang-Wei:
Thanks @seanmalloy . The build is green now: https://testgrid.k8s.io/sig-scheduling#post-scheduler-plugins-push-images

PR #533, opened 2023-03-03
doc: using --namespace option when installed by helm

czybjtu:
#### What type of PR is this?
/kind documentation

Fixes #https://github.com/kubernetes-sigs/scheduler-plugins/pull/519#issuecomment-1452922035

```release-note
users could install scheduler-plugins-scheduler in any custom namespace by helm
```


zwpaper:
gotcha, I will look into it in the following days

/assign

manifests/install/charts/as-a-second-scheduler/README.md
```diff
@@ -26,7 +26,8 @@ Quick start instructions for the setup and configuration of as-a-second-schedule
 ```bash
 $ git clone git@github.com:kubernetes-sigs/scheduler-plugins.git
 $ cd scheduler-plugins/manifests/install/charts
-$ helm install scheduler-plugins as-a-second-scheduler/
+# you can specify target namespace by `--namespaces` option
+$ helm install scheduler-plugins as-a-second-scheduler/ --create-namespace --namespace scheduler-plugins
```
Huang-Wei, line 30: I'd like to support --create-namespace and --namespace as a bonus, i.e., if these 2 parameters are not provided, the old command `helm install scheduler-plugins as-a-second-scheduler/` still functions well and defaults the namespace to `scheduler-plugins`. But now it defaults the current-namespace when user runs the command, which should be fixed.

cc @zwpaper

zwpaper: let me raise a PR later today, we should be in time for v0.25.x

zwpaper: @Huang-Wei I found a limitation in helm, `{{ .Release.Namespace }}`, which is the value of `-n` args in `helm install`, will be rendered to `default` no matter user omits the `-n` args or specifies `-n default`.

in short, we can not distinguish `-n default` from not specifying `-n` when using `helm install`.

how about we install the scheduler plugins to `scheduler-plugins` ns in both cases when specifying `-n default` or not specifying a ns?

czybjtu: > will be rendered to `default` no matter user omits the `-n` args or specifies `-n default`

{{ .Release.Namespace }} is not always to be rendered to `default` namespace. It depends on the `default namspace` in kubectl current-context.
{{ .Release.Namespace }} will be rendered to `test` namespace after you set current context namespace to `test`:
```bash
k config set-context --current --namespace=test
```

zwpaper: thanks for pointing that out @czybjtu, but it should be ok when users not specifying `-n`, no matter what the default ns is, we can install scheduler plugins in `scheduler-plugins` ns.

the problem is that when users specified `-n default`, can we still install in `scheduler-plugins` ns? 

helm chart can not detect whether `-n` is specified.

Huang-Wei: Here is my 2 cents:

- `helm install`: the whole installation dumps stuff to ns `scheduler-plugins`, and it creates the namespace `scheduler-plugins` if needed. This is aligned with the old behavior.
- `helm install -n <ns>`: install to `<ns>`. If `<ns>` doesn't exist, the install fails.
- `helm install -n <ns> --create-namespace`: install to `<ns>` and create `<ns>` if it doesn't exist.

Huang-Wei: > the problem is that when users specified -n default, can we still install in scheduler-plugins ns?

I'd say if `-n` is specified, we should install to that specified namespace.

Huang-Wei: > helm chart can not detect whether -n is specified.

I found a way to work it around: check whether `.Release.Namespace` is `default`. But yes, in this case, we cannot support installing the chart to `default` namespace until Helm support a built-in variable like `Args`.

Will send out a PR shortly.

zwpaper: I have reconsidered the installation method, and have a few different thoughts.

Let's step back to the why we want to raise this patch, we could keep the compatibility to the origin installation, but this not followed the helm way.

Helm chart is namespaced, it doesn't make that sense if we installed a chart under default ns, but all contents appear in another one, like `scheduler-plugins`.

I have checked some the other projects, like descheduler, nginx-ingress, they both encourage people to use their chart with namespace specified.

Ref: 
- https://kubernetes.github.io/ingress-nginx/deploy/#quick-start
- https://github.com/kubernetes-sigs/descheduler/blob/master/charts/descheduler/README.md

zwpaper: So maybe we should reconsider whether we want to keep the previous installation way, or switch to the new one with namespace recommended

Huang-Wei: @zwpaper as long as the new way is idiomatic, I'm fine with that. One question: is it conventional to save the efforts of creating namespace for the user? i.e., `helm install ... -n <ns>` should be good enough?

btw: the workaround I gave doesn't quite work, thanks @czybjtu for pointing out - the current-context matters.

zwpaper: Good point, as I know, most cases do NOT create ns for users, we should drop the `--create-namespace` in the example, and then add a line to mentions that user could also create the ns in the same time with `--create-namespace`

@czybjtu

manifests/install/charts/as-a-second-scheduler/README.md
```diff
@@ -26,7 +26,11 @@ Quick start instructions for the setup and configuration of as-a-second-schedule
 ```bash
 $ git clone git@github.com:kubernetes-sigs/scheduler-plugins.git
 $ cd scheduler-plugins/manifests/install/charts
-$ helm install scheduler-plugins as-a-second-scheduler/
+# Helm will install all components in your corrent-context namespace.
+$ helm install scheduler-plugins as-a-second-scheduler/ 
+# You can use `--namespace` parameter to specify the namespace you want to deploy. 
+# You can also use `--create-namespace` parameter to create it automatically if the specified namespace does not exist.
+$ helm install scheduler-plugins as-a-second-scheduler/ --create-namespace --namespace scheduler-plugins
```
zwpaper, line 33: I would prefer we provide the `scheduler-plugins` ns by default and provide some guidance to auto-create the ns.

it seems not worth mentioning the context namespace.

```suggestion
helm install scheduler-plugins as-a-second-scheduler/ --namespace scheduler-plugins
    ```

By specifying `--create-namespace` parameter, You can also create the namespace if it does not exist
```

ps, I have to indent the quote in the suggestions

zwpaper: WDYT @Huang-Wei

Huang-Wei: sounds good. I will mark this as an incompatible change in the release notes, and here let's just simply suggest:

```
helm install scheduler-plugins as-a-second-scheduler/ --namespace scheduler-plugins --create-namespace
```

to install the Helm chart.

Huang-Wei:
/approve

Thanks @czybjtu @zwpaper !

PR #536, opened 2023-03-05
helm: install chart to ns scheduler-pluings if `-n` is not provided

Huang-Wei:
#### What type of PR is this?

/kind regression

A follow-up of https://github.com/kubernetes-sigs/scheduler-plugins/pull/533#discussion_r1125581851.

With this PR, the command

```
helm install scheduler-plugins manifests/install/charts/as-a-second-scheduler/
```

works as before - the chart will be installed into the default namespace `default-plugins`.

And it also works well with PR #519 to support `-n` command:

```
helm install scheduler-plugins manifests/install/charts/as-a-second-scheduler/ -n foo --create-namespace
```

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: scheduler-config
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ ne .Release.Namespace "default" | ternary .Release.Namespace .Values.namespace }}
 data:
   scheduler-config.yaml: |
     apiVersion: kubescheduler.config.k8s.io/v1beta3
--- a/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
@@ -2,7 +2,7 @@ kind: Deployment
 apiVersion: apps/v1
 metadata:
   name: {{ .Values.controller.name }}
-  namespace: {{ .Release.Namespace }}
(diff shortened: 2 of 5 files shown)


Huang-Wei:
cc @zwpaper @czybjtu

zwpaper:
I have add one thoughts here, https://github.com/kubernetes-sigs/scheduler-plugins/pull/533#discussion_r1125583377, PTAL

let me check this PR and we can have more discussion

zwpaper:
like @czybjtu said in https://github.com/kubernetes-sigs/scheduler-plugins/pull/533#discussion_r1125471522, helm would render `.Release.Namespace` to some others if users changed the default ns in kubeconfig, makes this PR can not handle that case

czybjtu:
`ne .Release.Namespace "default"` cannot handle the case that users changed the default ns in kubeconfig by `k config set-context --current --namespace=test`

I wonder if we really need to support this feature that install chart to ns `scheduler-pluings` if -n is not provided. I think install chart to `current context` ns is also make sense if user **do not** provice `-n`.

Huang-Wei:
> like @czybjtu said in https://github.com/kubernetes-sigs/scheduler-plugins/pull/533#discussion_r1125471522, helm would render .Release.Namespace to some others if users changed the default ns in kubeconfig, makes this PR can not handle that case

> I wonder if we really need to support this feature that install chart to ns `scheduler-pluings` if -n is not provided. I think install chart to `current context` ns is also make sense if user **do not** provide `-n`.

This is a valid point. Thanks.

So can we get to a condense that in today's Helm, it's basically impossible to achieve the goal to "fall back to a defaulted namespace when -n is not specified"?

zwpaper:
> it's basically impossible to achieve the goal to "fall back to a defaulted namespace when -n is not specified"?

as for as I know, yes, that's why most of the charts' installation guides added a `-n package` in the helm installation examples

PR #545, opened 2023-03-12
doc: update ElasticQuota's max value

Gekko0114:
#### What type of PR is this?
/kind documentation

Updated the explanation of ElasticQuota's max value.

```release-note
NONE
```

### DIFF
--- a/kep/9-capacity-scheduling/README.md
+++ b/kep/9-capacity-scheduling/README.md
@@ -202,7 +202,8 @@ spec:
 ```
 
 The definitions of "min" and "max" have been explained above. Max is infinite
-by default, min is 0 by default. The sum of min in all ElasticQuota must be 
+(in particular `math.MaxInt64`) by default, min is 0 by default.
+The sum of min in all ElasticQuota must be 
 less than or equal to the total resource capacity of the cluster.The 
 ElasticQuota objects are created by a cluster administrator. One ElasticQuota
 object per namespace.


ffromani:
/ok-to-test

Hi! Thanks for your contribution!
I for myself do believe that clarifying that "infinite" actually means "maxint" is a useful and welcome addition, but I'd just link back to the golang spec rather than inline the actual value in our docs. Let's see what other reviewers think though.

Gekko0114:
Thanks for your comment !

> but I'd just link back to the golang spec 

Do you mean that I should simply paste the link like this spec?
https://cs.opensource.google/go/go/+/master:src/runtime/string.go;l=354?q=maxint64&ss=go%2Fgo
If I have misunderstood, please let me know!

kep/9-capacity-scheduling/README.md
```diff
@@ -202,7 +202,8 @@ spec:
 ```
 
 The definitions of "min" and "max" have been explained above. Max is infinite
-by default, min is 0 by default. The sum of min in all ElasticQuota must be 
+(math.MaxInt64: 9,223,372,036,854,775,807) by default, min is 0 by default.
```
Huang-Wei, line 205: Readers may not be interested in the concrete number, so:

```suggestion
(in particular `math.MaxInt64`) by default, min is 0 by default.
```

ffromani: @Gekko0114 (sorry for late answer) this is exactly what I tried to explain previously - sorry if I expressed myself poorly

Gekko0114: Thank you so much! I fixed it

Issue #467, opened 2022-12-20
[Trimaran]panic: version string "" doesn't match expected regular expression


wintle777:
I'm beginner and I don't know how to fix this error. Do you have any ideas?Thank you.

kubectl logs trimaran-6d9ddd6c9b-knqzm -n kube-system -c trimaran
panic: version string "" doesn't match expected regular expression: "^v(\d+\.\d+\.\d+)"
goroutine 1 [running]:
k8s.io/component-base/metrics.parseVersion({{0x0, 0x0}, {0x0, 0x0}, {0x23b378a, 0x0}, {0x20b5eb1, 0xb}, {0x0, 0x0}, ...})
	/home/nvidia/MY_PROJECT/scheduler-plugins-master/scheduler-plugins-master/vendor/k8s.io/component-base/metrics/version_parser.go:47 +0x274
k8s.io/component-base/metrics.newKubeRegistry({{0x0, 0x0}, {0x0, 0x0}, {0x23b378a, 0x0}, {0x20b5eb1, 0xb}, {0x0, 0x0}, ...})
	/home/nvidia/MY_PROJECT/scheduler-plugins-master/scheduler-plugins-master/vendor/k8s.io/component-base/metrics/registry.go:320 +0x119
k8s.io/component-base/metrics.NewKubeRegistry()
	/home/nvidia/MY_PROJECT/scheduler-plugins-master/scheduler-plugins-master/vendor/k8s.io/component-base/metrics/registry.go:335 +0x78
k8s.io/component-base/metrics/legacyregistry.init()
	/home/nvidia/MY_PROJECT/scheduler-plugins-master/scheduler-plugins-master/vendor/k8s.io/component-base/metrics/legacyregistry/registry.go:29 +0x1d

mofanke:
i faced the same issue ， https://github.com/kubernetes-sigs/scheduler-plugins/blob/release-1.22/doc/release-guide.md
seems need a tag ,not branch 
then
`make local-image`

wangchen615:
@wintle777 , thanks for raising the issue. The log message did not point to a particular line in the trimaran code, so I am unsure if it is just an issue with Trimaran. Have you tried other plugins?

At the same time, could you help share more info about the configuration of your Trimaran plugin? Did you run it as a secondary scheduler? And what metric provider are you using?

zwpaper:
[...]

PR #553, opened 2023-03-24
build: fallback version to based on date if no tags found

zwpaper:

when working on a fork without tags, like a single branch clone, we can not found a tag, and it will make the scheduler panic because not version set.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/e5a5dbd145e85146bc43c5ad11eecc22565d6ea3/Makefile#L24

may be related to https://github.com/kubernetes-sigs/scheduler-plugins/issues/467

this PR will fallback version to `v0.0.0` if not tags found

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ RELEASE_CONTROLLER_IMAGE:=controller:$(RELEASE_VERSION)
 # v20201009-v0.18.800-46-g939c1c0 - automated build for a commit(not a tag) and also a local build
 # v20200521-v0.18.800             - automated build for a tag
 VERSION=$(shell echo $(RELEASE_VERSION) | awk -F - '{print $$2}')
+VERSION:=$(or $(VERSION),v0.0.$(shell date +%Y%m%d))
 
 .PHONY: all
 all: build


Makefile
```diff
@@ -31,6 +31,7 @@ RELEASE_CONTROLLER_IMAGE:=controller:$(RELEASE_VERSION)
 # v20201009-v0.18.800-46-g939c1c0 - automated build for a commit(not a tag) and also a local build
 # v20200521-v0.18.800             - automated build for a tag
 VERSION=$(shell echo $(RELEASE_VERSION) | awk -F - '{print $$2}')
+VERSION:=$(or $(VERSION),v0.0.0)
```
ffromani, line 34: I agree a fix like this is beneficial for the project because it lowers the barrier to consume and experiment with.
Thinking about it, for internal snapshots/tests in RH we use the format `v0.0.$(date +%Y%m%d)` (we actually append also a sequential, but we can defer that) so it reads like `v0.0.20230324` for example.

I think it's worth mentioning but I have no real preference here.

ffromani (commented):
/approve

Makefile
zwpaper: sound great, applied the change to make it date based

PR #557, opened 2023-03-28
nrt: overreserve: drop custom indexer, use lister

### LINKED ISSUE #552: [noderesourcetopology]: remove/simplify the custom indexer
now that rebased on top of 1.26 (actually we needed 1.25) we can simplify or hopefully get rid of the custom replacement we had to have for the client-go indexer, and use the standard machinery. See https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/cache/indexer.go#L29 for details.

We should have no regression or change in behavior, just code simplification

ffromani:
#### What type of PR is this?

/kind cleanup

Optionally add one or more of the following kinds if applicable:

The overreserve cache needs to periodically list all the pods on a node to perform the synchronization check.
When the cache was introduced in 0f4ed7a9 , adding a thin layer over a indexer and manage a map of node -> slice of pod identifiers seemed like simple and nice. 
As usage and testing of the code increases, especially at scale, we are discovering more and more corner cases to handle.
So, we re-evaluate the whole idea and we drop the not-thin-anymore cache and we just use the simplest solution: in the resync code, we just use the battle-tested client-go lister, with trivial filtering over the pod properties to check if we need to consider on a node or not.

Fixes #552 

Performance (granted this is actually a concern, because there's no profiling evidence this was ever a hotspot) is not a concern yet, and the new code is way simpler and easier to reason about.

```release-note
NONE
```

This is a implementation detail the user has no direct control or visibility about.


pkg/noderesourcetopology/cache/overreserve_test.go
```diff
@@ -652,35 +657,38 @@ func dumpNRT(nrtObj *topologyv1alpha2.NodeResourceTopology) string {
...
+	return &fakePodNamespaceLister{
+		parent:    fpl,
+		namespace: namespace,
 	}
-	return objs, nil
 }
 
-func (fi *fakePodByNodeNameIndex) TrackReservedPod(pod *corev1.Pod, nodeName string)   {}
-func (fi *fakePodByNodeNameIndex) UntrackReservedPod(pod *corev1.Pod, nodeName string) {}
+func (fpnl *fakePodNamespaceLister) List(selector labels.Selector) ([]*corev1.Pod, error) {
```
Tal-or, line 685: Are this going to be implement at some point or they're just needed for the interface implementation?

Tal-or (commented):
Thank you for the PR.
I reviewed it deeply and it looks like a straight-forward change to me.
/lgtm

PR #558, opened 2023-03-28
Add integration tests for DiscardReservedNodes NRT cache type

PiotrProkop:
#### What type of PR is this?
/kind cleanup

Add integration tests for DiscardReservedNodes NRT cache type.
Enable DiscardReservedNodes cache when proper option in config is set.

Fixes #

```release-note
NONE
```


ffromani:
/retitle Add integration tests for DiscardReservedNodes NRT cache type

ffromani:
@PiotrProkop it's silly but still: can we use consistent naming for the feature, be it "nodelocking" or "DiscardReserved nodes nrt cache" or anything else you see most fit?

When I use the terms interchangeably (and happens often in NRT) I try to always use all aliases to make it trivial to find the change later in the future.

PiotrProkop:
> @PiotrProkop it's silly but still: can we use consistent naming for the feature, be it "nodelocking" or "DiscardReserved nodes nrt cache" or anything else you see most fit?
> 
> When I use the terms interchangeably (and happens often in NRT) I try to always use all aliases to make it trivial to find the change later in the future.

Yeah definitely, `DiscardReservedNodes` is the way to go, nodelocking came from my branch name :(

ffromani:
> > @PiotrProkop it's silly but still: can we use consistent naming for the feature, be it "nodelocking" or "DiscardReserved nodes nrt cache" or anything else you see most fit?
> > When I use the terms interchangeably (and happens often in NRT) I try to always use all aliases to make it trivial to find the change later in the future.
> 
> Yeah definitely, `DiscardReservedNodes` is the way to go, nodelocking came from my branch name :(

no problem! It's just to remove unnecessary friction when, in the future, we will have to dig in git/github history. A little extra work now is likely to save headaches in the future. I'll do the review ASAP.

PiotrProkop:
@ffromani I couldn't figure out proper test scenarios without reading logs of the scheduler :) cause this cache type will eventually act as `passthrough` cache when scheduler will retry scheduling. The scenarios here are trivial but I have tried to test the difference in bahaviour between `DiscardReservedNodes` and `Overreserve`.

test/integration/noderesourcetopology_cache_test.go
```diff
@@ -264,6 +268,109 @@ func TestTopologyCachePluginWithoutUpdates(t *testing.T) {
 						}).Obj(),
 			},
 		},
+		{
+			name: "GU pod: DiscardReservedNodes: allows scheduling on both Zones",
```
ffromani, line 272: IIUC this test wants to demonstrate that DiscardReservedNodes can use both NUMA zones of a node without needing a NRT object update, comparing with OverReserve which cannot (because it overallocates).

So don't exercise yet the fact that the DiscardedReservedNodes is *preventing* access to a node when it is supposed to do, right?

test/integration/noderesourcetopology_cache_test.go
```diff
@@ -264,6 +268,109 @@ func TestTopologyCachePluginWithoutUpdates(t *testing.T) {
...
+						}).
+					Zone(
+						topologyv1alpha2.ResourceInfoList{
+							noderesourcetopology.MakeTopologyResInfo(cpu, "32", "8"),
+							noderesourcetopology.MakeTopologyResInfo(memory, "64Gi", "10Gi"),
+						}).Obj(),
+			},
+		},
+		{
+			name: "GU pod: DiscardReservedNodes: scheduling works as expected, when deleting pods",
```
ffromani, line 321: Can we please clarify "works as expected?" :) You mean that the scheduler can make use of a node which was previously unable to accomodate a workload because it had already a workload running which consumed too many resources, and once the existing delete is issued the new workload can be scheduled without waiting for a NRT update?

If this doesn't fit in the test name/description (and I think it will hardly fit) please add it as code comment

ffromani (commented):
IIUC the intent (please check the inline questions) these integration tests are a good starting point. 
I acknowledge the challenge of adding elaborate tests - `OverReserve` shares a very similar fate. I think it would be nice to add a code comment with a TODO about the cases we wish to cover but we don't know how to implement yet (and what's the blocker now).
(and I'll do the same for OverReserve later)

test/integration/noderesourcetopology_cache_test.go
```diff
@@ -689,3 +798,52 @@ func mkPFP(nodeName string, pods ...*corev1.Pod) string {
 	klog.Infof("PFP for %q: %s", nodeName, st.Repr())
 	return pfp
 }
+
+func discardReservedSchedulerProfile() schedapi.KubeSchedulerProfile {
```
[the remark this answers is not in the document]
PiotrProkop: Yes.

test/integration/noderesourcetopology_cache_test.go
PiotrProkop: Yes, it's hard to check `preventing` access behaviour, cause Scheduler will retry scheduling when the node is locked and it will finally be successful.

test/integration/noderesourcetopology_cache_test.go
PiotrProkop: Sure, changing now.

test/integration/noderesourcetopology_cache_test.go
ffromani: ok, and there's no way right (besides awkwardly parsing the logs) now to even verify the retry count.
I think really need to start the observability conversation.

ffromani (commented):
/lgtm
/hold cancel

That's a good start for integration tests, we need some coverage, we need also the final wiring. We will need to start the monitoring/observability conversation so we can later add more and more complex integration tests for this and for OverReserve.

Issue #429, opened 2022-09-21
[coscheduling] Infinite scheduling loop of distributed tasks block all other pods


zhiyxu:
Recently, we encountered some infinite scheduling loop issues when using Coscheduling in our production environment. 

Suppose the following scenario, there are 3 instances of the distributed task, and each instance requires 4 GPU resources. The resources in the current cluster are as follows:

| Node  | Resource Left |
| ----- | ------------- |
| node1 | 6GPU          |
| node2 | 2GPU          |
| node3 | 2GPU          |
| node4 | 2GPU          |

Obviously, there is resource fragmentation problem in the cluster. Since the total resources of the cluster can [satisfy](https://github.com/kubernetes-sigs/scheduler-plugins/blob/d831125ae12cee913bd2ad71d8a10e3776a478a4/pkg/coscheduling/core/core.go#L306) 3 instances, instance will succeeded in the PreFilter stage, the scheduling process is as follows:

- Instance 1 is successfully scheduled and wait in permit stage. Add Instance 2 and 3 to ActiveQueue
- Instance 2 fails to schedule due to resource fragmentation, instance 1 will be [rejected](https://github.com/kubernetes-sigs/scheduler-plugins/blob/d831125ae12cee913bd2ad71d8a10e3776a478a4/pkg/coscheduling/coscheduling.go#L168) in the PostFilter stage
- Instance 3 is successfully scheduled and wait in permit stage. Add instance 1 and 2 to ActiveQueue
- Instance 1 fails to schedule, instance 3 will be [rejected](https://github.com/kubernetes-sigs/scheduler-plugins/blob/d831125ae12cee913bd2ad71d8a10e3776a478a4/pkg/coscheduling/coscheduling.go#L168) in the PostFilter stage
- Instance 2 is successfully scheduled and wait in permit stage. Add Instance 1 and 3 to ActiveQueue
- ......scheduling loop

Worse, whenever an instance is added to the ActiveQueue, according to the [QueueSort](https://github.com/kubernetes-sigs/scheduler-plugins/blob/d831125ae12cee913bd2ad71d8a10e3776a478a4/pkg/coscheduling/coscheduling.go#L117) extension point, the distributed instance will be queued in front of all later created Pods, and the above loop will block all other later created Pods, these pods will stuck in Pending state and don't have schedule opportunities at all.

[...]

PR #559, opened 2023-03-29
add podGroup backoff time for coscheduling

KunWuLuan:
Adding a backoff in co-scheduling (https://github.com/kubernetes-sigs/scheduler-plugins/issues/429#issuecomment-1254515086) helps to avoid infinite scheduling loop.

```release-note
Add a new coscheduling plugin argument `podGroupBackoffSeconds` to configure backoff timer for failed PodGroup. User needs to explicitly specify a positive integer to enable this feature.
```

### DIFF
--- a/apis/config/scheme/scheme_test.go
+++ b/apis/config/scheme/scheme_test.go
@@ -29,7 +29,7 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/apis/config/testing/defaults"
 
 	"sigs.k8s.io/scheduler-plugins/apis/config"
-	"sigs.k8s.io/scheduler-plugins/apis/config/v1"
+	v1 "sigs.k8s.io/scheduler-plugins/apis/config/v1"
 	"sigs.k8s.io/scheduler-plugins/apis/config/v1beta2"
 	"sigs.k8s.io/scheduler-plugins/apis/config/v1beta3"
 	"sigs.k8s.io/scheduler-plugins/pkg/coscheduling"
@@ -66,6 +66,7 @@ profiles:
   - name: Coscheduling
     args:
       permitWaitingTimeSeconds: 10
+      podGroupBackoffSeconds: 0
       deniedPGExpirationTimeSeconds: 3
   - name: NodeResourcesAllocatable
     args:
@@ -472,6 +473,7 @@ kind: KubeSchedulerConfiguration
 profiles:
 - schedulerName: scheduler-plugins
   pluginConfig:
+  - name: Coscheduling # Test argument defaulting logic
   - name: TopologicalSort
     args:
       namespaces:
@@ -488,6 +490,12 @@ profiles:
 					SchedulerName: "scheduler-plugins",
 					Plugins:       defaults.PluginsV1,
 					PluginConfig: []schedconfig.PluginConfig{
+						{
+							Name: coscheduling.Name,
+							Args: &config.CoschedulingArgs{
+								PermitWaitingTimeSeconds: 60,
+							},
+						},
 						{
 							Name: topologicalsort.Name,
 							Args: &config.TopologicalSortArgs{
@@ -728,6 +736,7 @@ profiles:
       apiVersion: kubescheduler.config.k8s.io/v1beta2
       kind: CoschedulingArgs
       permitWaitingTimeSeconds: 10
+      podGroupBackoffSeconds: 0
     name: Coscheduling
   - args:
       apiVersion: kubescheduler.config.k8s.io/v1beta2
@@ -781,6 +790,7 @@ profiles:
 								Name: coscheduling.Name,
 								Args: &config.CoschedulingArgs{
 									PermitWaitingTimeSeconds: 10,
+									PodGroupBackoffSeconds:   20,
 								},
 							},
 							{
@@ -868,6 +878,7 @@ profiles:
       apiVersion: kubescheduler.config.k8s.io/v1beta3
       kind: CoschedulingArgs
       permitWaitingTimeSeconds: 10
+      podGroupBackoffSeconds: 20
     name: Coscheduling
   - args:
       apiVersion: kubescheduler.config.k8s.io/v1beta3
@@ -1022,6 +1033,7 @@ profiles:
       apiVersion: kubescheduler.config.k8s.io/v1
       kind: CoschedulingArgs
       permitWaitingTimeSeconds: 10
+      podGroupBackoffSeconds: 0
     name: Coscheduling
   - args:
       apiVersion: kubescheduler.config.k8s.io/v1
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -61,6 +61,7 @@ type Manager interface {
 	DeletePermittedPodGroup(string)
 	CalculateAssignedPods(string, string) int
 	ActivateSiblings(pod *corev1.Pod, state *framework.CycleState)
+	BackoffPodGroup(string, time.Duration)
 }
 
 // PodGroupManager defines the scheduling operation called
@@ -74,6 +75,8 @@ type PodGroupManager struct {
 	scheduleTimeout *time.Duration
 	// permittedPG stores the podgroup name which has passed the pre resource check.
 	permittedPG *gochache.Cache
+	// backedOffPG stores the podgorup name which failed scheudling recently.
+	backedOffPG *gochache.Cache
 	// pgLister is podgroup lister
 	pgLister pglister.PodGroupLister
 	// podLister is pod lister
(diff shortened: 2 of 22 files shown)


test/integration/noderesourcetopology_cache_test.go
```diff
@@ -485,7 +485,7 @@ func TestTopologyCachePluginWithoutUpdates(t *testing.T) {
 					checkPod = podIsPending
 				}
 
-				err = checkPod(1*time.Second, 20, cs, p.pod.Namespace, p.pod.Name)
```
Huang-Wei, line 488: I'd expect to see a new (sub-)integration test to simulate what was described in the original issue.

Otherwise, I'd postpone the impl. to next release.

KunWuLuan: Have added a new integration test for the case

Huang-Wei: Thanks. But even without introducing the backoff, the integration test seems to be also passing.

```
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins|master⚡ ⇒  go test ./test/integration/... -run TestPodgroupBackoff -count 1
ok  	sigs.k8s.io/scheduler-plugins/test/integration	27.100s
weih@m1max:~/go/src/sigs.k8s.io/scheduler-plugins|master⚡ ⇒  gst
On branch master
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   test/integration/coscheduling_test.go
```

Could you update the test to pass and only pass with the backoff changes? in other words, the test is expected to fail on master.

KunWuLuan: No problem

KunWuLuan: Done

pkg/coscheduling/core/core.go
```diff
@@ -74,6 +75,8 @@ type PodGroupManager struct {
 	scheduleTimeout *time.Duration
 	// permittedPG stores the podgroup name which has passed the pre resource check.
 	permittedPG *gochache.Cache
+	// podGroupBackoff stores the podgorup name which failed scheudling recently.
+	podGroupBackoff *gochache.Cache
```
Huang-Wei, line 79: `backedOffPG` may read better (named consistently with permittedPG)

pkg/coscheduling/core/core.go
```diff
@@ -61,6 +61,7 @@ type Manager interface {
 	DeletePermittedPodGroup(string)
 	CalculateAssignedPods(string, string) int
 	ActivateSiblings(pod *corev1.Pod, state *framework.CycleState)
+	AddToPodGroupBackoff(string)
```
Huang-Wei, line 64: maybe just simply name it `BackoffPodGroup(string)`

pkg/coscheduling/coscheduling.go
```diff
@@ -166,6 +168,14 @@ func (cs *Coscheduling) PostFilter(ctx context.Context, state *framework.CycleSt
 			waitingPod.Reject(cs.Name(), "optimistic rejection in PostFilter")
 		}
 	})
+
+	pods, err := cs.frameworkHandler.SharedInformerFactory().Core().V1().Pods().Lister().Pods(pod.Namespace).List(
+		labels.SelectorFromSet(labels.Set{v1alpha1.PodGroupLabel: util.GetPodGroupLabel(pod)}),
+	)
+	if err == nil && len(pods) >= int(pg.Spec.MinMember) {
```
Huang-Wei, line 175: can we remove the check here? (given `assigned` < pg.Spec.MinMember per L150)

pkg/coscheduling/core/core.go
```diff
@@ -94,13 +94,16 @@ func NewPodGroupManager(pgClient pgclientset.Interface, snapshotSharedLister fra
...
-		podGroupBackoff:      gochache.New(10*time.Second, 10*time.Second),
+		backedOffPG:          gochache.New(10*time.Second, 10*time.Second),
 	}
 	return pgMgr
 }
 
-func (pgMgr *PodGroupManager) AddToPodGroupBackoff(pgName string) {
-	pgMgr.podGroupBackoff.Add(pgName, nil, 10*time.Second)
+func (pgMgr *PodGroupManager) BackoffPodGroup(pgName string, backoff time.Duration) {
+	if backoff == time.Duration(0) {
```
Huang-Wei, line 103: commented elsewhere: IMO we should return as early as possible.

pkg/coscheduling/coscheduling.go
Huang-Wei: bump.

apis/config/scheme/scheme_test.go
```diff
@@ -29,7 +29,7 @@ import (
 	"k8s.io/kubernetes/pkg/scheduler/apis/config/testing/defaults"
 
 	"sigs.k8s.io/scheduler-plugins/apis/config"
-	"sigs.k8s.io/scheduler-plugins/apis/config/v1"
+	v1 "sigs.k8s.io/scheduler-plugins/apis/config/v1"
```
Huang-Wei, line 32: no need to add v1.

apis/config/scheme/scheme_test.go
```diff
@@ -1022,6 +1024,7 @@ profiles:
       apiVersion: kubescheduler.config.k8s.io/v1
       kind: CoschedulingArgs
       permitWaitingTimeSeconds: 10
+      podGroupBackoffTime: 0
```
Huang-Wei, line 1027: I'd expect without explicitly specifying it, UT would also pass. Could you check if you want to add explicit defaulting logic? (convert nil from versioned args to 0 in internal args)

pkg/coscheduling/coscheduling.go
KunWuLuan: I think we can not  remove the check here.

`assigned < pg.Spec.MinMember` check whether the gang has been satisfied and `len(pods) >= int(pg.Spec.MinMember)` check whether there is enough pods to schedule.

When the first pod come, scheduler may list not all pods. We will reject the first pod in preFilter
https://github.com/kubernetes-sigs/scheduler-plugins/blob/6ef6f63a0826b6045dd955eac599c902e396457f/pkg/coscheduling/core/core.go#L153-L156
But we should not add this pod group to Backoff. If we add this pod group to Backoff, all following pods will be rejected.

apis/config/scheme/scheme_test.go
KunWuLuan: `podGroupBackoffTime: 0` means we will not backoff pod groups.
I think maybe we don't have to add explicit defaulting logic, because by default nil will be convert to 0

apis/config/scheme/scheme_test.go
KunWuLuan: done

pkg/coscheduling/core/core.go
KunWuLuan: we can check whether `cs.pgBackoff != nil` outside, It's already done

apis/config/scheme/scheme_test.go
Huang-Wei: That's correct. The previous comment was more about "how it behaves without explicitly setting the argument", and hence I'd like to see this UT passes w/o specifying it to 0.

pkg/coscheduling/coscheduling.go
```diff
@@ -81,7 +80,10 @@ func New(obj runtime.Object, handle framework.Handle) (framework.Plugin, error)
 		frameworkHandler: handle,
 		pgMgr:            pgMgr,
 		scheduleTimeout:  &scheduleTimeDuration,
-		pgBackoff:        &pgBackoff,
+	}
+	if args.PodGroupBackoffSeconds != 0 {
```
Huang-Wei, line 84: nit: fail it if it's negative

Huang-Wei:
One comment ^^ and https://github.com/kubernetes-sigs/scheduler-plugins/pull/559#discussion_r1253795201

pkg/coscheduling/coscheduling.go
KunWuLuan: done

apis/config/scheme/scheme_test.go
KunWuLuan: Ok, I added a test for the case

Huang-Wei: sry for confusing, I misread - this is the test for encode.

could you apply this diff - we want to test both defaulting logic (in decode), and encode for non-default logic: https://gist.github.com/Huang-Wei/d86f188b4e85b48ad4e3e4aa4bd427da

Huang-Wei (commented):
Final comments.

KunWuLuan:
No problem, but in this diff, we set default backoff time to 60, not to 0, if there is not any args in profile, is this same with our expectation?

Huang-Wei:
> No problem, but in this diff, we set default backoff time to 60, not to 0, if there is not any args in profile, is this same with our expectation?

That was "PermitWaitingTimeSeconds" :) not backoff time. PermitWaitingTimeSeconds defaults to 60 as-is; Backoff time defaults to 0. The logic is accurate.

Huang-Wei:
Added release note.

/lgtm
/approve

Issue #567, opened 2023-03-30
Delete useless options in controller


czybjtu:
```bash
./bin/controller -h
Usage of ./bin/controller:
      --burst int              burst of query apiserver. (default 10)
      --enableLeaderElection   If EnableLeaderElection for controller.
      --incluster              If controller run incluster.
      --kubeConfig string      Kube Config path if not run in cluster.
      --kubeconfig string      Paths to a kubeconfig. Only required if out-of-cluster.
      --masterUrl string       Master Url if not run in cluster.
      --metricsAddr string     Metrics server bind listen address. (default ":8080")
      --probeAddr string       Probe endpoint bind  address. (default ":8081")
      --qps int                qps of query apiserver. (default 5)
      --workers int            workers of scheduler-plugin-controllers. (default 1)
```
There are some options which should be discarded after we migrate controller to ctrl runtime, such as `enableLeaderElection`/`kubeConfig`/ `incluster`.
Meanwhile we should also discard options which were never been used, such as `workers`.

/assign

PR #569, opened 2023-03-31
Delete Useless options in controller

czybjtu:
#### What type of PR is this?
/kind cleanup

Fixes https://github.com/kubernetes-sigs/scheduler-plugins/issues/567

```release-note
Delete kubeConfig/incluster/masterUrl options in controller. 
Implentment --worker option.
```

### DIFF
--- a/cmd/controller/app/server.go
+++ b/cmd/controller/app/server.go
@@ -17,22 +17,9 @@ limitations under the License.
 package app
 
 import (
-	"context"
-	"os"
-
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/apimachinery/pkg/util/uuid"
-	"k8s.io/apiserver/pkg/server"
-	"k8s.io/client-go/informers"
-	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/rest"
-	restclient "k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/client-go/tools/leaderelection"
-	"k8s.io/client-go/tools/leaderelection/resourcelock"
-	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -40,8 +27,6 @@ import (
 
 	schedulingv1a1 "sigs.k8s.io/scheduler-plugins/apis/scheduling/v1alpha1"
 	"sigs.k8s.io/scheduler-plugins/pkg/controllers"
-	schedclientset "sigs.k8s.io/scheduler-plugins/pkg/generated/clientset/versioned"
-	schedformers "sigs.k8s.io/scheduler-plugins/pkg/generated/informers/externalversions"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


zwpaper:
hi @czybjtu, thanks for the PR, but maybe we should implement those options instead of deleting them 

WDYT @Huang-Wei ?

cmd/controller/app/server.go
```diff
@@ -129,53 +86,9 @@ func Run(s *ServerRunOptions) error {
...
-		if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
-			setupLog.Error(err, "unable to start manager")
-			panic(err)
-		}
-
-		select {}
-	}
-	schedInformerFactory.Start(stopCh)
-	coreInformerFactory.Start(stopCh)
-	if !s.EnableLeaderElection {
```
Huang-Wei, line 143: this seems to be totally disabling leader election?

Huang-Wei:
> hi @czybjtu, thanks for the PR, but maybe we should implement those options instead of deleting them

Yes, we need to bring the functionality back, which is part of the migration; otherwise, it's a breaking change as params like workers, leader elections are not supported.

@czybjtu are you going to remove and add the param/function back, via 2 PRs? If so, I'm leaning towards merging into one PR, so reviewers only need to review function changes while keep parameters unchanged (or with minimum changes).

cmd/controller/app/server.go
czybjtu: No need to implement this manually, the ctrl runtime can handle this instead. We could enable leader election when create `ctrl.NewManager`

czybjtu:
AFAIK, after we migrate controllers to controller-runtime, there is no need to implement `kubeConfig/incluster/masterUrl`  manually. The controller-runtime can handle this instead.
The `--wokers` option was never be implemented before, So I delete this option in this pr. Maybe it's better to implement it. I'll take a look later.

cmd/controller/app/server.go
czybjtu: https://github.com/kubernetes-sigs/scheduler-plugins/blob/aeb7ff27f5fa6559c588d7a133fe933d1fc65ca6/cmd/controller/app/server.go#L93-L100

zwpaper:
hi @czybjtu, sorry for the late reply again, I was kinda busy recently.

my concern is that:
1. ctrl-runtime automatically added a kubeconfig command line flag to the controller, it may be worth a comment in the Option struct.
2. the `kubeconfig` cmd flag was changed from `kubeConfig`, we should add this breaking change in the release note @Huang-Wei 
3. it seems that ctrl-runtime did not handle the `masterAddr` flag, and it also seems can be replaced by the `kubeconfig` flag, we may also drop it and notice it in the release note?
4. the `inCluster` flag seems to be reasonable to drop, as it can be detected by whether `kubeconfig` is specified.

ps, I only glanced at the PR, will try to find some free cycles to check it deeply.

Huang-Wei:
/ok-to-test

I'm not concerned with the features that didn't exist before, we can keep the behavior as-is as long as we can confirm some parameters didn't work (like masters).

For kubeconfig -> kubeConfig and changes alike, I'm less concerned as well. Let's just stick to ctrl-runtime's naming conversion and write a release note about that.

pkg/controllers/elasticquota_controller.go
```diff
@@ -173,5 +175,6 @@ func (r *ElasticQuotaReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		Watches(&source.Kind{Type: &v1.Pod{}}, &handler.EnqueueRequestForObject{}).
 		For(&schedv1alpha1.ElasticQuota{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: r.Workers}).
```
Huang-Wei, line 178: thanks for adding this.

Huang-Wei:
/approve

Leaving /lgtm to @zwpaper .

zwpaper:
/lgtm

thanks for the contribution @czybjtu!

PR #570, opened 2023-04-04
chart: fix service account name

kd7lxl:
#### What type of PR is this?

/kind bug

The scheduler name, and subsequently the scheduler service account name can be configured by a value, but the deployment still referenced a hardcoded service account name. This resulted in the deployment failing whenever the scheduler name was overridden.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/ac60529f9d8f8d444c952352876571d5b99d49d5/manifests/install/charts/as-a-second-scheduler/templates/serviceaccount.yaml#L4

https://github.com/kubernetes-sigs/scheduler-plugins/blob/ac60529f9d8f8d444c952352876571d5b99d49d5/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml#L41

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
@@ -15,7 +15,7 @@ spec:
       labels:
         app: scheduler-plugins-controller
     spec:
-      serviceAccountName: scheduler-plugins-controller
+      serviceAccountName: {{ .Values.controller.name }}
       containers:
       - name: scheduler-plugins-controller
         image: {{ .Values.controller.image }}
@@ -38,7 +38,7 @@ spec:
       labels:
         component: scheduler
     spec:
-      serviceAccountName: scheduler-plugins-scheduler
+      serviceAccountName: {{ .Values.scheduler.name }}
       containers:
       - command:
         - /bin/kube-scheduler


Huang-Wei:
Strictly speaking, it's not a bug :)

/remove-kind bug
/kind enhancement

Could you also update the release-note section to:

```
Helm: tie controller/scheduler's service account name to their variable name.
```

manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
```diff
@@ -38,7 +38,7 @@ spec:
       labels:
         component: scheduler
     spec:
-      serviceAccountName: scheduler-plugins-scheduler
+      serviceAccountName: {{ .Values.scheduler.name }}
```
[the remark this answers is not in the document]
kd7lxl: The service account's name is currently working, so expanding the scope to parameterize it would make this an enhancement. 😄

zwpaper:
hi @kd7lxl, I have found some time to look into the PR, and I found that you use the `scheduler.name` to ack as the serviceAccountName for the scheduler, but it seems that having a rbac related configuration would be better.

could you please add a `scheduler.serviceAccount.name` to specify the serviceAccountName, like

```yaml
scheduler:
  serviceAccount:
    name:
```

and later, we can raise another PR, or keep it in the one that would also be great, to implement the RBAC related in our chart.

ref: https://helm.sh/docs/chart_best_practices/rbac/

WDYT @Huang-Wei

zwpaper (changes requested):
/kind feature

kd7lxl:
> hi @kd7lxl, I have found some time to look into the PR, and I found that you use the `scheduler.name` to ack as the serviceAccountName for the scheduler, but it seems that having a rbac related configuration would be better.
> 
> could you please add a `scheduler.serviceAccount.name` to specify the serviceAccountName, like
> 
> ```yaml
> scheduler:
>   serviceAccount:
>     name:
> ```
> 
> and later, we can raise another PR, or keep it in the one that would also be great, to implement the RBAC related in our chart.
> 
> ref: https://helm.sh/docs/chart_best_practices/rbac/
> 
> WDYT @Huang-Wei

Sorry, I'm not sure what you're requesting. The chart already has RBAC, and already uses the parameterized name. That's why this was marked /kind bug. The names need to match everywhere for it to work.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/ac60529f9d8f8d444c952352876571d5b99d49d5/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml#L87

I would prefer to keep any enhancements, like the ability to customize the service account name separate from the deployment name, in a separate PR so that this one is limited to the bug fix.

manifests/install/charts/as-a-second-scheduler/templates/deployment.yaml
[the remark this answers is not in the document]
kd7lxl: Updated.

zwpaper:
Oops, sorry for the mistake...

what I say is that we should expose the rbac related in `values.yaml` like it mentioned here https://helm.sh/docs/chart_best_practices/rbac/.

it's ok for this PR to go with the enhancement with this `serviceAccountName`

.github/workflows/chart.yaml
```diff
@@ -0,0 +1,48 @@
...
+          python-version: '3.9'
+          check-latest: true
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.4.0
+
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(ct list-changed --target-branch ${{ github.event.repository.default_branch }})
```
zwpaper, line 33: we may need to specify the `--chart-dirs`

zwpaper (changes requested):

manifests/install/charts/as-a-second-scheduler/ci/default-values.yaml
zwpaper, line 1: why is there 2 empty default yaml?

.github/workflows/chart.yaml
kd7lxl: I set the working directory to achieve this. Need to run it to know if it works this way. Not sure how to trigger a test run in this project.

manifests/install/charts/as-a-second-scheduler/ci/default-values.yaml
kd7lxl: Oops, fixed.

kd7lxl:
I added a chart test with a test case for the name override feature that is currently broken in the default branch. If the test passes, we will know this PR fixes the problem. The test was implemented in Github Actions, and I'm not sure how to run it in this project. It appears the project does not yet utilize Github Actions, but from past experience, that's the easiest tool for implementing chart testing.

Huang-Wei:
> Can someone help?

Now it's running. Let's wait and see.

manifests/install/charts/as-a-second-scheduler/ci/default-values.yaml
Huang-Wei, line 1: this file is empty, is it intended?

.github/workflows/chart.yaml
```diff
@@ -0,0 +1,48 @@
+name: Lint and Test Chart
```
Huang-Wei, line 1: nit: maybe adding "helm" to it? like "Helm Chart Lint and Test"

Huang-Wei:
Thanks @kd7lxl for the Helm testing bits. It runs successfully.

To keep a linear git history, would you mind squashing the commits into two: one is core changes, the other for helm CI?

manifests/install/charts/as-a-second-scheduler/ci/default-values.yaml
kd7lxl: Yes, it tells `ct` to run a test case with the default values (no overrides).

kd7lxl:
> > Can someone help?
> 
> Now it's running. Let's wait and see.

Hmm, it looks like it [ran](https://github.com/kubernetes-sigs/scheduler-plugins/actions/runs/4631296377/jobs/8195809018?pr=570), but skipped the test cases. Need to figure that out.

zwpaper:
> Hmm, it looks like it [ran](https://github.com/kubernetes-sigs/scheduler-plugins/actions/runs/4631296377/jobs/8195809018?pr=570), but skipped the test cases. Need to figure that out.

maybe you need to define the output, and `need` dependency first?
https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs

Huang-Wei:
@kd7lxl if it's too much work, I'm fine with only focus on parameterize the service account in this PR, and follow up the CI bits in another PR. But it's up to you :)

kd7lxl:
> To keep a linear git history, would you mind squashing the commits into two: one is core changes, the other for helm CI?

> @kd7lxl if it's too much work, I'm fine with only focus on parameterize the service account in this PR, and follow up the CI bits in another PR. But it's up to you :)

Sorry, was away from the computer for a few days. Squashed and rebased now to just the service account name fix. I'll open another PR to add chart tests.

PR #554, opened 2023-03-24
Cleanup PostBind in coscheduling

czybjtu:
#### What type of PR is this?
/kind cleanup

#### What this PR does / why we need it:
this logic has been moved to PodGroup's controller.
#### Which issue(s) this PR fixes:

Fixes #https://github.com/kubernetes-sigs/scheduler-plugins/issues/531

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #574, opened 2023-04-10
Coscheduling: remove podgroup.scheduled

### LINKED ISSUE #562: PodGroupStatus.Scheduled is not updated due to the removal of PostBind
The removal of the [PostBind](https://github.com/kubernetes-sigs/scheduler-plugins/pull/554) function for `PodGroups` has caused the `podgroupstatus.scheduled` field to no longer update and remain at 0. 
After the [PR of integration test for `PodGroups`](https://github.com/kubernetes-sigs/scheduler-plugins/pull/549) is merged, the `podgroupstatus.scheduled` field can be safely removed as it will no longer be necessary.

Gekko0114:
#### What type of PR is this?
/kind bug
/kind cleanup

Optionally add one or more of the following kinds if applicable:
/kind api-change

The removal of the [PostBind](https://github.com/kubernetes-sigs/scheduler-plugins/pull/554) function for `PodGroups` has caused the `PodGroupStatus.Scheduled` field to no longer update and remain at 0.

I have removed the field `PodGroupStatus.Scheduled` as it is no longer required. Additionally, I have also removed unnecessary statuses from the `PodGroupPhase`.

Fixes #562 , #578

Please let me know if I have made any mistakes or if there are any improvements that could be made.
Your comments and feedback are appreciated.

```release-note
Remove podGroupStatus.Scheduled.
Remove two podGroupPhase (podGroupPreScheduling and podGroupScheduled)
```

### DIFF
--- a/kep/42-podgroup-coscheduling/README.md
+++ b/kep/42-podgroup-coscheduling/README.md
@@ -9,7 +9,7 @@
 - [Use Cases](#use-cases)
 - [Design Details](#design-details)
   - [PodGroup](#podgroup)
-	- [PodGroupPhase](#podgroupphase)
+    - [PodGroupPhase](#podgroupphase)
   - [Controller](#controller)
   - [Extension points](#extension-points)
     - [QueueSort](#queuesort)
--- a/kep/42-podgroup-coscheduling/README.md
+++ b/kep/42-podgroup-coscheduling/README.md
@@ -118,7 +118,7 @@ stateDiagram-v2
     Pending --> Scheduling: minMember pods are scheduling
     Scheduling --> Running: minMember pods are running
     Running --> Failed: at least one of the pods failed
-    Failed --> if_minMember: failed fixed
+    Failed --> if_minMember: failed resolved
     if_minMember --> Scheduling: minMember pods are scheduling
     if_minMember --> Running: minMember pods are running
     Running --> Finished: all pods successfully finished
--- a/kep/42-podgroup-coscheduling/README.md
+++ b/kep/42-podgroup-coscheduling/README.md
@@ -9,13 +9,13 @@
 - [Use Cases](#use-cases)
 - [Design Details](#design-details)
   - [PodGroup](#podgroup)
+	- [PodGroupPhase](#podgroupphase)
   - [Controller](#controller)
   - [Extension points](#extension-points)
     - [QueueSort](#queuesort)
     - [PreFilter](#prefilter)
     - [PostFilter](#postfilter)
     - [Permit](#permit)
-    - [PostBind](#postbind)
   - [Known Limitations](#known-limitations)
 <!-- /toc -->
 
@@ -64,10 +64,6 @@ type PodGroupStatus struct {
 
 	// The number of actively running pods.
 	// +optional
-	Scheduled uint32 `json:"scheduled"`
-
-	// The number of actively running pods.
-	// +optional
 	Running uint32 `json:"running"`
 
 	// The number of pods which reached phase Succeeded.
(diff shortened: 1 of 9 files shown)


pkg/controllers/podgroup_controller_test.go
```diff
@@ -146,47 +147,49 @@ func Test_Run(t *testing.T) {
...
 			previousPhase:     v1alpha1.PodGroupRunning,
-			desiredGroupPhase: v1alpha1.PodGroupPending,
+			desiredGroupPhase: v1alpha1.PodGroupRunning,
 		},
 	}
 	for _, c := range cases {
 		t.Run(c.name, func(t *testing.T) {
 			controller, kClient := setUp(ctx, c.podNames, c.pgName, c.podPhase, c.minMember, c.previousPhase, c.podGroupCreateTime, nil)
 			// 0 means not set
+			ps := makePods(c.podNames, c.pgName, c.podPhase, nil)
```
Gekko0114, line 163: This test code didn't work properly. I've Fixed it

pkg/controllers/podgroup_controller_test.go
```diff
@@ -146,47 +147,49 @@ func Test_Run(t *testing.T) {
...
 			if len(c.podNextPhase) != 0 {
```
Gekko0114, line 164: This test code didn't work properly if test case didn't have podNextPhase. I've Fixed it

zwpaper:
hi @Gekko0114, thanks for raising the PR, but I have reconsidered the scheduling status flow and found that there may be something we need to discuss before deleting the phases.

1. there are some cases making the PodGroup partially running, like some pods in the group failed and were rescheduling, then the PodGroup should be in a Scheduling status
2. in the case like `1.`, we have a running count less than expected, and would get some scheduling counts, so, the scheduled number could mean something?
3. we should be fine to drop the PostBind related, but I will raise an issue to discuss the schedule flow before deleting the status

cc @Huang-Wei WDYT?

zwpaper:
https://github.com/kubernetes-sigs/scheduler-plugins/issues/578
created a issue

cc @Huang-Wei

Gekko0114:
Sure. Let's discuss this topic on the issue #578 . Thanks,

Gekko0114:
@zwpaper, @denkensk
Based on the [discussion](https://github.com/kubernetes-sigs/scheduler-plugins/issues/578), I've updated this PR.
Could you take a look?

Gekko0114:
@zwpaper, @denkensk
Could you take a look?

zwpaper:
Sorry for the delay, but I was swamped whose days both personal and work, it may take some more time to be accessible, I will try to find some time to have a look at this recently

Gekko0114:
No problem, thank you so much.

pkg/controllers/podgroup_controller_test.go
```diff
@@ -159,34 +160,36 @@ func Test_Run(t *testing.T) {
 		t.Run(c.name, func(t *testing.T) {
 			controller, kClient := setUp(ctx, c.podNames, c.pgName, c.podPhase, c.minMember, c.previousPhase, c.podGroupCreateTime, nil)
 			// 0 means not set
+			ps := makePods(c.podNames, c.pgName, c.podPhase, nil)
```
zwpaper, line 163: this line should be in front of L162, as the comment is describing `len(c.podNextPhase) != 0`

pkg/controllers/podgroup_controller_test.go
zwpaper: Thanks for the update, it should fix the cases without `podNextPhase`.

but on second thought, the controller should always focus on the next phase, and it does not care about the previous pod phase, so we may have to update the test cases to adopt this.

it should be ok for this PR, and we will work on the testing later in a separate PR.

zwpaper (changes requested):
hi @Gekko0114, I have done the review, the PR is great, and only some small nit.

I will test the logic locally and then we will be good

pkg/controllers/podgroup_controller_test.go
```diff
@@ -85,20 +86,20 @@ func Test_Run(t *testing.T) {
...
 			podNames:          []string{"pod1", "pod2"},
 			podPhase:          v1.PodPending,
-			previousPhase:     v1alpha1.PodGroupScheduling,
-			desiredGroupPhase: v1alpha1.PodGroupScheduled,
+			previousPhase:     v1alpha1.PodGroupPending,
+			desiredGroupPhase: v1alpha1.PodGroupScheduling,
 		},
 		{
-			name:              "Group status convert from scheduling to succeed",
+			name:              "Group status convert from prescheduling to finished",
```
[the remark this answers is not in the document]
Gekko0114: It's typo, thanks!

pkg/controllers/podgroup_controller_test.go
Gekko0114: > This test code didn't work properly if test case didn't have podNextPhase. I've Fixed it

To be precise, in the test case without `podNextPhase`, we weren't testing anything.

> but on second thought, the controller should always focus on the next phase, and it does not care about the previous pod phase, so we may have to update the test cases to adopt this.

I agree. this test has `previousPhase`, `podPhase` and `podNextPhase` fields. It might be redundant.

> it should be ok for this PR, and we will work on the testing later in a separate PR.

Sure. Thanks for your review!

Gekko0114:
Hi @zwpaper,

Thanks for your review! I've fixed it. Could you have a look?

zwpaper:
Hi, @Gekko0114 really sorry for taking so long, I am back to Open Source finally.

I have test the scheduler locally, and with one more tiny request,

please also remove the postBind in README so that users can follow the readme to start the scheduler without meeting error:
https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/coscheduling/README.md?plain=1#L78-L80

and then we would be good to go!

Gekko0114:
Hi @zwpaper,

Thanks for your comment! I missed README. I've fixed it

kep/42-podgroup-coscheduling/README.md
```diff
@@ -83,6 +79,53 @@ type PodGroupStatus struct {
...
+```
+
+```mermaid
+stateDiagram-v2
+	state if_minMember <<choice>>
+    [*] --> Pending
+    Pending --> Scheduling: minMember pods are scheduling
+    Scheduling --> Running: minMember pods are running
+    Running --> Failed: at least one of the pods failed
+    Failed --> if_minMember: failed fixed
```
Huang-Wei, line 121:
```suggestion
    Failed --> if_minMember: failure resolved
```

Huang-Wei (commented):
LGTM. just  one nit on the diagram text.

And could you detail the changes in release-notes? (as the changes involved removal of two phases and one field in status)

/hold

kep/42-podgroup-coscheduling/README.md
Gekko0114: Thanks! Fixed it

Gekko0114:
@Huang-Wei ,

Thanks for your comments! I fixed it. Could you have a look again?

Huang-Wei:
/lgtm
/approve
/label tide/merge-method-squash

Thanks @Gekko0114 !

Gekko0114:
Hi @Huang-Wei 
Thanks!
This PR has hold tag.
Could you remove it if it's ok?

Huang-Wei:
Ah sure.

/unhold

PR #580, opened 2023-04-26
nrt: overreserve: minimize the pod listings

ffromani:
#### What type of PR is this?

/kind cleanup

In https://github.com/kubernetes-sigs/scheduler-plugins/pull/557 we replaced the custom indexer implementation with the simpler lister implementation.
The relisting works, but does unnecessary work relisting for each node which is processed, which is wasteful: even though client-go is optimized, let's avoid useless work, because we can relist the pods once per resync attempt.

Fixes N/A

N/A

```release-note
NONE
```

### DIFF
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -191,6 +191,13 @@ func (ov *OverReserve) Resync() {
 		return
 	}
 
+	// node -> pod identifier (namespace, name)
+	nodeToObjsMap, err := makeNodeToNamespacedNamesMap(ov.podLister, logID)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -206,6 +213,13 @@ func (ov *OverReserve) Resync() {
 			continue
 		}
 
+		objs, ok := nodeToObjsMap[nodeName]
+		if !ok || len(objs) == 0 {
```
PiotrProkop, line 217: `if !ok` shouldn't be enough?

ffromani: yes, and more idiomatic. I'll fix in the next upload.

PR #576, opened 2023-04-13
WIP: nodetopologymatch: improve and deflake integration tests

ffromani:
#### What type of PR is this?

/kind cleanup
/kind flake

#### What this PR does / why we need it:
Improve and deflake the NodeTopologyMatch integration tests

#### Which issue(s) this PR fixes:
Fixes #573 

#### Special notes for your reviewer:
WIP: no obvious reason yet for the flakes, I'll iterate over this PR as needed

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

PR #590, opened 2023-05-23
nodetopologymatch: allow to track only pods with exclusive resources

ffromani:
#### What type of PR is this?

/kind cleanup

Optimize both the foreign pods detection and the overreserve resync loop considering only pods which have exclusive resources assigned (vs considering all of them) because only exclusive resources have NUMA affinity, so only those contribute to per-NUMA accounting. This reduces churn, noise and computing power required. This is also expected to improve the scheduling performance by reducing the time for the resync loop to converge

Fixes N/A

```release-note
nodetopologymatch: track only pods with exclusive resources. Reduces churn and improves the resync latency.
```


ffromani:
/hold

1. still WIP
2. want to make more progress on https://github.com/kubernetes-sigs/scheduler-plugins/pull/576 before

ffromani:
/retest

looks like a infra glitch

ffromani:
/hold cancel

the PR is reviewable

pkg/noderesourcetopology/cache/foreign_pods.go
```diff
@@ -72,7 +74,7 @@ func IsForeignPod(pod *corev1.Pod) bool {
 		// nothing to do here - we know already about this pod
 		return false
 	}
-	return true
+	return resourcerequests.AreExclusiveForPod(pod)
```
ffromani, line 77: I need to doublecheck how this plays out with PFP compute method "all"

ffromani:
/hold

see inline comment

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
```diff
@@ -0,0 +1,176 @@
...
+									corev1.ResourceName("veryfast.io/fpga"): resource.MustParse("1"),
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedExclusive: true,
+			expectedNonNative: true,
+		},
```
PiotrProkop, line 174: I think we should add more complex tests, like mixing devs with native resources to verify that if container has non native resource it's still exclusive.

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return false
+}
+func AreExclusiveForPod(pod *corev1.Pod) bool {
+	qos := v1qos.GetPodQOS(pod)
+	return (areExclusiveForAnyContainer(qos, pod.Spec.InitContainers) ||
+		areExclusiveForAnyContainer(qos, pod.Spec.Containers))
+}
+
+func areExclusiveForAnyContainer(qos corev1.PodQOSClass, containers []corev1.Container) bool {
```
PiotrProkop, line 49: I am wondering if something like this wouldn't be more readeable:
```suggestion
	return areExclusiveForAnyContainer(qos, append(pod.Spec.InitContainers, pod.Spec.Containers...))
}

func areExclusiveForAnyContainer(qos corev1.PodQOSClass, containers []corev1.Container) bool {
```
wdyt?

pkg/noderesourcetopology/cache/store.go
```diff
@@ -198,37 +198,48 @@ func (cnt counter) Len() int {
 
 // podFingerprintForNodeTopology extracts without recomputing the pods fingerprint from
 // the provided Node Resource Topology object.
```
PiotrProkop, line 200: I would modify comment to explain what is second return value

PiotrProkop (commented):
small nits, otherwise looks good.

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
ffromani: yes, I need to add more tests here. Luckily various scenarios are very easy to test. So let me add more cases, this is too minimal.

pkg/noderesourcetopology/cache/store.go
ffromani: right! fixed

ffromani:
back to WIP as I revisit the approach and address reviewer's comments

pkg/noderesourcetopology/cache/foreign_pods.go
ffromani: it will addressed in the next upload. Spoiler: adding a config option to let the user control the behavior. This seems to be the only meaningful way to interact with the other changes in this PR.

pkg/noderesourcetopology/resourcerequests/exclusive.go
ffromani: I'm still thinking about this one (last pending comment). On one hand I don't like too much the temporary variable, on the other hand the way you suggest is pretty much the standard we use all across the codebase (including TM)

ffromani: Let's go with the flow:
```
1020 16:39:47 fromani@laptop ~/go/src/k8s.io/kubernetes $ git log --oneline -1
11067ceb328 (HEAD -> master, upstream/master, origin/master, origin/HEAD) Merge pull request #118306 from SataQiu/fix-kubeadm-20230529
1021 16:39:50 fromani@laptop ~/go/src/k8s.io/kubernetes $ git grep 'append(pod.Spec.Init'
pkg/kubelet/cm/cpumanager/cpu_manager.go:373:           for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/cpumanager/cpu_manager_test.go:183:              pod.Spec.InitContainers = append(pod.Spec.InitContainers, v1.Container{
pkg/kubelet/cm/cpumanager/policy_static.go:542: for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/cpumanager/policy_static_test.go:697:            for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/memorymanager/memory_manager.go:327:             for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/memorymanager/policy_static.go:375:      for _, ctn := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/topologymanager/scope.go:137:    for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/topologymanager/scope_container.go:53:   for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/cm/topologymanager/scope_pod.go:60: for _, container := range append(pod.Spec.InitContainers, pod.Spec.Containers...) {
pkg/kubelet/status/status_manager_test.go:708:                  pod.Spec.InitContainers = append(pod.Spec.InitContainers, v1.Container{
plugin/pkg/admission/limitranger/admission_test.go:168:         pod.Spec.InitContainers = append(pod.Spec.InitContainers, api.Container{
plugin/pkg/admission/serviceaccount/admission.go:390:                   pod.Spec.InitContainers[i].VolumeMounts = append(pod.Spec.InitContainers[i].VolumeMounts, volumeMount)
```

ffromani:
/hold

trying to avoid premature merges

ffromani:
currently hitting: https://github.com/kubernetes/test-infra/issues/29622

ffromani:
/hold cancel

all issues seems to be gone

pkg/noderesourcetopology/cache/store.go
```diff
@@ -197,38 +198,57 @@ func (cnt counter) Len() int {
 }
 
 // podFingerprintForNodeTopology extracts without recomputing the pods fingerprint from
-// the provided Node Resource Topology object.
-func podFingerprintForNodeTopology(nrt *topologyv1alpha2.NodeResourceTopology) string {
+// the provided Node Resource Topology object. Returns the expected fingerprint and the method to compute it.
+func podFingerprintForNodeTopology(nrt *topologyv1alpha2.NodeResourceTopology, method apiconfig.CacheResyncMethod) (string, apiconfig.CacheResyncMethod) {
```
PiotrProkop, line 202: maybe change signature to:
```go
func podFingerprintForNodeTopology(nrt *topologyv1alpha2.NodeResourceTopology, method apiconfig.CacheResyncMethod) (string, bool) {
```
and the second return param will be `wantsOnlyExclRes`.
Then when passing to `checkPodFingerprintForNode` we can pass only boolean and not `pfpMethod`

ffromani: good point. That was the original version of the code. I'm still torn between YAGNI and keep the signature simple (and don't leak `apiconfig` too much all around the codebase) and be generic and forward looking.
Now that you hihglight this, I'm leaning towards reverting to YAGNI and implement your suggestion

PiotrProkop (commented):
/lgtm

Issue #475, opened 2023-01-07
[proposal] add a schedule plugin that support pod expands and shrinks according to the order of the defined  logical node set 


fjding:
The kubernetes support pod-deletion-cost after v1.21， in my cloud scene， the user have the demands like this：
1、Define multiple  logical node set, a deployment workload can schedule pods according to this node set order, and shrink in the opposite order at the same time. 
2、At the same time, it also supports the maximum number of schedulable pods per node set。
BTW, I have implemented this feature and want to contribute it to the community, I hope everyone can discuss it together

KunWuLuan:
My company also have a similar plugin like yours. We can find a time to have a discuss.

fjding:
> My company also have a similar plugin like yours. We can find a time to have a discuss.

Hi,  we can collaborate on this proposal

fjding:
@Huang-Wei @ffromani @seanmalloy @denkensk 
Could you take a look at this proposal? We can discuss whether we need to create a keps

ffromani:
> @Huang-Wei @ffromani @seanmalloy @denkensk Could you take a look at this proposal? We can discuss whether we need to create a keps

I'll have a look later this week (beginning April 3 2023)

Huang-Wei:
It will help us understanding the motivation(s) if you can elaborate on the real-world use-cases.

---

> 1、Define multiple logical node set, a deployment workload can schedule pods according to this node set order, and shrink in the opposite order at the same time.

What do you mean by "node set order"? is that a priority field of the NodeSet CR?

How a deployment's replicas are expected to be scheduled onto the matching NodeSets? and is the scheduling directives a hard or soft constraint?

> 2、At the same time, it also supports the maximum number of schedulable pods per node set

Where is this max num defined?

fjding:
> It will help us understanding the motivation(s) if you can elaborate on the real-world use-cases.
> 
> > 1、Define multiple logical node set, a deployment workload can schedule pods according to this node set order, and shrink in the opposite order at the same time.
> 
> What do you mean by "node set order"? is that a priority field of the NodeSet CR?
> 
> How a deployment's replicas are expected to be scheduled onto the matching NodeSets? and is the scheduling directives a hard or soft constraint?
> 
> > 2、At the same time, it also supports the maximum number of schedulable pods per node set
> 
> Where is this max num defined?

Hi, Thank you for your attention！
The motivation：
[...]

PR #594, opened 2023-05-25
[proposal] add resource policy plugin

KunWuLuan:
/kind feature

Fixes #

```release-note

```


KunWuLuan:
@denkensk @Huang-Wei Hi, if you have time, you can help to review the kep, thanks very much. 😆

ffromani:
/ok-to-test

It's a bit of a crazy time for me but I'll try to also add my (non-binding) review

Huang-Wei:
I will try to review it by this week.

KunWuLuan:
Hi, @Huang-Wei , This PR is ready for review. PTAL. Thanks for your time. : )

KunWuLuan:
There is a known issue, when the number of nodes in cluster is larger than 100, scheduler will not find all feasible nodes unless `percentageOfNodesToScore` is set to 100. This can make resource policy lose efficacy.

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,144 @@
...
+## Motivation
+The machines in a Kubernetes cluster are typically heterogeneous, with varying CPU, memory, GPU, and pricing. To efficiently utilize the different resources available in the cluster, users can set priorities for machines of different types and configure resource allocations for different workloads. Additionally, they may choose to delete pods running on low priority nodes instead of high priority ones. 
+
+### Use Cases
+
+1. As a user of cloud services, there are some stable but expensive ECS instances and some unstable but cheaper Spot instances in my cluster. I hope that my workload can be deployed first on stable ECS instances, and during business peak periods, the Pods that are scaled out are deployed on Spot instances. At the end of the business peak, the Pods on Spot instances are prioritized to be scaled in.
+
+### Goals
+
+1. Delvelop a filter plugin to restrict the resource consumption on each unit for different workloads.
```
ffromani, line 32: typo: develop

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,144 @@
...
+PreFilter check if the current pods match only one resource policy. If not, PreFilter will reject the pod.
+If yes, PreFilter will get the number of pods on each unit to determine which units are available for the pod
+and write this information into cycleState.
+
+##### Filter
+Filter check if the node belongs to an available unit. If the node doesn't belong to any unit, we will return
+success if the strategy is `prefer`, otherwise we will return unschedulable.
+
+##### Score
+If `priority` and `weight` is set in resource policy, we will schedule pod based on `priority` first. For units with the same `priority`, we will spread pods based on `weight`.
```
ffromani, line 114: this may be a silly question, but hey.
So it seems to me that the actual scoring criteria is based on the tuple (priority, weight). Is there any reason to use two values instead of just one single value?
IOW, would it work if we use just `proiority`? What are the pros and cons of using 2 values here?

ffromani (commented):
initial pass (sorry for the long delay)
at glance looks sensible, will be asking questions to fully grasp the proposal

kep/594-resourcepolicy/README.md
KunWuLuan: I realized that this is a modification that I missed when removing weight earlier. I have already updated the kep

KunWuLuan:
@ffromani @Huang-Wei Hi, do you have any other questions?

ffromani:
The main blocker atm is that the KEP template is not fully filled (e.g.  PostFilter - do we need it? Do we have Graduation criterias? Do we need Production Readiness? [probably not] We probably need a few words in Feature enablement and rollback though)

From what I've read so far I have no major objections. About the architecture and the fitness of this plugin in the existing ecosystem I'd have to defer to someone more experience in the scheduling.

KunWuLuan:
Have update the kep

KunWuLuan:
for #475

KunWuLuan:
@ffromani @Huang-Wei Hi, please have a look when u have time. Thanks for your time. : )

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+    - [Resource Policy Controller](#resource-policy-controller)
+- [Known limitations](#known-limitations)
+- [Test plans](#test-plans)
+- [Graduation criteria](#graduation-criteria)
+- [Feature enablement and rollback](#feature-enablement-and-rollback)
+<!-- /toc -->
+
+## Summary
+This proposal introduces a plugin to allow users to specify the priority of different resources and max resource 
+consumption for workload on differnet resources.
```
Huang-Wei, line 27:
```suggestion
This proposal introduces a plugin that enables users to set priorities for various resources and define maximum resource consumption limits for workloads across different resources.
```

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+### Use Cases
+
+1. As a user of cloud services, there are some stable but expensive ECS instances and some unstable but cheaper Spot 
+instances in my cluster. I hope that my workload can be deployed first on stable ECS instances, and during business peak 
+periods, the Pods that are scaled out are deployed on Spot instances. At the end of the business peak, the Pods on Spot 
+instances are prioritized to be scaled in.
+
+### Goals
+
+1. Develop a filter plugin to restrict the resource consumption on each unit for different workloads.
```
Huang-Wei, line 44: could you elaborate on "unit"? do you mean one kind of resource?

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+2. Develop a score plugin to favor nodes matched by a high priority unit.
+3. Automatically setting deletion costs on Pods to control the scaling in sequence of workloads through a controller.
+
+### Non-Goals
+
+1. Modify the workload controller to support deletion costs. If the workload don't support deletion costs, scaling in 
+sequence will be random.
+2. When creating a ResourcePolicy, if the number of Pods has already violated the quantity constraint of the
```
Huang-Wei, line 52: could you reword this to be more readable? basically, just use a positive tone to describe it as a goal that is out of scope of this KEP.

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+        - value2
+  - name: unit3
+    priority: 4
+    maxCount: 20
+    nodeSelector:
+      matchLabels:
+        key1: value3
+```
+
+`Priority` define the priority of each unit. Pods will be scheduled on units with a higher priority.
```
Huang-Wei, line 108: is `Prirority` is under `units[*]`? if so, it'd read better to use a bullet to explain it under a top-level `units` API field.

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+    nodeSelector:
+      matchLabels:
+        key1: value3
+```
+
+`Priority` define the priority of each unit. Pods will be scheduled on units with a higher priority. 
+If all units have the same priority, resourcepolicy will only limit the max pod on these units.
+
+`Strategy` indicate how we treat the nodes doesn't match any unit. 
+If strategy is `required`, the pod can only be scheduled on nodes that match the units in resource policy.
```
Huang-Wei, line 112: In a multi-tenant cluster, given ResourcePolicy is a namespace-scoped CR, but nodes/node pools might be shared across tenants. Does it mean that if a tenant set a high-priority policy with strategy `required`, it could make other tenants who don't set or set low-priority policy get less chances to get their workloads scheduled?

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+If strategy is `prefer`, the pod can be scheduled on all nodes, these nodes not match the units will be 
+considered after all nodes match the units. So if the strategy is `required`, we will return `unschedulable` 
+for those nodes not match the units.
+
+`MatchLabelKeys` indicate how we group the pods matched by `podSelector` and `matchPolicy`, its behavior is like 
+`MatchLabelKeys` in `PodTopologySpread`.
+
+`matchPolicy` indicate if we should ignore some kind pods when calculate pods in certain unit.
+
+If `forceMaxNum` is set `true`, we will not try the next units when one unit is not full, this property have no effect
```
Huang-Wei, line 122: Could you reword this sentence a bit?

kep/594-resourcepolicy/README.md
```diff
@@ -0,0 +1,191 @@
...
+when `max` is not set in units.
+
+### Implementation Details
+
+#### Scheduler Plugins
+
+For each unit, we will record which pods were scheduled on it to prevent too many pods scheduled on it.
+
+##### PreFilter
```
Huang-Wei, line 131: Let’s avoid using `#####` for paragraph indentation. The maximum indentation level should be `####` in most cases.

kep/594-resourcepolicy/README.md
Huang-Wei: I think we missed a detailed explanation for `units`.

KunWuLuan: Updated the kep. Thanks

kep/594-resourcepolicy/README.md
KunWuLuan: Rewrote this sentence. Pods will be matched by the ResourcePolicy in same namespace when the `.spec.podSelector`. And if `.spec.matchPolicy.ignoreTerminatingPod` is `true`, pods with Non-Zero `.spec.deletionTimestamp` will be ignored.

kep/594-resourcepolicy/README.md
KunWuLuan: Ok, no problem.

Huang-Wei:
@KunWuLuan I guess the toc needs a refresh. Could you re-run `hack/update-toc.sh`?

13567436138:
where is implemention？it has been a year ago。

KunWuLuan:
I will give an implementation recently

PR #596, opened 2023-06-15
chore: pkg imported more than once

testwill:
#### What type of PR is this?
pkg is being imported more than once 
/kind cleanup

pkg is being imported more than once 
Fixes #

NONE
```release-note

```


zwpaper:
good catch! thanks

/release-note-none
/ok-to-test

PR #599, opened 2023-06-30
nrt: cache: add support for dedicated informer

### LINKED ISSUE #598: [NodeResourceTopology] the plugin should be able to list pods in terminal state
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

A key feature of the NodeResourceTopology plugin is the pessimistic caching, described in detail [here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/v0.25.7/pkg/noderesourcetopology/cache/docs/NUMA-aware-scheduler-side-reserve-plugin.md)

In a nutshell, to enable proper resync the scheduler needs to periodically evaluate the set of pods running on a node.
Here "running" however doesn't mean "running" in kubernetes sense (actually doing work) but in general "existing with resources assigned" because these pods consume resources, which can't thus be allocated to other containers until the pods are cleaned.

This includes pods in terminal state which are not immediately cleaned up. Kubernetes jobs are the main way to get in this state.
[The scheduler framework explicitly filters out such pods on its watcher](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/scheduler.go#L505). Thus, the scheduler side will never see pod in terminal phase.

On the node side, the kubelet reports through the podresources API pods in terminal phase until they are cleaned up by the system. The correctness of this behavior is being investigated **also** on kubelet side.

To the best of my knowledge, this is the only plugin which requires listing pod in terminal phase.
If so, it seems too risky and invasive to propose any scheduler framework change just to accomodate this need.
The best option seems to optionally allow this plugin to open a separate listwatch (through a new informer probably) to the apiserverver, allowing consumers of the plugin to choose between extra load to the apiserver (albeit arguably limited and predictable) or accepting the risk of scheduler misalignment.

1. deploy nfd-topology-updater to provide NodeResourceTopology data (or also https://github.com/k8stopologyawareschedwg/resource-topology-exporter or any other equivalent tool)
2. deploy a scheduler with a profile with NodeResourceTopology **and the cache** enabled (filter+reserve+cacheResyncPeriod option)
3. run a trivial job which just do "echo something" or sleeps for 0.1s and spans all the node 
4. observe all the pods managed by the profile with NodeResourceTopology **and the cache** enabled stay "Pending" forever, the plugin filters out all the nodes with "invalid topology data"

The NodeResourceTopology is meant to tolerate _temporary_ misalignment and _temporarily_ filter out nodes on which the misalignment is detected with "invalid node topology". The problem here is there is no chance to converge ever because the node and the scheduler will observe different set of pods which can never converge, because the scheduler filters out pods in terminal state, which are relevant in this case (and probably _only_ in this case)

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
# Any version >= 1.25
```

</details>

### Scheduler Plugins version

<details>
Any version with the scheduler cache enabled (so >= 0.25.7)
</details>

ffromani:
#### What type of PR is this?

/kind bug
/kind api-change

add optional dedicated informer for the NodeResourceTopology scheduler plugin.
For its reconciliation mechanism to converge, the plugin needs to list pods in terminal phase which are not deleted yet

Fixes #598 

Technically API change but it is backward compatible (optional fields) and done like many others we did in the past. Still labeled for transparency

```release-note
NONE
```


ffromani:
@Huang-Wei @dankensk PTAL when you have time. Would be adding a separate, optional, disabled by default informer acceptable? (please check the related issue for context) I'm also investigating from kubelet side about if/how terminal pods should be reported.

pkg/noderesourcetopology/cache/store.go
```diff
@@ -254,10 +254,6 @@ func makeNodeToPodDataMap(podLister podlisterv1.PodLister, logID string) (map[st
 		return nodeToObjsMap, err
 	}
 	for _, pod := range pods {
-		if pod.Status.Phase != corev1.PodRunning {
-			// we are interested only about nodes which consume resources
-			continue
-		}
```
ffromani, line 260: this is because:

- shared informer: we get non-terminal phase pods anyway, the filter will always pass, so we can remove it
- dedicated informer: we DO WANT also pods in terminal phase, so this filter does more harm than good

when/if moving to non-POC PR (if the overall direction is OK) I'll document the above moving the change in its own commit conveying this explanation

ffromani:
@Huang-Wei @zwpaper since I mentioned this PR when reviewing #655, any design objections from you folks? note this is still on hold as we're still checking if the fix is better done on kubelet or not. But would this PR raise concern from the scheduling side?

ffromani:
> @Huang-Wei @zwpaper since I mentioned this PR when reviewing #655, any design objections from you folks? note this is still on hold as we're still checking if the fix is better done on kubelet or not. But would this PR raise concern from the scheduling side?

gentle ping :) no major objections I guess?

ffromani:
assuming no design blockers, @Tal-or @swatisehgal PTAL
cc @PiotrProkop for awareness

ffromani:
/hold

to give time to ppl to review

pkg/noderesourcetopology/cache/store.go
Tal-or: But what about other phases? will we get pods in `PodPending` and `PodUnknown` phases?

pkg/noderesourcetopology/podprovider/podprovider.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func NewFromHandle(handle framework.Handle, cacheConf *apiconfig.NodeResourceTopologyCache) (k8scache.SharedIndexInformer, podlisterv1.PodLister) {
+	dedicated := WantsDedicatedInformer(cacheConf)
+	if !dedicated {
+		podHandle := handle.SharedInformerFactory().Core().V1().Pods() // shortcut
+		return podHandle.Informer(), podHandle.Lister()
+	}
+
+	podInformer := coreinformers.NewFilteredPodInformer(handle.ClientSet(), metav1.NamespaceAll, 0, cache.Indexers{}, nil)
```
Tal-or, line 51: Sorry for my ignorant but how this custom informers knows to provide pods in terminal phase as well?

ffromani: because by default the informers watch objects without filtering. However,  [the scheduler framework explicitly filters out such pods on its watcher](https://github.com/kubernetes/kubernetes/blob/master/pkg/scheduler/scheduler.go#L31). Thus, the scheduler side will never see pod in terminal phase if it uses the framework-provided informer.

swatisehgal (commented):
The overall direction seems reasonable to me. Will do a proper review  in the coming days.

>I'm also investigating from kubelet side about if/how terminal pods should be reported.

Any conclusion on this? Perhaps we should also mention that fixing this in Kubelet is a possible alternative in the PR description. Do we want to position this in a way that the plan is to fix this in Kubelet in the long run, and this is a workaround in the short term? Can we add more details about that in the PR description.

zwpaper:
hi @ffromani sorry for the late reply, I was really busy on those days.

I was not able to get some time to look into the changes, it would be better for other reviewers to take it.

ffromani:
gentle ping @swatisehgal @Tal-or

swatisehgal (commented):
Thanks for your patience and your work on this!

I am happy to proceed with the current approach as the introduction of informer mode is a nice addition to current implementation and provides us a way to resolve the inconsistency in the pod data obtained from podresource API and that visible to scheduler. 

But, I do feel that in the long term resolving this on the kubelet side would be a more appropriate fix as exposing resources of pods that are terminal does not seem appropriate to me. Irrespective of how we go about things from kubelet point of view, I think we are a way forward to handle this for now as well as in the future with some minor adjustments to default mode if needed.

I left comments suggesting a minor addition to tests and why IMO we should reevaluate the current default value of the mode.

Tal-or:
Thank you for working on this, LGTM (not adding official label since I see there're other comments)

ffromani:
> Thanks for your patience and your work on this!
> 
> I am happy to proceed with the current approach as the introduction of informer mode is a nice addition to current implementation and provides us a way to resolve the inconsistency in the pod data obtained from podresource API and that visible to scheduler.
> 
> But, I do feel that in the long term resolving this on the kubelet side would be a more appropriate fix as exposing resources of pods that are terminal does not seem appropriate to me. Irrespective of how we go about things from kubelet point of view, I think we are a way forward to handle this for now as well as in the future with some minor adjustments to default mode if needed.

Thank you for the valuable feedback. You raise (and raised, now that I remember) a great point that makes me still torn about this approach. The problem is indeed the represantation of completed pod.

The core issue is: a completed (either failed or succeeded) pod should accounted for? IOW, should resources accounted to completed pods be reused if need arises? There are valid points for both approaches.

What is more and more evident, anyway is the fact I'm still ruminating about the merit of the approach. This on itself is a strong indicator we should NOT bake in this workaround in the main codebase. Should the workaround be needeed, there are ways to add in the 3rd party codebases.

At the same time, out of practicality, the issue we're trying to fix here is real and affect real-world deployments.
Hence I'm pivoting this PR to remove all but the parts which make integration of an external informer possible, and propose this as next step.

swatisehgal:
> The core issue is: a completed (either failed or succeeded) pod should accounted for? IOW, should resources accounted to completed pods be reused if need arises? There are valid points for both approaches.

IMO the answer to your question should be NO (in the long term) but I understand that we are not in a perfect world and we can't achieve that immediately. Solving this in the scheduler obviously brings us to a state better than we are currently at. To be a bit more explicit, the way I see this working out in my head is as follows:

1. We merge your current PR that adds support for informer modes (default:dedicated and shared) in the scheduler plugin: we end up in a state where both the pod resource API are monitoring all pods including terminal pods. This is not ideal but resolves the reconciliation issue.
2.  https://github.com/kubernetes/kubernetes/issues/119423 is fixed in kubelet either as a bug to fully remove terminal pods from the podresource API list endpoint or by allowing some sort of filtering.
3. The exporters (NFD/RTE consumes podresource API updates) and in the scheduler plugin we default back to shared mode as now both scheduler plugin and podresource API would be accounting pods that are non-terminal.

> What is more and more evident, anyway is the fact I'm still ruminating about the merit of the approach. This on itself is a strong indicator we should NOT bake in this workaround in the main codebase. Should the workaround be needeed, there are ways to add in the 3rd party codebases.
> 
> At the same time, out of practicality, the issue we're trying to fix here is real and affect real-world deployments. Hence I'm pivoting this PR to remove all but the parts which make integration of an external informer possible, and propose this as next step.

Just want to state that I am happy with the PR in its current form with some minor updates but we can totally explore a more streamlined PR if you think that would be better.

ffromani:
> > The core issue is: a completed (either failed or succeeded) pod should accounted for? IOW, should resources accounted to completed pods be reused if need arises? There are valid points for both approaches.
> 
> IMO the answer to your question should be NO (in the long term) but I understand that we are not in a perfect world and we can't achieve that immediately. Solving this in the scheduler obviously brings us to a state better than we are currently at. To be a bit more explicit, the way I see this working out in my head is as follows:
> 
>     1. We merge your current PR that adds support for informer modes (default:dedicated and shared) in the scheduler plugin: we end up in a state where both the pod resource API are monitoring all pods including terminal pods. This is not ideal but resolves the reconciliation issue.
> 
>     2. [PodresourceAPI reports about resources of pods in terminal phase kubernetes/kubernetes#119423](https://github.com/kubernetes/kubernetes/issues/119423) is fixed in kubelet either as a bug to fully remove terminal pods from the podresource API list endpoint or by allowing some sort of filtering.
> 
>     3. The exporters (NFD/RTE consumes podresource API updates) and in the scheduler plugin we default back to shared mode as now both scheduler plugin and podresource API would be accounting pods that are non-terminal (the ideal state!).
> 
> 
> > What is more and more evident, anyway is the fact I'm still ruminating about the merit of the approach. This on itself is a strong indicator we should NOT bake in this workaround in the main codebase. Should the workaround be needeed, there are ways to add in the 3rd party codebases.
> > At the same time, out of practicality, the issue we're trying to fix here is real and affect real-world deployments. Hence I'm pivoting this PR to remove all but the parts which make integration of an external informer possible, and propose this as next step.
> 
> Just want to state that I am happy with the PR in its current form with some minor updates but we can totally explore a more streamlined PR if you think that would be better.

Thanks for sharing your insights! I was contemplating the best course of action based on your feedback.
On one hand, this is a workaround on the scheduler side, there's no question about it.
On the other hand, without this workaround the caching mechanism of the plugin can end up in unrecoverable stall with basic functionalities of kubernetes are in use, like jobs. The only fix is to restart the scheduler (!!!) when stalls are detected, and detecting a stall is not trivial in the first place.

All of this is bad UX at best, and a crippling blow to the plugin functionality in all the scenarios.

The best course of action would be agree about a standard in the kubernetes, or at least node, community, but this is likely to happen anytime soon - https://github.com/kubernetes/kubernetes/issues/119423 got little traction to begin with, and in all fairness this is a nontrivial effort.

This will lead us to a state on which the plugin functionality is severely diminuished until the oldest support version (current-2 if not mistaken) is fixed, which will takes at least months for now.

Because all of these reasons, and because reviewers and maintainers are at worst neutral if not mildly positive out of practical matters about this change, I'm removing the hold and moving forward.

Granted, we should, no we must keep working toward a conclusive and clean fix. This change can't be the long term solution, even though "temporary" solutions tend to stick around forever. Let's steer the boat for once.

ffromani:
/unhold

begrudgingly. It's a workaround. But we can't keep a cripple plugin until we have a N-2 fixed version. We will talk about at very least kube 1.32

apis/config/v1/defaults.go
```diff
@@ -89,6 +89,8 @@ var (
 
 	defaultResyncMethod = CacheResyncAutodetect
 
+	defaultInformerMode = CacheInformerShared
+
```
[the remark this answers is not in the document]
ffromani: Yes, the default is exactly to both keep the existing behavior AND to not require the awkward dedicated informer
You have the point that if we go as far as merging this change we should not be so shy about it and default to a working solution. I'll change that.

pkg/noderesourcetopology/cache/store_test.go
```diff
@@ -588,9 +588,19 @@ func TestMakeNodeToPodDataMap(t *testing.T) {
 					Spec: corev1.PodSpec{
 						NodeName: "node1",
 					},
+					Status: corev1.PodStatus{
+						Phase: corev1.PodFailed,
+					},
```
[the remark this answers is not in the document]
ffromani: sure, adding

ffromani:
addressed all comments and fixed a subtle bug on which the pods were **not** filtered according to the new option value. Now they are.

swatisehgal (commented):
/lgtm

Thanks for addressing all my review comments.

/assign @Huang-Wei
for approval

swatisehgal:
/hold
to allow time in case @Huang-Wei would like to review and prevent inadvertent merge.

ffromani:
> /hold to allow time in case @Huang-Wei would like to review and prevent inadvertent merge.

Looks like the bot was faster. We can revert if needed.

swatisehgal:
> Looks like the bot was faster. We can revert if needed.

Ah! Didn't expect that! Next time will be more careful.

@Huang-Wei Please let us know if you have concerns here.

Huang-Wei:
> @Huang-Wei Please let us know if you have concerns here.

No concern from my side.

PR #600, opened 2023-07-18
coscheduling list should be in namespace

xiaoanyunfei:
#### What type of PR is this?
/kind bug

Fixes # coscheduling list should be in namespace

```release-note

```


PiotrProkop:
@xiaoanyunfei can you add unit test for this change?

PR #604, opened 2023-07-25
Documents update for release v0.26.7

Huang-Wei:
#### What type of PR is this?

/kind documentation

Update document and Helm chart for release 0.26.

Fixes #

- [ ] TODO: cherrypick this to `release-1.26` once merged

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ changed.
 
 | Scheduler Plugins | Compiled With k8s Version | Container Image                                           | Arch           |
 |-------------------|---------------------------|-----------------------------------------------------------|----------------|
+| v0.26.7           | v1.26.7                   | registry.k8s.io/scheduler-plugins/kube-scheduler:v0.26.7  | AMD64<br>ARM64 |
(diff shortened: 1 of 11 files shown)


Huang-Wei:
/unhold

/assign @zwpaper
Could you take a look?

zwpaper:
/lgtm 

Also, there is one more thing I would like to mention, how about we add a helm chart asset in the release? As the release page would always be one of the first beginnings when users come to us, they could easily start with downloading the Chart.

Huang-Wei:
> how about we add a helm chart asset in the release?

I'm thinking of hosting the helm artifacts using github pages. Please open an issue to track it. For now, users have to install from the sub-directory (`manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml`) which is not good UX.

zwpaper:
> I'm thinking of hosting the helm artifacts using github pages.

I mistakenly remembered that we could not create a project GitHub page, the GitHub page would definitely be a better place for hosting the chart.

https://github.com/kubernetes-sigs/scheduler-plugins/issues/607 created an issue and we can discuss it there.

PR #611, opened 2023-08-05
build: bump builder image to go 1.20

zwpaper:

we bump go to 1.20, but not the dockerfile, this PR update it.

fix https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-scheduler-plugins-push-images/1687662851197505536

```release-note
bump builder image to go1.20
```

### DIFF
--- a/build/controller/Dockerfile
+++ b/build/controller/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ARG ARCH
-FROM golang:1.19
+FROM golang:1.20
 
 WORKDIR /go/src/sigs.k8s.io/scheduler-plugins
 COPY . .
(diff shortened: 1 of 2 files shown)


build/controller/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Kubernetes Authors.
+# Copyright 2023 The Kubernetes Authors.
```
Huang-Wei, line 1: In k8s' codebase, it's conventional to keep the original date instead of updating it for each year.

build/scheduler/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Kubernetes Authors.
+# Copyright 2023 The Kubernetes Authors.
```
Huang-Wei, line 1: ditto.

Huang-Wei (commented):
LGTM except for a nit.

zwpaper:
ok, fixed

Tal-or:
/lgtm
Thanks!

Issue #607, opened 2023-07-27
We should host the helm chart on our github page


zwpaper:
### Area

- [ ] Scheduler
- [ ] Controller
- [X] Helm Chart
- [ ] Documents

### What happened?

users can only use the scheduler plugins Helm Chart from the source code

### What did you expect to happen?

install scheduler plugins Helm Chart directly from the internet(github page)

we can use https://github.com/helm/chart-releaser to create the release and index for the Chart,
it would:

1. create a chart artifacts to GitHub Release
2. create an `index.yaml` which we could used to hosted on GitHub Page

### Kubernetes version

<details>

```console
$ kubectl version
# 1.26.7
```

</details>

### Scheduler Plugins version

<details>
0.26.7
</details>

Huang-Wei:
I'm not sure this is still valid: https://medium.com/@mattiaperi/create-a-public-helm-chart-repository-with-github-pages-49b180dbb417. Jus FYI.

Huang-Wei:
Seems Github pages has evolved a lot since last time I used it:

## build from source

![image](https://github.com/kubernetes-sigs/scheduler-plugins/assets/1425903/77607c2f-0c4c-4e1e-9f34-44a663661e0a)

## build from github actions

<img width="968" alt="image" src="https://github.com/kubernetes-sigs/scheduler-plugins/assets/1425903/af96017c-081c-477f-b778-c3c2d67d6c8b">

@zwpaper you can explore either option to build

zwpaper:
GitHub Action may be a good option, we can use the markdown docs in repo to build a website.

I will look into this after I fix the 1.27 bug

Huang-Wei:
> I will look into this after I fix the 1.27 bug

Yup, that's definitely a higher priority one.

zwpaper:
/reopen

we should leave this issue open until host the index on out netlify page

zwpaper:
/close 

we have hosted the chart on page

PR #616, opened 2023-08-07
site: init hugo site by docs

zwpaper:
#### What type of PR is this?

/kind documentation

1. init GitHub Pages by using the docs
2. create GitHub Action to deploy the Page
3. Skip generate and verify the toc in site, as the Page already has toc

This PR is related to https://github.com/kubernetes-sigs/scheduler-plugins/issues/607, but deploy a page first, later I will fire a PR for the Helm Chart after this one.

1. we need to decide the domain for the page, `scheduler-plugins.sigs.k8s.io` would be the obvious choice, but it seems to be a bit too long.
4. update the domain in Pages Settings
5. I have set up a demo in my fork: https://sched.kweizh.com/

@Huang-Wei 

BTW, the page theme learned from https://kwok.sigs.k8s.io/

```release-note
init github page support
```


zwpaper:
@Huang-Wei we need some advice and help from you to decide whether to continue and how to

Huang-Wei:
Thanks @zwpaper for working on this!

> we need to decide the domain for the page, scheduler-plugins.sigs.k8s.io would be the obvious choice, but it seems to be a bit too long.

scheduler-plugins.sigs.k8s.io doesn't look bad. I'm fine with it.

> update the domain in Pages Settings
I will wait for this PR to be merged, and then submit a PR to create a subdomain under sigs.k8s.io, like [this one](https://github.com/kubernetes/k8s.io/issues/4581)

Two questions:

- Are you going to introduce [netlify](https://github.com/kubernetes-sigs/kwok/blob/main/netlify.toml) into doc PR's CI pipeline?
- Are you going to work on de-dupe the docs among `site/` and non-site locations? so we have a single place for all documents.

zwpaper:
@Huang-Wei thanks for the review.

> scheduler-plugins.sigs.k8s.io doesn't look bad. I'm fine with it.

ok, I will update the PR to the domain.

> I will wait for this PR to be merged, and then submit a PR to create a subdomain under sigs.k8s.io, like https://github.com/kubernetes/k8s.io/issues/4581

ok, I can also help with that after we are on the same page.

> Are you going to introduce [netlify](https://github.com/kubernetes-sigs/kwok/blob/main/netlify.toml?rgh-link-date=2023-08-09T21%3A28%3A08Z) into doc PR's CI pipeline?

No, GitHub has built-in Page support, I am using the GitHub Pages to set up the site.

> Are you going to work on de-dupe the docs among site/ and non-site locations? so we have a single place for all documents.

yes, but after the site is online, so that we can avoid the gap between the site and non-site docs, because we may need some time to set up and make sure the site works as expected.

I will update the PR, and then we should be good to go.

zwpaper:
@Huang-Wei I have updated the PR to set the site to scheduler-plugins.sigs.k8s.io, we may need your help first to update the Page setting before merging this PR, I have appended the steps here https://github.com/kubernetes-sigs/scheduler-plugins/issues/620

zwpaper:
just find out it looks like all sigs projects use netlify to set up Pages, is there something I have not caught up on?

should we change to Netlify?

Huang-Wei:
> just find out it looks like all sigs projects use netlify to set up Pages, is there something I have not caught up on?
> 
> should we change to Netlify?

I think so, which allows us to identify tricky issues (html rendering, etc.) during PR review. I noticed kqueue also used [it](https://github.com/kubernetes-sigs/kueue/blob/main/netlify.toml). So it's a good practice I think.

If we use netlify, so basically we will get a netlify domain name, and in the k8s.io DNS request, we point scheduler-plugins.sigs.k8s.io to the netlify one (instead of the pages URL of github).

> @Huang-Wei I have updated the PR to set the site to scheduler-plugins.sigs.k8s.io, we may need your help first to update the Page setting before merging this PR, I have appended the steps here https://github.com/kubernetes-sigs/scheduler-plugins/issues/620

With that being said, some of the steps are no longer needed. I will do whatever needed in the Github side after you switch to netlify.

Huang-Wei:
Per my understanding, when using netlify it will take over the docs CI, so the Pages setting can be left as-is and github-action is also not needed. See https://github.com/kubernetes-sigs/kueue/tree/main/.github and https://github.com/kubernetes-sigs/kueue/blob/main/netlify.toml

zwpaper:
BTW, found the Netlify standard, FYI if anyone is interested. https://github.com/kubernetes/community/blob/master/github-management/subproject-site-requests.md#subproject-domain-request.

@Huang-Wei it seems that we should merge this PR before requesting the Netlify setup.

I have updated the PR to use netlify, PTAL, and I will fire the netlify and DNS requests after this one merged.

Huang-Wei:
/approve

Thanks @zwpaper again!

PR #622, opened 2023-08-17
Kubelet configuration change without immediate cache updates

### LINKED ISSUE #621: Kubelet configuration change without immediate cache updates
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

func (ov *OverReserve) GetCachedNRTCopy(nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
	ov.lock.Lock()
	defer ov.lock.Unlock()
		return nil, false
	}

	if nrt == nil {
		return nil, true
	}
	nodeAssumedResources, ok := ov.assumedResources[nodeName]
	if !ok {
		return nrt, true
	}

	klog.V(6).InfoS("nrtcache NRT", "logID", klog.KObj(pod), "vanilla", stringify.NodeResourceTopologyResources(nrt))
	nodeAssumedResources.UpdateNRT(klog.KObj(pod).String(), nrt)

	klog.V(5).InfoS("nrtcache NRT", "logID", klog.KObj(pod), "updated", stringify.NodeResourceTopologyResources(nrt))
	return nrt, true
}

When obtaining NRTs from the cache during the filter and score stages, it is assumed that there are changes in the topologyManagerPolicy and topologyManagerScope in the computational node/var/lib/kubelet/config.yaml. Then the topologyPolicies in the NRT resources reported by NFD also changed. However, the cache will not be updated immediately, as the condition for cache updates is that NodesMaybeOverReserved can return nodes. At this point, the filter stage will enter the wrong function based on the cache configuration. For example, if there is a computing node node1, and the topologyManagerPolicy in the NRT cache is singleNUMPodLevel, but the topologyManagerScope in config.yaml is modified to Container, the filter stage will still enter the singleNUMPodLevelHandler function for operation instead of the singleNUMAContainerLevelHandler.

There should be a way to obtain the latest topologyPolicies when the filter and score stages retrieve cached NRT data, avoiding entering incorrect functions for operations

_No response_

_No response_

### Kubernetes version

[root@master1 ~]# kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short. Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.13", GitCommit:"49433308be5b958856b6949df02b716e0a7cf0a3", GitTreeState:"clean", BuildDate:"2023-04-12T12:15:50Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/arm64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.13", GitCommit:"49433308be5b958856b6949df02b716e0a7cf0a3", GitTreeState:"clean", BuildDate:"2023-04-12T12:08:36Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/arm64"}

### Scheduler Plugins version

master

283713406:
#### What type of PR is this?
/kind bug

When obtaining cached NRT data in the filter and score stages, compare the topologyPolicies in the NRT cache with the topologyPolicies of the real-time NRT resources in the environment. If there are differences, update the topologyPolicies in the cache to the latest value. This can obtain the latest topologyPolicies and avoid entering incorrect functions for operations

Fixes #621

"NONE" 
```release-note

```

### DIFF
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -87,6 +87,19 @@ func (ov *OverReserve) GetCachedNRTCopy(nodeName string, pod *corev1.Pod) (*topo
 	if nrt == nil {
 		return nil, true
 	}
+
+	if ov.ShouldUpdateCacheNow(nodeName, nrt) {
+		nrtActual, err := ov.nrtLister.Get(nodeName)
+		if err != nil || nrtActual == nil {
+			return nil, true
+		}
+		ov.nrts.UpdateAttributesAndTopologyPolicies(nrtActual)
+		nrt = ov.nrts.GetNRTCopyByNodeName(nodeName)
+		if nrt == nil {
+			return nil, true
+		}
+	}
+
 	nodeAssumedResources, ok := ov.assumedResources[nodeName]
 	if !ok {
 		return nrt, true
@@ -263,6 +276,23 @@ func (ov *OverReserve) FlushNodes(logID string, nrts ...*topologyv1alpha2.NodeRe
 		ov.nodesMaybeOverreserved.Delete(nrt.Name)
 		ov.nodesWithForeignPods.Delete(nrt.Name)
 	}
+}
+
+func (ov *OverReserve) ShouldUpdateCacheNow(nodeName string, nrtCache *topologyv1alpha2.NodeResourceTopology) bool {
+	nrtActual, err := ov.nrtLister.Get(nodeName)
+	if err != nil || nrtActual == nil {
+		return false
+	}
+
+	if len(nrtActual.TopologyPolicies) == 0 || len(nrtCache.TopologyPolicies) == 0 {
+		return false
+	}
+
+	if nrtActual.TopologyPolicies[0] != nrtCache.TopologyPolicies[0] {
+		return true
+	}
+
+	return false
 }
 
 func InformerFromHandle(handle framework.Handle) (k8scache.SharedIndexInformer, podlisterv1.PodLister) {
--- a/pkg/noderesourcetopology/cache/overreserve_test.go
+++ b/pkg/noderesourcetopology/cache/overreserve_test.go
@@ -367,6 +367,76 @@ func TestGetCachedNRTCopyReserveRelease(t *testing.T) {
 	}
 }
 
+func TestGetCachedNRTCopyShouldUpdateCacheNow(t *testing.T) {
+	fakeClient := faketopologyv1alpha2.NewSimpleClientset()
+	fakeInformer := topologyinformers.NewSharedInformerFactory(fakeClient, 0).Topology().V1alpha2().NodeResourceTopologies()
+	fakePodLister := &fakePodLister{}
+
+	nrtCache := mustOverReserve(t, fakeInformer.Lister(), fakePodLister)
+
+	nodeTopologies := makeDefaultTestTopology()
+	for _, obj := range nodeTopologies {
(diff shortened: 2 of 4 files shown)


PiotrProkop:
Thanks for the PR! @ffromani can you take a look?

ffromani:
Thanks @283713406 for filing the issue and providing the PR. After some thinking, I agree we have a issue here worth fixing.
As explained commenting #621 , we are not bothered by the kubelet config changing _during a schedule cycle_, we expect changed mid-cycle to be so rare we pretty much decide to ignore such occurrences.
However you highlight a related issue: we fail to notice kubelet configuration changes (the bits we care reported by NRT data) and we update them incidentally when resyncing the cache.
These updates are expected to be rare, but still failing to noticing them is a different issue and worth looking.

I'm not keen of adding logic in `GetCachedNRTCopy` for something we rate so infrequent (and, must be noted, for something we _don't support yet_ if happens so frequently that the scheduling performance suffers because of it).
Additionally, adding logic into `GetCachedNRTCopy` to bypass the cache management doesn't fit too well with the existing model.

So, I'd change the `Resync` function to handle this case and generalize the "need update" condition. If the attributes in the client-go cache doesn't match the cached copy, we update our cached copy. This should be done integrating with the existing checks. Perhaps the safest way is to do this update only if the expected pod set matches - otherwise we're going to update the full object anyway.
I'll follow up later with more details.

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -265,6 +278,23 @@ func (ov *OverReserve) FlushNodes(logID string, nrts ...*topologyv1alpha2.NodeRe
 	}
 }
 
+func (ov *OverReserve) ShouldUpdateCacheNow(nodeName string, nrtCache *topologyv1alpha2.NodeResourceTopology) bool {
+	nrtActual, err := ov.nrtLister.Get(nodeName)
+	if err != nil || nrtActual == nil {
+		return false
+	}
+
+	if len(nrtActual.TopologyPolicies) == 0 || len(nrtCache.TopologyPolicies) == 0 {
```
ffromani, line 287: in addition to the comment already made, `TopologyPolicies` is deprecated: we should stop checking it (different PR of course) and check the top-level NRT attributes instead.

283713406:
> Thanks @283713406 for filing the issue and providing the PR. After some thinking, I agree we have a issue here worth fixing. As explained commenting #621 , we are not bothered by the kubelet config changing _during a schedule cycle_, we expect changed mid-cycle to be so rare we pretty much decide to ignore such occurrences. However you highlight a related issue: we fail to notice kubelet configuration changes (the bits we care reported by NRT data) and we update them incidentally when resyncing the cache. These updates are expected to be rare, but still failing to noticing them is a different issue and worth looking.
> 
> I'm not keen of adding logic in `GetCachedNRTCopy` for something we rate so infrequent (and, must be noted, for something we _don't support yet_ if happens so frequently that the scheduling performance suffers because of it). Additionally, adding logic into `GetCachedNRTCopy` to bypass the cache management doesn't fit too well with the existing model.
> 
> So, I'd change the `Resync` function to handle this case and generalize the "need update" condition. If the attributes in the client-go cache doesn't match the cached copy, we update our cached copy. This should be done integrating with the existing checks. Perhaps the safest way is to do this update only if the expected pod set matches - otherwise we're going to update the full object anyway. I'll follow up later with more details.

PR #628, opened 2023-08-31
Disk IO Aware Scheduling KEP

### LINKED ISSUE #624: need to add disk IO aware scheduling to avoid the PODs competing for disk IO resource
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

The current Kubernetes scheduler only supports cpu, memory, and storage aware scheduling. It does not support disk IO resource-aware scheduling. It could happen that the pods scheduled onto a node compete for the disk IO resource, resulting in performance degradation (the noisy neighbor problem). There is an increasing demand for adding disk IO resource-aware scheduling to Kubernetes to avoid or mitigate the noisy neighbor problem. 

A disk IO scheduler plugin can be developed to support disk IO resource-aware scheduling. The plugin would check the new pod's requested disk IO BW against each node's real-time available disk IO BW when making the scheduling decision. 

A disk IO scheduler plugin KEP proposal and corresponding implementation 

_No response_

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

ichbinblau:
#### What type of PR is this?
/kind feature

Propose a scheduler plugin which makes the scheduling decision based on node’s available disk IO capacity.

Fixes #624

Co-authored by @cathyhongzhang 

```release-note
None
```


ffromani:
/cc

At glance, it seems there are remarkable conceptual similarities at architecture level between this work and the noderesourcetopology plugin.

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+The aggregator consolidates the IO metrics from each worker node, reports a consolidated list of real-time available disk IO capacity to the scheduler plugin. It is also responsible for converting each new pod’s disk IO BW request to a normalized value using the disk’s IO Calculation Model so as to match the disk’s normalized available IO capacity. Since the normalization function is disk type and vendor specific, this component is outside the scope of the scheduler plugin’s implementation.
+
+### IO Calculation Model
+
+The IO Calculation Model is responsible for converting the disk’s available IO capacity and each new pod’s IO request to normalized values. Since the normalization function is disk type and vendor specific, this component is outside the scope of the scheduler plugin’s implementation.
+
+### Disk IO Scheduler Plugin
+
+We leverage the K8s scheduler framework to add the disk IO scheduler plugin.
+When the scheduler plugin starts, it subscribes to the IO Driver with a long streaming connection. The IO Driver updates the scheduler plugin about each node disk’s normalized available IO capacity and the scheduler plugin stores the info in the plugin’s local cache.
```
Tal-or, line 84: Since the communication here done between the scheduler and a 3rd party component (and not the API server) are there any safety mechanism against timeout/slow responses, etc.?

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+During the filter phase, the scheduler plugin sends the PodSpec with a disk IO BW request (as shown below) to the IO Driver and gets back a normalized disk IO BW needed by this POD. It then loops through each node in the candidate node list and checks this needed disk IO request of the POD against each node’s available disk IO capacity saved in the local cache to generate an updated candidate list.
+
+``` yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ga_pod
+  annotations:
+     blockio.kubernetes.io/throughput: |
```
Tal-or, line 98: Does the annotation should be added by the scheduler or this is something added by the user?

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+For the Least Allocated policy:
+$$ score = {R \over T} $$
+
+#### Reserve Plugin
+
+During the reserve phase, the scheduler plugin updates the selected node’s available disk IO capacity by deducting the pod’s needed disk IO resource. In addition, it adds the pod to the ReservedPods list, tags the list with a Generation, and then notifies the IO Driver of the new ReservedPods list. The Generation, similar to the Kubernetes’s [ResourceVersion](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions), ensures that only the update on the latest generation ReservedPod list is saved to the scheduler plugin’s cache.
+
+Whenever there is any change to the IO metric collection context, the Generation increases by 1. If the IO Driver reports data based on an older generation than what is saved in the scheduler plugin’s cache, the update will be discarded.
+<p align="center"><img src="images/reserve.png" title="Reserve Phase" width="600" class="center"/></p>
+The API protocol between the disk IO scheduler plugin and the IO Driver is outlined below:
```
ffromani, line 131: Which factors made you decide for this form of communication instead of using a CRD or a new kubernetes datatype in general to enable communication between components? Could you perhaps elaborate about pros/cons of the different approaches?

ffromani: Which service (a new daemonset perhaps?) will provide the data?

ichbinblau:
> /cc
> 
> At glance, it seems there are remarkable conceptual similarities at architecture level between this work and the noderesourcetopology plugin.

Thanks for your comments. 
The total IO bandwidth capacity of a disk is not fixed and changes dynamically based on the characteristics of the workloads running on it, such as the block size and read/write ratio of the workload’s disk access. The normalized disk IO bandwidth of a pod is influenced by various disk IO attributes. Each disk vendor might have its own normalize method. The noderesourcetopology plugin based on CRD cannot support the function calls (EstimateRequest is to normalize a pod's request). Moreover, unlike the noderesourcetopology plugin, the IO driver reports the remaining disk IO BW to the scheduler plugin rather than the total/allocable capacity,  resulting in different accounting methods.

kep/624-disk-io-aware-scheduling/README.md
ichbinblau: The safety mechanism is similar to scheduler extender. And the normalize function will be called when the pod has the disk IO specification. Moreover, unlike the scheduler extender, the IO driver will only be called once at the filter hook point.

kep/624-disk-io-aware-scheduling/README.md
ichbinblau: The annotation is specified by a user.

kep/624-disk-io-aware-scheduling/README.md
ichbinblau: > Which service (a new daemonset perhaps?) will provide the data?

The implementation is vendor-specific. One possible reference implementation involves an aggregator running on the control plane as a service to consolidate data collected from all IO metrics collectors, while IO metrics collectors run on worker nodes as daemonsets.

ichbinblau: > Which factors made you decide for this form of communication instead of using a CRD or a new kubernetes datatype in general to enable communication between components? Could you perhaps elaborate about pros/cons of the different approaches?

In our scenario, the disk IO request of a pod specified in the pod spec must be normalized. To obtain the normalized IO bandwidth, the scheduler plugin needs to call the EstimateRequest API and obtain normalized IO BW from IO driver. Using a Custom Resource Definition (CRD) has the advantage of allowing reuse of Kubernetes’ existing security mechanisms. However, the disadvantage is that using the CRD mechanism to obtain the normalized IO bandwidth can increase scheduling latency.

ffromani:
> The noderesourcetopology plugin based on CRD cannot support the function calls (EstimateRequest is to normalize a pod's request).

Thanks for elaborating. Could, in general, the vendor provide a service which updates a kubernetes datatype (e.g. a CRD) with the normalized data? I'd be interesting to know if this approach is technically feasible or if there are blockers.

In general it seems to me the approach you are following here lean towards the pull model: the consumer, e.g. the scheduler, pulls data when it needs that. I'd be interested in exploring if the push model is feasible and how practical it is.
So, for example, if vendors can provide exporters which push the data to the apiserver, which the scheduler then consumes when it needs.

Historically, the project AFAIR tried hard to avoid direct dependencies between control plane components (scheduler) and nodes.

> Moreover, unlike the noderesourcetopology plugin, the IO driver reports the remaining disk IO BW to the scheduler plugin rather than the total/allocable capacity, resulting in different accounting methods.

Sure, it's unlikely you can just use the noderesorcetopology data type and, going this direction, you will likely need another CRD.

That said, the similarities between this proposal and noderesourcetopology seems that in both cases the scheduler cannot fully predict the resource allocation, and it has to sync with the current node state to keep good scheduling performances.
In other words, in both cases some agent in the node contributes to the final resource allocation decision, and the scheduler does not have the complete picture of a node (unlike the core kubernetes scheduler plugins and resources) and needs to sync after scheduling decisions with the nodes to get the real picture.

ffromani:
> Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.

@ichbinblau I think the root cause of this is that the PR description (first message here) is not following the proper PR template. For a KEP we would probably not need the release note, but I think is worth trying to follow the PR template (https://github.com/openshift-kni/scheduler-plugins/blob/master/.github/PULL_REQUEST_TEMPLATE.md)

ichbinblau:
> > Adding the "do-not-merge/release-note-label-needed" label because no release-note block was detected, please follow our [release note process](https://git.k8s.io/community/contributors/guide/release-notes.md) to remove it.
> 
> @ichbinblau I think the root cause of this is that the PR description (first message here) is not following the proper PR template. For a KEP we would probably not need the release note, but I think is worth trying to follow the PR template (https://github.com/openshift-kni/scheduler-plugins/blob/master/.github/PULL_REQUEST_TEMPLATE.md)
Thanks for the suggestion. It works like a charm.

ichbinblau:
> Thanks for elaborating. Could, in general, the vendor provide a service which updates a kubernetes datatype (e.g. a CRD) with the normalized data? I'd be interesting to know if this approach is technically feasible or if there are blockers.
> 
> In general it seems to me the approach you are following here lean towards the pull model: the consumer, e.g. the scheduler, pulls data when it needs that. I'd be interested in exploring if the push model is feasible and how practical it is. So, for example, if vendors can provide exporters which push the data to the apiserver, which the scheduler then consumes when it needs.
> 
> Historically, the project AFAIR tried hard to avoid direct dependencies between control plane components (scheduler) and nodes.

Thanks for the prompt reply.  The proposal involves both the push model and the pull model. 
The EstimateRequest API normalizes the Pod disk IO request specified in the pod annotation which estimates how much disk capacity the workload would take. The normalization method varies by disk vendors. For instance, one pod disk IO request is 10Mbps and the read/write ratio is 2:1. The normalized disk IO request (obtained from IO Driver) would be 15Mbps taking the read/write ratio and many other factors into consideration. It is a pull model and takes place once per pod. 
The subscribe API is implemented through a gRPC long streaming. The scheduler watches the actual available disk IO capacity from IO driver and updates its local cache. This is a push model. 
> 
> Sure, it's unlikely you can just use the noderesorcetopology data type and, going this direction, you will likely need another CRD.
> 
> That said, the similarities between this proposal and noderesourcetopology seems that in both cases the scheduler cannot fully predict the resource allocation, and it has to sync with the current node state to keep good scheduling performances. In other words, in both cases some agent in the node contributes to the final resource allocation decision, and the scheduler does not have the complete picture of a node (unlike the core kubernetes scheduler plugins and resources) and needs to sync after scheduling decisions with the nodes to get the real picture.

For the subscribe API, another CRD applies.

kep/624-disk-io-aware-scheduling/README.md
ffromani: If I got @Tal-or right, I think he's referring to the fact that adhering closely to the standard kubernetes model, with node agents which push updates to objects (be them CRDs or native objects) to the apiserver, and with the client (scheduler) which watches these changes, brings for free the reliability and scalability guarantees ensured by the project framework. Sure, we need to adhere to the project's best practices, but once that is done the ecosystem basically solves that problem for free.
Diverging by that model means that to a possibly large extent the reimplementor has to take care of these concerns. This may be warranted by the specific case, but I think it should be clearly outlined why the k8s framework is not good enough and why changes to enable the specific use case are undesirable or even impossible.

cathyhongzhang: We thought about using CRD before. But going through the CRD route has two drawbacks: 1) In the pod scheduling flow, the disk IO scheduler plugin needs to pass the disk IO info specified in the annotation of the pod spec to the vendor-specific IO driver and get back a normalized disk IO BW value. Using CRD to support this bi-directional communication would involve creating two CRDs and two watches, which introduces long latency to the pod scheduling flow. 2) A node's disk IO available capacity would change dynamically upon the workloads' real-time IO access characteristics such as block size and IO read/write ratio. If we go through the CRD route, the real-time disk IO available capacity update will inject too much traffic into the API server and degrade the API server performance. Using a direct communication channel between the disk IO scheduler plugin and the vendor-specific IO driver greatly helps mitigate these two issues

ffromani: Thansk @cathyhongzhang for adding extra context, this helps. I recommend to add this information to the KEP text proper. I for myself I'll think a bit more about the problem and try to suggest alternatives. For alpha stage out-of-tree plugin I think the gRPC approach can work, taking into account the fact that scalability and communication issues would need to be pretty much reimplemented.  I'll defer to more experienced scheduler approver the final architectural review.

cathyhongzhang: @ffromani We will add this information to the KEP. Thanks for your review comments!

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+1. In the pod scheduling flow, the disk IO scheduler plugin needs to pass the disk IO info specified in the annotation of the pod spec to the vendor-specific IO driver and get back a normalized disk IO BW value. Using CRD to support this bi-directional communication would involve creating two CRDs and two watches, which introduces long latency to the pod scheduling flow. 
+2. A node's disk IO available capacity would change dynamically upon the workloads' real-time IO access characteristics such as block size and IO read/write ratio. If we choose the CRD route, the real-time disk IO available capacity update will inject too much traffic into the API server and degrade the API server's performance. Using a direct communication channel between the disk IO scheduler plugin and the vendor-specific IO driver greatly helps to mitigate these two issues.
```
Huang-Wei, line 84: Using API to unify all components if the core of k8s ecosystem. We shouldn't arbitrarily plumb connections among any two components directly. Furthermore, treat scheduler as one component in the big picture, its functionality is also used by other ones like cluster autoscaler. If this design is plumbed into scheduler, it'd mean it needs to be carry over to those components as well.

> If we choose the CRD route, the real-time disk IO available capacity update will inject too much traffic into the API server and degrade the API server's performance.

Did you benchmark the API server's performance?

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+During the filter phase, the scheduler plugin sends the PodSpec with a disk IO BW request (as shown below) to the IO Driver and gets back a normalized disk IO BW needed by this POD. It then loops through each node in the candidate node list and checks this needed disk IO request of the POD against each node’s available disk IO capacity saved in the local cache to generate an updated candidate list.
+
+``` yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ga_pod
+  annotations:
+     blockio.kubernetes.io/throughput: |
```
Huang-Wei, line 98: is this spec vendor-specific? or general terms that most vendors would agree to.

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+- Define flexible communication APIs between disk IO aware scheduler plugin and the vendor-specific IO Driver
+
+### Non-Goals
+
+- The implementation of the disk IO Driver due to the distinct characteristics of each disk device.
+
+## Proposal
+
+The disk IO aware scheduler plugin would implement the filter, score and reserve hook points of the scheduler framework. At the filter stage, it would obtain each node’s available disk IO capacity from the IO Driver at run time, update the available disk IO capacity of each node in its local cache, and filter out nodes which do not have enough IO capacity from the node candidate list. At the score stage, it would prioritize the node candidates based on a scoring policy, such as most allocated policy. At the reserve stage it would notify the node IO Driver of the new pod so that the IO Driver can start collecting the disk IO metrics of this new pod.
```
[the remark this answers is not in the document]
cathyhongzhang: @Huang-Wei Thanks for your review! Let me try to clarify. The BW a pod demands is a fixed value specified by the user in the annotation of the pod spec. You are right that a node's BW capacity could change dynamically upon different combinations of workloads running on it. But at the time point when the scheduler plugin performs the filtering operation upon scheduling a new pod, the combination of workloads at each node is known and the capacity of each node at that time point can be calculated. Therefore the scheduler plugin can check whether the disk IO capacity of each node based on the new workload combination is larger than the total disk IO requirements of all the pods including the to-be-scheduled pod, and filter out nodes that do not meet the requirement.

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Aggregator
+
+The aggregator consolidates the IO metrics from each worker node, reports a consolidated list of real-time available disk IO capacity to the scheduler plugin. It is also responsible for converting each new pod’s disk IO BW request to a normalized value using the disk’s IO Calculation Model so as to match the disk’s normalized available IO capacity. Since the normalization function is disk type and vendor specific, this component is outside the scope of the scheduler plugin’s implementation.
+
+### IO Calculation Model
+
+The IO Calculation Model is responsible for converting the disk’s available IO capacity and each new pod’s IO request to normalized values. Since the normalization function is disk type and vendor specific, this component is outside the scope of the scheduler plugin’s implementation.
```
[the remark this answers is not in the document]
cathyhongzhang: Through annotation in the pod spec

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Disk IO Scheduler Plugin
+
+We leverage the K8s scheduler framework to add the disk IO scheduler plugin.
+When the scheduler plugin starts, it subscribes to the IO Driver with a long streaming connection. The IO Driver updates the scheduler plugin about each node disk’s normalized available IO capacity and the scheduler plugin stores the info in the plugin’s local cache. We thought about using CRD mechanism before. But going through the CRD route, it has two drawbacks: 
+1. In the pod scheduling flow, the disk IO scheduler plugin needs to pass the disk IO info specified in the annotation of the pod spec to the vendor-specific IO driver and get back a normalized disk IO BW value. Using CRD to support this bi-directional communication would involve creating two CRDs and two watches, which introduces long latency to the pod scheduling flow.
```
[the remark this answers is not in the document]
cathyhongzhang: The reason that the scheduler plugin needs to communicate with the IO driver is as follows: the disk IO BW specified in the workload's pod spec is a value when the workload performs the disk IO operation using the workload's associated block size and read/write ratio. Different workloads could have different block sizes and read/write ratios. It is known that a workload performing the same disk IO operation but using different block sizes or different read/write ratios would consume different disk IO BW. Therefore to be consistent, we need to normalize all the pods' disk IO BW requirements. Because the normalizing parameters are disk vendor-dependent, the scheduler plugin needs to interface with a vendor-specific driver to get the normalized IO BW value. The communication is gRPC-based and should be fast given the small amount of msg communication.

kep/624-disk-io-aware-scheduling/README.md
cathyhongzhang: Our original design is for each node to send the real-time disk IO telemetry info (such as real-time disk IO BW usage of each pod and real-time available disk IO capacity of each node) to the API server. However, we found out that it injects too much traffic into the API server which degrades the API server performance. Therefore we optimize it by adding an intermediate aggregator that performs telemetry traffic aggregation and batch transport and letting the scheduler plugin handle the interaction with the aggregator. Another consideration of the new design is that this telemetry info changes dynamically, there is no need to persist this info in the API server/ETCD, and therefore no need to go through the CRD route.

kep/624-disk-io-aware-scheduling/README.md
cathyhongzhang: The scheduler plugin is not vendor-specific, but the IO driver is vendor-specific. That is what the KEP defines a generic interface between the scheduler plugin and the vendor-specific IO driver

kep/624-disk-io-aware-scheduling/README.md
Huang-Wei: Per scheuduling's perspective, the minimum needed info is the Node's available disk IO capacity, I'm skeptical that sort of into plus a reasonable sync internal would cause a downgrade on API server.

kep/624-disk-io-aware-scheduling/README.md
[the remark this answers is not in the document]
cathyhongzhang: OK, I see your point. For best-effort workloads, yes, we allow it to run in a non-optimal way and do not apply filtering. But for critical workloads that need a kind of guaranteed disk IO performance (this is the requirement we got from customers like Pinterest), we need to make sure the node has enough disk IO BW for those critical workloads so that those critical workloads do not compete for disk IO resources since when that happens, noisy neighbor problem could arise which may lead to serious performance degradation of those critical workloads.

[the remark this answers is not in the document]
Huang-Wei: in that case, separate them with two annotations: "hard" and "soft", and only invoke Filter for hard requirement.

kep/624-disk-io-aware-scheduling/README.md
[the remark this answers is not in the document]
cathyhongzhang: I think you are right. We will investigate this path. Thanks!

kep/624-disk-io-aware-scheduling/README.md
cathyhongzhang: The impact also depends on the number of nodes in a cluster, especially if it is a large cluster with thousands of nodes. We will re-evaluate the impact and see if it is OK to feed the telemetry info through the API server route.

kep/624-disk-io-aware-scheduling/README.md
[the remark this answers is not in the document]
cathyhongzhang: Yes, we have done that in the annotation. If the user specifies the requested disk IO BW value, it is guaranteed workload, i.e. "hard" requirement. If the user does not specify the requested disk IO BW value, then it is the best effort workload, i.e. "soft".

kep/624-disk-io-aware-scheduling/README.md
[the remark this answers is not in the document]
cathyhongzhang: @Huang-Wei Thanks for the suggestion. This path works. We will update the KEP to have the scheduler plugin get the disk-specific normalization function from the IO driver upon startup time and cache it. Then at the pod scheduling cycle, the scheduler plugin will invoke the normalization function directly without talking to the gRPC server.

kep/624-disk-io-aware-scheduling/README.md
ichbinblau: > Using API to unify all components if the core of k8s ecosystem. We shouldn't arbitrarily plumb connections among any two components directly. Furthermore, treat scheduler as one component in the big picture, its functionality is also used by other ones like cluster autoscaler. If this design is plumbed into scheduler, it'd mean it needs to be carry over to those components as well.
> 
> > If we choose the CRD route, the real-time disk IO available capacity update will inject too much traffic into the API server and degrade the API server's performance.
> 
> Did you benchmark the API server's performance?

Hi, Wei. We did benchmark API server's performance. The test is performed in k8s v1.25.  As scheduling accuracy depends on the node metric update frequency, our update interval is 1s. We measured the P99 latency and failed request rate when sending createPod requests to API Server along with concurrent CR updates to API server. Telling from our test result, when there are 1500 simultaneous node status updates within a one-second interval, the P99 latency will be over 0.5s and the rate of failed requests will exceed 10%. Do you think it is an expected behavior?

cathyhongzhang: @Huang-Wei Our test shows that the larger the cluster, the higher the failure rate. If the 10% failure rate for a cluster of 1,500 nodes is acceptable, we can change our design and feed the telemetry info via the API server route. I feel the failure rate is kind of high if we go for the API server route.

Huang-Wei: I cannot speak for the upper bound of API Server/etcd, but updating interval to 1s sounds short. (ref: the default heartbeat interval is much higher: https://kubernetes.io/docs/reference/node/node-status/#heartbeats)

> we can change our design and feed the telemetry info via the API server route. I feel the failure rate is kind of high if we go for the API server route.

Generally I've seen similar solutions to impose delta resource changes to Node or CR objects when necessary, which serves as an input to make the resource aware by scheduler, and I believe this is the right approach to proceed with.

kep/624-disk-io-aware-scheduling/README.md
```diff
@@ -0,0 +1,257 @@
...
+```
+For the Most Allocated policy:
+
+$$ score = { T - R \over T} $$
+For the Least Allocated policy:
+$$ score = {R \over T} $$
+
+#### Reserve Plugin
+
+During the reserve phase, the scheduler plugin updates the selected node’s available disk IO capacity by deducting the pod’s needed disk IO resource. In addition, it adds the pod to the ReservedPods list, tags the list with a Generation, updates the new ReservedPods list to the CR in the API server and then the IO Driver which is watching the CR would be notified of the change. The Generation, similar to the Kubernetes’s [ResourceVersion](https://kubernetes.io/docs/reference/using-api/api-concepts/#resource-versions), ensures that only the update on the latest generation ReservedPod list is saved to the scheduler plugin’s cache.
```
Huang-Wei, line 231: What's the motivation to persist the in-memory reserving-pods list to CR? Ideally, I'd expect the _transient_ reserve operation (in scheduler) is transparent to any component that reads the CR, so we don't need to issue API call to update the CR in a scheduling cycle. (iow, issuing blocking API calls in a scheduling flow is an anti-pattern)

Huang-Wei (commented):
@ichbinblau could you rebase the master so that CI can be green? Thanks!

Overall the latest version looks good except for the reservedPods part, see comment below. I'm thinking of get rid of this spec field in the CR.

ichbinblau:
> @ichbinblau could you rebase the master so that CI can be green? Thanks!
> 
Yes, the branch has been rebased to master.

kep/624-disk-io-aware-scheduling/README.md
ichbinblau: In our scenario, the IO Driver monitors and reports real-time available disk IO to the scheduler plugin for pod scheduling decisions. The accuracy of this report is crucial and can be affected by the "ReservedPods list." In some cases, the IO Driver's "ReservedPods list" may not match the updated list in the scheduler cache during a new pod's reservation time window. This discrepancy can lead to overprovisioning on a node, as the IO Driver doesn't consider the new pod's disk IO consumption.
To address this issue, we implement a solution by tagging the "ReservedPods list" with a generation. At the reserve hook point, the scheduler plugin increments the generation, tags the new "ReservedPods list" with the incremented generation, and notifies the IO Driver of the updated "ReservedPods list" via the CR. This ensures that the IO Driver considers the new pod's IO consumption when reporting real-time available disk IO. The CR acts as a vital communication link, preventing incorrect reports of real-time disk IO and ensuring accurate scheduling decisions. Its timely update enables the scheduler's cache to reflect the most current and accurate information about available disk IO.                                                                 BTW, the API call to update the "ReservedPods list" via the CR is asynchronous and non-blocking.

ichbinblau: > What's the motivation to persist the in-memory reserving-pods list to CR? Ideally, I'd expect the _transient_ reserve operation (in scheduler) is transparent to any component that reads the CR, so we don't need to issue API call to update the CR in a scheduling cycle. (iow, issuing blocking API calls in a scheduling flow is an anti-pattern)

In our scenario, the IO Driver monitors and reports real-time available disk IO to the scheduler plugin for pod scheduling decisions. The accuracy of this report is crucial and can be affected by the "ReservedPods list." In some cases, the IO Driver's "ReservedPods list" may not match the updated list in the scheduler cache during a new pod's reservation time window. This discrepancy can lead to overprovisioning on a node, as the IO Driver doesn't consider the new pod's disk IO consumption.
To address this issue, we implement a solution by tagging the "ReservedPods list" with a generation. At the reserve hook point, the scheduler plugin increments the generation, tags the new "ReservedPods list" with the incremented generation, and notifies the IO Driver of the updated "ReservedPods list" via the CR. This ensures that the IO Driver considers the new pod's IO consumption when reporting real-time available disk IO. The CR acts as a vital communication link, preventing incorrect reports of real-time disk IO and ensuring accurate scheduling decisions. Its timely update enables the scheduler's cache to reflect the most current and accurate information about available disk IO.                                                                 BTW, the API call to update the "ReservedPods list" via the CR is asynchronous and non-blocking.

Huang-Wei: > This discrepancy can lead to overprovisioning on a node, as the IO Driver doesn't consider the new pod's disk IO consumption.

Thanks for the explanation. Once you prepare for the PR, I'd like this case to be implemented as an integration test so that it can prove the "ReservedPods list" is mandatory.

Huang-Wei:
/lgtm
/approve

Thanks @ichbinblau !

PR #630, opened 2023-08-31
rename hugo config to config.yaml

zwpaper:
#### What type of PR is this?
/kind documentation

default config file name is config.yaml before hugo 109.

and the theme we using needs sass extension, so switch the config name back to `config.yaml`

Fixes #

ref: https://github.com/kubernetes/org/issues/4387

I have tested in my netlify, it works without problem.

```release-note
NONE
```


zwpaper:
the site works now https://github.com/kubernetes-sigs/scheduler-plugins/pull/630#issuecomment-1700369171

/assign @Huang-Wei

Huang-Wei:
Awesome! thanks @zwpaper

/approve
/lgtm

PR #631, opened 2023-08-31
[noderesourcetopology] update NRT when attributes change

### LINKED ISSUE #621: Kubelet configuration change without immediate cache updates

ffromani:
#### What type of PR is this?

/kind bug

One of the key assumptions we took when designing the NodeResourceTopology (NRT) plugin is that the kubelet config of the worker node changes VERY rarely, if at all, during the cluster lifetime. As rule of thumb, it was expected to change with a frequency of like once every quarter (3 months) or so, and likely less often. So the event of changing during a scheduling cycle was deemed extremely low.

However,  we fail to notice kubelet configuration changes (the bits we care reported by NRT data) and we update them incidentally when resyncing the cache.
These updates are expected to be rare, but still failing to noticing them is a different issue because it can lead to unnecessarily bad scheduling decision.

Simply put, failing to noticing these updates is a gap, which we address with this PR

Fixes #621 

still WIP, more test ongoing and upcoming

```release-note
NONE
```


ffromani:
the integration test is pretty much done but thare are known issues I'm investigating.

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -181,14 +216,22 @@ func (ov *OverReserve) NodesMaybeOverReserved(lh logr.Logger) []string {
...
 		nodes.Incr(node)
 	}
 
+	// always use local copies
+	configChangeNodes := ov.nodesWithAttrUpdate.Clone()
+	configChangeCount := configChangeNodes.Len()
+
 	if nodes.Len() > 0 {
-		lh.V(4).Info("found dirty nodes", "foreign", foreignCount, "discarded", nodes.Len()-foreignCount, "total", nodes.Len())
+		lh.Info("found dirty nodes", "foreign", foreignCount, "discarded", overreservedCount, "configChange", configChangeCount, "total", nodes.Len())
```
Tal-or, line 229: Why did we changed from `V(4)` to `Info`

Tal-or (commented):
This new functionality is opt-in in order to support backward compatibility.
But IIUC, the current behavior might causes wrong scheduling decision,  would it be wise to keep the default ("buggy") behavior as an option at all?

pkg/noderesourcetopology/cache/overreserve.go
ffromani: debug change which I forgot to revert, fixing

ffromani:
> Thank you for working on this.
> 
> This new functionality is opt-in in order to support backward compatibility. But IIUC, the current behavior might causes wrong scheduling decision, would it be wise to keep the default ("buggy") behavior as an option at all?

That's a good point. We add a extra watch, but previously the behavior was arguably buggy. Let me switch the defaults.

Tal-or:
/hold
/lgtm 

Hold letting other reviews time to chime in

ffromani:
@PiotrProkop howdy! Do you have any comments by any chance?

PiotrProkop:
@ffromani sorry I was on vacations. Looks good to me!

Issue #629, opened 2023-08-31
Upgrade 1.25 dependencies to v1.25.12


zwpaper:
### Area

- [X] Scheduler
- [X] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?

we currently released scheduler plugins with dependencies 1.25.7.

it would be great to have the upgrade, although 0.25.7 may work seamlessly with 1.25.12.

### What did you expect to happen?

the newest 1.25 release is 1.25.13, and we have some internal usage, so we would like to upgrade the dependencies to 1.25.12.

can we upgrade the 1.25 release to 1.25.12?

if so I would like to fire a PR for this and be thankful to have it happen.

/assign @Huang-Wei 

### Kubernetes version

<details>

```console
$ kubectl version
v1.25.12
```

</details>

### Scheduler Plugins version

<details>
v0.25.7
</details>

Huang-Wei:
@zwpaper sure, go for it. Afterwards I can create a tag for you. And let me know if you want to build a new image.

zwpaper:
thanks, I will fire a PR in days, and then we could build new images so that others could also use them if needed.

Huang-Wei:
https://github.com/kubernetes/k8s.io/pull/5798 has been created

PR #632, opened 2023-09-02
deps: upgrade 1.25 deps to 0.25.12

zwpaper:

upgrade 1.25 release dependencies to 1.25.12

Fixes https://github.com/kubernetes-sigs/scheduler-plugins/issues/629

```release-note
upgrade target kubernetes dependencies to v0.25.12
```


zwpaper:
Should I update the docs in the same PR or wait for the image to be built and create a new one?

Huang-Wei:
> 
Should I update the docs in the same PR or wait for the image to be built and create a new one?

You can do that after the image is out.

PR #643, opened 2023-09-27
KEP: Error Pods Rescheduling

### LINKED ISSUE #640: A new scheduler plugin for rescheduling scheduled pods which encountered some errors to other nodes
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

The current kubernetes scheduler can't support rescheduling already scheduled pods (encounter some errors) to other nodes. for example, a pod may be OOM killed because of the pod's increasing usage of memory and node's insufficient system memory. The in-place restart strategy in pod does not solve the problem of pod exceptions and requires manual intervention to reschedule pods to other nodes. This limits the automatic fault tolerance of the system and reduces the reliability and high availability of the service.

A pod rescheduling plugins  reschedules pods (encounter some errors) to other nodes, excluding nodes that the pod has already scheduled

A pod rescheduling scheduler plugin KEP proposal and corresponding implementation

_No response_

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

freelizhun:
#### What type of PR is this?

/kind feature

Propose a scheduler plugin  to rescheduling pods which encountered some erros

Fixes #640 

None

```release-note
NONE
```


kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+
+
+
+
+
+## Summary
+
+This proposal describes a solution to rescheduling pods which encountered some erros.
```
ffromani, line 43: let's clarify which errors, giving a few key examples

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+#### Story 2
+
+After a pod is submitted to the Kubernetes cluster and successfully scheduled, it may encounter startup issues due to mismatches in hardware resources on the node, 
+such as NUMA or device-related information. In such cases, we need to manually reschedule the pod to a different node. But this pod may schedule to the same node with the default scheduler.
+
+## Design Details
+
+Our scheme is to design two components as outlined in the diagram and described below: Rescheduler-controller and Podrescheduling. 
+Users can configure the `scheduling-retries` (maximum rescheduling retry count) in the metadata annotations of workload resources, such as Deployment and StatefulSet. 
+Rescheduler-controller is responsible for monitoring pods with abnormal states in the cluster. For each abnormal pod,
```
ffromani, line 83: using annotations is fine for starters. Do you plan to introduce a new data structure (CRD object) to hold these informations later on?

The approach of recording possible/discarded target nodes is reminiscent of what DRA does: https://github.com/kubernetes/enhancements/pull/3064/files#diff-8fa237d276346416c2aafa209f721e43c9762f59cabec234eafafe01694de3ee

It's worth exploring how this feature can integrate with DRA (maybe leveraging some of the existing data structures?) and/or if it conflicts with it, and, if conflicts, how manage that

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+the rescheduler-controller checks the current scheduling count (defined as `currentReschedulingTimes`) recorded in the metadata annotations of its associated workload resource. 
+If the current scheduling count does not exceed the user-configured `scheduling-retries` (maximum rescheduling retry count), 
+it increments the current scheduling count by 1 and records the current node where the pod is located in the scheduled nodes list (eg: defined as `deployScheduledHosts`). 
+Then, rescheduler-controller deletes the abnormal pod, and this pod's associated workload resource will create a new pod automatically. 
+Rescheduler-controller registers a MutateWebHook in the kube-apiserver, this MutateWebHook will patch scheduled nodes list info to this newly created pod's annotations (defined as `sigs.k8s.io/scheduled-hosts`) when this newly created pod comes to kube-apiserver admission point.  
+Podrescheduling is the scheduler-plugins that implement the PreFilter extension point of kube-scheduler to prefilter out the scheduled nodes list recorded in this pod annotation and schedule this pod to other nodes.
```
ffromani, line 89: can the functionality of the Mutating Webhook be moived in the rescheduler controller? IOW, can we increase the scheduling count AND record the current running node in the same component and at the same time? If not, please elaborate why

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+The MutateWebHook registers with the kube-apiserver and intercepts incoming pods within the cluster. Subsequently, the MutateWebHook patches the list of scheduled nodes for these pods, which is retrieved from the annotations of the pods' associated workload resources, into the metada annotations of the pods themselves.
+
+#### Scheduled nodes list and current rescheduling times definition
+
+for pods' workload resources，like Deployment
+
+```golang
+type DeployInfo struct {
+	CurrentReschedulingTimes int `json:"currentReschedulingTimes, omitempty"`
+	DeployScheduledHosts []string `json:"deployScheduledHosts, omitempty"`
```
ffromani, line 110: why do we need a different name for Deployment/higher order constructs and for pods?

ffromani (commented):
initial review

kep/640-error-pods-rescheduling/README.md
freelizhun: > can the functionality of the Mutating Webhook be moived in the rescheduler controller? 

Yes, Mutating Webhook belongs to the rescheduler controller as we described in the picture.

> IOW, can we increase the scheduling count AND record the current running node in the same component and at the same time?

Yes, we record scheduling count and current running node at the same time in a struct like this :
```golang
type DeployInfo struct {
	CurrentReschedulingTimes int `json:"currentReschedulingTimes, omitempty"`
	DeployScheduledHosts []string `json:"deployScheduledHosts, omitempty"`
}
```
We create or update `DeployInfo` in rescheduler controller.

kep/640-error-pods-rescheduling/README.md
freelizhun: As we described in [**A concrete use case**](https://github.com/kubernetes-sigs/scheduler-plugins/blob/70a94308603bf3917bb2b7c174cf3bdbb99b564d/kep/640-error-pods-rescheduling/README.md#a-concrete-use-case), the ListFunc component of the rescheduler-controller delete error pod, so this pod disappear and will be created automatically by it's associated workload resource like deployment. We need a struct like DeployInfo  to record this pod's `currentReschedulingTimes` and `scheduledhosts`

kep/640-error-pods-rescheduling/README.md
freelizhun: > Do you plan to introduce a new data structure (CRD object) to hold these informations later on?

Thanks for your suggestion.
Yes, In the next versions, i will develop a CRD (Custom Resource Definition) to replace the current use of annotations.

I store information in the annotations of the associated workload resources (like deployment) for pods, ensuring there are no conflicts with DRA.

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+it increments the current scheduling count by 1 and records the current node where the pod is located in the scheduled nodes list (eg: defined as `deployScheduledHosts`). 
+Then, rescheduler-controller deletes the abnormal pod, and this pod's associated workload resource will create a new pod automatically. 
+Rescheduler-controller registers a MutateWebHook in the kube-apiserver, this MutateWebHook will patch scheduled nodes list info to this newly created pod's annotations (defined as `sigs.k8s.io/scheduled-hosts`) when this newly created pod comes to kube-apiserver admission point.  
+Podrescheduling is the scheduler-plugins that implement the PreFilter extension point of kube-scheduler to prefilter out the scheduled nodes list recorded in this pod annotation and schedule this pod to other nodes.
+
+![design](./design.png)
+
+### Rescheduler-controller
+
+The rescheduler-controller is comprised of two components, as illustrated in the above diagram: ListFunc and MutateWebHook.
```
PiotrProkop, line 95: Maybe I am just peaky but why `ListFunc` component is also responsible for deleting pods?
It should be either `ListAndDeleteFunc` components or it should be divided into `ListFunc` and `DeleteFunc` components.

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+The ListFunc serves as a periodic monitoring program responsible for listing pods in abnormal states within the cluster. 
+It is also in charge of deleting these abnormal pods and updating annotations in the metadata of these pods' workload resources to record the scheduled nodes list. 
+The ListFunc triggers pod rescheduling by deleting pods. Additionally, for pods without an owner reference definition, the ListFunc maintains a map (for updating pod's metada annotations) in memory to store scheduled hosts, allowing for direct code-based resubmission of these pods to the cluster.
+
+The MutateWebHook registers with the kube-apiserver and intercepts incoming pods within the cluster. Subsequently, the MutateWebHook patches the list of scheduled nodes for these pods, which is retrieved from the annotations of the pods' associated workload resources, into the metada annotations of the pods themselves.
```
Huang-Wei, line 101: What if we directly impose additional podAntiAffinity field to the ownerReference's podTemplate / orphan pod's spec? In this case, we can simplify the entire workflow and simplify the design to be a controller-only solution.

kep/640-error-pods-rescheduling/README.md
freelizhun: @PiotrProkop Thanks for your review! I will update `ListFunc` to `ListAndDeleteFunc`.

kep/640-error-pods-rescheduling/kep.yaml
```diff
@@ -0,0 +1,14 @@
+title: Error Pods Rescheduling
+kep-number: 640
+authors:
+  - "@freelizhun"
+owning-sig: sig-scheduling
+reviewers:
+  - "@Huang-Wei"
+  - "@ahg-g"
+  - "@alculquicondor"
```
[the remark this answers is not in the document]
freelizhun: @Huang-Wei Thanks for your review! I will replace names with actual reviewers.

kep/640-error-pods-rescheduling/README.md
freelizhun: > What if we directly impose additional podAntiAffinity field to the ownerReference's podTemplate / orphan pod's spec? 
This can result in pod's ownerReference

This will cause changes to the pod's `ownerReference` definition, but how does this additional podAntiAffinity field prevent the pod from being scheduled to the previously erroneous node?

ffromani:
A couple weeks ago we start a conversation on the #wg-batch channel on k8s's slack about tackling a very similar problem. I'm writing a design proposal, which is a stage just before a KEP, [here](https://docs.google.com/document/d/1-wJhiNy84w7tzFdo9HqwTu5DrVSuXFLGTUv8FBiRAAc/edit?usp=sharing), and later we will evaluate if move towards a KEP with the goal of helping DRA and/or specialized schedulers like NUMA-aware scheduling.

I'm mentioning this work here because of the significant overlaps. If scope and timelines align, we can probably join forces in a single proposal.

I'll be reviewing again this work shortly considering the aforementioned. Some design decisions we are evaluating may help here.

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+
+
+
+
+
+## Summary
+
+This proposal outlines a solution for rescheduling pods that have encountered errors, such as pods being OOM-killed or experiencing errors related to NUMA or devices.
```
ffromani, line 43: We're working on a [similar design](https://docs.google.com/document/d/1-wJhiNy84w7tzFdo9HqwTu5DrVSuXFLGTUv8FBiRAAc/edit?usp=sharing) spawned from DRA beta graduation conversation and it's very hard to implement this approach without kubelet changes, and even with that we need to take into account races with existing controllers.

Because this proposal is against scheduler-plugins, this suggests you are trying to avoid core kubernetes changes. Is my understanding right?

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Goals
+
+* Reschedule already scheduled pods (encounter some errors) to other nodes automatically.
+* Implement pod rescheduling plugins in PreFilter extension point to prefilter nodes that the pod has already scheduled.
+* Pod rescheduling plugin in this proposal has no conflict with default scheduler's plugins.
+
+### Non-Goals
+
+* Reschedule pods controlled by daemonset or other workload resources which don't exist in default k8s controller.
```
ffromani, line 63: is this to avoid changes to existing controllers? (sorry if asked previously, if so just link the answer please)

kep/640-error-pods-rescheduling/README.md
ffromani: Elaborating further, this seems quite like a separate controller which kinda adopts and manages pods which aren't owned by a controller (e.g. pod belonging to a daemonset, deployment...). Would it be the case?

If it would be the case, it becomes IMO a hard sell to have this separate controller in parallel to the existing ones, because the obvious question would become "why don't change kubernetes core to make it work with the existing controllers?" which would lead us in the vicinity of https://docs.google.com/document/d/1-wJhiNy84w7tzFdo9HqwTu5DrVSuXFLGTUv8FBiRAAc/edit#heading=h.5mgn2182hlz6

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+such as NUMA or device-related information. In such cases, we need to manually reschedule the pod to a different node. But this pod may schedule to the same node with the default scheduler.
+
+## Design Details
+
+Our scheme is to design two components as outlined in the diagram and described below: Rescheduler-controller and Podrescheduling. 
+Users can configure the `scheduling-retries` (maximum rescheduling retry count) in the metadata annotations of workload resources, such as Deployment and StatefulSet. 
+Rescheduler-controller is responsible for monitoring pods with abnormal states in the cluster. For each abnormal pod, 
+the rescheduler-controller checks the current scheduling count (defined as `currentReschedulingTimes`) recorded in the metadata annotations of its associated workload resource. 
+If the current scheduling count does not exceed the user-configured `scheduling-retries` (maximum rescheduling retry count), 
+it increments the current scheduling count by 1 and records the current node where the pod is located in the scheduled nodes list (eg: defined as `deployScheduledHosts`).
```
ffromani, line 86: logic seems fine, but for the sake of clarity I'd mention earlier and more explicitly that the most important extra annotations are both `currentReschedulingTimes` and `deployScheduledHosts`. The flow of the tests initially highlights the former and only later introduces the latter on the go.
Additionally I'd use a consistent style between `camelCase` and `dash-separate-words`.

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+Users can configure the `scheduling-retries` (maximum rescheduling retry count) in the metadata annotations of workload resources, such as Deployment and StatefulSet. 
+Rescheduler-controller is responsible for monitoring pods with abnormal states in the cluster. For each abnormal pod, 
+the rescheduler-controller checks the current scheduling count (defined as `currentReschedulingTimes`) recorded in the metadata annotations of its associated workload resource. 
+If the current scheduling count does not exceed the user-configured `scheduling-retries` (maximum rescheduling retry count), 
+it increments the current scheduling count by 1 and records the current node where the pod is located in the scheduled nodes list (eg: defined as `deployScheduledHosts`). 
+Then, rescheduler-controller deletes the abnormal pod, and this pod's associated workload resource will create a new pod automatically. 
+Rescheduler-controller registers a MutateWebHook in the kube-apiserver, this MutateWebHook will patch scheduled nodes list info to this newly created pod's annotations (defined as `sigs.k8s.io/scheduled-hosts`) when this newly created pod comes to kube-apiserver admission point.
```
ffromani, line 88: I think I get why you want to do like this, but deleting a pod and recreating it overlaps with descheduler and other components. We need to clarify why descheduler is not good enough and/or how can be extended. Needed to do very similarly in https://docs.google.com/document/d/1-wJhiNy84w7tzFdo9HqwTu5DrVSuXFLGTUv8FBiRAAc/edit#heading=h.5mgn2182hlz6 - Please check "discarded alternatives" in the end.

kep/640-error-pods-rescheduling/README.md
```diff
@@ -0,0 +1,166 @@
...
+type PodInfo struct {
+	CurrentReschedulingTimes int `json:"currentReschedulingTimes, omitempty"`
+	PodScheduledHosts []string `json:"podScheduledHosts, omitempty"`
+}
+```
+
+### Podrescheduling
+
+Podrescheduling implements the PreFilter interface of the scheduler framework. It extracts the list of nodes already scheduled from the "`sigs.k8s.io/scheduled-hosts`" field in the pod's metadata annotations. 
+Then, it filters out these already scheduled nodes and returns a list of nodes in the cluster where the pod has not yet been run.
```
ffromani, line 125: Mostly curious, why in `PreFilter` and not in `Filter`?

ffromani (commented):
Mostly general comments about the design and the possible interactions with a very similar work recently started from DRA conversations and linked multiple times.

Would you be interested in cooperating in such effort? Full disclosure: is not yet clear if the _proposal_ can be sent in time for 1.30 because EU/NA holidays. I'm aiming to be ready in time for 1.30 but I can't make promises. This effort aims to change core k8s so it can take a while.

I'll have a deeper look at the implementation ASAP.

kep/640-error-pods-rescheduling/README.md
freelizhun: Thanks for your review! Yes, we don't need to change any core kubernetes code, and just implement a scheduler-plugin.

kep/640-error-pods-rescheduling/README.md
freelizhun: We don't need to change existing controllers. We reschedule thoses pods which controlled by existing k8s default controllers like deployment or statefulset, but don't reschedule pods controlled by user's newly defined controllers.

As for the daemonset controller, it ensures that all nodes run a copy of a pod always, so we don't reschedule thoses pods controlled by damoneset.

kep/640-error-pods-rescheduling/README.md
freelizhun: Ok, i will use a consistent style between `camelCase` and `dash-separate-words`.

kep/640-error-pods-rescheduling/README.md
freelizhun: In `PreFilter` phase, we can preFilter many nodes at a time, because all the already scheduled nodes are recorded in `sigs.k8s.io/scheduled-hosts`.
But in `Filter` phase, we can only filter one node at a time. So we implement Podrescheduling in `PreFilter` phase.

kep/640-error-pods-rescheduling/README.md
freelizhun: Ok, i will read the contents of this link carefully.

freelizhun: The descheduler project pays attentio to **already running pods**, but our proposal focus on **pods running normally at first but fails after some time**.

ffromani: That would make sense to me. If this is the case also for your proposal, please make sure to write it down very clearly that in your usecase you cannot use descheduler for the reasons above.

kep/640-error-pods-rescheduling/README.md
ffromani: OK, so this proposal seems a subset of the kubelet-triggers-reschedule one which we're working on. Avoiding any change to kubernetes core is a pretty severe limitation.
In general I think it would be good to start with this approach in the scheduler-plugins and see how it performs. 
We can later generalize (or try to) the solution incorporating the lessons learned here. And if things go well, maybe have something in kubernetes core someday in the future.

freelizhun: Ok, thanks for your reply. As you mentioned above, [here](https://docs.google.com/document/d/1-wJhiNy84w7tzFdo9HqwTu5DrVSuXFLGTUv8FBiRAAc/edit#heading=h.iyqdsaanqjtg), i have a great interest in cooperating in this proposal.
> Mostly general comments about the design and the possible interactions with a very similar work recently started from DRA conversations and linked multiple times.
> 
> Would you be interested in cooperating in such effort? Full disclosure: is not yet clear if the _proposal_ can be sent in time for 1.30 because EU/NA holidays. I'm aiming to be ready in time for 1.30 but I can't make promises. This effort aims to change core k8s so it can take a while.
> 
> I'll have a deeper look at the implementation ASAP.

So how do I get involved in this project? Thank you.

kep/640-error-pods-rescheduling/README.md
freelizhun: Ok, i will clarify it clearly in my KEP.

kep/640-error-pods-rescheduling/README.md
ffromani: Hi! for starters, please join the #wg-batch channel on k8s slack and perhaps join the conversations here (I'll be updating regularly but please be aware of the upcoming EU/NA holidays and limited availability).
I'll open the aforementioned design doc for comments as soon as the remaining gaps are open.

freelizhun: Ok, thanks for your reply. I have joined the #wg-batch channel on k8s slack.
I have updated the latest KEP version, is there anything else I need to change about this KEP?
Thanks a lot for your review!

ffromani: I will have another pass as soon as I can, but generally speaking I don't have any major objection.

Please note however that while I for myself I'm ok with the approach outlined here, it seems to me that basically this proposal is adding a new controller for pods which kinda overlaps with deployment controller - it manages pod retries after all, reminiscent of how a deployment would do. Or, in general, there are similarities. For this reason I think more review, especially from more senior contributors of the project is required.

freelizhun: Ok, thanks for your reply. The Rescheduler-controller will not be pushed to this project, but only podrescheduling scheduler-plugins be pushed.

Issue #485, opened 2023-01-26
migrate controllers and clients to leverage controller-runtime


zwpaper:
Scheduler Plugins is working on migrating controllers and clients to leverage `controller-runtime`, 

I have walked through the code base quickly and met some questions, so I am opening this issue to
make sure I am doing things right and figure out how can we work on this.

there should be 2 parts we should work on:
1. controllers: https://github.com/kubernetes-sigs/scheduler-plugins/tree/056fa7df192d4b49101650f37873338e3d8b96fb/pkg/controller, which is WIP with https://github.com/kubernetes-sigs/scheduler-plugins/tree/056fa7df192d4b49101650f37873338e3d8b96fb/pkg/controllers and not yet being used
2. generated clients: used everywhere in controllers and schedulers

my questions are:
1. is the 2 parts above right?
2. should we use a new command args to choose the legacy or controller-runtime controllers
3. should all the generated clients used in controllers and schedulers be migrated to controller-runtime clients?

# break down

## Controllers

- [x] Pod Group: @zwpaper https://github.com/kubernetes-sigs/scheduler-plugins/pull/506 
- [x] Elastic Quota: @czybjtu https://github.com/kubernetes-sigs/scheduler-plugins/pull/527

## clients

this is generated by the following cmd:

```sh
rg -c -l -g '!vendor' -g '!pkg/generated' -g '!*_test.go' 'pkg/generated/' . | sort
```

- [x] ./cmd/controller/app/server.go:2 https://github.com/kubernetes-sigs/scheduler-plugins/pull/569
- [x] ./pkg/capacityscheduling/capacity_scheduling.go:3 #665
- [x] ./pkg/controller/elasticquota.go:3 https://github.com/kubernetes-sigs/scheduler-plugins/pull/527
- [x] ./pkg/controller/podgroup.go:3: @zwpaper https://github.com/kubernetes-sigs/scheduler-plugins/pull/506 
- [x] ./pkg/coscheduling/core/core.go:3 #652
- [x] ./pkg/coscheduling/coscheduling.go:2 #652
- [x] ./pkg/networkaware/networkoverhead/networkoverhead.go:2 @zwpaper https://github.com/kubernetes-sigs/scheduler-plugins/pull/522
- [x] ./pkg/networkaware/topologicalsort/topologicalsort.go:1 @zwpaper https://github.com/kubernetes-sigs/scheduler-plugins/pull/522
[...]

PR #655, opened 2023-11-01
migrate nrt to ctrl runtime client

zwpaper:
#### What type of PR is this?
/kind cleanup

Fixes #

part of https://github.com/kubernetes-sigs/scheduler-plugins/issues/485

after deleting the generated vendor, CI found noderesourcetopology crd file changed, v1alpha1 version was deleted, I checked and found nowhere that used the v1alpha1, so I deleted the version.

now we only have noderesourcetopology v1alpha2

```release-note
migrate noderesourcetopology to ctrl runtime client
```


zwpaper:
@Huang-Wei CI passed, PTAL

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -17,66 +17,67 @@ limitations under the License.
...
+	if client == nil || podLister == nil {
 		return nil, fmt.Errorf("nrtcache: received nil references")
 	}
 
 	resyncMethod := getCacheResyncMethod(cfg)
 
-	nrtObjs, err := nrtLister.List(labels.Everything())
-	if err != nil {
+	nrtObjs := &topologyv1alpha2.NodeResourceTopologyList{}
+	if err := client.List(context.Background(), nrtObjs); err != nil {
```
Huang-Wei, line 63: TODO: same as coscheduling, we can pass-in a context in the future

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -17,66 +17,67 @@ limitations under the License.
...
 		nodesWithForeignPods:   newCounter(),
-		nrtLister:              nrtLister,
 		podLister:              podLister,
 		resyncMethod:           resyncMethod,
 	}
 	return obj, nil
 }
 
-func (ov *OverReserve) GetCachedNRTCopy(nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
+func (ov *OverReserve) GetCachedNRTCopy(ctx context.Context, nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
```
Huang-Wei, line 80: It seems ctx is not used in this function.

Huang-Wei (commented):
Some nits. LGTM overall.

/approve
/lgtm

Huang-Wei:
/hold

cc @ffromani @swatisehgal @Tal-or for awareness.

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -17,66 +17,67 @@ limitations under the License.
...
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	apiconfig "sigs.k8s.io/scheduler-plugins/apis/config"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/stringify"
-
-	"github.com/k8stopologyawareschedwg/podfingerprint"
 )
 
 type OverReserve struct {
+	ctrlclient.Client
```
Tal-or, line 43: Is there a reason for embedding the client?

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -43,32 +45,26 @@ const (
 )
 
 func initNodeTopologyInformer(tcfg *apiconfig.NodeResourceTopologyMatchArgs, handle framework.Handle) (nrtcache.Interface, error) {
-	topoClient, err := topoclientset.NewForConfig(handle.KubeConfig())
+	scheme := runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
```
Tal-or, line 49: I would handle this error instead of panic

pkg/noderesourcetopology/pluginhelpers.go
```diff
@@ -43,32 +45,26 @@ const (
...
+	utilruntime.Must(topologyv1alpha2.AddToScheme(scheme))
```
Tal-or, line 50: ditto

pkg/noderesourcetopology/score_test.go
```diff
@@ -128,13 +128,18 @@ func initTest(policy topologyv1alpha2.TopologyManagerPolicy, nodes ...*topologyv
 	}
 
 	// init topology lister
-	fakeClient := faketopologyv1alpha2.NewSimpleClientset()
-	fakeInformer := topologyinformers.NewSharedInformerFactory(fakeClient, 0).Topology().V1alpha2().NodeResourceTopologies()
+	fakeClient, err := tu.NewFakeClient()
+	if err != nil {
+		panic(err)
```
Tal-or, line 133: This is a test code, so `panic` is fine

Tal-or (commented):
Thank you for the PR.
I left some notes from my side, and wait for @ffromani and @swatisehgal to review this  PR as well.

pkg/noderesourcetopology/cache/overreserve.go
Huang-Wei: Did you mean to make it private?

pkg/noderesourcetopology/pluginhelpers.go
Huang-Wei: this error in practice won't happen IMO. If we treat it as an error, @Tal-or do you mean we can find a way to proceed?

pkg/noderesourcetopology/cache/overreserve.go
ffromani: I guess (and I'd agree with) make it consistent with `DiscardReserved` struct above

pkg/noderesourcetopology/cache/overreserve.go
ffromani: yup, either pass it through `ov.Get` (which will become `ov.client.Get`?) or discard it please (applies to similar places elsewhere in the code)

pkg/noderesourcetopology/cache/passthrough.go
```diff
@@ -17,27 +17,31 @@ limitations under the License.
...
 	"k8s.io/klog/v2"
 
-	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
-	listerv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/listers/topology/v1alpha2"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type Passthrough struct {
-	lister listerv1alpha2.NodeResourceTopologyLister
+	client ctrlclient.Client
```
ffromani, line 32: likewise, we're not using embedding here so better be consistent in `OverReserve` as well

test/integration/noderesourcetopology_cache_test.go
```diff
@@ -477,7 +481,13 @@ func TestTopologyCachePluginWithoutUpdates(t *testing.T) {
 			testCtx.Ctx, testCtx.CancelFn = context.WithCancel(context.Background())
 
 			cs := clientset.NewForConfigOrDie(globalKubeConfig)
-			extClient := versioned.NewForConfigOrDie(globalKubeConfig)
+			scheme := runtime.NewScheme()
+			utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+			utilruntime.Must(topologyv1alpha2.AddToScheme(scheme))
```
ffromani, line 486: while this is probably OK, it seems wasteful to do each time. Can we do it once when test runs?

test/integration/noderesourcetopology_cache_test.go
```diff
@@ -657,8 +667,14 @@ func TestTopologyCachePluginWithUpdates(t *testing.T) {
 	testCtx := &testContext{}
 	testCtx.Ctx, testCtx.CancelFn = context.WithCancel(context.Background())
 
+	scheme := runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(topologyv1alpha2.AddToScheme(scheme))
```
ffromani, line 672: ditto

test/integration/noderesourcetopology_test.go
```diff
@@ -127,7 +131,14 @@ func TestTopologyMatchPlugin(t *testing.T) {
 	testCtx.Ctx, testCtx.CancelFn = context.WithCancel(context.Background())
 
 	cs := clientset.NewForConfigOrDie(globalKubeConfig)
-	extClient := versioned.NewForConfigOrDie(globalKubeConfig)
+
+	scheme := runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(topologyv1alpha2.AddToScheme(scheme))
+	extClient, err := ctrlclient.New(globalKubeConfig, ctrlclient.Options{Scheme: scheme})
```
ffromani, line 138: ditto

ffromani (commented):
there could be some impl details to iterate over, but design wise moving to controller-runtime seems OK. Doesn't help us in general in our journey towards k8s core merge (or its proposal, at least), but its' manageable.

What puzzles me more is if this approach is still compatible with https://github.com/kubernetes-sigs/scheduler-plugins/pull/599 .
Note that #599 is not necessarily something which I want to merge, but still the problem it attacks is real, and we didn't reach a conclusion which side should carry a fix. In addition, it would be nice to have the option to workaround the issue _also_ on scheduler side if needed.

pkg/noderesourcetopology/pluginhelpers.go
Tal-or: @Huang-Wei Not sure if it's possible to proceed, but at least throwing a detailed error which bubble up and explained what happen, would be more user friendly.

pkg/noderesourcetopology/cache/overreserve.go
zwpaper: it is because we use the ctx here: https://github.com/kubernetes-sigs/scheduler-plugins/pull/655/files#diff-d469fc389aadc64511b0fcbf7af645119b04c14a1cc18f5d2307755df1c41a2eR57

and this function is part of the cache Interface here: https://github.com/kubernetes-sigs/scheduler-plugins/pull/655/files#diff-4792874d567196ef3ce188f2f3ccda73d46457d99d95111e04a0072dcf183375R35

so we have to change it to both DiscardReserved and OverReserve

BTW, Passthrough also used the ctx in GetCachedNRTCopy: https://github.com/kubernetes-sigs/scheduler-plugins/pull/655/files#diff-5747013d93a36239aaa4d659c8de3b9a2ef08bbd0f842377c25a1e810007b3e6R44

pkg/noderesourcetopology/cache/overreserve.go
zwpaper: I don't have a strong preference for this, but yes, we should keep it consistent, changed it to align with DiscardReserved.

pkg/noderesourcetopology/pluginhelpers.go
zwpaper: This is how the controller runtime client can know what to do with the types, this is a common usage from kube-builder, which is one of the most famous libs using controller runtime.

you can check: https://book.kubebuilder.io/cronjob-tutorial/empty-main

image

test/integration/noderesourcetopology_cache_test.go
zwpaper: the tests run independently, how can we init once and run cross all the tests? maybe we have to refactor to use a test.M?

Huang-Wei:
> What puzzles me more is if this approach is still compatible with https://github.com/kubernetes-sigs/scheduler-plugins/pull/599 .

@ffromani this PR won't block #599 as #599 touched scheduler config API, and the introduced informer are imposed on core api objects (pods).

test/integration/noderesourcetopology_cache_test.go
Huang-Wei: I sort of agree with @zwpaper, in the perspective to making each UT hermetic.

pkg/noderesourcetopology/pluginhelpers.go
Huang-Wei: My 2 cents is still that this error is not recoverable, and based on my exp. of writing controllers, this doesn't ever happen at all (pretty static behavior IMHO). So I'm more skewed with this pattern of panic immediately- if the error really happens.

ffromani: I agree with @Huang-Wei . The error is not recoverable, so it's fine to use `utilruntime.Must` and fail badly in the (I reckon) exceptionally rare case on which AddToScheme fails.

The only thing which doesn't fully click to me is why we need to fix the schema _in this precise code location_. Disclosure: I didn't review the latest changes, so this is possibly already improved.

zwpaper: Hi @ffromani, I did not update the codes here.

we must handle the scheme somewhere, as the main scheduler, we added in main.go, and for this nrt plugin, I add it here when initing it, so that we are not adding this scheme if nrt plugin is not used.

as for `utilruntime.Must`, WDYT @Tal-or ?

pkg/noderesourcetopology/cache/passthrough.go
zwpaper: I have changed the `OverReserve` to private client to keep it consistent.

pkg/noderesourcetopology/pluginhelpers.go
Huang-Wei: @ffromani raised a very good point. The AddToScheme logic should only happen once per its usage:

- for running scheduler, it's should in main's init() or equivalent place. @zwpaper I know this is generic code problem, but if you can add this logic in this PR, it'd be much appreciated 🙇‍♂️ 
- for UTs, it's fine to init in each UT as we treat each UT as a client

Tal-or: > as for utilruntime.Must, WDYT @Tal-or ?
@zwpaper since the other reviewers are in favor of `utilruntime.Must` that's ok

zwpaper: @Huang-Wei moved all scheme initial into `init` func, PTAL

```console
 rg -B 1 -g '!vendor/' -g '!*_test.go' -- utilruntime.Must
pkg/noderesourcetopology/plugin.go
45-func init() {
46:	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
47:	utilruntime.Must(topologyv1alpha2.AddToScheme(scheme))

pkg/networkaware/topologicalsort/topologicalsort.go
45-func init() {
46:	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
47:	utilruntime.Must(agv1alpha.AddToScheme(scheme))

pkg/networkaware/networkoverhead/networkoverhead.go
66-func init() {
67:	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
68-
69:	utilruntime.Must(agv1alpha1.AddToScheme(scheme))
70:	utilruntime.Must(ntv1alpha1.AddToScheme(scheme))

pkg/generated/clientset/versioned/scheme/register.go
54-	v1.AddToGroupVersion(Scheme, schema.GroupVersion{Version: "v1"})
55:	utilruntime.Must(AddToScheme(Scheme))

pkg/generated/clientset/versioned/fake/register.go
54-	v1.AddToGroupVersion(scheme, schema.GroupVersion{Version: "v1"})
55:	utilruntime.Must(AddToScheme(scheme))

cmd/controller/app/server.go
37-func init() {
38:	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
39-
40:	utilruntime.Must(schedulingv1a1.AddToScheme(scheme))

apis/config/scheme/scheme.go
44-func AddToScheme(scheme *runtime.Scheme) {
45:	utilruntime.Must(config.AddToScheme(scheme))
46:	utilruntime.Must(v1.AddToScheme(scheme))
47:	utilruntime.Must(v1beta2.AddToScheme(scheme))
48:	utilruntime.Must(v1beta3.AddToScheme(scheme))

apis/scheduling/scheme/scheme.go
31-func AddToScheme(scheme *runtime.Scheme) {
32:	utilruntime.Must(schedv1alpha1.AddToScheme(scheme
```

pkg/networkaware/topologicalsort/topologicalsort_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/rand"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/client-go/kubernetes/scheme"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
```
Huang-Wei, line 31: nit: revert this?

Huang-Wei (commented):
LGTM overall.

A potential future improvement is to instantiate the `scheme` once (in main.go) for non-test code, and then pass to all plugins that need it using a WithScheme option. But this doesn't block on this PR's merge.

pkg/noderesourcetopology/cache/overreserve.go
ffromani: right, my bad. It would be nice to change it as
`func (ov *OverReserve) GetCachedNRTCopy(_ context.Context, nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
}
`
because `ctx` is indeed not needed in the body, but this is borderline a nit and totally not blocking

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -208,8 +210,8 @@ func (ov *OverReserve) Resync() {
 
 	var nrtUpdates []*topologyv1alpha2.NodeResourceTopology
 	for _, nodeName := range nodeNames {
-		nrtCandidate, err := ov.nrtLister.Get(nodeName)
-		if err != nil {
+		nrtCandidate := &topologyv1alpha2.NodeResourceTopology{}
+		if err := ov.client.Get(context.Background(), types.NamespacedName{Name: nodeName}, nrtCandidate); err != nil {
```
ffromani, line 214: I wonder if we should use the same context for all the `Get`s - and I think we should (and we should also make it cancellable as further extension). But this change is better served in a followup PR.

ffromani (commented):
LGTM
(non-binding to let other tagged reviewers review)

pkg/noderesourcetopology/cache/overreserve.go
Huang-Wei: yup, other PRs also have some leftovers. The general idea would be use context wherever you can:

1. the chained calls in PreFitler/Filter can use the context that is provided by scheduler framework
2. for calls outside PreFilter/Filter/..., we should instantiate one at the start of main() (or make some change in scheduler framework to expose it), and pass it to all plugins

zwpaper: created an issue to track this https://github.com/kubernetes-sigs/scheduler-plugins/issues/664

pkg/networkaware/topologicalsort/topologicalsort_test.go
zwpaper: the scheme name is conflict with the `var scheme` for init schemes, changed to keep consistence

PR #657, opened 2023-11-02
Instantiate one NRT informer across NRT plugin instances

### LINKED ISSUE #656: Multiple informers for the same resource(NodeResourceTopologies) may be serving at the same time
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

# Why could this happen?
Once we configure plugin `NodeResourceTopologyMatch` in multi profiles(must more than one), then informers about NodeResourceTopologies may start multi times with different ShardInformerFactory.

Since informer about NodeResourceTopologies is created within this plugin's func New() [here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/plugin.go#L121-L125) as following:
```go
func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error) {
	...
	nrtCache, err := initNodeTopologyInformer(tcfg, handle)
	...
```
And, informer about NodeResourceTopologies created from the `brandNew Factory` within func `initNodeTopologyInformer` [here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/pluginhelpers.go#L52-L54), as following:
```go
func initNodeTopologyInformer(tcfg *apiconfig.NodeResourceTopologyMatchArgs, handle framework.Handle) (nrtcache.Interface, error) {
	...
	topologyInformerFactory := topologyinformers.NewSharedInformerFactory(topoClient, 0)
	nodeTopologyInformer := topologyInformerFactory.Topology().V1alpha2().NodeResourceTopologies()
	nodeTopologyLister := nodeTopologyInformer.Lister()
        ...
}
```
Then, When we create scheduler profile one bye one, The func `New()` would be called many times, since plugin NodeResourceTopologyMatch exists in multiple profiles. Then multiple informers about NodeResourceTopologies will be produced.

# What are the negative effects?
- Data inconsistencies among those informers, even if the time difference is small.
- Waste of local storage: as multiple same indexer with the same content will exist

# How to prevent it?
Creating NewSharedInformerFactory When we create Scheduler instead of creating plugins. And we pass the created Informers about CRD from NewSharedInformerFactory to those needed plugins while it started.

_No response_

@Huang-Wei  Please see if this issue makes sense, thanks.

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

NoicFank:
#### What type of PR is this?
/kind bug

extract common custom sharedInformerFactory among plugins

Fixes #656 

```release-note
NONE

```

### DIFF
--- a/pkg/noderesourcetopology/plugin.go
+++ b/pkg/noderesourcetopology/plugin.go
@@ -31,6 +31,8 @@ import (
 
 	topologyapi "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology"
 	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	listerv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/listers/topology/v1alpha2"
+	"sync"
 )
 
 const (
@@ -106,6 +108,16 @@ func (tm *TopologyMatch) Name() string {
 	return Name
 }
 
+var (
+	// once guards singletonNodeTopologyLister to be initialized only once.
+	once sync.Once
+	// singletonNodeTopologyLister is a shared NRT lister across different NRT plugin instances.
+	// It could happen while NRT plugin is registered in multiple profiles.
+	singletonNodeTopologyLister listerv1alpha2.NodeResourceTopologyLister
+	// initializeListerErr indicates the error in the initialization singletonNodeTopologyLister.
+	initializeSingletonNodeTopologyListerErr error
+)
+
 // New initializes a new plugin and returns it.
 func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error) {
 	klog.V(5).InfoS("Creating new TopologyMatch plugin")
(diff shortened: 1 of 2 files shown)


pkg/noderesourcetopology/plugin.go
```diff
@@ -118,7 +119,7 @@ func New(args runtime.Object, handle framework.Handle) (framework.Plugin, error)
 		return nil, err
 	}
 
-	nrtCache, err := initNodeTopologyInformer(tcfg, handle)
+	nrtCache, err := initNodeTopologyInformer(tcfg, handle, topologyHandle)
```
Huang-Wei, line 122: I may prefer to use once.Do() to ensure initNodeTopologyInformer() is only called once, and then share the variable across different plugin instances.

NoicFank: yes, the idea is good, updated. PTAL, thanks.

pkg/noderesourcetopology/plugin.go
```diff
@@ -106,6 +108,16 @@ func (tm *TopologyMatch) Name() string {
 	return Name
 }
 
+var (
+	// once guards singletonNodeTopologyLister to be initialized only once.
+	once sync.Once
+	// singletonNodeTopologyLister is a shared NRT lister across different NRT plugin instances.
+	// It could happen while NRT plugin is registered in multiple profiles.
+	singletonNodeTopologyLister listerv1alpha2.NodeResourceTopologyLister
```
Huang-Wei, line 116: @NoicFank would you on-hold this change for a while? as another PR is refactoring the code use controller-runtime client, and once it's merged, this lister is no longer needed.

NoicFank: Okay! maybe I need to close this PR?

BTW, does using dynamic client may cause performance issues with scheduling, compared to using CRD informer?

Huang-Wei: for lister, it shouldn't - it's also retrieving objs from in-memory cache.

NoicFank: Ok, that's great.

Issue #662, opened 2023-11-10
Release 0.27


Huang-Wei:
```[tasklist]
# Tasks
- [x] cut new branch https://github.com/kubernetes-sigs/scheduler-plugins/tree/release-1.27
- [x] https://github.com/kubernetes/test-infra/pull/31260
- [x] tag cutting: https://github.com/kubernetes-sigs/scheduler-plugins/releases/tag/v0.27.8
- [x] https://github.com/kubernetes/k8s.io/pull/6096
- [ ] #669
- [x] Release page: https://github.com/kubernetes-sigs/scheduler-plugins/releases/tag/v0.27.8
- [x] Announcement email sent to sig-scheduling mailing group: https://groups.google.com/g/kubernetes-sig-scheduling/c/BczC-5aAlOw/m/l3qSLDrFAAAJ
```

## Refs: 

- [release process](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/RELEASE.md)
- [🔒 staging images](https://console.cloud.google.com/gcr/images/k8s-staging-scheduler-plugins/GLOBAL/kube-scheduler?gcrImageListsize=30&pli=1)
- [testgrid:post-scheduler-plugins-push-images](https://testgrid.k8s.io/sig-scheduling#post-scheduler-plugins-push-images)

Fixes #651

tenzen-y:
Hi, @Huang-Wei. Thanks for moving this forward. I'm looking forward to the new release.
BTW, do you have any estimated day to cut a new release?

Huang-Wei:
@tenzen-y today I'm gonna raise a final PR to migrate the typed client usage in capacityscheduling pkg to controller-runtime. After this is done, I will get the new release cut. ETA is end of this week.

Thanks for the patience ;)

tenzen-y:
> @tenzen-y today I'm gonna raise a final PR to migrate the typed client usage in capacityscheduling pkg to controller-runtime. After this is done, I will get the new release cut. ETA is end of this week.
> 
> Thanks for the patience ;)

Thanks for your effort!

Huang-Wei:
@tenzen-y you're good to go use v0.27.8 release.

tenzen-y:
> @tenzen-y you're good to go use v0.27.8 release.

Thank you so much!

PR #670, opened 2023-11-17
bump k8s deps to v1.28.4

Huang-Wei:
#### What type of PR is this?

/kind cleanup

With completion of #662, we should bump k8s deps in `master` to v1.28.

Fixes #

```release-note
bump k8s deps to v1.28.4
```


Huang-Wei:
The CI failure is valid. It hits an issue/regression on [RunScorePlugins](https://github.com/kubernetes/kubernetes/blob/v1.28.4/pkg/scheduler/framework/runtime/framework.go#L1019) of upstream:

```go
numPlugins := len(f.scorePlugins) - state.SkipScorePlugins.Len()
```

If a plugin implements both PreScore and Score, numPlugins' could be negative if:

- **only** Score is disabled
- `Skip` is returned in PreScore

Huang-Wei:
cc @denkensk

Huang-Wei:
(@denkensk was in a business trip)

@ffromani @zwpaper could you help review this PR? we should get it to master as early as we can. 🙇‍♂️

cmd/scheduler/main_test.go
```diff
@@ -497,18 +497,18 @@ profiles:
...
+			name:            "single profile config - PodState - v1beta3",
 			flags:           []string{"--config", podStateConfigFile},
 			registryOptions: []app.Option{app.WithPlugin(podstate.Name, podstate.New)},
 			wantPlugins: map[string]*config.Plugins{
 				"default-scheduler": {
-					PreEnqueue: defaults.PluginsV1beta2.PreEnqueue,
-					QueueSort:  defaults.PluginsV1beta2.QueueSort,
-					Bind:       defaults.PluginsV1beta2.Bind,
-					PostFilter: defaults.PluginsV1beta2.PostFilter,
+					PreEnqueue: defaults.ExpandedPluginsV1beta3.PreEnqueue,
```
ffromani, line 505: why are we moving from `Plugins*` to `ExpandedPlugins*`? I see other occurrences in the old code, but still curious

Huang-Wei: v1beta2 was removed in 1.28, so we have to turn to v1beta3. While in v1beta3, the variable `defaults.PluginsV1beta3` refers to a special plugin list with `MultiPoint` extension point:

```go
var PluginsV1beta3 = &config.Plugins{
	MultiPoint: config.PluginSet{...}
}
```

 so we have to use `ExpandedPluginsV1beta3` instead.

test/util/fake.go
```diff
@@ -146,45 +167,54 @@ func (n *nominator) add(pi *framework.PodInfo, nominatingInfo *framework.Nominat
...
 	}
-	delete(n.nominatedPodToNode, p.UID)
+	delete(npm.nominatedPodToNode, p.UID)
 }
 
 // UpdateNominatedPod updates the <oldPod> with <newPod>.
-func (n *nominator) UpdateNominatedPod(oldPod *v1.Pod, newPodInfo *framework.PodInfo) {
-	n.Lock()
-	defer n.Unlock()
+func (npm *nominator) UpdateNominatedPod(logger klog.Logger, oldPod *v1.Pod, newPodInfo *framework.PodInfo) {
```
denkensk, line 211: Do we need to replace 'n' with 'npm'? I recall that 'npm' is an abbreviation for 'nominatePodMap '. or just copy from upstream？

Huang-Wei: I didn't notice that - just copied from upstream.

Huang-Wei:
Thanks @ffromani and @denkensk .

/unhold

cmd/scheduler/main_test.go
ffromani: thanks for the clarification

PR #672, opened 2023-11-18
Coscheduling: Remove the postBind filter in the KubeSchedulerConfiguration

tenzen-y:
#### What type of PR is this?

/kind cleanup

I removed the PostBind filter in the KubeSchedulerConfiguration because the PostBind filter was removed in the https://github.com/kubernetes-sigs/scheduler-plugins/pull/554.

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/coscheduling/scheduler-config.yaml
+++ b/manifests/coscheduling/scheduler-config.yaml
@@ -24,9 +24,6 @@ profiles:
     reserve:
       enabled:
       - name: Coscheduling
-    postBind:
-      enabled:
-      - name: Coscheduling
   pluginConfig:
   - name: Coscheduling
     args:


tenzen-y:
I think that it would be better to cherry-pick this to the release-0.26 and release-0.27 branches.

Huang-Wei:
> I think that it would be better to cherry-pick this to the release-0.26 and release-0.27 branches.

Actually specifying a non-existent extension point would fail scheduler's startup. But yes, let's fix the documents. And, we should turn to using `MultiPoint` to simplify the scheduler config.

/lgtm
/approve

tenzen-y:
> > I think that it would be better to cherry-pick this to the release-0.26 and release-0.27 branches.
> 
> Actually specifying a non-existent extension point would fail scheduler's startup. But yes, let's fix the documents. And, we should turn to using `MultiPoint` to simplify the scheduler config.
> 
> /lgtm /approve

That makes sense.

Issue #682, opened 2023-12-04
Group of unschedulable pods causes scheduler (coscheduling) to be blocked


fg91:
### Area

- [X] Scheduler
- [X] Controller
- [X] Helm Chart
- [ ] Documents

### What happened?

We use the coscheduling plugin for ML training workloads.

Due to quota and resource restrictions, it regularly happens that the required nodes cannot be scaled up and a group of pending pod remains in the cluster.

Once that happens, sometimes we observe that the scheduler even fails to schedule pods for which resources would exist. We don't even see events on the respective pods.

This behaviour coincides with the scheduler emitting millions of log lines within a short period of time saying:

> `"To-activate pod does not exist in unschedulablePods or backoffQ"`

<img width="745" alt="Screenshot_2023-12-04_at_16_31_03" src="https://github.com/kubernetes-sigs/scheduler-plugins/assets/36511035/9ef0369c-f755-44f3-9441-d087552b8b27">

<img width="1120" alt="Screenshot_2023-12-04_at_16_30_01" src="https://github.com/kubernetes-sigs/scheduler-plugins/assets/36511035/33decd00-d3d9-4e47-b157-028f132b16a5">

At the same time, we observe a high increase in CPU usage:

<img width="732" alt="Screenshot 2023-12-04 at 16 34 14" src="https://github.com/kubernetes-sigs/scheduler-plugins/assets/36511035/29ba99fe-b09b-412d-8720-3ab23f0103de">

---

To make the scheduler schedule pods again, sometimes it's enough to restart the scheduler. Sometimes, however, we need to manually delete the pods/pod groups referenced in those millions of log lines.

---

To us it seems as if the issue isn't just starvation but generally the coscheduler not doing anything meaningful anymore.

### What did you expect to happen?

We would expect the coscheduler to continue to schedule pods (especially those not part of a pod group) for which resources do exist.

### How can we reproduce it (as minimally and precisely as possible)?

Unfortunately we haven't been able to create a minimal working example.

We tried debugging this issue by running the scheduler locally with a visual debugger attached. We did hit [the line logging the error](https://github.com/kubernetes/kubernetes/blob/55f2bc10435160619d1ece8de49a1c0f8fcdf276/pkg/scheduler/internal/queue/scheduling_queue.go#L611) with the debugger but don't know how to proceed from here to debug this further. We would be very keen to get hints.

### Kubernetes version

<details>

```console
[...]

PR #684, opened 2023-12-10
[DO NOT MERGE] experimental fix for coscheduling issue

Huang-Wei:
DO NOT MERGE.

This is an experimental fix for #682.


Huang-Wei:
Close in favor of #700

PR #685, opened 2023-12-12
NodeResourceTopology: add tests for score plugin with partial node data

ffromani:
#### What type of PR is this?

Add test for the NodeResourceTopology score plugin to ensure correct behavior with partial data available (some nodes have NRT data, some nodes don't)

/kind cleanup

Since the very beginning the NRT filter plugin wanted to deal with partial data (the filter plugin silently allows nodes without NRT data!) and we need to make sure the score plugin works as expected in this scenario. Let's start with test coverage.

Fixes N/A

N/A

```release-note
NONE
```

### DIFF
--- a/pkg/noderesourcetopology/score_test.go
+++ b/pkg/noderesourcetopology/score_test.go
@@ -41,80 +41,9 @@ const (
 
 type nodeToScoreMap map[string]int64
 
-func initTest(policy topologyv1alpha2.TopologyManagerPolicy, nodes ...*topologyv1alpha2.NodeResourceTopology) (map[string]*v1.Node, ctrlclient.Client) {
-	nodeTopologies := make([]*topologyv1alpha2.NodeResourceTopology, 3)
+func initTest(nodeTopologies []*topologyv1alpha2.NodeResourceTopology) (map[string]*v1.Node, ctrlclient.Client) {
 	nodesMap := make(map[string]*v1.Node)
 
-	if len(nodes) == 0 {
-		// noderesourcetopology objects
-		nodeTopologies[0] = &topologyv1alpha2.NodeResourceTopology{
-			ObjectMeta:       metav1.ObjectMeta{Name: "Node1"},
-			TopologyPolicies: []string{string(policy)},
-			Zones: topologyv1alpha2.ZoneList{
-				topologyv1alpha2.Zone{
-					Name: "node-0",
-					Type: "Node",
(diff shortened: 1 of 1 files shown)


ffromani:
/cc @PiotrProkop 

because I'm changing the LeastNUMA testcases. Tests should behave the same, though (no intended change of behavior).

ffromani:
/cc @Tal-or 

because he did major contributions to the score plugin

pkg/noderesourcetopology/score_test.go
```diff
@@ -133,7 +144,7 @@ func TestNodeResourceScorePlugin(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		nodesMap, lister := initTest(defaultNUMANodes(topologyv1alpha2.SingleNUMANodeContainerLevel))
+		nodesMap, lister := initTest(defaultNUMANodes(topologyv1alpha2.SingleNUMANodeContainerLevel), nrtPassthrough)
```
Tal-or, line 147: I believe more option would introduce in the future.
Hence, I would suggest to use functional options: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

Tal-or (commented):
LGTM
one comment inside

pkg/noderesourcetopology/score_test.go
Tal-or: The comment is not in place, I was referring to `defaultNUMANodes` function

ffromani:
/hold

to give time to all reviewers

pkg/noderesourcetopology/score_test.go
```diff
@@ -536,6 +472,144 @@ func TestNodeResourceScorePluginLeastNUMA(t *testing.T) {
...
+			wantedRes: nodeToScoreMap{},
+			requests:  pRequests,
+			strategy:  balancedAllocationScoreStrategy,
+			nrtFilter: func(nrt *topologyv1alpha2.NodeResourceTopology) *topologyv1alpha2.NodeResourceTopology {
+				return nil
+			},
+		},
+		{
+			name:      "One node with NRT data, MostAllocated strategy",
+			wantedRes: nodeToScoreMap{"Node1": 27},
```
PiotrProkop, line 531: it took me a while to spot you are leveraging the Go famous default value of 0 for nonexistent keys in map 😄

ffromani:
/hold cancel

got enough reviews, thanks!

Issue #690, opened 2023-12-26
KEP - GPUTopologyAware plugin


kerthcet:
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?

GPU communication bandwidth is affected a lot by the P2P link type, broader known as PCIe vs NVLink . As an user, I hope my GPU accelerated workloads can be placed with the best effort allocations between GPUs and nodes
across the cluster, which will speed up the experimental a lot.

/kind feature

### What did you expect to happen?

Best effort GPU allocation among a set of GPUs 

### Kubernetes version

None

### Scheduler Plugins version

None

kerthcet:
Will revisit this once finished other work, FYI: DRA will support this IIRC.

PR #691, opened 2023-12-26
[WIP] KEP: introduce GPUTopologyAware plugin

kerthcet:
#### What type of PR is this?
/kind feature

Xref https://github.com/kubernetes-sigs/scheduler-plugins/issues/690

```release-note
None
```


kerthcet:
/assign @Huang-Wei 
cc @lengrongfu

kep/690-gpu-topology-aware-scheduling/README.md
```diff
@@ -0,0 +1,473 @@
...
+    maxScore = limitQuota * (limitQuota - 1) * 1800 / 2 // 1800 is the maximum score between two GPUs, on behalf of 18 NVLinks
+    finalScore = originalScore * 100 / maxScore
+    ```
+
+A potential issue here is if the score of the link type is too small, like `P2PLinkCrossCPU` maps to 10, then we may loss the precision here
+and got 0 score finally.
+
+**In summary**
+
+Based on the same scoring algorithm, kube-scheduler will find the best-fit node, but no need to tell the device plugin which GPUs are they,
```
lianghao208, line 282: IIIUC, kube-scheduler reuses NVIDIA/go-gpuallocator algorithm to calculate the best GPU device id for allocation and score the nodes accordingly. But scheduler can't 100% sure which GPU will be used eventually, it "guesses" the result depends on the topology information reported by NFD, and device plugin will make its own decision, am I right? 
If so, does scheduler get to know which GPU has already been used?

kerthcet: Missed that part, will add it. Thanks.

KunWuLuan:
Scheduler may allocate pod in different sequence with device plugin, will this cause problem? For example, scheduler may allocate 4 GPUs for pod A and allocate 2 GPUs for another pod, but kubelet may handle pod with 2 GPUs first.

KunWuLuan:
Maybe we can implement a new plugin to ensure scheduler make same decision with kubelet, this new plugin can be a common plugin and can be used by many other plugins, like gpu topology or rdma topology. Then gpu topology just need to do its own decisions.

PR #692, opened 2023-12-28
CI: add releasing helm chart

zwpaper:
#### What type of PR is this?
/kind feature

release a helm chart when creating a tag/release as artifact, then we would host our helm chart repo on netlify page

Fixes https://github.com/kubernetes-sigs/scheduler-plugins/issues/607

this will create the helm chart as release artifact when creating v-prefix releases.

But I was thinking, the chart may change more frequently, maybe we should create some helm-chart-v-prefix releases? 

```release-note
release helm chart as artifact
```

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -1,0 +1,28 @@
+name: Releases
+on:
+  push:
+    tags:
+    - v*
+
+jobs:
+  releases:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Huang-Wei:
> But I was thinking, the chart may change more frequently, maybe we should create some helm-chart-v-prefix releases?

Let's stick to tag name for now, and change if needed in the future.

/lgtm
/approve

PR #700, opened 2024-01-30
Avoid unnecessary requeue operations in coscheduling

### LINKED ISSUE #682: Group of unschedulable pods causes scheduler (coscheduling) to be blocked

Huang-Wei:
#### What type of PR is this?

/kind bug

The current coscheduling code aggressively actives a pod's PodGroup siblings in `Permit()`. For example, if a deployment doesn't reach the minMember, all its pods are in pending state. Once it's scaled up to reach the minMember, it'd trigger `ActivateSiblings()` N times - this causes a ton of unnecessary re-queue operations which cause the CPU usage to spike, and may relates with potential starvation reported in #682.

Fixes #682

```release-note
Performance fix to eliminate unnecessary re-queue actions in coscheduling plugin
```

### DIFF
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -48,12 +48,22 @@ const (
 	PodGroupNotFound Status = "PodGroup not found"
 	Success          Status = "Success"
 	Wait             Status = "Wait"
+
+	preFilterStateKey = "PreFilterCoscheduling"
 )
+
+type PreFilterState struct {
+	Activate bool
+}
+
+func (s *PreFilterState) Clone() framework.StateData {
+	return &PreFilterState{Activate: s.Activate}
+}
 
 // Manager defines the interfaces for PodGroup management.
 type Manager interface {
 	PreFilter(context.Context, *corev1.Pod) error
-	Permit(context.Context, *corev1.Pod) Status
+	Permit(context.Context, *framework.CycleState, *corev1.Pod) Status
 	GetPodGroup(context.Context, *corev1.Pod) (string, *v1alpha1.PodGroup)
 	GetCreationTimestamp(*corev1.Pod, time.Time) time.Time
 	DeletePermittedPodGroup(string)
@@ -105,6 +115,13 @@ func (pgMgr *PodGroupManager) BackoffPodGroup(pgName string, backoff time.Durati
 func (pgMgr *PodGroupManager) ActivateSiblings(pod *corev1.Pod, state *framework.CycleState) {
 	pgName := util.GetPodGroupLabel(pod)
 	if pgName == "" {
+		return
+	}
+
+	// Only proceed if it's explicitly requested to activate sibling pods.
+	if c, err := state.Read(preFilterStateKey); err != nil {
+		return
+	} else if s, ok := c.(*PreFilterState); !ok || !s.Activate {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


Huang-Wei:
cc @denkensk

pkg/coscheduling/core/core.go
```diff
@@ -209,6 +226,19 @@ func (pgMgr *PodGroupManager) Permit(ctx context.Context, pod *corev1.Pod) Statu
 	if int32(assigned)+1 >= pg.Spec.MinMember {
 		return Success
 	}
+
+	if assigned == 0 {
```
kerthcet, line 230: Based on my understanding, when `assigned == 0`, we should and only should trigger the `activating`, then in `ActivateSiblings`, once `Activate is Ture`, we should start the trigger but now we're the opposite. And what's more, shouldn't we set the `Activate to False` in the follow up.

kerthcet (commented):
One question about the implementation in the comment.

pkg/coscheduling/core/core.go
Huang-Wei: > when assigned == 0, we should and only should trigger the activating, then in ActivateSiblings, once Activate is Ture, we should start the trigger but now we're the opposite.

That's the current case. Why do you think it's the opposite.

> shouldn't we set the Activate to False in the follow up.

Nope b/c every scheduling cycle comes with its own CycleState instance.

pkg/coscheduling/core/core.go
```diff
@@ -48,12 +48,22 @@ const (
 	PodGroupNotFound Status = "PodGroup not found"
 	Success          Status = "Success"
 	Wait             Status = "Wait"
+
+	preFilterStateKey = "PreFilterCoscheduling"
```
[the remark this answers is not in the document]
Huang-Wei: Yes, good catch...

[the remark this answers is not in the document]
Huang-Wei: Updated. PTAL.

pkg/coscheduling/core/core.go
kerthcet: Make sense.
/lgtm
/hold cancel

PR #704, opened 2024-02-28
update the second scheduler install docs

runzhliu:
git clone by https would be more general.

#### What type of PR is this?

/kind documentation

Fixes #

```release-note

```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/README.md
+++ b/manifests/install/charts/as-a-second-scheduler/README.md
@@ -24,7 +24,7 @@ Quick start instructions for the setup and configuration of as-a-second-schedule
 #### Install chart using Helm v3.0+
 
 ```bash
-$ git clone git@github.com:kubernetes-sigs/scheduler-plugins.git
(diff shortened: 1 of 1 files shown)


manifests/install/charts/as-a-second-scheduler/README.md
```diff
@@ -24,7 +24,7 @@ Quick start instructions for the setup and configuration of as-a-second-schedule
 #### Install chart using Helm v3.0+
 
 ```bash
-$ git clone git@github.com:kubernetes-sigs/scheduler-plugins.git
+$ git clone https://github.com/kubernetes-sigs/scheduler-plugins.git
```
Huang-Wei, line 27: I don't think https is the conventional way to git clone.

runzhliu: The prerequisite for using ssh is that your machine has been configured with ssh to access github. After referring to the document, the user will most likely need to change it to https. It is better to give an example of https directly in the document, BTW, this is what I experienced.

Issue #581, opened 2023-04-26
Preemption for Coscheduling


KunWuLuan:
I noticed that when using coscheduling, preempt only one pod a time is not enough.

So I am considering about implement a plugin to support preempt more than one pod a time.

I have implemented a prototype of batch preemption. The batch preemption plugin (will be called BP) will interact with other plugin. BP provide an interface with 2 functions for other plugins: 1. to query the list of pods ready for preemption and 2. do preemption if current pod schedule successfully or preempt successfully.
If a pod find some victims on a node, the plugin will record the result and do nothing.If some other plugins decide to do the preemption after a successful scheduling or successful preemption, they can call the interface, then BP will preempt all victims it has recorded. 
Of course, the prototype has already taken into account the situation where the Pod in the record is preempted again by other Pods, or the Pod in the record is rescheduled. This is a bit complicated. If necessary, I will add it later

The point is that I believe the user interaction in the prototype is unreasonable.
In the prototype, BP's behavior is hard coded. I think we may need a CRD for BP to decide when to do preemption. Do you have some ideas?

JasonChen86899:
It's a good idea I feel like there can be two ways

1. Create your own plugin similar to what you mentioned
2. Modify the preemptive logic of the scheduling framework to support plug-in transparent parameters to it

shadowdsp:
Hi @KunWuLuan , is there any ideas on this issue?
Currently, we are encountering the problem that after postfilter checks the entire podgroup, it cannot set nominated node in batches.

KunWuLuan:
> Hi @KunWuLuan , is there any ideas on this issue? Currently, we are encountering the problem that after postfilter checks the entire podgroup, it cannot set nominated node in batches.

@shadowdsp Yes, currently the preemption plugin will only set nominated node for the pod that triggers the preemption. We have a custom plugin to solve the problem, but currently it seems that the plugin is for the specific situation.

Issue #560, opened 2023-03-29
[coscheduling] support match policy for podgroup


KunWuLuan:
For GPU topology, all my tasks need to be scheduled at the same time to ensure the best performance, so I hope that gang can only be satisfied when the number of pods waiting on permit is larger than the min-available.
For task of elastic training, the new workers'' arrival is random. So I hope the gang can be satisfied when the total number of the running pods and waiting pods is large than the min-available.
In some specific scenarios, the completed task Pod may not exist due to node recycling. At this time, it is impossible to count the number of completed tasks. In this case, OnceResourceSatisfied can be used.

We can use `scheduling.sigs.x-k8s.io/pod-group-matchpolicy` to declare which pods should be considered in permit.

Huang-Wei:
> We can use `scheduling.sigs.x-k8s.io/pod-group-matchpolicy` to declare which pods should be considered in permit.

IMO a field in PodGroup API is a better place?

cc @denkensk

KunWuLuan:
A new property in spec is also feasible I think

denkensk:
>  So I hope the gang can be satisfied when the total number of the running pods and waiting pods is large than the min-available.

Isn't that what we have now?

> scheduling.sigs.x-k8s.io/pod-group-matchpolicy

What value should we fill in？ @KunWuLuan

KunWuLuan:
> Isn't that what we have now?

yes, we have already support this case, and I think this can be the default behavior

> What value should we fill in？ @KunWuLuan

I think we can support three values for three cases I mentioned above, like: "only-waiting", "waiting-and-running" and "oncesatisfied".

Or we can let user choose the pod status they need, like: "scheduled", "running", "succeed", "failed", "oncesatisfied". For example, they can do nothing or use "scheduling.sigs.x-k8s.io/pod-group-matchpolicy"="scheduled,running" to achieve the same behavior as what we have now.  @denkensk

denkensk:
> I think we can support three values for three cases I mentioned above, like: "only-waiting", "waiting-and-running" and "oncesatisfied".

> Or we can let user choose the pod status they need, like: "scheduled", "running", "succeed", "failed", "oncesatisfied". For example, they can do nothing or use "scheduling.sigs.x-k8s.io/pod-group-matchpolicy"="scheduled,running" to achieve the same behavior as what we have now.


[...]

PR #708, opened 2024-03-28
cache assigned pod count

### LINKED ISSUE #707: [Coscheduling] CalculateAssignedPods cost too much time
PodGroup manager will try every node and every pod in CalculateAssignedPods()

https://github.com/kubernetes-sigs/scheduler-plugins/blob/cd3e4fb738722a7e07b10e4be93eff7285f39f73/pkg/coscheduling/core/core.go#L246-L264

When there are 5k or more nodes and 50000+ pods in cluster, CalculateAssignedPods() will cost 30+ms. 
We should cache the number of assigned pods to save time in scheduling cycle.

KunWuLuan:
#### What type of PR is this?

/kind feature

This PR will enhance the speed of the Coscheduling plugin in counting Pods that have already been assumed.

Fix #707

```release-note
Speedup the calculation of assumed pods number in coscheduling plugin.
```

### DIFF
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -28,8 +28,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	informerv1 "k8s.io/client-go/informers/core/v1"
 	listerv1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -64,10 +66,11 @@ func (s *PermitState) Clone() framework.StateData {
 type Manager interface {
 	PreFilter(context.Context, *corev1.Pod) error
 	Permit(context.Context, *framework.CycleState, *corev1.Pod) Status
+	Unreserve(context.Context, *corev1.Pod)
 	GetPodGroup(context.Context, *corev1.Pod) (string, *v1alpha1.PodGroup)
+	GetAssignedPodCount(string) int
 	GetCreationTimestamp(context.Context, *corev1.Pod, time.Time) time.Time
 	DeletePermittedPodGroup(context.Context, string)
-	CalculateAssignedPods(context.Context, string, string) int
 	ActivateSiblings(ctx context.Context, pod *corev1.Pod, state *framework.CycleState)
 	BackoffPodGroup(string, time.Duration)
 }
@@ -87,7 +90,32 @@ type PodGroupManager struct {
 	backedOffPG *gocache.Cache
 	// podLister is pod lister
 	podLister listerv1.PodLister
+	// assignedPodsByPG stores the pods assumed or bound for podgroups
+	assignedPodsByPG map[string]sets.Set[string]
 	sync.RWMutex
+}
+
+func AddPodFactory(pgMgr *PodGroupManager) func(obj interface{}) {
+	return func(obj interface{}) {
+		p, ok := obj.(*corev1.Pod)
+		if !ok {
+			return
+		}
+		if p.Spec.NodeName == "" {
+			return
+		}
+		pgFullName, _ := pgMgr.GetPodGroup(context.Background(), p)
+		if pgFullName == "" {
+			return
+		}
+		pgMgr.RWMutex.Lock()
+		defer pgMgr.RWMutex.Unlock()
+		if assigned, exist := pgMgr.assignedPodsByPG[pgFullName]; exist {
+			assigned.Insert(p.Name)
+		} else {
+			pgMgr.assignedPodsByPG[pgFullName] = sets.New(p.Name)
+		}
+	}
 }
 
 // NewPodGroupManager creates a new operation object.
@@ -99,8 +127,41 @@ func NewPodGroupManager(client client.Client, snapshotSharedLister framework.Sha
 		podLister:            podInformer.Lister(),
 		permittedPG:          gocache.New(3*time.Second, 3*time.Second),
 		backedOffPG:          gocache.New(10*time.Second, 10*time.Second),
+		assignedPodsByPG:     map[string]sets.Set[string]{},
 	}
+	podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: AddPodFactory(pgMgr),
+		DeleteFunc: func(obj interface{}) {
+			switch t := obj.(type) {
+			case *corev1.Pod:
+				pod := t
+				if pod.Spec.NodeName == "" {
+					return
+				}
+				pgMgr.Unreserve(context.Background(), pod)
+				return
+			case cache.DeletedFinalStateUnknown:
+				pod, ok := t.Obj.(*corev1.Pod)
+				if !ok {
+					return
+				}
+				if pod.Spec.NodeName == "" {
+					return
+				}
+				pgMgr.Unreserve(context.Background(), pod)
+				return
+			default:
+				return
+			}
+		},
+	})
 	return pgMgr
+}
+
+func (pgMgr *PodGroupManager) GetAssignedPodCount(pgName string) int {
+	pgMgr.RWMutex.RLock()
+	defer pgMgr.RWMutex.RUnlock()
+	return len(pgMgr.assignedPodsByPG[pgName])
 }
 
 func (pgMgr *PodGroupManager) BackoffPodGroup(pgName string, backoff time.Duration) {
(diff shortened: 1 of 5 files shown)


pkg/coscheduling/core/core.go
```diff
@@ -89,10 +98,37 @@ func NewPodGroupManager(client client.Client, snapshotSharedLister framework.Sha
...
 		permittedPG:          gochache.New(3*time.Second, 3*time.Second),
 		backedOffPG:          gochache.New(10*time.Second, 10*time.Second),
+		podgroups:            map[string]PodGroupInfo{},
 	}
+	podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			switch t := obj.(type) {
+			case *corev1.Pod:
+				pod := t
+				pgMgr.Unreserve(context.Background(), pod)
```
Huang-Wei, line 108: PodDelete event consists of 3 types of events:

- Pod failed
- Pod completed (successfully)
- Pod get deleted

but for completed Pod, we should still count them as part of gang, right? could you also help if integration test covers this case?

Huang-Wei (commented):
Could you help fix the CI failures?

pkg/coscheduling/core/core.go
KunWuLuan: When pod completed, it will be removed from NodeInfo. CalculateAssignedPods will count pods in NodeInfo, so we did not count completed pods previously.

KunWuLuan: Ok, I will see if integration test covers this case

pkg/coscheduling/core/core.go
```diff
@@ -203,15 +239,34 @@ func (pgMgr *PodGroupManager) Permit(ctx context.Context, pod *corev1.Pod) Statu
...
 	// The number of pods that have been assigned nodes is calculated from the snapshot.
 	// The current pod in not included in the snapshot during the current scheduling cycle.
-	if int32(assigned)+1 >= pg.Spec.MinMember {
+	if len(pgInfo.assigned) >= int(pg.Spec.MinMember) {
 		return Success
 	}
 	return Wait
 }
 
+func (pgMgr *PodGroupManager) Unreserve(ctx context.Context, pod *corev1.Pod) {
```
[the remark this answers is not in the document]
KunWuLuan: Yes, thanks

pkg/coscheduling/core/core.go
```diff
@@ -203,15 +267,40 @@ func (pgMgr *PodGroupManager) Permit(ctx context.Context, pod *corev1.Pod) Statu
...
 	// The number of pods that have been assigned nodes is calculated from the snapshot.
 	// The current pod in not included in the snapshot during the current scheduling cycle.
-	if int32(assigned)+1 >= pg.Spec.MinMember {
+	if len(assigned) >= int(pg.Spec.MinMember) {
 		return Success
 	}
 	return Wait
 }
 
+// Unreserve removes assigned pod from asssigned pod map from PodGroupInfo when schedule or bind failed.
```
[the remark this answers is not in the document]
Huang-Wei: Unreserve invalidates assigned pod from `assignedPodsByPG` when schedule or bind failed.

pkg/coscheduling/core/core.go
```diff
@@ -77,9 +80,34 @@ type PodGroupManager struct {
 	backedOffPG *gochache.Cache
 	// podLister is pod lister
 	podLister listerv1.PodLister
+	//
```
Huang-Wei, line 83: add a comment?

pkg/coscheduling/core/core.go
Huang-Wei: > so we did not count completed pods previously.

True. I'm wondering if we fix this glitch in this PR - in DeleteFunc(), **additionally** check if the Pod is completed, if so, do NOT invalidate it from the `assignedPodsByPG` cache. WDYT?

pkg/coscheduling/core/core.go
```diff
@@ -89,10 +117,46 @@ func NewPodGroupManager(client client.Client, snapshotSharedLister framework.Sha
...
 	return pgMgr
 }
 
+func (pgMgr *PodGroupManager) GetAssignedPodCount(pgName string) int {
+	pgMgr.RWMutex.RLock()
+	defer pgMgr.RWMutex.RUnlock()
+	if assigned, exist := pgMgr.assignedPodsByPG[pgName]; exist {
+		return len(assigned)
+	}
+	return 0
```
Huang-Wei, line 157: nit: if not found, the zero value of pgMgr.assignedPodsByPG[pgName] is nil, we can simply:

```go
return len(pgMgr.assignedPodsByPG[pgName])
```

hack/install-envtest.sh
```diff
@@ -32,6 +32,6 @@ version=$(cat ${SCRIPT_ROOT}/go.mod | grep 'k8s.io/kubernetes' | grep -v '=>' |
 
 GOPATH=$(go env GOPATH)
 TEMP_DIR=${TMPDIR-/tmp}
-go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+go install sigs.k8s.io/controller-runtime/tools/setup-envtest@release-0.17
```
Huang-Wei, line 35: you can drop this change now and rebase master.

Huang-Wei (commented):
I forgot one thing about the cache's consistency during one scheduling cycle - we will need to:

- snapshot the pg->podNames map at the beginning of the scheduling cycle (PreFilter), so that we can treat it as source of truth during the whole scheduling cycle
- support preemption
  - implement the Clone() function
  - for each PodAddition dryrun, if the pod is hit, add it
  - for each PodDeletion dryrun, if the pod is hit, remove it

pkg/coscheduling/core/core.go
KunWuLuan: https://github.com/kubernetes-sigs/scheduler-plugins/issues/560#issuecomment-1495316959

We have discussed in this issues about whether we should count completed pods. 
Is there new situation to count completed pods?

hack/install-envtest.sh
KunWuLuan: done

pkg/coscheduling/core/core.go
KunWuLuan: fixed

pkg/coscheduling/core/core.go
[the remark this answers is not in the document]
KunWuLuan: fixed

pkg/coscheduling/core/core.go
KunWuLuan: added a comment

KunWuLuan:
> I forgot one thing about the cache's consistency during one scheduling cycle - we will need to:
> 
> * snapshot the pg->podNames map at the beginning of the scheduling cycle (PreFilter), so that we can treat it as source of truth during the whole scheduling cycle
> * support preemption
>   
>   * implement the Clone() function
>   * for each PodAddition dryrun, if the pod is hit, add it
>   * for each PodDeletion dryrun, if the pod is hit, remove it

We only check the number of pods assigned in Permit, so I think there is no inconsistency during one scheduling cycle.

And postFilter will not check Permit plugin, so implementation of PodAddition and PodDeletion will have no effect on preemption, right?

What we can do is return framework.Unschedulable if the PodDeletion will make a podgroup rejected, but I think it is not enought for preemption of coscheduling.

I think support preemption for coscheduling is complecated, maybe in another issue. 
We can determine the expected behaviro for preemption of coscheduling. WDYT?
#581

pkg/coscheduling/core/core.go
Huang-Wei: I see. It seems restart the whole Job is more conventional for now, then let's postpone the idea until new requirement emerges.

Huang-Wei:
> And postFilter will not check Permit plugin, so implementation of PodAddition and PodDeletion will have no effect on preemption, right?

Yes, the current preemption skeleton code assumes each plugin only use PreFilter to pre-calculate state. But for coscheduling, PreFilter can fail early (upon inadequate quorum).

I think scheduler framework should open up a hook for out-of-tree plugin to choose whether or not to run PreFilter as part of the preemption; otherwise, out-of-tree plugin has to rewrite the PostFilter impl. to hack that part.

> I think support preemption for coscheduling is complecated, maybe in another issue.
We can determine the expected behaviro for preemption of coscheduling. WDYT?

Let's consolidate all the cases and use a new PR to try to tackle it. Thanks.

Huang-Wei:
@KunWuLuan are you ok with postpone this PR's merge after I cut release for v0.28, so that we have more time for soak testing.

And could you add a release-note to highlight it's a performance enhancement?

pkg/coscheduling/core/core.go
KunWuLuan: sure

KunWuLuan:
> @KunWuLuan are you ok with postpone this PR's merge after I cut release for v0.28, so that we have more time for soak testing.
> 
> And could you add a release-note to highlight it's a performance enhancement?

Ok, no problem.

KunWuLuan:
> > And postFilter will not check Permit plugin, so implementation of PodAddition and PodDeletion will have no effect on preemption, right?
> 
> Yes, the current preemption skeleton code assumes each plugin only use PreFilter to pre-calculate state. But for coscheduling, PreFilter can fail early (upon inadequate quorum).
> 
> I think scheduler framework should open up a hook for out-of-tree plugin to choose whether or not to run PreFilter as part of the preemption; otherwise, out-of-tree plugin has to rewrite the PostFilter impl. to hack that part.
> 
> > I think support preemption for coscheduling is complecated, maybe in another issue.
> > We can determine the expected behaviro for preemption of coscheduling. WDYT?
> 
> Let's consolidate all the cases and use a new PR to try to tackle it. Thanks.

Ok. I will try to design a preemption framework in postFilter, and if implementation in postFilter is enough, I will create a new pr to track the kep. Otherwise I will try to open a discuss in kubernetes/scheduling-sigs.

Huang-Wei:
@KunWuLuan could you resolve the conflicts? and it'll be good to be merged afterwards

KunWuLuan:
@Huang-Wei Hi, I have resolved the conflicts and make the tests passed. Please have a look when you have time. Thanks

Huang-Wei (commented):
/lgtm
/approve

Thanks @KunWuLuan

Issue #667, opened 2023-11-17
Inappropriate error logging in capacity_scheduling.go


Huang-Wei:
https://github.com/kubernetes-sigs/scheduler-plugins/blob/fc5b4b296f6725c8ead647b9d9cca32f0118c642/pkg/capacityscheduling/capacity_scheduling.go#L404-L407

`ErrorS` doesn't seem appropriate as once a Pod is unschedulable, it will hit this line and error out. Changing to V(5) InfoS makes more sense.

/good-first-issue

googs1025:
/close 
already fixed

PR #726, opened 2024-05-01
fix log print and some nit issue

googs1025:
#### What type of PR is this?
/kind cleanup

- fix some nit issue

Fixes #
https://github.com/kubernetes-sigs/scheduler-plugins/issues/667

```release-note

```

### DIFF
--- a/pkg/noderesourcetopology/README.md
+++ b/pkg/noderesourcetopology/README.md
@@ -262,7 +262,7 @@ zones:
          $ kubectl get noderesourcetopologies.topology.node.k8s.io
        ```
 
-    1. Alternatively, in case you are just interested in simply testing the scheduler plugin, use the manifest in the manifest directory to deploy the CRD and CRs as follows:
+    2. Alternatively, in case you are just interested in simply testing the scheduler plugin, use the manifest in the manifest directory to deploy the CRD and CRs as follows:
 
          1. Deploy the Custom Resource Definition manifest
 
@@ -270,14 +270,14 @@ zones:
             $ kubectl create -f crd.yaml
             ```
 
-         1. Check if the noderesourcetopologies.topology.node.k8s.io CRD is created
+         2. Check if the noderesourcetopologies.topology.node.k8s.io CRD is created
 
             ```bash
              $ kubectl get crd
              $ kubectl get noderesourcetopologies.topology.node.k8s.io
             ```
 
-         1. Deploy the CRs representative of the hardware topology of the worker-node-A and worker-node-B if CRs haven't been created using RTE or NFD as mentioned above:
+         3. Deploy the CRs representative of the hardware topology of the worker-node-A and worker-node-B if CRs haven't been created using RTE or NFD as mentioned above:
 
             ```bash
              $ kubectl create -f worker-node-A.yaml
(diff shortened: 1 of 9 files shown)


pkg/noderesourcetopology/README.md
```diff
@@ -262,22 +262,22 @@ zones:
          $ kubectl get noderesourcetopologies.topology.node.k8s.io
        ```
 
-    1. Alternatively, in case you are just interested in simply testing the scheduler plugin, use the manifest in the manifest directory to deploy the CRD and CRs as follows:
+    2. Alternatively, in case you are just interested in simply testing the scheduler plugin, use the manifest in the manifest directory to deploy the CRD and CRs as follows:
```
ffromani, line 265: markdown renderer should automatically convert the list index without the need for humans to enforce the correct ordering: https://www.markdownguide.org/basic-syntax/#ordered-lists

pkg/trimaran/resourcestats.go
```diff
@@ -121,7 +121,7 @@ func GetResourceLimits(pod *v1.Pod) *framework.Resource {
 	})
 }
 
-// GetEffectiveResource: calculate effective resources of a pod (CPU and Memory)
+// GetEffectiveResource : calculate effective resources of a pod (CPU and Memory)
```
ffromani, line 124: why the extra space?

ffromani (commented):
/ok-to-test

I'm not a fan of commits which fixes so many different issues in one go - unnecessarily hard to describe and bisect. But I won't block it either, let's see what other reviewers think.
However it seems some of the proposed changes can be simplified, please check inline comments

googs1025:
> /ok-to-test
> 
> I'm not a fan of commits which fixes so many different issues in one go - unnecessarily hard to describe and bisect. But I won't block it either, let's see what other reviewers think. However it seems some of the proposed changes can be simplified, please check inline comments

Yes, normally I wouldn't submit a pull request like that. I just thought that these changes were all trivial, so it might be more convenient to merge them into a single pull request.

pkg/noderesourcetopology/README.md
googs1025: got it. done

pkg/trimaran/resourcestats.go
googs1025: I noticed that the method annotations in this directory were all like that. I simply reverted the changes to maintain code cleanliness. I have already rolled back the changes.
https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/trimaran/resourcestats.go#L77
```go
// GetResourceRequested : calculate the resource requests of a pod (CPU and Memory)
func GetResourceRequested(pod *v1.Pod) *framework.Resource {
	return GetEffectiveResource(pod, func(container *v1.Container) v1.ResourceList {
		return container.Resources.Requests
	})
}

// GetResourceLimits : calculate the resource limits of a pod (CPU and Memory)
func GetResourceLimits(pod *v1.Pod) *framework.Resource {
	return GetEffectiveResource(pod, func(container *v1.Container) v1.ResourceList {
		return container.Resources.Limits
	})
}

```

googs1025:
@ffromani /PTAL thanks a lot

googs1025:
@Huang-Wei /PTAL thanks!

Huang-Wei (commented):
I agree with @ffromani that PRs should be organized by similar themes: like typo fixes, klog standardization, code de-duplication, etc.

But I don't want to block the merge either... Just next time please separate the changes instead of submitting a spaghetti-like one.

/lgtm
/approve

Issue #732, opened 2024-05-07
update-codegen.sh does not work on M1 mac out of the box


aariv1:
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

codegen

### What happened?

This is more of an annoyance but when you execute ./hack/update-codegen.sh on m1 mac it by default fails because realpath fails if the path you are requesting does not exist. The error in this case is simply:
`realpath: ./hack/tools: No such file or directory`

### What did you expect to happen?

Expectation is that you could run the update-codegen.sh without needing to create the tools directory beforehand

For example something like this: 

`TOOLS_DIR=$(realpath ./hack)
TOOLS_DIR+="/tools"`

### How can we reproduce it (as minimally and precisely as possible)?

On at least m1 mach clone the repo and execute hack/update-codegen.sh

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>
release-1.28
</details>

Huang-Wei:
@aariv1 does adding the following line work for you locally? (if so, feel free to raise a PR)

```diff
diff --git a/hack/update-codegen.sh b/hack/update-codegen.sh
index 6303a29c..0f626256 100755
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -19,6 +19,7 @@ set -o nounset
 set -o pipefail
 
 SCRIPT_ROOT=$(dirname "${BASH_SOURCE[@]}")/..
+mkdir -p "{SCRIPT_ROOT}/hack/tools/bin"
 
 TOOLS_DIR=$(realpath ./hack/tools)
 TOOLS_BIN_DIR="${TOOLS_DIR}/bin"
```

googs1025:
@aariv1 I encountered the same issue on my Mac, but when I used Ubuntu, everything worked fine. Maybe you could try using Linux to see if it helps.

aariv1:
Adding `mkdir -p "${SCRIPT_ROOT}/hack/tools/bin"` does work (although the codegen is still broken https://github.com/kubernetes-sigs/scheduler-plugins/issues/729). I'll raise a pr anyway. Thank you

clubanderson:
applied this fix and then received when the script ran 

```
${CONTROLLER_GEN} object:headerFile="hack/boilerplate/boilerplate.generatego.txt" \
  paths="./apis/scheduling/..."
```

[...]

PR #733, opened 2024-05-08
Fix update-codegen.sh on macOS

aariv1:
#### What type of PR is this?

/kind bug

realpath on macOS fails if the file or directory referenced does not exist. hack/update-codegen.sh fails by default on mac because the hack/tools path does not exist when we try to set it to variable with realpath `TOOLS_DIR=$(realpath ./hack/tools)`. This fix simply creates the hack/tools and hack/tools/bin directories before they are referenced

Fixes [#](https://github.com/kubernetes-sigs/scheduler-plugins/issues/732)

```release-note
NONE
```

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -19,6 +19,7 @@ set -o nounset
 set -o pipefail
 
 SCRIPT_ROOT=$(dirname "${BASH_SOURCE[@]}")/..
+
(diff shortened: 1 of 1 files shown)


googs1025:
@aariv1 Let's sign cla first

Huang-Wei:
@aariv1 could you sign the CLA? otherwise we have no way to merge this PR.

PR #736, opened 2024-05-14
docs: anyone should be any

vsoch:
Problem: the comment and associated docs should say if there are not resources available needed for the group, the scheduler should not start any.

Solution: update all references anyone->any.

It could also be stated "not start any one" but "not start any" is a more natural way to say it

#### What type of PR is this?

/kind documentation

It's a subtle word choice mistake.

None - did not open an issue for something so small.

```release-note
NONE
```


ffromani:
/ok-to-test
/lgtm

I'm not a native english speaker, but this reads better and AFAIK is indeed the correct way

Issue #734, opened 2024-05-11
replace deprecated methods


googs1025:
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?
I see that the `wait.Poll `method is still widely used in unit tests and integration tests. This method has been deprecated in favor of `PollUntilContextTimeout`.
We should replace this method
```
// Deprecated: This method does not return errors from context, use PollUntilContextTimeout.
// Note that the new method will no longer return ErrWaitTimeout and instead return errors
// defined by the context package. Will be removed in a future release.
func Poll(interval, timeout time.Duration, condition ConditionFunc) error {
	return PollWithContext(context.Background(), interval, timeout, condition.WithContext())
}
```

### What did you expect to happen?

replace `Poll` method to `PollUntilContextTimeout`

### Kubernetes version

None
<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

None
<details>

</details>

PR #739, opened 2024-05-16
replace deprecated method

googs1025:
#### What type of PR is this?
/kind cleanup

`Poll` method has been deprecated in favor of `PollUntilContextTimeout`.

Fixes #https://github.com/kubernetes-sigs/scheduler-plugins/issues/734#event-12776086054

None

```release-note
None
```

### DIFF
--- a/pkg/controllers/elasticquota_controller_test.go
+++ b/pkg/controllers/elasticquota_controller_test.go
@@ -203,8 +203,7 @@ func TestElasticQuotaController_Run(t *testing.T) {
 					t.Errorf("reconcile: (%v)", err)
 				}
 			}
-
-			err := wait.Poll(200*time.Millisecond, 1*time.Second, func() (done bool, err error) {
+			err := wait.PollUntilContextTimeout(ctx, 200*time.Millisecond, 1*time.Second, false, func(ctx context.Context) (done bool, err error) {
 				for _, v := range c.want {
 					eq := &v1alpha1.ElasticQuota{
 						ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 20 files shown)


googs1025:
@Huang-Wei @ffromani Could you help me review this PR? I'm not sure if this change is needed

pkg/controllers/elasticquota_controller_test.go
```diff
@@ -203,25 +203,25 @@ func TestElasticQuotaController_Run(t *testing.T) {
...
-						ObjectMeta: metav1.ObjectMeta{
-							Name:      v.Name,
-							Namespace: v.Namespace,
-						},
-					}
-					err := kClient.Get(ctx, client.ObjectKeyFromObject(eq), eq)
-					if err != nil {
-						return false, err
+			err := wait.PollUntilContextTimeout(ctx, 200*time.Millisecond, 1*time.Second,
+				false, func(ctx context.Context) (done bool, err error) {
```
ffromani, line 207: can we keep the function signature in just one (admittedly long) line? that would make the review much easier because only that line should change

ffromani: Here and everywhere else

ffromani (commented):
this change my be not strictly needed but the proposed and suggested replacement seems better according to every metric so I'd be in favor of this change

pkg/controllers/elasticquota_controller_test.go
googs1025: Okay, thanks for reminding this, it helps the readability of the review. done!

ffromani (commented):
/lgtm
thanks for the updates

Huang-Wei (commented):
/approve

PR #752, opened 2024-05-27
[noderesourcetopology] rewrite accounting of numa-affine resources with scope=container

### LINKED ISSUE #751: [noderesourcetopology] incorrect accounting of hostlevel resources (ephemeral storage) with scope=container
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

After #735 and #646 we performed further investigation in how we handle host-level resources, like ephemeral storage.
"host-level resources" are resources which don't have NUMA affinity so don't contribute to the noderesourcetopology decisions, at least in theory.
In #646 we added a fix to ignore the ephemeral storage (being a common example of hostlevel resource) in the filtering proper, which was indeed a needed fix; this however uncovered another issue in how we do resource accounting for hostlevel resources; When dealing with multiple containers, each requiring ephemeral storage, a couple of flaws in the `subtractFromNUMA` function causes:
1. When processing the first container the resource to be incorrectly subtracted from the counters for a NUMA cell. But since it is a hostlevel resource, the counters go negative, an [occurrence the code warns for, but does not treat as error](https://github.com/kubernetes-sigs/scheduler-plugins/blob/v0.28.9/pkg/noderesourcetopology/filter.go#L243)
2. When processing the subsequent containers, we now have unexpected entries in the per-NUMA resource counters, which cause [the guard to fail](https://github.com/kubernetes-sigs/scheduler-plugins/blob/v0.28.9/pkg/noderesourcetopology/filter.go#L120) and, since the resource cannot have NUMA affinity, to end up with [incorrect NUMA availability bitmask, leading to bad scheduling decision](https://github.com/kubernetes-sigs/scheduler-plugins/blob/v0.28.9/pkg/noderesourcetopology/filter.go#L142)

The proper fix is to review the way we treat these resources from the ground up, adding explicit handling all along the way.

host-level resources like ephemeral storage should never contribute to the noderesourcetopology scheduling decisions

1. configure topology manager with scope=container
2. make sure the kubelet exposes the ephemeral storage resource
3. create a pod which have multiple (2+) container each of them requires ephemeral storage; set the overall resource requests in such a way that there is suitable worker nodes
4. observe the scheduler incorrectly discarding all the suitable worker nodes

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

ffromani:
#### What type of PR is this?

/kind bug

Rewrite and fix, adding a bunch of tests and better logs along the way, how we treat non-NUMA-affine aka host-level resources. The most prominent example is ephemeral storage, but we aim to cover the general case.

Fixes #751

N/A

```release-note
Rewrite handling of ephemeral storage and other non-NUMA-affine resources
```


ffromani:
note: at this stage the integration tests are expected to fail - for the right reasons though :)

ffromani:
> note: at this stage the integration tests are expected to fail - for the right reasons though :)

like this, the expected failures:
```
    --- FAIL: TestTopologyMatchPlugin/[tier1]_multi_containers_all_requiring_ephemeral_storage_with_good_allocation_-_fit (20.03s)
        noderesourcetopology_test.go:1992: Start-topology-match-test [tier1] multi containers all requiring ephemeral storage with good allocation - fit
        noderesourcetopology_test.go:2002: Creating Pod "topology-aware-scheduler-pod-20"
        noderesourcetopology_test.go:2016: pod "topology-aware-scheduler-pod-20" to be scheduled, error: context deadline exceeded
        noderesourcetopology_test.go:2019: Pod topology-aware-scheduler-pod-20 scheduled
        noderesourcetopology_test.go:2028: Pod topology-aware-scheduler-pod-20 is expected on node [fake-node-1], but found on node 
        noderesourcetopology_test.go:2058: Case [tier1] multi containers all requiring ephemeral storage with good allocation - fit finished
    --- FAIL: TestTopologyMatchPlugin/[tier1]_multi_containers_some_requiring_ephemeral_storage_with_good_allocation_-_fit (20.03s)
        noderesourcetopology_test.go:1992: Start-topology-match-test [tier1] multi containers some requiring ephemeral storage with good allocation - fit
        noderesourcetopology_test.go:2002: Creating Pod "topology-aware-scheduler-pod-21"
        noderesourcetopology_test.go:2016: pod "topology-aware-scheduler-pod-21" to be scheduled, error: context deadline exceeded
        noderesourcetopology_test.go:2019: Pod topology-aware-scheduler-pod-21 scheduled
        noderesourcetopology_test.go:2028: Pod topology-aware-scheduler-pod-21 is expected on node [fake-node-1], but found on node 
        noderesourcetopology_test.go:2058: Case [tier1] multi containers some requiring ephemeral storage with good allocation - fit finished
    --- FAIL: TestTopologyMatchPlugin/[tier1]_multi_containers_one_requiring_ephemeral_storage_with_good_allocation_-_fit (20.03s)
        noderesourcetopology_test.go:1992: Start-topology-match-test [tier1] multi containers one requiring ephemeral storage with good allocation - fit
        noderesourcetopology_test.go:2002: Creating Pod "topology-aware-scheduler-pod-22"
        noderesourcetopology_test.go:2016: pod "topology-aware-scheduler-pod-22" to be scheduled, error: context deadline exceeded
        noderesourcetopology_test.go:2019: Pod topology-aware-scheduler-pod-22 scheduled
        noderesourcetopology_test.go:2028: Pod topology-aware-scheduler-pod-22 is expected on node [fake-node-1], but found on node 
        noderesourcetopology_test.go:2058: Case [tier1] multi containers one requiring ephemeral storage with good allocation - fit finished
```

ffromani:
/hold

more testing ongoing

ffromani:
/hold cancel
ready for review

pkg/noderesourcetopology/filter.go
```diff
@@ -55,32 +54,38 @@ func singleNUMAContainerLevelHandler(lh logr.Logger, pod *v1.Pod, zones topology
...
 
 		// subtract the resources requested by the container from the given NUMA.
 		// this is necessary, so we won't allocate the same resources for the upcoming containers
-		subtractFromNUMA(lh, nodes, numaID, container)
+		err := subtractResourcesFromNUMANodeList(clh, nodes, numaID, qos, container.Resources.Requests)
+		if err != nil {
+			// this is an internal error which should never happen
+			return framework.NewStatus(framework.Error, "inconsistent resource accounting", err.Error())
+		}
+		clh.V(4).Info("app container placed", "NUMACandidate", numaID)
```
PiotrProkop, line 87: s/app container placed/container aligned 
somehow the `app` prefix doesn't sound right for me.

pkg/noderesourcetopology/numaresources.go
```diff
@@ -52,3 +133,87 @@ func isNUMAAffineResource(resource corev1.ResourceName) bool {
...
+		return true
+	}
+	return numaQuantity.Cmp(quantity) >= 0
+}
+
+// subtractResourcesFromNUMANodeList finds the correct NUMA ID's resources and always subtract them from `nodes` in-place.
+func subtractResourcesFromNUMANodeList(lh logr.Logger, nodes NUMANodeList, numaID int, qos corev1.PodQOSClass, containerRes corev1.ResourceList) error {
+	logEntries := []any{"numaCell", numaID}
+
+	for i := 0; i < len(nodes); i++ {
```
PiotrProkop, line 148:
```suggestion
	for _, node := range nodes{
```
?

pkg/noderesourcetopology/numaresources.go
```diff
@@ -52,3 +133,87 @@ func isNUMAAffineResource(resource corev1.ResourceName) bool {
...
+				lh.V(4).Info("ignoring QoS-depending exclusive request", "resource", resName, "QoS", qos)
+				continue
+			}
+			if resQty.IsZero() {
+				lh.V(4).Info("ignoring zero-valued request", "resource", resName)
+				continue
+			}
+			nResQ, ok := nodeRes[resName]
+			if !ok {
+				if isAffine {
```
PiotrProkop, line 168: what if `cpu` or `memory` is not in NRT? One may configure nfd not to advertise it(like me for example) :)

ffromani: good point, need to think about this scenario

pkg/noderesourcetopology/filter.go
ffromani: Gotcha, let me review how we call them elsewhere in the code.

pkg/noderesourcetopology/numaresources.go
ffromani: I think we can downgrade this error to warning (log).

PiotrProkop:
Looks good now, can you just squash 3rd and 4th commit into one?

Issue #760, opened 2024-06-14
request: remove vendoring from this repo


jaypipes:
### Area

- [X] Scheduler
- [X] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?

We mirror the scheduler-plugins repository into our private build servers in order to comply with our secure software supply chain policies. There is governance and vulnerability automation that scans the repository for CVEs and licensing issues. Unfortunately, this repository uses Go vendoring which means two things:

1. When we pull in new upstream code, we need to pull in sometimes millions of lines of code (most of which is in the `vendor/` folder) making it difficult if not impossible to know what outside of the dependencies has changed.
2. Our governance and CVE automation scans all the vendored code and alerts on everything it finds, and we need to manually change the go.mod file to include "fixed" versions of dependencies (many times there isn't, frankly, a usable version of a dependency because of incompatible 0-series versions, but we still have to struggle through trying to update the dependency anyway)

If this repo didn't use vendoring, our life would be a whole lot easier because a) there'd be a whole lot less code in the repository and Git commits and b) we could take advantage of modern Go module tooling that essentially obviates the need for vendoring entirely.

In the topologyawarewg and networkplumbingwg I've already merged PRs that remove vendoring from a bunch of their repos and that's helped tremendously in simplifying life for those of us who have to build components using internal build tooling. Would it be possible to remove vendoring from the scheduler-plugins repo?

### What did you expect to happen?

n/a

### Kubernetes version

all

### Scheduler Plugins version

all

zwpaper:
we may need some input from @Huang-Wei before we work on this, 

WDYT? @Huang-Wei

Huang-Wei:
Let's use next release's branch (v1.30) to exercise this.

binacs:
/remove-lifecycle stale

/assign

Considering this issue has not been updated for a long time, I opened another pull request to resolve it. Please refer to https://github.com/kubernetes-sigs/scheduler-plugins/pull/834

PR #761, opened 2024-06-20
fix: remove vendoring from this repo

jdtuhui:
#### What type of PR is this?
/kind bug

 Remove vendor folder,   Use go.mod to download dependent code

Fixes #
https://github.com/kubernetes-sigs/scheduler-plugins/issues/760

```release-note

```


jaypipes:
@Huang-Wei @zwpaper thoughts on this PR? :)

googs1025:
@jdtuhui we should fix `EasyCLA` first.
Also, I think this is not a kind bug classification, but a cleanup classification, please let me know if I am wrong.

Huang-Wei:
We used to add `vendor/` intentionally b/c of some code (zz_openapi*.go) are excluded in git, but needed to make codegen functional. So it ends up developer having to locally install necessary toolkit to do codegen, which is quite cumbersome.

But due to recent refactoring on codegen scripts (both this repo and upstream), those zz_openapi*.go might not be needed, and hence we can get rid of the whole vendor/ folder.

cc @zwpaper @denkensk @ffromani

ffromani:
> We used to add `vendor/` intentionally b/c of some code (zz_openapi*.go) are excluded in git, but needed to make codegen functional. So it ends up developer having to locally install necessary toolkit to do codegen, which is quite cumbersome.
> 
> But due to recent refactoring on codegen scripts (both this repo and upstream), those zz_openapi*.go might not be needed, and hence we can get rid of the whole vendor/ folder.
> 
> cc @zwpaper @denkensk @ffromani

I for myself I don't have strong opinions on the matter.  If there is user demand to remove `vendor/` and we have options projectwise to do so, then why not?

zwpaper:
I would prefer dropping the `vendor/` dir if we do not need it for codegen anymore, that would reduce unnecessary changes for PRs, and we can focus more on our core logic.

PR #768, opened 2024-07-24
increase the timeout of cloudbuild

Huang-Wei:
#### What type of PR is this?

/kind cleanup

Increase the timeout of cloudbuild from 2400s to 3000s.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,7 +1,7 @@
 # See https://cloud.google.com/cloud-build/docs/build-config
 
 # this must be specified in seconds. If omitted, defaults to 600s (10 mins)
-timeout: 2400s
+timeout: 3000s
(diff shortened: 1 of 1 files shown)


prb112:
apologies, I didn't realize I missed so many elements. I'm following along for future. Please let me know if I need to correct anything, Paul

Huang-Wei:
Hmm... increase to 3000s still timed out: https://storage.googleapis.com/kubernetes-jenkins/logs/post-scheduler-plugins-push-images/1816262430691430400/build-log.txt

Given we're trying to support more archs, we need to come out with a solution to reduce the overall image build time.

Huang-Wei:
@prb112 do you have some buffer to see seek any option that can help?

PR #769, opened 2024-07-25
use high memory machine to speedup cloudbuild

Huang-Wei:
#### What type of PR is this?

/kind feature

We now build 4 platforms:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/1c436c586f3772cc66e29cc7c2a83aa4a7ec63e9/Makefile#L20

in a distroless build. Which seems to be pretty slow and [constantly time out](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-scheduler-plugins-push-images/1816225617260056576):

```
#31 [linux/amd64 stage-0 4/4] RUN RELEASE_VERSION=v20240724-v0.28.8-71-g9ae8f214 make build-scheduler
#31 DONE 1290.7s
```

^ shows for one single platform, it takes 21.5 min.

This PR tries to pin a high memory machine for cloudbuild. Hope it works.

PS: this is inspired by https://github.com/kubernetes-sigs/kueue/pull/2303.

Fixes #

```release-note
NONE
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -6,6 +6,7 @@ timeout: 3000s
 # or any new substitutions added in the future.
 options:
   substitution_option: ALLOW_LOOSE
+  machineType: 'N1_HIGHCPU_8'
 steps:
   # see https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing
   - name: 'gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20240718-5ef92b5c36'


Huang-Wei:
Build still [timed out](https://storage.googleapis.com/kubernetes-jenkins/logs/post-scheduler-plugins-push-images/1816341747207245824/build-log.txt) (but building linux/amd64 is way faster 1290s -> 260s):

```
#29 [linux/amd64 stage-0 4/4] RUN RELEASE_VERSION=v20240725-v0.28.8-75-g8ac621c8 make build-scheduler
#29 DONE 260.9s

#32 [linux/ppc64le stage-0 4/4] RUN RELEASE_VERSION=v20240725-v0.28.8-75-g8ac621c8 make build-scheduler
#32 DONE 2547.9s <=== building linux/ppc64le takes roughly 2287s (2547.9s - 260.9s)

#30 [linux/arm64 stage-0 4/4] RUN RELEASE_VERSION=v20240725-v0.28.8-75-g8ac621c8 make build-scheduler
#30 DONE 2657.2s <=== building linux/ppc64le takes roughly 109s (2657.2s - 2547.9s)

#31 [linux/s390x stage-0 4/4] RUN RELEASE_VERSION=v20240725-v0.28.8-75-g8ac621c8 make build-scheduler
TIMEOUT
ERROR: context deadline exceeded
make: *** [Makefile:66: release-image] Terminated
```

Huang-Wei:
@prb112 It looks like building ppc64le and s390x takes an incredible amount of time. It I do the math correct, it would need roughly 2287s X 2 = 4574s. However, the current setting (although very high) still cannot afford it:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/8ac621c809eb583a725f1fdf053b7a8719b84f2e/cloudbuild.yaml#L4

Huang-Wei:
@prb112 the upper limit is 3600s (I think). I will revert the changes in https://github.com/kubernetes-sigs/scheduler-plugins/blob/1c436c586f3772cc66e29cc7c2a83aa4a7ec63e9/Makefile#L20

to:

```
PLATFORMS ?= linux/amd64,linux/arm64
```

Given the math above, I think it might be possible add one of `linux/s390x` **OR** `linux/ppc64le`, but not both. Let me know if you'd like add ONE platform, or want to dedicate some time to find solutions to significantly reduce the build time for these two platforms.

Huang-Wei:
well, hack/build-images.sh cannot get the `DOCKER_BUILDX_CMD` env variable defined in cloudbuild.yaml; also, the Makefile entry of `push-release-image` needs a revamp: "docker manifest push" should be updated to "/buildx-entrypoint build ... --push"

https://github.com/kubernetes-sigs/scheduler-plugins/blob/0759578037216574d5af0be6e8e03ddac02dd35f/Makefile#L77

Huang-Wei:
Once ^^ issue gets resolved. I think the build time can be greatly reduced and hence we should be able to afford all 4 platforms. (the [analysis result](https://github.com/kubernetes-sigs/scheduler-plugins/pull/769#issuecomment-2249542781) was still based on using `docker build` instead of `/buildx-entrypoint build`)

prb112:
I'll look to see what I can do.

Huang-Wei:
> I'll look to see what I can do.

The environment variables defined in cloudbuild.yaml are valid in Makefile, we will have to come out with a way to pass them to hack/build-images.sh; if that's a lot of hassling, we may have to flatten logic of hack/build-images.sh into Makefile.

prb112:
can do. will get to it today.

Huang-Wei:
> can do. will get to it today.

Thanks. I may have some buffer this week to thoroughly look into it.

Huang-Wei:
We're good now.

PR #788, opened 2024-08-01
[Feature] Add a Filter plugin to ensure that non-GPU pods are not sch…

Zeel-Patel:
…eduled on GPU nodes

This plugin is widely being used at Uber and thought can be useful for open source community.

Fixes #

```release-note

```

### DIFF
--- a/pkg/gpumanagementfilter/gpu_management_filter.go
+++ b/pkg/gpumanagementfilter/gpu_management_filter.go
@@ -1,0 +1,129 @@
+package gpumanagementfilter
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubernetes/pkg/scheduler/framework"
+)
+
+const (
+	// PluginName is the name of the plugin used in the plugin registry and configurations.
+	PluginName = "GPUManagementFilter"
+
+	//preFilterStateKey is the key in CycleState to NodeResourcesFit pre-computed data.
+	// Using the name of the plugin helps avoid collisions with other plugins.
+	preFilterStateKey = "PreFilter" + PluginName
+
+	//devicePluginContainerImage is the nvidia provided kubernetes device plugin image.
+	devicePluginContainerImage = "nvcr.io/nvidia/k8s-device-plugin"
+
+	// ResourceNvidiaGPU is nvidia GPU resource
+	ResourceNvidiaGPU = "nvidia.com/gpu"
+)
+
+type GPUManagementFilter struct {
+	profileName string
+}
+
+var (
+	_ framework.FilterPlugin    = &GPUManagementFilter{}
+	_ framework.PreFilterPlugin = &GPUManagementFilter{}
+)
+
+// Name returns name of the plugin. It is used in logs, etc.
+func (p *GPUManagementFilter) Name() string {
+	return PluginName
+}
+
+// preFilterState computed at PreFilter and used in Filter.
+type preFilterState struct {
+	isPodRequestingGPUs     bool
+	isPodNvidiaDevicePlugin bool
+}
+
+// Clone implements the StateData interface.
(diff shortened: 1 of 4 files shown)


pkg/gpumanagementfilter/gpu_management_filter.go
```diff
@@ -0,0 +1,129 @@
...
+
+	//preFilterStateKey is the key in CycleState to NodeResourcesFit pre-computed data.
+	// Using the name of the plugin helps avoid collisions with other plugins.
+	preFilterStateKey = "PreFilter" + PluginName
+
+	//devicePluginContainerImage is the nvidia provided kubernetes device plugin image.
+	devicePluginContainerImage = "nvcr.io/nvidia/k8s-device-plugin"
+
+	// ResourceNvidiaGPU is nvidia GPU resource
+	ResourceNvidiaGPU = "nvidia.com/gpu"
```
PiotrProkop, line 25: I think we should move both of those values so they can be changed via configuration file. Some clusters may also have more than one gpu resource on the node, so it would be great if we could provide a list of gpu resources.

PiotrProkop:
Overall I think  this could be a generic plugin that filters out pod not requesting some specific resource from nodes advertising those resources with except for `device-plugin` of `dra-plugin`(in the future).

pkg/gpumanagementfilter/gpu_management_filter.go
```diff
@@ -0,0 +1,129 @@
+package gpumanagementfilter
```
googs1025, line 1: Can we have a KEP document?

Zeel-Patel:
> Overall I think this could be a generic plugin that filters out pod not requesting some specific resource from nodes advertising those resources with except for `device-plugin` of `dra-plugin`(in the future).

@PiotrProkop sure, can you provide more info on what you mean by 
` except for `device-plugin` of `dra-plugin`(in the future).`

PiotrProkop:
> > Overall I think this could be a generic plugin that filters out pod not requesting some specific resource from nodes advertising those resources with except for `device-plugin` of `dra-plugin`(in the future).
> 
> @PiotrProkop sure, can you provide more info on what you mean by `except for`device-plugin`of`dra-plugin`(in the future).`

Same idea what this func does:
```go
func containsNvidiaDevicePluginImage(containerList []v1.Container) bool {
	for _, containerItem := range containerList {
		if strings.Contains(containerItem.Image, devicePluginContainerImage) {
			return true
		}
	}
	return false
}
```
make this filter plugin not fail device-plugins that provides given resource for other pods :)

pkg/gpumanagementfilter/gpu_management_filter.go
googs1025: +1, Using configuration files is better than hard-coding. Maybe we don't need to limit ourselves to GPU resources, can we extend it to different resources?

pkg/gpumanagementfilter/gpu_management_filter.go
Zeel-Patel: Sure, let me write a KEP doc first

ffromani:
/cc

+1 for having a design document/KEP
I'm interested in this topic

ffromani:
> Overall I think this could be a generic plugin that filters out pod not requesting some specific resource from nodes advertising those resources with except for `device-plugin` of `dra-plugin`(in the future).

strong agree. It seems there is a easy (and safe?) generalization path for this filter plugin. We can have allowlist/denylist in the plugin.
But I wonder if the same functionality can be achieved using existing scheduling tools. On top of my head: taints/tolerations? What extra value a filter plugin will bring?

Zeel-Patel:
@PiotrProkop @googs1025 @zwpaper @ffromani 
https://github.com/kubernetes-sigs/scheduler-plugins/pull/812

Here is the PR with the KEP doc

Huang-Wei:
IMHO having a separate plugin is sort of an overkill. If we don't want to non-GPU workloads to land on GPU nodes, we can simply taint GPU nodes, and only GPU workloads can land on them (with proper tolerations).

PR #798, opened 2024-09-04
WIP: nrt: log: introduce and use "generation" for cache

ffromani:
#### What type of PR is this?

/kind feature

Improves debuggability of the overreserve cache

We introduce the concept of "generation" which is an opaque monotonically increasing integer similar in spirit to the `resourceVersion` kube API field.
Every time the internal state of the cache is updated, which happens only in the resync loop by design, we increment the generation.

GetCachedNRTCopy will also return the generation of the data being used, so we have now an uniform way to correlate readers and writer of the cache, and we gain better visibility of the data being used.

Fixes N/A

N/A

```release-note
NONE
```


ffromani:
Howdy @Tal-or @PiotrProkop could you PTAL? Changes to `discardreserved` are trivial and minimal, but still I'd like a review

pkg/noderesourcetopology/logging/logging.go
```diff
@@ -33,6 +31,7 @@ const (
 	KeyFlow          string = "flow"
 	KeyContainer     string = "container"
 	KeyContainerKind string = "kind"
+	KeyGeneration    string = "gen"
```
ffromani, line 34: or `generation`? size of log entries is a concern too

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -97,30 +98,33 @@ func NewOverReserve(ctx context.Context, lh logr.Logger, cfg *apiconfig.NodeReso
 	return obj, nil
 }
 
-func (ov *OverReserve) GetCachedNRTCopy(ctx context.Context, nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, bool) {
+func (ov *OverReserve) GetCachedNRTCopy(ctx context.Context, nodeName string, pod *corev1.Pod) (*topologyv1alpha2.NodeResourceTopology, CachedNRTInfo) {
```
PiotrProkop, line 101:
```suggestion
func (ov *OverReserve) GetCachedNRTCopy(ctx context.Context, nodeName string, pod *corev1.Pod) (nrt *topologyv1alpha2.NodeResourceTopology, info CachedNRTInfo) {
```

and then just change all return to `return nrt, info` ?

pkg/noderesourcetopology/cache/cache.go
```diff
@@ -24,17 +24,29 @@ import (
 	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 )
 
+type CachedNRTInfo struct {
+	// Generation"is akin to the object resourceVersion and represents
```
PiotrProkop, line 28:
```suggestion
	// Generation is akin to the object resourceVersion and represents
```

PiotrProkop (commented):
some small nits, overall looks good!

pkg/noderesourcetopology/cache/overreserve.go
ffromani: I'm not a fan of the named returns but for silly reasons, but this case seems interesting I'll check. Thanks!

pkg/noderesourcetopology/cache/cache.go
ffromani: thanks for catching! will fix

pkg/noderesourcetopology/cache/overreserve.go
ffromani: OK, tried out. Looks very nice in the `overreserved` impl, but doesn't look so great in the `discardreserved` and `passthrough` impl, IMO leading to a slightly more convoluted code than we have now. I value the fact implementation across the implementations is as consistent as could be, so I think overall I'd like more the current approach and not using the named parameters just yet.

pkg/noderesourcetopology/logging/logging.go
Tal-or: I would prefer readability over compactness

ffromani: fair enoug, "generation" it is

PR #799, opened 2024-09-16
Ensure consistent casing in Dockerfiles

LavredisG:
#### What type of PR is this?
/kind cleanup

It will eliminate the warning we get from the incosistent casing of the "FROM" and "as" keywords
laying in the Dockerfiles when we use the Makefile to build the images.


Huang-Wei (commented):
/lgtm
/approve

PR #804, opened 2024-09-29
fix: remove vendoring from this repo

jdtuhui:
#### What type of PR is this?
/kind cleanup 

Remove vendor folder, Use go.mod to download dependent code

https://github.com/kubernetes-sigs/scheduler-plugins/issues/760


Huang-Wei:
@jdtuhui I think this PR needs some non-trivial work to get all things working - not only tests, but also the Docker builds, codegen, etc. Please let me know if you need some help.

Huang-Wei:
@jdtuhui given we're gonna cut a release in the next week. I guess we can finish this in the new release. WDYT?

Huang-Wei:
Close in favor of #834

PR #805, opened 2024-09-30
fix: fix wrong PreemptionToleration policy example

odeinjul:
#### What type of PR is this?
/kind documentation

Fixes the wrong PreemptionToleration policy example.

```release-note
NONE
```


Huang-Wei (commented):
/lgtm
/approve

Issue #743, opened 2024-05-18
Migrate to contextual logging


zwpaper:
### Area

- [X] Scheduler
- [X] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?

k/k is migrating to contextual logging, we should also follow up to add more context when logging

https://github.com/kubernetes/enhancements/issues/3077

### What did you expect to happen?

use contextual logging

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>
v0.28.9
</details>

ffromani:
noderesourcetopology migrated in #710 and #725

googs1025:
@ffromani @zwpaper Do other plugins also need log modification? If necessary, I will take it

zwpaper:
Hi @googs1025, thanks for the interest, I think it would be great if we can migrate all logs to contextual logging since the k/k is working hard on it.

feel free to take it and raise PRs for it!

googs1025:
@zwpaper thanks for reply. I want to confirm some details first. Do we need to replace the original `klog.ErrorS(err, "xxxxx")` with `lh := klog.FromContext(ctx).WithValues(logging.KeyPod, klog.KObj(pod), logging.KeyPodUID, logging.PodUID(pod), logging.KeyNode, nodeName)`?  Is my understanding correct?

The following code
```go
	lh := klog.FromContext(ctx).WithValues(logging.KeyPod, klog.KObj(pod), logging.KeyPodUID, logging.PodUID(pod), logging.KeyNode, nodeName)

	lh.V(4).Info(logging.FlowBegin)
	defer lh.V(4).Info(logging.FlowEnd)

	nodeTopology, ok := tm.nrtCache.GetCachedNRTCopy(ctx, nodeName, pod)
	if !ok {
		lh.V(2).Info("invalid topology data")
		return framework.NewStatus(framework.Unschedulable, "invalid node topology data")
	}
	if nodeTopology == nil {
		return nil
	}
```

ffromani:
> @zwpaper thanks for reply. I want to confirm some details first. Do we need to replace the original `klog.ErrorS(err, "xxxxx")` with `lh := klog.FromContext(ctx).WithValues(logging.KeyPod, klog.KObj(pod), logging.KeyPodUID, logging.PodUID(pod), logging.KeyNode, nodeName)`? Is my understanding correct?
> 
> The following code
> 
> ```go
[...]

PR #813, opened 2024-10-29
feature: use contextal logging 

googs1025:
- use contextal logging

Due to the large scope, I may only be able to complete part of the plugins work in this PR.
- coscheduling
- capacityscheduling
- networkaware
- sysched

```release-note
None
```

### DIFF
--- a/pkg/capacityscheduling/capacity_scheduling.go
+++ b/pkg/capacityscheduling/capacity_scheduling.go
@@ -125,6 +125,7 @@ func New(ctx context.Context, obj runtime.Object, handle framework.Handle) (fram
 		podLister:         handle.SharedInformerFactory().Core().V1().Pods().Lister(),
 		pdbLister:         getPDBLister(handle.SharedInformerFactory()),
 	}
+	logger := klog.FromContext(ctx)
 
 	client, err := client.New(handle.KubeConfig(), client.Options{Scheme: scheme})
 	if err != nil {
@@ -187,7 +188,7 @@ func New(ctx context.Context, obj runtime.Object, handle framework.Handle) (fram
 			},
 		},
 	)
-	klog.InfoS("CapacityScheduling start")
+	logger.Info("CapacityScheduling start")
 	return c, nil
 }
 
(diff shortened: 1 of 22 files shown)


pkg/coscheduling/core/core.go
```diff
@@ -211,6 +213,8 @@ func (pgMgr *PodGroupManager) PreFilter(ctx context.Context, pod *corev1.Pod) er
 
 // Permit permits a pod to run, if the minMember match, it would send a signal to chan.
 func (pgMgr *PodGroupManager) Permit(ctx context.Context, state *framework.CycleState, pod *corev1.Pod) Status {
+	lh := klog.FromContext(ctx)
```
googs1025, line 216: `klog.FromContext(ctx)` 
When using klog.FromContext(ctx), do I need to add WithValues ​​to give it more information? 🤔

`klog.FromContext(ctx).WithValues("PodGroupManager", "Unreserve")`

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -125,6 +125,7 @@ func New(ctx context.Context, obj runtime.Object, handle framework.Handle) (fram
 		podLister:         handle.SharedInformerFactory().Core().V1().Pods().Lister(),
 		pdbLister:         getPDBLister(handle.SharedInformerFactory()),
 	}
+	logger := klog.FromContext(ctx).WithName(Name).WithValues("Phase", "New")
```
Huang-Wei, line 128: To ease reviewing, I'm inclined to keep this PR merely as a "migration" PR - i.e., keep new bits in a follow-up, only migrate klog.InfoS / klog.ErrorS.

pkg/capacityscheduling/capacity_scheduling.go
```diff
@@ -480,25 +491,27 @@ func (p *preemptor) SelectVictimsOnNode(
 	pod *v1.Pod,
 	nodeInfo *framework.NodeInfo,
 	pdbs []*policy.PodDisruptionBudget) ([]*v1.Pod, int, *framework.Status) {
+
+	logger := klog.FromContext(ctx).WithName(Name).WithValues("Phase", "SelectVictimsOnNode")
```
Huang-Wei, line 495: ditto. Would you mind folding the "changes" in a follow-up PR?

googs1025: ok, I will revert this

googs1025: done

pkg/capacityscheduling/capacity_scheduling.go
googs1025: done

googs1025:
> Thanks for the PR @googs1025 !
> 
> One question: do you plan to work on some script to enforce contextual logging into CI? (I may assume we may leverage from k/k)

Ok, this is something I need to look into, I'm not familiar with this part of the script. If you could point me to some help, that would be great :)

Huang-Wei (commented):
/lgtm
/approve

Huang-Wei:
When migrating to structural logging, we composed a script hack/verify-structured-logging.sh to enforce it - which was based on k8s.io/klog/hack/tools/logcheck.

For contextual logging, we may proceed with a similar approach.

PR #814, opened 2024-10-31
Allow specifying the scheduler container entrypoint in the helm chart

chotiwat:
#### What type of PR is this?

/kind feature

The current chart template doesn't support the older kube-scheduler image which doesn't specify ENTRYPOINT in the image. This PR allows customizing the container entrypoint through the new `scheduler.command` value.

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/README.md
+++ b/manifests/install/charts/as-a-second-scheduler/README.md
@@ -45,9 +45,10 @@ scheduler-plugins-scheduler    1/1     1            1           7s
 The following table lists the configurable parameters of the as-a-second-scheduler chart and their default values.
 
 | Parameter                 | Description                 | Default                                                                                         |
-|---------------------------|-----------------------------|-------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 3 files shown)


Huang-Wei:
By "older kube-scheduler image", do you mean the v0.29.7 image? If so, we changed back the default entrypoint to /bin/scheduler (#794) - which is available in v0.29.8 and master.

And we will keep the path unchanged in the future, for backwards compatibility. Did it resolve your issue?

xref: https://kubernetes.slack.com/archives/C09TP78DV/p1723546975114219

chotiwat:
@Huang-Wei By older images, I mean the ones before v0.29, e.g. https://github.com/kubernetes-sigs/scheduler-plugins/blob/v0.27.9/build/scheduler/Dockerfile#L28. I'd like to use the latest chart with an older image version to match my org's k8s clusters.

Huang-Wei:
ok, I see the motiviation.

/ok-to-test

PR #815, opened 2024-10-31
Fix leader election config in the helm chart

chotiwat:
#### What type of PR is this?

/kind bug

The default leader election for kube-scheduler will try to acquire the `kube-system/kube-scheduler` lease, which is what the default scheduler uses. This prevents our custom scheduler from ever acquiring the leader lease until the leader loses the lease, at which point it could cause undesirable behavior by stealing the lease from the default scheduler. We should be using a separate lease if we want to run the custom scheduler alongside the default scheduler.

Fixes #

```release-note
NONE
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
@@ -10,6 +10,8 @@ data:
     kind: KubeSchedulerConfiguration
     leaderElection:
       leaderElect: {{ .Values.scheduler.leaderElect }}
+      resourceNamespace: {{ .Release.Namespace }}
+      resourceName: {{ .Values.scheduler.name }}
     profiles:
     # Compose all plugins in one profile
     - schedulerName: {{ .Values.scheduler.name }}


manifests/install/charts/as-a-second-scheduler/templates/configmap.yaml
```diff
@@ -10,6 +10,8 @@ data:
     kind: KubeSchedulerConfiguration
     leaderElection:
       leaderElect: {{ .Values.scheduler.leaderElect }}
+      resourceNamespace: {{ .Release.Namespace }}
+      resourceName: {{ .Values.scheduler.name }}
```
chotiwat, line 14: Based on:
- https://github.com/aws-samples/custom-scheduler-eks/blob/01a576e2aa45b1bf779ea1750a541b6ecb6beb9e/deploy/charts/custom-scheduler-eks/templates/configmap.yaml#L31-L32
- https://kubernetes.io/docs/reference/config-api/kube-scheduler-config.v1/#LeaderElectionConfiguration

Huang-Wei:
@chotiwat thanks for the PR. Could you add a brief stanza in the `release-notes` section in the PR description?

chotiwat:
Done @Huang-Wei

PR #818, opened 2024-11-03
Dynamically add required RBAC in Helm chart

EladDolev:
#### What type of PR is this?

/kind feature

Dynamically adds required RBAC permissions for enabled plugins

```release-note
Dynamically add required RBAC bits in Helm chart
```

### DIFF
--- a/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
+++ b/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
@@ -62,20 +62,23 @@ rules:
 - apiGroups: ["topology.node.k8s.io"]
   resources: ["noderesourcetopologies"]
   verbs: ["get", "list", "watch"]
-# resources need to be updated with the scheduler plugins used
 - apiGroups: ["scheduling.x-k8s.io"]
   resources: ["podgroups", "elasticquotas", "podgroups/status", "elasticquotas/status"]
   verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
-# for network-aware plugins add the following lines
-#- apiGroups: [ "appgroup.diktyo.x-k8s.io" ]
-#  resources: [ "appgroups" ]
-#  verbs: [ "get", "list", "watch", "create", "delete", "update", "patch" ]
-#- apiGroups: [ "networktopology.diktyo.x-k8s.io" ]
-#  resources: [ "networktopologies" ]
-#  verbs: [ "get", "list", "watch", "create", "delete", "update", "patch" ]
-#- apiGroups: ["security-profiles-operator.x-k8s.io"]
-#  resources: ["seccompprofiles", "profilebindings"]
-#  verbs: ["get", "list", "watch", "create", "delete", "update", "patch"]
+{{- /* resources need to be updated with the scheduler plugins used */}}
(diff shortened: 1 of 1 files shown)


manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml
```diff
@@ -107,6 +107,9 @@ rules:
 - apiGroups: ["topology.node.k8s.io"]
   resources: ["noderesourcetopologies"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["scheduling.k8s.io"]
+  resources: ["priorityclasses"]
+  verbs: ["get", "list", "watch"]
```
Huang-Wei, line 112: Could you elaborate why priorityClass permission is needed in controller?

I searched through the code, it's only needed on a plugin that is not enabled by default in scheduler plugins:

```
⇒  ag "priorityclasses"
test/integration/preemption_toleration_test.go
208:			if _, err := cs.SchedulingV1().PriorityClasses().Create(testCtx.Ctx, tt.priorityClass, metav1.CreateOptions{}); err != nil {

pkg/preemptiontoleration/preemption_toleration.go
96:		priorityClassLister: fh.SharedInformerFactory().Scheduling().V1().PriorityClasses().Lister(),

pkg/preemptiontoleration/preemption_toleration_test.go
219:	pcInformer := informersFactory.Scheduling().V1().PriorityClasses().Informer()
229:		informersFactory.Scheduling().V1().PriorityClasses().Lister(),
```

EladDolev: Yes @Huang-Wei you're right, we are using the Preemption Toleration plugin :-)

What would you think would be the best way to tackle this ?
Those are readonly permissions, so is it fine having it here ?
If it is not here on this file, then as a user, am I supposed to use YAML files external to the chart ?
I am also happy to introduce support to the chart for allowing the user to add arbitrary RBAC permissions

Huang-Wei: I think we can introduce the changes but comment them out like what other plugins did:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/dc97958b489ec1c370d7d0e334e88c1cf07ac8b2/manifests/install/charts/as-a-second-scheduler/templates/rbac.yaml#L69-L78

EladDolev: Thank you @Huang-Wei 
TBH I'm not happy to do that as I want to see a chart that can be used OOTB as opposed to forcing the users to fork it or deploy additional files with it
Another thing I can do is iterate through the plugins which are enabled via the the `plugins.enabled` parameter, and add permissions accordingly. Would that be a good idea ?

Huang-Wei: Definitely, that would be the desirable solution - auto-inject the needed RBAC bits upon the enable/disable of particular plugin.

EladDolev: would appreciate it if you could review again @Huang-Wei

Huang-Wei (commented):
Thanks @EladDolev .

/lgtm
/approve

PR #823, opened 2024-11-05
feature: use contextal logging for integration

googs1025:
#### What type of PR is this?
/kind cleanup

```release-note
None
```

### DIFF
--- a/pkg/noderesourcetopology/cache/attr_watch.go
+++ b/pkg/noderesourcetopology/cache/attr_watch.go
@@ -18,7 +18,7 @@ package cache
 
 import (
 	"context"
-
+	"fmt"
 	"github.com/go-logr/logr"
 
 	"k8s.io/apimachinery/pkg/watch"
@@ -72,7 +72,7 @@ func (wt Watcher) ProcessEvent(ev watch.Event) bool {
 
 	nrtObj, ok := ev.Object.(*topologyv1alpha2.NodeResourceTopology)
 	if !ok {
-		wt.lh.Info("unexpected object %T", ev.Object)
+		wt.lh.Info("unexpected object", "kind", fmt.Sprintf("%T", ev.Object))
 		return false
 	}
 
--- a/test/integration/coscheduling_test.go
+++ b/test/integration/coscheduling_test.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler"
 	schedapi "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	fwkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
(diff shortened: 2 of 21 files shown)


googs1025:
@Huang-Wei 
/PTAL :) 

It is expected that there will be another PR to complete the CI script access 🤔

pkg/noderesourcetopology/cache/attr_watch.go
```diff
@@ -72,7 +73,7 @@ func (wt Watcher) ProcessEvent(ev watch.Event) bool {
 
 	nrtObj, ok := ev.Object.(*topologyv1alpha2.NodeResourceTopology)
 	if !ok {
-		wt.lh.Info("unexpected object %T", ev.Object)
+		wt.lh.Info(fmt.Sprintf("unexpected object %T", ev.Object))
```
Huang-Wei, line 76: ditto, remove fmt.XYZ

test/integration/coscheduling_test.go
```diff
@@ -356,14 +355,14 @@ func TestCoschedulingPlugin(t *testing.T) {
 			defer cleanupPods(t, testCtx, tt.pods)
 			// Create Pods, we will expect them to be scheduled in a reversed order.
 			for i := range tt.pods {
-				klog.InfoS("Creating pod ", "podName", tt.pods[i].Name)
+				t.Logf("Creating pod: %s", tt.pods[i].Name)
```
Huang-Wei, line 358: Thanks!

Huang-Wei (commented):
some nits.

pkg/noderesourcetopology/cache/attr_watch.go
googs1025: I suggest this approach as bellow.  🤔 

`wt.lh.Info("unexpected object", "kind", fmt.Sprintf("%T", ev.Object))`

Because we use with original writing and will get logcheck output similar to the following

```bash
/home/ubuntu/scheduler-plugins/pkg/noderesourcetopology/cache/attr_watch.go:75:3: logging function "Info" should not use format specifier "%T"
/home/ubuntu/scheduler-plugins/pkg/noderesourcetopology/cache/attr_watch.go:81:43: Key positional arguments are expected to be inlined constant strings. Please replace &{logging KeyNode} provided with string value.
/home/ubuntu/scheduler-plugins/pkg/noderesourcetopology/cache/attr_watch.go:89:38: Key positional arguments are expected to be inlined constant strings. Please replace &{logging KeyNode} provided with string value.
/home/ubuntu/scheduler-plugins/pkg/noderesourcetopology/cache/discardreserved.go:82:33: Key positional arguments are expected to be inlined constant strings. Please replace &{logging KeyPod} provided with string value.
/home/ubuntu/scheduler-plugins/pkg/noderesourcetopology/cache/discardreserved.go:82:65: Key positional arguments are expected to be inlined constant strings. Please replace &{logging KeyPodUID} provided with string value.
```

Huang-Wei: sounds good given `%T` is a bit special here - we cannot interpret it easily w/o using `%`

Huang-Wei (commented):
/lgtm
/approve

PR #831, opened 2024-11-09
chore: renew verify conextual structured script

googs1025:
#### What type of PR is this?
/kind cleanup

```release-note
None
```

### DIFF
--- a/hack/verify-structured-logging.sh
+++ b/hack/verify-structured-logging.sh
@@ -46,9 +46,33 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
 
 ret=0
 echo -e "\nRunning structured logging static check on all packages"
-GOOS=linux    logcheck "${packages[@]}" || ret=$?
-GOOS=windows  logcheck "${packages[@]}" || ret=$?
+
+# Function to run logcheck and capture output
+run_logcheck() {
+  local os="$1"
+  shift
+  local output
+  output=$(GOOS="$os" logcheck "$@" 2>&1)
+  # TODO: Add more checks as needed
+  # Currently, we only check for unstructured logging
+  # more detail: https://github.com/kubernetes-sigs/scheduler-plugins/pull/831#discussion_r1835553079
+  errors=$(echo "$output" | grep 'unstructured logging function')
+  if [[ $? -eq 0 ]]; then
+      echo "Error: Unstructured logging detected. Please fix the issues."
+      echo "$errors"
+      return 1
+  fi
+}
+
+# Run logcheck for Linux
+echo "Running logcheck for Linux..."
+if ! run_logcheck linux "${packages[@]}"; then
+  ret=1
+fi
 
 if [ $ret -eq 0 ]; then
   echo "Structured logging static check passed on all packages :)"
+else
+  echo "Structured logging static check failed. Please fix the issues."
+  exit 1
 fi


pkg/sysched/sysched.go
```diff
@@ -134,6 +134,9 @@ func (sc *SySched) getSyscalls(logger klog.Logger, pod *v1.Pod) sets.Set[string]
 			if len(ns) > 0 && len(name) > 0 {
 				syscalls, err := sc.readSPOProfileCR(name, ns)
 				if err != nil {
+					klog.ErrorS(err, "aaaa")
```
googs1025, line 137: Just to test whether the script works as expected

googs1025:
use this as example: https://github.com/kubernetes-sigs/scheduler-plugins/pull/831#discussion_r1835238891
ci workflow works fine


https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_scheduler-plugins/831/pull-scheduler-plugins-verify/1855090645207617536

googs1025:
When this script is merged, we can only use conextual logs like `logger := klog.FromContext(ctx)`. If using methods like klog.InfoS, we will get an error in ci workflow.

hack/verify-structured-logging.sh
```diff
@@ -46,9 +50,35 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
...
+  ret=1
+fi
 
 if [ $ret -eq 0 ]; then
   echo "Structured logging static check passed on all packages :)"
-fi
+else
+  echo "Structured logging static check failed. Please fix the issues."
+  exit 1
+fi
```
Huang-Wei, line 84: nit: keep the newline :)

hack/verify-structured-logging.sh
```diff
@@ -46,9 +50,35 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
...
-GOOS=linux    logcheck "${packages[@]}" || ret=$?
-GOOS=windows  logcheck "${packages[@]}" || ret=$?
+
+# Function to run logcheck and capture output
+run_logcheck() {
+  local os="$1"
+  shift
+  local output
+  output=$(GOOS="$os" logcheck "$@" 2>&1)
+  echo "$output"
```
Huang-Wei, line 60: My 2 cents:

- remove this line here as for current code base, it reports a lot of violations. I may choose to resolve them in phases
- in Phase 1: treat violations with string "unstructured logging function" as fatal error, dump the problematic entries under L62 
- in Phase 2: fix all violations, and hence don't need to filter out "unstructured logging function"

hack/verify-structured-logging.sh
```diff
@@ -46,9 +50,35 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
...
+echo "Running logcheck for Linux..."
+if ! run_logcheck linux "${packages[@]}"; then
+  ret=1
+fi
+
+# Run logcheck for Windows
+echo "Running logcheck for Windows..."
+if ! run_logcheck windows "${packages[@]}"; then
+  ret=1
+fi
```
Huang-Wei, line 77: I think previous code to run against **both** linux and windows is a fault. We can remove running against windows.

hack/verify-structured-logging.sh
```diff
@@ -46,9 +50,35 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
...
-GOOS=windows  logcheck "${packages[@]}" || ret=$?
+
+# Function to run logcheck and capture output
+run_logcheck() {
+  local os="$1"
+  shift
+  local output
+  output=$(GOOS="$os" logcheck "$@" 2>&1)
+  echo "$output"
+  if echo "$output" | grep -q 'unstructured logging function'; then
```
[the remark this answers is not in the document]
Huang-Wei: yup, we can target it as the goal of phase 1.

hack/verify-structured-logging.sh
googs1025: done

hack/verify-structured-logging.sh
googs1025: removed.

In addition, I added todo items to record these things 🤔

hack/verify-structured-logging.sh
```diff
@@ -46,9 +50,31 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
...
+run_logcheck() {
+  local os="$1"
+  shift
+  local output
+  output=$(GOOS="$os" logcheck "$@" 2>&1)
+  # TODO: Add more checks as needed
+  # Currently, we only check for unstructured logging
+  # more detail: https://github.com/kubernetes-sigs/scheduler-plugins/pull/831#discussion_r1835553079
+  if echo "$output" | grep -q 'unstructured logging function'; then
+    echo "Error: Unstructured logging detected. Please fix the issues."
```
Huang-Wei, line 64: I think we need to use a variable in L63 to hold the errorous entries, so as to dump below L64, right? (pseudo code below, please adjust/evalaute)

```bash
errors=$(echo "$output" | grep -q 'unstructured logging function')
if [[ $? -ne 0 ]]; then
   echo "Error blabla"
   echo "$errors"
```

hack/verify-structured-logging.sh
```diff
@@ -26,9 +26,13 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 source "${KUBE_ROOT}/hack/lib/init.sh"
 source "${KUBE_ROOT}/hack/lib/util.sh"
 
+export GOPATH=${GOPATH:-$HOME/go}
+export GOBIN=${GOBIN:-$GOPATH/bin}
+export PATH=$GOBIN:$PATH
```
Huang-Wei, line 31: do we need this?

hack/verify-structured-logging.sh
```diff
@@ -26,9 +26,13 @@ KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
...
+
 kube::golang::verify_go_version
 
-GO111MODULE=on go install k8s.io/klog/hack/tools/logcheck@latest
+GO111MODULE=on go install sigs.k8s.io/logtools/logcheck@latest
```
Huang-Wei, line 35: we can stick to k8s.io/klog in this PR. I have a WIP #832 taking care of this.

hack/verify-structured-logging.sh
```diff
@@ -46,9 +46,33 @@ done < <(echo "${all_packages[@]}" | tr " " "\n")
...
+run_logcheck() {
+  local os="$1"
+  shift
+  local output
+  output=$(GOOS="$os" logcheck "$@" 2>&1)
+  # TODO: Add more checks as needed
+  # Currently, we only check for unstructured logging
+  # more detail: https://github.com/kubernetes-sigs/scheduler-plugins/pull/831#discussion_r1835553079
+  errors=$(echo "$output" | grep 'unstructured logging function')
+  if [[ $? -eq 0 ]]; then
```
Huang-Wei, line 60: nope, it should ensure the return code of previous command not equal 0 - i.e., `-ne`

googs1025: ok got this

Huang-Wei: sry my bad 🤦

`grep` returns 0 to indicate there is a hit - in our case, it's an error. So your previous version is correct.

googs1025: ok

hack/verify-structured-logging.sh
googs1025: removed

Huang-Wei (commented):
/lgtm
/approve

PR #837, opened 2024-12-10
add kep 4986 sakkara cluster topology

atantawi:
#### What type of PR is this?
/kind feature

Add kep for Sakkara plugin, a hierarchical cluster topology group scheduler

Fixes #4986
https://github.com/kubernetes/enhancements/issues/4986

```release-note

```


13567436138:
when to implement this plugin？

atantawi:
> when to implement this plugin？

I am planning to raise a PR and put the code in there shortly.

13567436138:
@atantawi awesome

googs1025:
/cc

I will try to understand this plugin.

13567436138:
How short
I can not wait

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+
+Want to deploy a job with 8 pods, equally partitioned into two partitions, where each partition is placed in separate rack, and the pods belonging to a partition are packed on nodes within the rack. Partitioning improves availability, and packing improves performance.
+
+![before-1](images/use-cases/before-1.png)
+
+![after-1](images/use-cases/after-1.png)
+
+2- **User B:**
+
+Want to place a job with 6 pods packed on same physical server, as much as possible, where pods on the same server are spread across nodes in that server, but in the range [2,4]. Having all pods on the same physical server improves communication among the pods. And, having less than 2 pods in a node spreads the pods too thinly, and having more than 4 may leads to congestion.
```
googs1025, line 55: initial thought for me is: can similar problems be solved by using existing `affinity` or `anti-affinity` or `Topology Spread Constraints`?  But this seems to be "multi-level", and there are many things to consider. 🤔

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+
+## Design Details
+
+![architecture](images/architecture.png)
+
+![plugin state transitions](images/state-transitions.png)
+
+### Cluster topology
+
+The cluster topology is specified as a tree where the leaves are the nodes in the cluster.
```
googs1025, line 91: I am not sure whether clusters need such a complex definition (at least I do not have such a complex definition internally... this might be my problem). In my work, there is only a one-dimensional level, which is described as "node pool". And I optimize scheduling based on this dimension. 🤔

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+      }
+    }
+```
+
+- *namespace*: The topology configMap is deployed in a special namespace to avoid modifications by unauthorized users. The name of the namespace is defined using the argument `topologyConfigMapNameSpace` of the Sakkara scheduler plugin in the `KubeSchedulerConfiguration`. If not specified, it defaults to the `default` namespace. Here we use namespace sakkara.
+- *resource-names*: Resources to be considered when placing the group.
+- *level-names*: Names of the levels as an array of strings ordered top to bottom (excluding the root).
+- *tree*: String specification of the tree.
+
+### Group constraints
```
googs1025, line 136: one question: will the number of configmaps for `group constraints` increase as more group settings are added? Can we merge configs into the same one? More configs sometime mean more difficult maintenance

atantawi:
> How short I can not wait

https://github.com/kubernetes-sigs/scheduler-plugins/pull/877

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+If no cluster topology is specified, Sakkara assumes a flat topology with only one level, i.e. a tree with a root and all nodes in the cluster as leaves, with two resources: CPU and memory.
+
+The tree is specified in a ConfigMap with a label `"sakkara.topology":`.
+It may be updated at any time. Updates take effect during the following group placement cycle.
+
+An example with a two-level tree of racks and nodes follows.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
```
Huang-Wei, line 103: Not a blocker for the KEP, but I highly recommend to start with CRD since day 1.

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+  resource-names: |
+    [ "cpu", "memory", "nvidia.com/gpu" ]
+  level-names: |
+    [ "rack", "node" ]
+  tree: |
+    {
+      "rack-0": {
+        "node-0": {},
+        "node-1": {},
+        "node-2": {}
```
Huang-Wei, line 121: How do we present the topological info if the cluster has a couple of thousands nodes?

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+metadata:
+  name: pod-6
+  labels:
+    sakkara.group.name: group-a
+    sakkara.member.rank: "2"
+    sakkara.member.retries: "1"
+    sakkara.member.status: Bound
+```
+
+The rank, status, and number of placement trials for the pod are provided as labels. This way, an (init) container in the pod may read the rank label and configure itself accordingly.
```
Huang-Wei, line 352: It's an anti-pattern to update a pod's labels to store extra info. Update a pod is expensive, no need to say updating a gang of pods.

kep/4986-sakkara-cluster-topology/README.md
```diff
@@ -0,0 +1,536 @@
...
+
+### Notes/Constraints/Caveats
+
+### Risks and Mitigations
+
+## Design Details
+
+![architecture](images/architecture.png)
+
+![plugin state transitions](images/state-transitions.png)
```
Huang-Wei, line 87: After reading the Design section, I still cannot figure out the exact responsibility of each phase. Like, which phase does the "pre-scheduling" to update the pod's label, what if the "pre-scheduling" result is in conflict with the real-time scheduling decision, how preemption works, etc.

It's quite essential to illustrate: by given a group of pods carrying the group constraints, in an e2e manner how they're processed in each phase of a scheduling cycle.

kep/4986-sakkara-cluster-topology/README.md
atantawi: There is only one configMap (alternatively, a CR or an extension of the PodGroup) per group. Group constraints is a key in the map (or a spec).

kep/4986-sakkara-cluster-topology/README.md
atantawi: Agreed. Should we introduce a new CR or extend the PodGroup?

kep/4986-sakkara-cluster-topology/README.md
atantawi: Good question! We need a tool to generate/update the tree definition. Or, we should consider an alternative way to code the tree. One possibility is to have a node (a leaf in the cluster topology tree) code the path to the root as labels. Problem is the plugin will need to create the tree every time a group needs to be scheduled. (Unless it's cached and updated whenever node labels change.)

kep/4986-sakkara-cluster-topology/README.md
atantawi: The most important data is the rank. Having it as a pod label simplifies the logic of the container getting it initially as an environment variable. Alternatively, rank values for all pods in the group may be written in the CR for the group. The challenge is for the containers to get them?

dom4ha:
/cc @johnbelamaric
There are a few aspects that we should discuss regarding the overall approach to topology aware scheduling

1. Resource management in scheduler becomes more complex, mainly because of rapid DRA development. Some of the resource types that are being introduced already enforce topology constraints implicitly (see https://github.com/kubernetes/enhancements/pull/5069). This mean that we cannot assume that there is only one topology.
1. DRA becomes a new method to define and manage resource (especially dynamically allocated), but also, it introduces a new way of expressing resource requirements and evolve toward covering all type of resources that pods may need to run.
1. We (and a few other parties) are currently exploring a possibility of pre-allocating (reserving) resources. There is a [proposal](https://docs.google.com/document/d/1LvA9_H4vyNbp173DUm3p9_fpsLYjjpujyRPO3TAOy7g/edit?tab=t.0#heading=h.cwif4pjaz63a) for making the first such attempt to leverage DRA framework for that. Note that reservation not necessarily need to consider final Pods at the time of allocation, but may need to consider topology (directly or indirectly)
1. More fine grained allocations (not proposed yet anywhere) could become dynamically scheduled/allocated (somehow) and will not leave too much space for scheduler to decide on the final pods placement.
1. Scheduler might play the crucial role in dynamic scheduling/allocation of such resources/reservations, but this is a very new and vague idea not discussed/brainstormed on the community forum yet (except the last SIG meeting). Note however that ResourceClaims themselves are in fact a sort of reservation method already.
1. Finally, any TAS implementation makes sense when we talk about group/gang scheduling. This concept is not well defined yet at Kubernetes level, so before we provide solutions for it, we need to introduce such well thought-out concept, which may be a challenge in itself. I know that gang-scheduling can be modeled in the current scheduler framework, but not sure if this is sufficient for long term maintainabliity.

I believe that TAS scheduling is important and it's great that you are sharing your approach and offering contribution. I believe that we should think how the proposal interacts with the aspects that I mentioned (frankly, I know questions but not answers yet). I'm especially curious whether the reservation ideas is something that we want to pursue (as community).

To scope the discussion for now, my major concern is whether the proposal that is based on topology scheduling enclosed into a single plugin is capable to take into consideration external constraints (including implicit topology constraints). IIUC the solver knows the topology, knows the group of pods, but does it know about external constraints put by other plugins?

PR #840, opened 2024-12-16
fix(util): fix the problem of pod overhead not being calculated

googs1025:
#### What type of PR is this?
/kind bug
/kind cleanup

- fix the problem of overhead not being calculated

Fixes None

```release-note
None
```

### DIFF
--- a/pkg/util/resource.go
+++ b/pkg/util/resource.go
@@ -47,6 +47,7 @@ func ResourceList(r *framework.Resource) v1.ResourceList {
 // - the sum of all app containers(spec.Containers) request for a resource.
 // - the effective init containers(spec.InitContainers) request for a resource.
 // The effective init containers request is the highest request on all init containers.
+// Additionally, if the Pod specifies overhead, it will also be included in the final calculation.
 func GetPodEffectiveRequest(pod *v1.Pod) v1.ResourceList {
 	initResources := make(v1.ResourceList)
 	resources := make(v1.ResourceList)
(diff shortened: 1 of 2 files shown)


pkg/util/resource.go
```diff
@@ -73,5 +74,14 @@ func GetPodEffectiveRequest(pod *v1.Pod) v1.ResourceList {
 		}
 		resources[name] = quantity
 	}
+	// Add pod overhead if present
+	if pod.Spec.Overhead != nil {
```
Huang-Wei, line 78: Overhead is a map, so we can directly iterate it w/o this if check.

pkg/util/resource_test.go
```diff
@@ -119,6 +141,9 @@ func TestGetPodEffectiveRequest(t *testing.T) {
 					},
 				})
 			}
+			if tt.podOverheadRequest != nil {
```
Huang-Wei, line 144: ditto. We can assign it directly w/o the if check.

Huang-Wei (commented):
A nit. LGTM otherwise.

pkg/util/resource_test.go
googs1025: done

pkg/util/resource.go
googs1025: done

Huang-Wei (commented):
/lgtm
/approve

PR #845, opened 2024-12-24
feat: add kep

LY-today:
### What would you like to be added?

**What is your proposal**:
The NodeResourcesFit plug-in of native k8s can only adopt a type of strategy for all resources, such as MostRequestedPriority and LeastRequestedPriority. However, in industrial practice, this design does not apply to some scenarios. For example: In AI scenarios, businesses that apply for GPUs prefer to occupy the entire GPU machine first to prevent GPU fragmentation; businesses that apply for CPU & MEM are prioritized and dispersed to non-GPU machines to prevent excessive consumption of CPU & MEM on GPU machines, resulting in real tasks of applying for GPUs. Pending due to insufficient non-GPU resources
. It is therefore hoped that both strategies can be extended to address this business need.

**Why is this needed**:
There are related descriptions above

**Is there a suggested solution, if so, please add it**:
### plugin-one
config：
```
resources: 
  nvidia.com/gpu:
    type: MostAllocated
    weight: 2
  cpu:
    type: LeastAllocated
    weight: 1
  memory:
    type: LeastAllocated
    weight: 1
```
config description：
image

node score:
```
finalScoreNode = [(weight1 * resource1) + (weight2 * resource2) + … + (weightN* resourceN)] /(weight1+weight2+ … +weightN)
```

 ### plugin-two
config：
```
resources: 
- nvidia.com/gpu 
```
config description：
image

node score:
```
finalScoreNode = (allocatablesResourcesNum - requestsResourcesNum) * framework.MaxNodeScore / allocatablesResourcesNum
```

### Why is this needed?

It’s introduced above


LY-today:
@googs1025 KEP

LY-today:
> @googs1025 KEP

@googs1025 Can you help advance this MR?

googs1025:
> > @googs1025 KEP
> 
> @googs1025 Can you help advance this MR?

Thanks for the invite, I'll handle this on weekend :)

LY-today:
> > > @googs1025 KEP
> > 
> > 
> > @googs1025 Can you help advance this MR?
> 
> Thanks for the invite, I'll handle this on weekend :)

thank you for your time

kep/625-node-resource-fit-plus-scoring/README.md
```diff
@@ -0,0 +1,114 @@
+# Node Resource Fit plus Scheduling
```
googs1025, line 1: This seems very similar to an old plugin. Can you help to tell the difference or integrate it?
FYI: https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/48-node-resources-allocatable-scoring

kep/625-node-resource-fit-plus-scoring/images/img1.png
googs1025, line 1: It is not recommended to use screenshots of tables and pictures.

kep/625-node-resource-fit-plus-scoring/README.md
```diff
@@ -0,0 +1,114 @@
...
+  - [Graduation Criteria](#graduation-criteria)
+    - [Alpha](#alpha)
+    - [Beta](#beta)
+- [Implementation History](#implementation-history)
+<!-- /toc -->
+
+
+## Summary
+
+The NodeResourcesFit plug-in of native k8s can only adopt a type of strategy for all resources, such as MostRequestedPriority and LeastRequestedPriority. However, in industrial practice, this design does not apply to some scenarios. For example: In AI scenarios, businesses that apply for GPUs prefer to occupy the entire GPU machine first to prevent GPU fragmentation; businesses that apply for CPU & MEM are prioritized and dispersed to non-GPU machines to prevent excessive consumption of CPU & MEM on GPU machines, resulting in real tasks of applying for GPUs. Pending due to insufficient non-GPU resources
```
googs1025, line 23: In the scenario of using gpu nodes, which are scarce resources, we should directly filter out the gpu nodes. Shouldn't this reduce the score? In addition, IIUC, gpu nodes (or other devices) are labeled (based on gpu-operator or nfd), and are generally filtered in this way.

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: @googs1025 The old version of the plug-in cannot distinguish between different strategies for resource use and can only treat them equally.

kep/625-node-resource-fit-plus-scoring/images/img1.png
LY-today: ok

LY-today: I'm going to make adjustments

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: Affinity strategies or nodeSelector require labeling nodes in advance, which is costly for cluster maintainers. The advantage of the strategy is to reduce this maintenance operation

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: Older version policies can only use one policy for different resources. Not suitable for complex resource scenarios, such as AI

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: Understand, labels can indeed be printed to distinguish machine types. Of course, it can also be done using the Affinity strategy. But what I want to say is that the above process has costs at the industrial practice level. 100 heterogeneous resources require the maintenance cost of 100 sets of labels.

LY-today: @googs1025 If you think that maintenance cost is not something that k8s needs to consider, then indeed the second expansion strategy does not need to be incorporated.

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: @googs1025 Under the AI ​​cluster. It is hoped that GPU tasks will be clustered and CPU tasks will be scattered. The old version of the strategy is not supported.

LY-today:
@googs1025 Hello, do you have any clear plans for these two plugins?

kep/625-node-resource-fit-plus-scoring/README.md
googs1025: As I mentioned, it seems to be very similar to the previous nodeResourcesAllocatable, and I don't think it needs to be extended with a new plugin. If it is possible, can it be integrated into the original plugin? 🤔

LY-today: @googs1025 What you mean is that you agree with this strategy, but the implementation method is changed to modify the original nodeResourcesAllocatable strategy.

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: @googs1025 Does the ScarceResourceAvoidance strategy have a clear conclusion? Accept or not?

googs1025: This is not for me to decide and can be left to other maintainers to suggest.

LY-today: @googs1025 Thank you for your feedback. Can you help me advance it and let other students review it?

kep/625-node-resource-fit-plus-scoring/README.md
LY-today: > @googs1025 What you mean is that you agree with the design of the NodeResourcesFitPlus strategy, but you want to implement it by modifying the original nodeResourcesAllocatable strategy?

@googs1025 Do I understand correctly?

LY-today:
@swatisehgal @zwpaper Please check

ffromani (commented):
initial review

LY-today:
@Huang-Wei Regarding plugin-2, how should I modify KEP? Is there any reference? Or is there something you don’t understand?

PR #855, opened 2025-01-13
fix(QOSSort): refine queue sorting logic by adding inqueue timestamp

googs1025:
#### What type of PR is this?
/kind bug

- After comparing with the k/k queuesort package, we seem to be missing timestamp comparison.
https://github.com/kubernetes/kubernetes/blob/35d6959ace07a4edc2a17da3bd4d0fcd625caa83/pkg/scheduler/framework/plugins/queuesort/priority_sort.go#L40-L49

Fixes #None

```release-note
None
```

### DIFF
--- a/test/integration/qos_test.go
+++ b/test/integration/qos_test.go
@@ -19,6 +19,9 @@ package integration
 import (
 	"context"
 	"fmt"
+	"reflect"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


googs1025:
friendly ping @Huang-Wei

test/integration/qos_test.go
```diff
@@ -114,19 +116,33 @@ func TestQOSPlugin(t *testing.T) {
...
 	// We will expect them to be scheduled in a reversed order.
 	t.Logf("Start to create 3 Pods.")
-	for i := range pods {
-		t.Logf("Creating Pod %q", pods[i].Name)
-		_, err = cs.CoreV1().Pods(ns).Create(testCtx.Ctx, pods[i], metav1.CreateOptions{})
-		if err != nil {
-			t.Fatalf("Failed to create Pod %q: %v", pods[i].Name, err)
-		}
+	// Concurrently create all Pods.
+	var wg sync.WaitGroup
```
googs1025, line 120: There seems to be some unstable flakes here, I still need to think about how to deal with it. 🤔

googs1025:
> @googs1025 this plugin was created just for demonstration purpose, and not enabled by default (if not in any case):
> 
> https://github.com/kubernetes-sigs/scheduler-plugins/blob/1e1b6b0332b84e83c629da6b51e43d2c3bfde3f1/cmd/scheduler/main.go#L63-L66
> 
> So it's not required to "sync" with the upstream plugin.

This pr is not to sync k/k upstream code. We are currently using the qos plugin internally. When we used it, we found that the sorting seemed to be a bit missing. The original logic only considered priority and qos type. IIUC, this is not to sync with the upstream, but to complete the missing part.

Huang-Wei:
> We are currently using the qos plugin internally. When we used it, we found that the sorting seemed to be a bit missing.

ok, good to know that.

/lgtm
/approve

PR #861, opened 2025-01-14
flake: fix qos sort plugin flaky integration test

googs1025:
#### What type of PR is this?
/kind flake

- During local testing, I found that the integration test was unstable. I guess that the creation and entry into the scheduling queue are not simultaneous, this causes unstable results.

Fixes #

```release-note
None
```

### DIFF
--- a/test/integration/qos_test.go
+++ b/test/integration/qos_test.go
@@ -98,18 +98,22 @@ func TestQOSPlugin(t *testing.T) {
 	// Make pods[1] Burstable.
 	pods[1].Spec.Containers[0].Resources = v1.ResourceRequirements{
 		Requests: v1.ResourceList{
+			v1.ResourceCPU:    *resource.NewMilliQuantity(50, resource.DecimalSI),
 			v1.ResourceMemory: *resource.NewQuantity(100, resource.DecimalSI),
 		},
 		Limits: v1.ResourceList{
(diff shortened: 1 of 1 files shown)


test/integration/qos_test.go
```diff
@@ -98,18 +98,22 @@ func TestQOSPlugin(t *testing.T) {
...
 		},
 		Limits: v1.ResourceList{
+			v1.ResourceCPU:    *resource.NewMilliQuantity(100, resource.DecimalSI),
 			v1.ResourceMemory: *resource.NewQuantity(200, resource.DecimalSI),
 		},
 	}
 	// Make pods[2] Guaranteed.
 	pods[2].Spec.Containers[0].Resources = v1.ResourceRequirements{
 		Requests: v1.ResourceList{
+			v1.ResourceCPU:    *resource.NewMilliQuantity(100, resource.DecimalSI),
```
googs1025, line 112: After some investigation, I found that the problem should be here: Guaranteed pod definition needs to be specified both cpu mem limit and request.  
FYI: https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#criteria

googs1025:
@Huang-Wei friendly ping :)

Huang-Wei (commented):
/lgtm
/approve

PR #881, opened 2025-02-26
add Cache to coscheduling plugin

KunWuLuan:
#### What type of PR is this?

/kind bug

No cache is registered for client in coscheduling, each time the plugin call GetPodGroup, it will send a api request to apiserver, which cause a huge latency.

Fixes #

```release-note

```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -29,6 +29,7 @@ import (
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
+	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/scheduler-plugins/apis/config"
@@ -74,7 +75,17 @@ func New(ctx context.Context, obj runtime.Object, handle framework.Handle) (fram
 	_ = clientscheme.AddToScheme(scheme)
 	_ = v1.AddToScheme(scheme)
 	_ = v1alpha1.AddToScheme(scheme)
-	client, err := client.New(handle.KubeConfig(), client.Options{Scheme: scheme})
+	ccache, err := ctrlcache.New(handle.KubeConfig(), ctrlcache.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		return nil, err
+	}
+	go ccache.Start(ctx)
+	ccache.WaitForCacheSync(ctx)
+	client, err := client.New(handle.KubeConfig(), client.Options{
+		Scheme: scheme,
+		Cache:  &client.CacheOptions{Reader: ccache}})
 	if err != nil {
 		return nil, err
 	}


pkg/coscheduling/coscheduling.go
```diff
@@ -71,10 +71,18 @@ func New(ctx context.Context, obj runtime.Object, handle framework.Handle) (fram
 	}
 
 	scheme := runtime.NewScheme()
-	_ = clientscheme.AddToScheme(scheme)
-	_ = v1.AddToScheme(scheme)
 	_ = v1alpha1.AddToScheme(scheme)
```
ffromani, line 74: why were the others `AddToScheme` removed?

pkg/coscheduling/coscheduling.go
```diff
@@ -71,10 +71,18 @@ func New(ctx context.Context, obj runtime.Object, handle framework.Handle) (fram
...
+		Scheme: scheme,
+	})
+	if err != nil {
+		return nil, err
+	}
+	go ccache.Start(ctx)
+	ccache.WaitForCacheSync(ctx)
+	client, err := client.New(handle.KubeConfig(), client.Options{
+		Scheme: scheme,
+		Cache:  &client.CacheOptions{Reader: ccache}})
```
ffromani, line 85: makes intuitively sense to me but I don't know if the `coscheduling` plugin makes assumption about the freshness of the data, so the cache can have unintended consequences. It would be better to have someone else more familiar with the coscheduling codebase to review.

ffromani (commented):
provisional LGTM, but I for myself would need to get more familiar with the `coscheduling` codebase

pkg/coscheduling/coscheduling.go
KunWuLuan: I don't find anywhere to call list or get or write to these resources.
And pods' schedule can still be completed after remove these AddToScheme.
Removing these AddToScheme can save memories because cache will not cache pods and nodes anymore.

ffromani: Which would be fine to me (if confirmed correct, haven't looked too deeply into this) but this seems unrelated to adding a cache; thus I'd move this specific change in its own commit if not its own pull request.

pkg/coscheduling/coscheduling.go
KunWuLuan: Ok, thanks. I think @denkensk @Huang-Wei are familiar with this plugin. Can you spend some time having a look at this pr?

pkg/coscheduling/coscheduling.go
KunWuLuan: Ok, I understood. I will move this to another pull request.

KunWuLuan:
And I believe I am familiar with coscheduling and elasticquota plugin. I can help to improve them it needed.

Chaunceyctx:
+1, there are not `cache` in `client`, so every `Get` or `List` request will bypass the cache and will be send to kube-apiserver directly. I noticed that we previously used the indexer of pg informer, but later switched to the current approach to adapt to the `controller-runtime` framework. It's possible that the cache initialization was overlooked during the adaptation process.

Huang-Wei:
> It's possible that the cache initialization was overlooked during the adaptation process.

Yup, we [migrated](https://github.com/kubernetes-sigs/scheduler-plugins/pull/652) to use controller-runtime client, but wasn't aware of the _breaking_ [change](https://github.com/kubernetes-sigs/controller-runtime/pull/2150) of controller-runtime.

@KunWuLuan could you help create a PR to consolidate changes need for all plugins? (not only coscheduling)

wzshiming:
In 0.30.6, I create 100 Pod as a group
image

With this patch, it behaves normally.

Huang-Wei:
> With this patch, it behaves normally

It'd be good to paste a picture after the fix ;)

ffromani:
thanks for the explanation. I must admit I didn't realize the client-go cache was disabled, and I kinda understood we were talking about an extra cache in layers above (this is what I meant in https://github.com/kubernetes-sigs/scheduler-plugins/pull/881#discussion_r1971085922). Thanks for catching this, we should by all means update all plugins and restore the cache

wzshiming:
> > With this patch, it behaves normally
> 
> It'd be good to paste a picture after the fix ;)

sure, 

image

Huang-Wei:
https://github.com/kubernetes-sigs/scheduler-plugins/pull/881#issuecomment-2734466705

Kindly ping @KunWuLuan .

KunWuLuan:
@Huang-Wei Sorry, I was busy for work in last month. Yes I can help to do that.

Huang-Wei:
@KunWuLuan I may release 0.31 soon and I do hope to contain this fix, so I'm going to create a generic fix based on this PR.

Huang-Wei:
Close in favor of #895

PR #882, opened 2025-02-27
remove useless scheme in coscheduling client

KunWuLuan:
#### What type of PR is this?
/kind cleanup

image

In #881, I added a cache for client in coscheduling plugin, this may cause listing of core resources like pods and nodes.

Client in coscheduling will only be used to get and list podgroups. Remove these code can save memory by preventing the listing of core resources like pods and nodes any more.

Fixes #

```release-note

```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -24,7 +24,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	clientscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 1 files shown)


ffromani (commented):
Am I missing something? this seems the same change as #881 - adding cache

Huang-Wei:
> Am I missing something? this seems the same change as #881 - adding cache

I had the same question... this PR looks the same as #881.

The changes are valid. Per https://github.com/kubernetes-sigs/scheduler-plugins/pull/881#issuecomment-2734466705, @KunWuLuan could you help consolidate changes needed for all plugins in one PR?

KunWuLuan:
@Huang-Wei @ffromani  Sorry I wanted to remove the useless scheme in this pr and it seems that I have some mistakes with the commits. I will correct this pr today.

ffromani:
/reopen

still valid

PR #883, opened 2025-03-26
chore: add more contextual logging info for plugins

googs1025:
#### What type of PR is this?
/kind cleanup

- add more info in contextual logging

We instead init the contextual logging of the entire plugin in the `New` and store it in the struct. In addition, we inject the ExtensionPoint name into each plugin ExtensionPoint, such as filter score ....
I understand that this will make our debug more efficient.

Fixes #

```release-note
None
```


Huang-Wei (commented):
/lgtm
/approve

Issue #874, opened 2025-01-25
follow up: coscheduling queue sort plugin starves pods


googs1025:
Since the issue https://github.com/kubernetes-sigs/scheduler-plugins/issues/110 has been open for a long time and has not been resolved, I would like to open a new issue as a follow up. 

There is no very depth research, but there are roughly two problems I think:
- The sort plugins use InitialAttemptTimestamp to sort, which can easily cause pod starvation for pods enqueue later than pggroup's pod.  Not all pods enqueued are managed by the podgroup, and there may be a large number of other pods. Therefore, it seems inappropriate to use only InitialAttemptTimestamp. based on https://github.com/kubernetes-sigs/scheduler-plugins/issues/110#issue-747926399

- When considering only InitialAttemptTimestamp enqueue in podgroup, there seems to be other problems: for example, podgroup should also consider minMember, we should satisfy most of the podgroup first, and then satisfy the pods of other numbers in the podgroup.

According to the above, can it be modified like this:

1. Priority Comparison:
First, compare the priorities of the two Pods.
2. PodGroup Check:
 to get the PodGroup info for each Pod.
 If one Pod does not belong to any PodGroup and the other does:
 We use podInfo timestamp sorting.
 If both Pods belong to PodGroups, proceed to the next step.

3. MinMember Comparison:
For Pods belonging to PodGroups, check whether each PodGroup has met its minMember requirement.
Give priority to PodGroups that have not yet met their minMember.
If both PodGroups do not meet their minMember or both meet their minMember, compare the creation time of the PodGroups.

googs1025:
similar issue: https://github.com/kubernetes-sigs/scheduler-plugins/issues/110#issuecomment-2611919472

googs1025:
@Huang-Wei @ffromani Could you please help me see if there is any problem with this idea? 🤔

Huang-Wei:
FYI: another proposal https://github.com/kubernetes-sigs/scheduler-plugins/pull/661 (closed due to inactivity)

Bowser1704:
> We use podInfo timestamp sorting.
If both Pods belong to PodGroups, proceed to the next step.

It maybe cause potential issues. Consider this scenario:

* podC: t4 (with pgc t0)
* podB: t2 (with pgb t1)
* podA: t3 (without pg)

so:

podC > podB, podB > podA, podA > podC

It's a circle, the sorting order is non-deterministic.

Bowser1704:
[...]

PR #884, opened 2025-04-01
feat: Optimize Pod QoS sort for coscheduling

dongjiang1989:
#### What type of PR is this?

/kind feature

1. Compare the priorities of Pods.
2. Compare the Qos of Pods.
3. Compare the initialization timestamps of PodGroups or Pods.
4. Compare the keys of PodGroups/Pods: <namespace>/<podname>.

Fixes #

```release-note
Optimize Pod QoS sort for coscheduling
```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
+	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


dongjiang1989:
Is the following order better?

1. Compare the priorities of Pods.
2. Compare the initialization timestamps of PodGroups or Pods.
3. Compare the Qos of Pods.
4. Compare the keys of PodGroups/Pods: /.

pkg/coscheduling/coscheduling.go
```diff
@@ -124,14 +125,26 @@ func (cs *Coscheduling) Name() string {
...
+// 2. Compare the Qos of Pods.
+// 3. Compare the initialization timestamps of PodGroups or Pods.
+// 4. Compare the keys of PodGroups/Pods: <namespace>/<podname>.
 func (cs *Coscheduling) Less(podInfo1, podInfo2 *framework.QueuedPodInfo) bool {
 	prio1 := corev1helpers.PodPriority(podInfo1.Pod)
 	prio2 := corev1helpers.PodPriority(podInfo2.Pod)
 	if prio1 != prio2 {
 		return prio1 > prio2
 	}
+
```
googs1025, line 137: Why do we need to obtain Qos type judgment in coscheduling? 🤔
If I understand correctly, pod have a priority defined by their own priority.
In QoS, it is used as a basis for judging whether to evict, and there is no need to sort QoS.

dongjiang1989: Hmmm.... Before comparing `CreateTime`, it may make more sense to compare the `QosClass` set by the user. 

Or, Compare `CreateTime` first and then compare `QosClass`?

googs1025: I understand createTime seems to be enough, because this change will not solve the issue: https://github.com/kubernetes-sigs/scheduler-plugins/issues/874

Issue #885, opened 2025-04-09
[NodeResourceTopology] Nodes with != `single-numa-node` topology-policy are automatically filtered-in


Tal-or:
### Area

- [x] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?

When nodes gets to the `Filter` stage in the NodeResoruceTopology scheduler,
the scheduler checks the topology policy of the node.

If the topology is != `single-numa-node` the scheduler consider the node to be a good candidate for scheduling (filtered-in) 
see: https://github.com/kubernetes-sigs/scheduler-plugins/blob/10d0e805b1937a7530312b18c8b61b0d48eff4d5/pkg/noderesourcetopology/filter.go#L227

It means that the node passed without any checks (skipped the scheduler logic) which is increasing the chances for a pod that is being scheduled on such node to hit TAE.

### What did you expect to happen?

The logic should be reversed, any node that is not configured with  `single-numa-node` policy should be filtered out (NOT in).

It makes more sense this way because a user that requests the pod to be scheduled with  NodeResoruceTopology scheduler is expecting the pod to avoid TAE. 

Reversing the logic will lean the balance towards pod get pending instead of hitting TAE.

### How can we reproduce it (as minimally and precisely as possible)?

1. Configure node with topology policy != `single-numa-node`
2. Schedule pod on this node (using node-selector) and specify under the pod `schedulerName` that the pod should schedule using NodeResoruceTopology scheduler
3. pod get scheduled and if there're not enough resources in a single numa it will hit TAE.

### Kubernetes version

Any version

### Scheduler Plugins version

Any version

googs1025:
/assign 
I will take a look this.

PR #886, opened 2025-04-16
bugfix(noderesourcetopology): reject non-single-numa-node policies in NodeResourceTopology plugin

googs1025:
#### What type of PR is this?
/kind bug

- reject non-single-numa-node policies in NodeResourceTopology plugin

Fixes https://github.com/kubernetes-sigs/scheduler-plugins/issues/885

```release-note

```


pkg/noderesourcetopology/filter.go
```diff
@@ -180,6 +180,11 @@ func singleNUMAPodLevelHandler(lh logr.Logger, pod *v1.Pod, zones topologyv1alph
 	return nil
 }
 
+func rejectNonSingleNUMANodeHandler(lh logr.Logger, _ *v1.Pod, _ topologyv1alpha2.ZoneList, nodeInfo *framework.NodeInfo) *framework.Status {
```
googs1025, line 183: we add another handler to handle this case

googs1025:
I'm not sure if my understanding is wrong. We have the following policy.

```go
           // RestrictedTopologyManagerPolicy is a mode in which kubelet only allows
	// pods with optimal NUMA node alignment for requested resources
	RestrictedTopologyManagerPolicy = "restricted"
	// BestEffortTopologyManagerPolicy is a mode in which kubelet will favour
	// pods with NUMA alignment of CPU and device resources.
	BestEffortTopologyManagerPolicy = "best-effort"
	// NoneTopologyManagerPolicy is a mode in which kubelet has no knowledge
	// of NUMA alignment of a pod's CPU and device resources.
	NoneTopologyManagerPolicy = "none"
	// SingleNumaNodeTopologyManagerPolicy is a mode in which kubelet only allows
	// pods with a single NUMA alignment of CPU and device resources.
	SingleNumaNodeTopologyManagerPolicy = "single-numa-node"
```
Do we only have the `none` case to handle (not sure)? Perhaps a better approach is for us to provide a `prefilter` `prescore` plugin that directly `skip status` when`policy == none`.

googs1025:
@ffromani friendly ping 😄 

If you have some time, could you give me some advice? I don't know much about this plugin, but I'd love to learn more, and it would be great if I could help.

ffromani:
Thanks for the fix. Codewise looks a good direction. The problem is designwise. We need to review the use cases and decide how do we want to treat these scenarios. There were reasons why the code currently auto-passes node without alignment requirements. We need to recover the past context, resume the conversation, re-evaluate and update the docs.

ffromani:
> @googs1025 what if we have cluster with mixed configuration and some nodes have `single-numa-node` policy enabled and some don't? Then we would always filter out nodes without `single-numa-node`?

this ^ mixed nodes was the contention point also back in time. Both approaches have pros and cons.

Our (RH) experience is that users _tend_ to have homogeneous cluster, but in the general case we support mixed clusters.

PiotrProkop:
> > @googs1025 what if we have cluster with mixed configuration and some nodes have `single-numa-node` policy enabled and some don't? Then we would always filter out nodes without `single-numa-node`?
> 
> this ^ mixed nodes was the contention point also back in time. Both approaches have pros and cons.
> 
> Our (RH) experience is that users _tend_ to have homogeneous cluster, but in the general case we support mixed clusters.

From my experience, we sometimes have mixed configuration due to node configuration changes etc. but we still want max cluster utilization.

ffromani:
I think both scenarios are valid and there are users for both scenarios. We can probably add an option to introduce the new behavior: reject-if-missing-data and keep the existing behavior allow-if-missing-data for backward compatibility.

ffromani:
> I think both scenarios are valid and there are users for both scenarios. We can probably add an option to introduce the new behavior: reject-if-missing-data and keep the existing behavior allow-if-missing-data for backward compatibility.

for clarity: the default should be the existing behavior

pkg/noderesourcetopology/filter.go
```diff
@@ -223,14 +228,17 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState *framework.Cycle
...
-	if conf.Scope == kubeletconfig.PodTopologyManagerScope {
+	switch conf.Scope {
+	case kubeletconfig.PodTopologyManagerScope:
 		return singleNUMAPodLevelHandler
-	}
-	if conf.Scope == kubeletconfig.ContainerTopologyManagerScope {
+	case kubeletconfig.ContainerTopologyManagerScope:
 		return singleNUMAContainerLevelHandler
+	default:
+		return nil
```
shajmakh, line 242: if the scope is not explicitly set it should be container by default according to
https://kubernetes.io/docs/tasks/administer-cluster/topology-manager/#topology-manager-scopes

Issue #887, opened 2025-04-16
feature: add more plugins args validation


googs1025:
Many plugins have their own args, but most of them don't have good validation of args. I will track this feature in this issue. 

some validation here: 
https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/apis/config/validation

PR #888, opened 2025-04-20
add validate for NodeResourcesAllocatable and Coscheduling plugin

googs1025:
#### What type of PR is this?
/kind feature
/kind cleanup

Fixes https://github.com/kubernetes-sigs/scheduler-plugins/issues/887

```release-note
None
```

### DIFF
--- a/apis/config/validation/validation_pluginargs.go
+++ b/apis/config/validation/validation_pluginargs.go
@@ -17,18 +17,33 @@ limitations under the License.
 package validation
 
 import (
+	"fmt"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	schedconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 
 	"sigs.k8s.io/scheduler-plugins/apis/config"
 )
 
-var validScoringStrategy = sets.NewString(
-	string(config.MostAllocated),
-	string(config.BalancedAllocation),
-	string(config.LeastAllocated),
-	string(config.LeastNUMANodes),
+var (
+	supportNodeResourcesMode sets.Set[string]
+	validScoringStrategy     sets.Set[string]
 )
+
+func init() {
+	supportNodeResourcesMode = sets.New[string](
+		string(config.Least),
+		string(config.Most),
+	)
+
+	validScoringStrategy = sets.New[string](
+		string(config.MostAllocated),
+		string(config.BalancedAllocation),
+		string(config.LeastAllocated),
+		string(config.LeastNUMANodes),
+	)
+}
 
 func ValidateNodeResourceTopologyMatchArgs(path *field.Path, args *config.NodeResourceTopologyMatchArgs) error {
 	var allErrs field.ErrorList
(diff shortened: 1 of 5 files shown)


apis/config/validation/validation_pluginargs.go
```diff
@@ -46,3 +49,56 @@ func validateScoringStrategyType(scoringStrategy config.ScoringStrategyType, pat
...
+			allErrs = append(allErrs, field.Invalid(p.Index(i).Child("weight"), resource.Weight, msg))
+		}
+	}
+	return allErrs
+}
+
+var supportNodeResourcesMode = sets.NewString(
+	string(config.Least),
+	string(config.Most),
+)
```
ffromani, line 67: this should probably be done in `init` or similar initialization function

apis/config/validation/validation_pluginargs.go
```diff
@@ -46,3 +49,56 @@ func validateScoringStrategyType(scoringStrategy config.ScoringStrategyType, pat
...
+
+func validateNodeResourcesModeType(mode config.ModeType, path *field.Path) *field.Error {
```
ffromani, line 69: if `ModeType` is an enum-like type (is it right?) how can it be different than the enumerated value? isn't the golang type system preventing this?

pkg/noderesources/allocatable.go
```diff
@@ -19,15 +19,14 @@ package noderesources
...
 	"fmt"
-	"math"
-
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog/v2"
 	schedulerconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/framework"
-
+	"math"
```
ffromani, line 27: why moving the import?

ffromani (commented):
overall seems OK, some inline comments

apis/config/validation/validation_pluginargs.go
googs1025: IIUC, we do not have enum restrictions on the configuration file, users may fill in any policy. 🤔 

```go

// NodeResourcesAllocatableArgs holds arguments used to configure NodeResourcesAllocatable plugin.
type NodeResourcesAllocatableArgs struct {

	// Resources to be considered when scoring.
	// Allowed weights start from 1.
	// An example resource set might include "cpu" (millicores) and "memory" (bytes)
	// with weights of 1<<20 and 1 respectfully. That would mean 1 MiB has equivalent
	// weight as 1 millicore.
	Resources []schedconfig.ResourceSpec `json:"resources,omitempty"`

	// Whether to prioritize nodes with least or most allocatable resources.
	Mode ModeType `json:"mode,omitempty"`
}
```

pkg/noderesources/allocatable.go
googs1025: removed

googs1025:
@Huang-Wei @ffromani /PTAL 😄

ffromani:
@googs1025 I'm generally in favor and at glance this seems like a good change. I will have a proper review ASAP. Could you please tag a coscheduling plugin author for review?

Huang-Wei (commented):
/lgtm
/approve

PR #895, opened 2025-04-27
Compose a cached reader as a cacheOption when initializing a controller-runtime client

Huang-Wei:
#### What type of PR is this?

/kind bug

An extended version of #881.

Fixes #

@ffromani I didn't fix the client construction logic in `pkg/noderesourcetopology`:

```
test/integration/noderesourcetopology_test.go
142:	extClient, err := ctrlclient.New(globalKubeConfig, ctrlclient.Options{Scheme: scheme})

test/integration/noderesourcetopology_cache_test.go
500:			extClient, err := ctrlclient.New(globalKubeConfig, ctrlclient.Options{Scheme: scheme})
1115:	extClient, err := ctrlclient.New(globalKubeConfig, ctrlclient.Options{Scheme: scheme})

pkg/util/client_util.go
22:	c, err := client.New(config, client.Options{

pkg/noderesourcetopology/pluginhelpers.go
47:	client, err := ctrlclient.NewWithWatch(handle.KubeConfig(), ctrlclient.Options{Scheme: scheme})
```

b/c `pkg/noderesourcetopology/pluginhelpers.go` calls a special `NewWithWatch` from controller-runtime. So I'm not sure if you want a cache-backed reader to bake into the client. Please let me know if you want to incorporate the fix into this plugin as well.

```release-note
Compose a cached reader as a cacheOption when initializing a controller-runtime client
```

### DIFF
--- a/pkg/util/client_util.go
+++ b/pkg/util/client_util.go
@@ -1,0 +1,29 @@
+package util
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// NewClientWithCachedReader returns a controller runtime Client with cache-baked client.
+func NewClientWithCachedReader(ctx context.Context, config *rest.Config, scheme *runtime.Scheme) (client.Client, cache.Cache, error) {
+	ccache, err := cache.New(config, cache.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+	go ccache.Start(ctx)
+	ccache.WaitForCacheSync(ctx)
+	c, err := client.New(config, client.Options{
+		Scheme: scheme,
+		Cache: &client.CacheOptions{
+			Reader: ccache,
+		},
+	})
+	return c, ccache, err
+}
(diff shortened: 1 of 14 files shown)


KunWuLuan:
@Huang-Wei Thanks. 

/lgtm

pkg/util/client_util.go
```diff
@@ -0,0 +1,29 @@
...
+// NewClientWithCachedReader returns a controller runtime Client with cache-baked client.
+func NewClientWithCachedReader(ctx context.Context, config *rest.Config, scheme *runtime.Scheme) (client.Client, cache.Cache, error) {
+	ccache, err := cache.New(config, cache.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		return nil, nil, err
+	}
+	go ccache.Start(ctx)
+	ccache.WaitForCacheSync(ctx)
```
googs1025, line 21: Do we need to verify that the sync was successful?
```go
if !ccache.WaitForCacheSync(ctx) {
    return nil, nil, fmt.Errorf("failed to sync cache")
}
```

pkg/util/client_util.go
```diff
@@ -0,0 +1,29 @@
...
+	c, err := client.New(config, client.Options{
```
googs1025, line 22: do we need check err here?

pkg/util/client_util.go
Huang-Wei: Good catch, could you help raise a follow-up PR?

pkg/util/client_util.go
Huang-Wei: We return the err, so should be ok to leave it for caller to handle

pkg/util/client_util.go
Huang-Wei: And could you possibly manually draft a cherrypick PR to `release-1.30` as auto-cherrypick has conflicts (https://github.com/kubernetes-sigs/scheduler-plugins/pull/895#issuecomment-2833185111). TIA!

PR #896, opened 2025-04-27
Automated cherry pick of #895: Compose a cached reader as a cacheOption when initializing a

googs1025:
Cherry pick of #895 on release-1.30.

#895: Compose a cached reader as a cacheOption when initializing a

For details on the cherry pick process, see the [cherry pick requests](https://git.k8s.io/community/contributors/devel/sig-release/cherry-picks.md) page.

```release-note
Compose a cached reader as a cacheOption when initializing a controller-runtime client
```

PR #897, opened 2025-04-27
clientutil: verify if cache is synced

Huang-Wei:
#### What type of PR is this?

/kind cleanup

A follow-up of https://github.com/kubernetes-sigs/scheduler-plugins/pull/895#discussion_r2062363236

It's manually patched to release-1.30: https://github.com/kubernetes-sigs/scheduler-plugins/pull/896#discussion_r2062479356

```release-note
NONE
```

### DIFF
--- a/pkg/util/client_util.go
+++ b/pkg/util/client_util.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"context"
+	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/rest"
@@ -19,6 +20,9 @@ func NewClientWithCachedReader(ctx context.Context, config *rest.Config, scheme 
 	}
 	go ccache.Start(ctx)
 	ccache.WaitForCacheSync(ctx)
+	if !ccache.WaitForCacheSync(ctx) {
+		return nil, nil, fmt.Errorf("failed to sync cache")
+	}
 	c, err := client.New(config, client.Options{
 		Scheme: scheme,
 		Cache: &client.CacheOptions{


pkg/util/client_util.go
```diff
@@ -19,6 +20,9 @@ func NewClientWithCachedReader(ctx context.Context, config *rest.Config, scheme
 	}
 	go ccache.Start(ctx)
 	ccache.WaitForCacheSync(ctx)
+	if !ccache.WaitForCacheSync(ctx) {
+		return nil, nil, fmt.Errorf("failed to sync cache")
+	}
```
tenzen-y, line 25:
```suggestion
	if !ccache.WaitForCacheSync(ctx) {
		return nil, nil, fmt.Errorf("failed to sync cache")
	}
```

Shouldn't we call the synchronizing cache only once?

Huang-Wei: We don't put a global cacheReader to include all potential CR schemes; instead, every plugin registers its own schema and hence instantiate its own cacheReader and call its own WaitForCacheSync.

@tenzen-y Do you see some side effect here?

tenzen-y: No, I do not observe side effects in the actual cluster. I was just wondering why we should check the cache sync twice.

Huang-Wei: > I was just wondering why we should check the cache sync twice.

What do you mean by "twice" here? (to ensure I fully understand your concern)

Huang-Wei: Ah sry... my bad

tenzen-y: Both line 22 and 23 execute waiting cache sync: https://github.com/Huang-Wei/scheduler-plugins/blob/65d77ef02af42b23b760c4c25b71cdab6ce5f4af/pkg/util/client_util.go#L22-L23

Huang-Wei: there is a typo...

Huang-Wei: https://github.com/kubernetes-sigs/scheduler-plugins/pull/899 will fix this.

Issue #890, opened 2025-04-25
Chart.yaml for 0.30.6 still uses 0.29.7


chotiwat:
### Area

- [ ] Scheduler
- [ ] Controller
- [x] Helm Chart
- [ ] Documents

### What happened?

See title.

https://github.com/kubernetes-sigs/scheduler-plugins/blob/v0.30.6/manifests/install/charts/as-a-second-scheduler/Chart.yaml#L15-L23

### What did you expect to happen?

The `version` and `appVersion` should match release version.

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

Huang-Wei:
I think I [cherrypicked](https://github.com/kubernetes-sigs/scheduler-plugins/pull/821) changes to `release-1.30`, but not pin it to tag `v0.30.6`, which is a process flaw - I should have cut the tag after the cherrypick PR.

I will refine the process a bit. Appologize for the confusion @chotiwat . Given that _usually_ a tag is not mutable, I'd prefer not to re-pin the right docs to `v0.30.6`. I will fix it in the next 0.31 release.

Huang-Wei:
@googs1025 could you help review #893? After that, I will:

- cut a tag `v0.30.12` to trigger Helm's chart publishing process
- submit a PR to promote the v0.30.12 image to `registry.k8s.io`

After these get completed, user can install v0.30.12 via `helm install --repo https://scheduler-plugins.sigs.k8s.io scheduler-plugins scheduler-plugins` - where the Helm tgz contains the correct values.yaml.

googs1025:
thanks for this quick fix 😄

Huang-Wei:
@googs1025 #900 is the last PR to fix this issue. After that PR gets merged, the latest Helm chart version will be pinned to 0.30.12.

chotiwat:
Thanks all for the quick fix and CR!

PR #900, opened 2025-04-28
helm: update the helm chart's index to include 0.30.12

Huang-Wei:
#### What type of PR is this?

/kind documentation
/area helm

Related with #890.

This PR introduces the following changes:

- update the index of Helm repo (https://scheduler-plugins.sigs.k8s.io) to pin `0.30.12` as the latest version
- remove 0.30.6 (which was misconfigured to point to 0.29.X)

```release-note
Index of Helm repo (https://scheduler-plugins.sigs.k8s.io) is updated to pin `0.30.12` as the latest version of release-1.30.
```

### DIFF
--- a/site/static/index.yaml
+++ b/site/static/index.yaml
@@ -2,15 +2,15 @@ apiVersion: v1
 entries:
   scheduler-plugins:
   - apiVersion: v2
-    appVersion: 0.30.6
-    created: "2024-11-03T16:22:31.375685-08:00"
+    appVersion: 0.30.12
+    created: "2025-04-27T17:23:15.733298-07:00"
     description: deploy scheduler plugin as a second scheduler in cluster
-    digest: ec1cb141128e3439e4f49f7da72e9919c999b243b5ff412c720bb7273cfbffe6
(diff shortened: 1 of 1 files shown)


site/static/index.yaml
```diff
@@ -2,15 +2,15 @@ apiVersion: v1
 entries:
   scheduler-plugins:
   - apiVersion: v2
-    appVersion: 0.30.6
```
googs1025, line 5: do we need to change here: https://scheduler-plugins.sigs.k8s.io/#compatibility-matrix ?

googs1025: and also README.md

googs1025:
/unhlod 

https://github.com/kubernetes-sigs/scheduler-plugins/pull/900#discussion_r2062766462
We can make change in other pr. this is not a block for this pr :)

site/static/index.yaml
Huang-Wei: yes, I'm gonna update in #898, and mark ti as ready to review soon.

Huang-Wei: @googs1025 #898 is ready to review

PR #905, opened 2025-05-01
feature: add MetricsUpdateIntervalSeconds args option in trimaran plugins

googs1025:
#### What type of PR is this?
/kind feature

This change introduces a new field `MetricsUpdateIntervalSeconds`, in the TrimaranSpecArgs. This allows users to define their own interval for periodic metrics updates instead of relying on a hardcoded value.

Benefits:
- Enhanced flexibility: Users can now customize how often metrics are updated based on their specific requirements.
- Better user experience: Providing configuration options improves the overall usability of the collector."

Fixes #

```release-note

```


apis/config/types.go
```diff
@@ -88,6 +88,8 @@ type TrimaranSpec struct {
 	MetricProvider MetricProviderSpec
 	// Address of load watcher service
 	WatcherAddress string
+	// Interval in seconds for periodic metrics updates. Default is 30 seconds if not set.
+	MetricsUpdateIntervalSeconds int32
```
ffromani, line 92: usually optional values are pointers to let disambiguate between "value not set" and "value explicitely set to zero". In this case, does the zero value make sense? do we have the possible ambiguity? if not we can probably use the value (vs the pointer) but the API side will need more consideration.

pkg/trimaran/collector.go
```diff
@@ -69,6 +69,11 @@ func NewCollector(logger klog.Logger, trimaranSpec *pluginConfig.TrimaranSpec) (
 		client, _ = loadwatcherapi.NewLibraryClient(opts)
 	}
 
+	metricsUpdateIntervalSeconds := trimaranSpec.MetricsUpdateIntervalSeconds
+	if metricsUpdateIntervalSeconds == 0 {
```
ffromani, line 73: `int32` can be negative. Do we have some validation preventing negative values?

ffromani (commented):
thanks for the PR. I see the reason and makes sense, but more code means more maintenance burden, so: do we have user request for this change or is it a "just in case" change?
in addition, I think we need unit tests and please please integration tests somehow

PR #908, opened 2025-05-19
nrt: log: return updated generation value for cache

shajmakh:
#### What type of PR is this?
/kind bug
A bug in the logs

The logs show misaligned generation value when tackling the same node in overreserve.go and filter.go logs where the former is always up to date while the latter is never updated (generation=0):

```
1 overreserve.go:310] "NodeTopology podset fingerprint mismatch" logger="nrtcache.resync" generation=3 node="node-1"
...
1 filter.go:202] "invalid topology data" logger="FilterWithNominatedPods.Filter.NodeResourceTopologyMatch" pod="default/test-6f798bf799-tj7bc" pod="default/test-6f798bf799-tj7bc" podUID="f221ce7c-09c9-4462-a67a-95f5fdec3fb7" node="node-1" generation=0
```

The only cache implementation that defines and uses "generation" is OverReserve thus this commit updates the method GetCachedNRTCopy() to return updated generation value in the returned CachedNRTInfo.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -340,7 +340,7 @@ func (ov *OverReserve) Resync() {
 }
 
 // FlushNodes drops all the cached information about a given node, resetting its state clean.
-func (ov *OverReserve) FlushNodes(lh logr.Logger, nrts ...*topologyv1alpha2.NodeResourceTopology) {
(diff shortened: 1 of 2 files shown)


ffromani:
thanks for the updates

shajmakh:
@ffromani Thanks for the valuable review and test guidance!

PR #911, opened 2025-05-24
Update scheduler-plugins to K8s 1.32.5

andrewbatesnetflix:
#### What type of PR is this?
/kind feature

This PR upgrades the `scheduler-plugins` project to Kubernetes **1.32.5**. 

Here's a rough checklist of changes that were made:
- Added an `upgrade-k8s.sh` script, then used that to target the new k8s versions in the `go.mod` file.
- Upgraded the Golang version to 1.23, the `controller-gen` version to `v0.18.0`, and the `controller-runtime` version to `v0.20.0`, all of which are in-line with 1.32.5. 
- Ran the `update-codegen.sh` script to re-generate files at the new k8s version.

Once that was finished, numerous code changes needed to be made to support the new k8s API changes. Most changes were relatively straightforward; adding/removing parameters to adhere to new method signatures, etc.

Here are a few notable callouts that weren't as straightforward:
- Refactored test usage of `framework.NodeToStatusMap` to use the new `framework.NodeToStatusReader` interface.
- Numerous tests needed to now initialize scheduler metrics via `metrics.Register()` in order to run. 
- `peaks.go` was refactored to use the new `resource.GetResourceRequestQuantity` method. 
- During testing, I noticed that the `cross-namespace preemption` test within `TestPostFilter` was flaking. The root cause was that `OrderedScoreFuncs` was returning `nil`, which was causing a nil pointer dereference panic. I refactored this method to instead return an empty, valid slice instead.

##### Testing
`make unit-test`, and this package is staged to be integrated into our Scheduler at Netflix. Unit tests are passing there as well.

```release-note
Some of the interface changes made in this PR will be backwards-incompatible with k8s 1.31.x and below. This change should only be integrated with Schedulers at 1.32.x.
```


go.mod
```diff
@@ -16,26 +16,27 @@ require (
...
+	k8s.io/component-helpers v0.32.5
 	k8s.io/klog/v2 v2.130.1
-	k8s.io/kube-scheduler v0.31.8
-	k8s.io/kubernetes v1.31.8
-	k8s.io/utils v0.0.0-20240711033017-18e509b52bc8
-	sigs.k8s.io/controller-runtime v0.19.2
+	k8s.io/kube-scheduler v0.32.5
+	k8s.io/kubernetes v1.32.5
+	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
+	sigs.k8s.io/controller-runtime v0.20.0
```
Huang-Wei, line 30: Maybe use the latest v0.20 patch version? i.e., `v0.20.4`.

config/crd/bases/scheduling.x-k8s.io_podgroups.yaml
```diff
@@ -4,7 +4,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-sigs/scheduler-plugins/pull/50
-    controller-gen.kubebuilder.io/version: v0.16.5
+    controller-gen.kubebuilder.io/version: v0.18.0
```
Huang-Wei, line 7: `v0.17.3` seems to be the "exact" match: https://github.com/kubernetes-sigs/controller-tools/blob/v0.17.3/go.mod#L16

hack/update-codegen.sh
```diff
@@ -23,7 +23,7 @@ SCRIPT_ROOT=$(dirname "${BASH_SOURCE[@]}")/..
 TOOLS_DIR=$(realpath ./hack/tools)
 TOOLS_BIN_DIR="${TOOLS_DIR}/bin"
 GO_INSTALL=$(realpath ./hack/go-install.sh)
-CONTROLLER_GEN_VER=v0.16.5
+CONTROLLER_GEN_VER=v0.18.0
```
Huang-Wei, line 26: v0.17.3 is the exact match: https://github.com/kubernetes-sigs/controller-tools/blob/v0.17.3/go.mod#L16

hack/upgrade-k8s.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
Huang-Wei, line 1:
```suggestion
#!/usr/env/bin bash
```

is more compatible and conventional.

hack/verify-crdgen.sh
```diff
@@ -27,7 +27,7 @@ CRD_OPTIONS="crd"
 
 # Download controller-gen locally
 CONTROLLER_GEN="${GOPATH}/bin/controller-gen"
-go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.5
+go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.18.0
```
Huang-Wei, line 30:
```suggestion
go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.17.3
```

pkg/trimaran/peaks/peaks_test.go
```diff
@@ -36,8 +30,14 @@ import (
...
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/queuesort"
 	"k8s.io/kubernetes/pkg/scheduler/framework/runtime"
+	"k8s.io/kubernetes/pkg/scheduler/metrics"
 	st "k8s.io/kubernetes/pkg/scheduler/testing"
 	tf "k8s.io/kubernetes/pkg/scheduler/testing/framework"
+	"math"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
```
Huang-Wei, line 40: Ditto.

pkg/podstate/pod_state_test.go
```diff
@@ -19,16 +19,16 @@ package podstate
...
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/informers"
 	clientsetfake "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/ktesting"
+	"k8s.io/kubernetes/pkg/scheduler/metrics"
+	"testing"
+	"time"
```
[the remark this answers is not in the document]
Huang-Wei: seems these changes were added back in the latest commit.

Huang-Wei:
@andrewbatesnetflix could you revert the change in the the commit [`91ac97d` (#911)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/911/commits/91ac97ddfe7367bb4e108a87d04ab7988be13a55)

pkg/podstate/pod_state_test.go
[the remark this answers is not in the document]
andrewbatesnetflix: Hey @Huang-Wei  - the CI bot was failing, complaining about a lack of `./hack/update-gofmt.sh`. Running that script I believe is what re-arranged these, but let me try.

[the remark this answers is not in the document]
andrewbatesnetflix: @Huang-Wei See this output to see what I mean: [link](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_scheduler-plugins/911/pull-scheduler-plugins-verify/1926684919166144512). Back to square one.

The prescribed solution in the logs is to run `./hack/update-gofmt.sh`, but undoes exactly what I just reverted. 

Please let me know if I'm missing something obvious.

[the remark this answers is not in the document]
Huang-Wei: ah, I see the issue here. It seems `gofmt` would re-order the imports if they are not separated by empty lines.

```diff
diff --git a/pkg/podstate/pod_state_test.go b/pkg/podstate/pod_state_test.go
index f5833196..d9676849 100644
--- a/pkg/podstate/pod_state_test.go
+++ b/pkg/podstate/pod_state_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"testing"
 	"time"
+
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -28,13 +29,13 @@ import (
 	clientsetfake "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/klog/v2"
 	"k8s.io/klog/v2/ktesting"
-	"k8s.io/kubernetes/pkg/scheduler/metrics"
-
 	"k8s.io/kubernetes/pkg/scheduler/framework"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/defaultbinder"
 	"k8s.io/kubernetes/pkg/scheduler/framework/plugins/queuesort"
 	frameworkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
+	"k8s.io/kubernetes/pkg/scheduler/metrics"
 	tf "k8s.io/kubernetes/pkg/scheduler/testing/framework"
+
 	testutil "sigs.k8s.io/scheduler-plugins/test/util"
 )
 
diff --git a/pkg/trimaran/peaks/peaks_test.go b/pkg/trimaran/peaks/peaks_test.go
index 0292473a..b6156e5b 100644
--- a/pkg/trimaran/peaks/peaks_test.go
+++ b/pkg/trimaran/peaks/peaks_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"testing"
+
 	"github.com/paypal/load-watcher/pkg/watcher"
 	"github.com/stretchr/testify/assert"
 	v1 "k8s.io/api/core/v1"

```

^^ would fix it.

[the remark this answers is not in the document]
andrewbatesnetflix: Yup! Looks like that did it. Diff looks better now, and the CI tool is passing. LMK if anything still looks off.

PR #942, opened 2025-12-12
Add NodeMetadata scoring plugin

alikhil:
#### What type of PR is this?

/kind feature

This PR introduces new score plugin which allows you to influence scheduling decisions based on custom node metadata. It supports two types of metadata:

1. **Numeric values** - Score nodes based on numeric labels/annotations
2. **Timestamps** - Score nodes based on timestamp labels/annotations

##### Use Cases

1. **Cost Optimization** - Prefer cheaper nodes
2. **Hardware Generation** - Prefer newer or older hardware
3. **Maintenance Windows** - Avoid recently maintained or prefer well-maintained nodes
4. **Custom Priorities** - Implement business-specific node prioritization
5. **Node Aging** - Distribute workloads based on node age
6. **Performance Tiers** - Score nodes by performance benchmarks

##### Behavior Notes

- If a node doesn't have the specified metadata key, it receives a score of 0 (lowest)
- Scores are normalized to the range [0, 100] (framework.MinNodeScore to framework.MaxNodeScore)
- The plugin can be combined with other scoring plugins using weighted scores
- Invalid metadata values (e.g., unparseable timestamps) result in a score of 0 for that node

##### Arguments

The plugin accepts the following configuration parameters:

| Parameter | Type | Required | Description |
|-----------|------|----------|-------------|
| `metadataKey` | string | Yes | The name of the label or annotation to use for scoring |
| `metadataSource` | string | Yes | Where to read metadata from: `"Label"` or `"Annotation"` |
| `metadataType` | string | Yes | Type of metadata value: `"Number"` or `"Timestamp"` |
| `scoringStrategy` | string | Yes | How to score nodes (see below) |
| `timestampFormat` | string | Conditional | Go time format string (required when `metadataType` is `"Timestamp"`) |

##### Example Score by Last Update Time (Newest Timestamp)

Prefer nodes that were updated most recently:

```yaml
apiVersion: kubescheduler.config.k8s.io/v1
profiles:
  - schedulerName: default-scheduler
    plugins:
      score:
        enabled:
          - name: NodeMetadata
    pluginConfig:
      - name: NodeMetadata
        args:
          metadataKey: "node.example.com/last-maintenance"
          metadataSource: "Annotation"
          metadataType: "Timestamp"
          scoringStrategy: "Newest"
          timestampFormat: "2006-01-02T15:04:05Z07:00"  # RFC3339
```

Node annotations:
```yaml
apiVersion: v1
kind: Node
metadata:
  name: node1
  annotations:
    node.example.com/last-maintenance: "2025-12-01T10:00:00Z"
---
apiVersion: v1
kind: Node
metadata:
  name: node2
  annotations:
    node.example.com/last-maintenance: "2025-12-03T10:00:00Z"
```

Result: `node2` gets a higher score than `node1` (more recently maintained).

Fixes #

```release-note
Add NodeMetadata scoring plugin
```


Huang-Wei:
@alikhil could you create a PR to author a KEP? (similar examples can be referred at https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep

alikhil:
@Huang-Wei here is a PR for KEP - https://github.com/kubernetes-sigs/scheduler-plugins/pull/974

My former coworkers at Criteo will rebase this PR soon. I no longer have access to the [fork repo](https://github.com/criteo-forks/scheduler-plugins/tree/alik-master)

josqu4red:
Branch rebased, please let me know if further updates are needed.

Thanks for following up on this @alikhil, much appreciated.

EDIT: working on a proper rebase

alikhil:
@Huang-Wei tests are passing now, could you re-review please?

pkg/nodemetadata/node_metadata.go
```diff
@@ -0,0 +1,241 @@
...
+			}
+			scores[i].Score = framework.MinNodeScore
+		}
+		logger.V(10).Info("All scores equal, normalized to MinNodeScore: ", "scores", scores, "pod", pod.Name)
+		return nil
+	}
+
+	// Normalize scores to [MinNodeScore, MaxNodeScore] range
+	oldRange := maxScore - minScore
+	newRange := framework.MaxNodeScore - framework.MinNodeScore
```
fanyang-real: Do you want to consider to normalize the valid nodes to be [minScore+1, maxScore] and reserve minScore to the invalid node so that you can avoid mixing invalid nodes with lowest nodes? Basically,
```
newRange := framework.MaxNodeScore - (framework.MinNodeScore + 1)
 // for valid nodes:
scores[i].Score = ((scores[i].Score-minScore)*newRange)/oldRange + (framework.MinNodeScore + 1)
 // for invalid:
scores[i].Score = framework.MinNodeScore
```

fanyang-real (approved):
node score logic part lgtm. thanks!

pkg/nodemetadata/node_metadata.go
```diff
@@ -0,0 +1,240 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Huang-Wei:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

pkg/nodemetadata/node_metadata.go
```diff
@@ -0,0 +1,240 @@
...
+
+	for i := range scores {
+		if scores[i].Score == invalidScore {
+			continue
+		}
+		if validCount == 0 {
+			minScore = scores[i].Score
+			maxScore = scores[i].Score
+		}
+		validCount++
```
Huang-Wei: If we initialize minScore, maxScore as MaxInt64, MinInt64, we should be able to eliminate `validCount` along with the `if` check.

pkg/nodemetadata/node_metadata.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+		if scores[i].Score > maxScore {
+			maxScore = scores[i].Score
+		}
+	}
+
+	if validCount == 0 {
+		for i := range scores {
+			scores[i].Score = framework.MinNodeScore
+		}
```
Huang-Wei, line 176: If we fall into this loop, that means all scores get assigned `invalidScore`, right? If so, we can just leave them as-is - re-assign them all to MinNodeScore is basically no-op.

pkg/nodemetadata/node_metadata_test.go
```diff
@@ -0,0 +1,668 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Huang-Wei:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

apis/config/validation/validation_pluginargs.go
```diff
@@ -113,3 +113,57 @@ func ValidateCoschedulingArgs(args *config.CoschedulingArgs, _ *field.Path) erro
...
+
+	// Validate compatibility between MetadataType and ScoringStrategy
+	if args.MetadataType == config.MetadataTypeNumber {
+		if args.ScoringStrategy == config.ScoringStrategyNewest || args.ScoringStrategy == config.ScoringStrategyOldest {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("scoringStrategy"),
+				args.ScoringStrategy, "scoringStrategy \"Newest\" and \"Oldest\" are only valid for metadataType \"Timestamp\""))
+		}
+	}
+
+	if args.MetadataType == config.MetadataTypeTimestamp {
```
Huang-Wei, line 158: In this case, what'd happen if TimestampFormat is an illegal string?

pkg/nodemetadata/node_metadata.go
josqu4red: Indeed but the invalid score is meant to be internal, and is changed to MinNodeScore to respect the "public" semantics.

apis/config/validation/validation_pluginargs.go
josqu4red: Currently an invalid format will lead to all nodes being assigned an invalid score then normalized to MinScore.

Huang-Wei: then I may prefer to add a follow-up to break early upon illegal timestamp.

remram44:
I am late to the party, but I would like to say that setting the default score (for non-labeled nodes) to something else than 0 would be useful. For example 100 or 1000. This would allow one to both prioritize certain nodes (by setting the value above the default) and de-prioritize nodes (by setting the value below the default).

The problem is that a label value has to start with an alphanumerical character, so **negative numbers are not possible**. So if the NodeMetadata default is 0, we can only get one of those behaviors:
- if I use `scoringStrategy: Highest`, I can only prioritize labeled nodes over non-labeled nodes
- if I use `scoringStrategy: Lowest`, I can only de-prioritize labeled nodes over non-labeled nodes

> `kubectl label node worker01 node.example.com/priority=-90`
> `error: invalid label value: "node.example.com/priority=-90": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')`

alikhil:
@remram44, you’re right, this is a real gap for label-based numeric metadata.

I opened [#1003](https://github.com/kubernetes-sigs/scheduler-plugins/pull/1003) to address it by adding an optional `defaultValue`. It lets operators assign a baseline to nodes missing the configured metadata, so unlabeled nodes can be ranked above or below labeled ones as needed.

Thank you for the feedback!

Issue #713, opened 2024-04-12
[Coscheduling] pending podgroup might lead to starvation by persistently assuming pods


lianghao208:
### Area

- [X] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### What happened?

- A podgroup with 10 min member, 9 pods have passed the filter and already assumed, stuck in permit stage, waiting for the 10th pod to pass the filter. 
- Unfortunately, the 10th pod failed to pass the filter, but the 9 pods has already consumed the resources(e.g. CPU, memory) in scheduler cache due to the `assume` stage.

In this case, all 10 pods cannot be scheduled successfully, but 9 pods have persistently consumed the nodes resources without running on them, which result in the waste of resources and starvation for other pods.

### What did you expect to happen?

Can we set a timeout for conscheduling permit plugin? After the maximum timeout, the whole podgroup will be rejected and release the resources for other pods. And all the pods from this podgroup will become unschedulable(placed in backoff queue) even if some of them has already in permit `wait` status.

I checked the kep of coscheduling, it has mentioned the `MaxScheduleTime` design, but it doesn't seem like it has been implemented yet.
https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/kep/42-podgroup-coscheduling/README.md?plain=1#L168

Not sure if I miss something, AFAIC, I do remember there was a MaxScheduleTime for podgroup, but I can't find it from the code base now.

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

Huang-Wei:
This has been somehow considered:

https://github.com/kubernetes-sigs/scheduler-plugins/blob/e3484e3150df761c22e1e7faeed4bd13b0a08dd2/pkg/coscheduling/coscheduling.go#L168-L174

In your case, it happens to be 10%, so the gang waits until timeout - the timeout is calculated by [global plugin's timeout and individual PodGroup's timeout](https://github.com/kubernetes-sigs/scheduler-plugins/blob/23e980c0e53bcbdaf08c9ae142f09fb0f92ee71b/pkg/util/podgroup.go#L64-L67).

If the PodGroup requires 9 as min member, and the 9th pod failed, then the gap (1/9) is less than 10%, so it will proactively reject the previous 8 waiting pods:

[...]

Issue #110, opened 2020-11-21
coscheduling queue sort plugin starves pods


mateuszlitwin:
Currently `coscheduling` plugin is using `InitialAttemptTimestamp` to compare pods of the same priority. If there are enough pods with early `InitialAttemptTimestamp` which cannot be scheduled then pods with later `InitialAttemptTimestamp` will get starved - scheduler will never attempt to schedule them. This is because scheduler will re-queue "early" pods before "later" pods are attempted. Normal scheduler is using time when pod was inserted into the queue, so this situation cannot occur.

Huang-Wei:
This sounds a reasonable optimization. @denkensk @cwdsuzhou thoughts?

denkensk:
@mateuszlitwin @Huang-Wei 
We may talk about it In the beginning. https://github.com/kubernetes/enhancements/pull/1463#discussion_r376465798
If we use the LastFailureTimestamp as the Normal scheduler. It will lead to undefined behavior in the heap.

Huang-Wei:
Ah true, I fail to notice that point.



> This is because scheduler will re-queue "early" pods before "later" pods are attempted.

@mateuszlitwin The failed PodGroup with an earlier timestamp will go through an internal backoff period, so that latter PodGroup is actually able to get scheduled, isn't it? If not, are you able to compose a simple test case to simulate this starvation?

cwdsuzhou:
> @mateuszlitwin @Huang-Wei
> We may talk about it In the beginning. [kubernetes/enhancements#1463 (comment)](https://github.com/kubernetes/enhancements/pull/1463#discussion_r376465798)
> If we use the LastFailureTimestamp as the Normal scheduler. It will lead to undefined behavior in the heap.

+1 for this

mateuszlitwin:
Might be hard to design a simple test. 

Issue occurred multiple times in the production environment where we had 100s of pods pending and 1000s of nodes to check. I observed that newer, recently created, pods were not attempted by the scheduler (based on lack of scheduling events and relevant logs), however older pods were attempted on a regular basis (but could not be scheduled because their scheduling constraints), at least once every sync. The issue went away when I disabled the coscheduling queue sort.

Maybe a test like this would reproduce the issue: 

- create say 500 pods that are unschedulable
- then create a single pod that could be scheduled (it's timestamp in the queue will be greater than other 500 pods)
- generate some fake events in the cluster to move pods from backoff/unschedulable queues back to the active queue

[...]

PR #949, opened 2026-03-09
Fix head-of-line blocking in coscheduling QueueSort

### LINKED ISSUE #874: follow up: coscheduling queue sort plugin starves pods

Huang-Wei:
#### What type of PR is this?

/kind bug

The coscheduling plugin's `Less()` QueueSort function sorts pods by their PodGroup's immutable `CreationTimestamp`. A large unschedulable PodGroup at the head of the queue blocks all newer, potentially schedulable PodGroups from being attempted, since the failed PodGroup always sorts first among same-priority groups.

Fix this by tracking the last time a PodGroup's scheduling attempt failed. When sorting, GetCreationTimestamp() now returns the failure timestamp instead of the immutable CreationTimestamp, effectively moving the failed PodGroup behind newer same-priority PodGroups or plain pods.

Fixes #874, #110, #713

Potential good-to-have follow-ups:

- [ ] Rename `GetCreationTimestamp` to `GetDerivedAttemptTimestamp`
- [ ] Enhance QueuingHint for coscheduling
- [ ] A period goroutine (e.g., every 10 mins) to check PodGroup's existence, and invalidate it in `lastFailedSchedulePG` if needed

```release-note
Fix head-of-line blocking in coscheduling QueueSort
```

### DIFF
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -93,7 +93,7 @@ type PodGroupManager struct {
 	backedOffPG *gocache.Cache
 	// lastFailedSchedulePG stores the last time a PodGroup's scheduling attempt failed.
 	// Used by GetCreationTimestamp to prevent head-of-line blocking.
-	lastFailedSchedulePG *gocache.Cache
+	lastFailedSchedulePG sync.Map
 	// podLister is pod lister
 	podLister listerv1.PodLister
 	// assignedPodsByPG stores the pods assumed or bound for podgroups
@@ -133,8 +133,8 @@ func NewPodGroupManager(client client.Client, snapshotSharedLister framework.Sha
 		podLister:            podInformer.Lister(),
 		permittedPG:          gocache.New(3*time.Second, 3*time.Second),
 		backedOffPG:          gocache.New(10*time.Second, 10*time.Second),
-		lastFailedSchedulePG: gocache.New(30*time.Minute, 10*time.Minute),
-		assignedPodsByPG:     map[string]sets.Set[string]{},
+		// lastFailedSchedulePG is a sync.Map, zero-value ready.
+		assignedPodsByPG: map[string]sets.Set[string]{},
 	}
 	podInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: AddPodFactory(pgMgr),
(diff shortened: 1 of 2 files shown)


Huang-Wei:
I reproduced the HOL issue locally using a one-node Kind cluster (12 CPU and ~1 CPU used for system pods):

- in **pkg/coscheduling/coscheduling.go**, inject `time.Sleep(100 * time.Millisecond)` in PreFilter() to mimic large cluster (5000 nodes) scheduling latency per pod
- **hol-blocker.yaml** - so it can occupy 0.23 X 50 = 11.5 CPUs 

```yaml
apiVersion: scheduling.x-k8s.io/v1alpha1
kind: PodGroup
metadata:
  name: pg1
spec:
  minMember: 50
  scheduleTimeoutSeconds: 30
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: pause
spec:
  replicas: 50
  selector:
    matchLabels:
      app: pause
  template:
    metadata:
      labels:
        app: pause
        scheduling.x-k8s.io/pod-group: pg1
    spec:
      schedulerName: coschduling
      containers:
      - name: pause
        image: registry.k8s.io/pause:3.6
        resources:
          requests:
            cpu: 230m
```
- **1-deploy.yaml** - a one-pod Deployment
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: pause2
spec:
  replicas: 1
  selector:
    matchLabels:
      app: pause2
  template:
    metadata:
      labels:
        app: pause2
    spec:
      schedulerName: mkube-allocator
      containers:
      - name: pause
        image: registry.k8s.io/pause:3.6
        resources:
          requests:
            cpu: "1"
```
- k apply -f hol-blocker.yaml && sleep 35 &&  k apply -f 1-deploy.yaml
- observe

## Without the fix

## With the fix

PR #953, opened 2026-03-18
feat: Add configurable pgRejectThreshold to Coscheduling plugin

ravikumarrvit:
#### What type of PR is this?

/kind feature

The PostFilter extension point previously used a hardcoded 10% threshold
to decide whether to reject a PodGroup when scheduling fails. If the
percentage of unassigned pods was at or below 10%, PostFilter would
optimistically skip rejection and let remaining pods retry. This
threshold was not configurable or documented.

This commit:
    - Adds PodGroupRejectPercentage (int32, 0-100) to CoschedulingArgs
      with a default of 10 (preserving existing behavior)
    - Replaces the hardcoded 0.1 in PostFilter with the configurable value
      (converted from percentage to fraction internally)
    - Adds validation ensuring the value is between 0 and 100
    - Adds tests for the new parameter (defaults, validation, PostFilter
      behavior at various percentage values)
    - Documents both podGroupBackoffSeconds and PodGroupRejectPercentage
      in the plugin docs and KEP

NONE

The newly introduced `PodGroupRejectPercentage` defaults to 10 (i.e. 10%) so there is no change in status quo unless the user chooses to configure this new setting differently for co-scheduling.

```release-note
Introduce the new PodGroupRejectPercentage config knob to support both toggle of the rejection w/ backoff feature as well granularly control the percentage of pods (not-assigned to nodes) that can lead to rejection w/ backoff.
```

### DIFF
--- a/apis/config/v1/types.go
+++ b/apis/config/v1/types.go
@@ -32,6 +32,11 @@ type CoschedulingArgs struct {
 	PermitWaitingTimeSeconds *int64 `json:"permitWaitingTimeSeconds,omitempty"`
 	// PodGroupBackoffSeconds is the backoff time in seconds before a pod group can be scheduled again.
 	PodGroupBackoffSeconds *int64 `json:"podGroupBackoffSeconds,omitempty"`
+	// PodGroupRejectPercentage is the percentage (0-100) of unassigned pods relative to
+	// minMember below which PostFilter will not reject the PodGroup.
+	// Default: 10 (10%). Set to 0 to always reject on any failure.
+	// Set to 100 to never reject (disable PostFilter group rejection).
+	PodGroupRejectPercentage *int32 `json:"podGroupRejectPercentage,omitempty"`
 }
 
 // ModeType is a type "string".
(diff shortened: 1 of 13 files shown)


apis/config/v1/types.go
```diff
@@ -32,6 +32,11 @@ type CoschedulingArgs struct {
 	PermitWaitingTimeSeconds *int64 `json:"permitWaitingTimeSeconds,omitempty"`
 	// PodGroupBackoffSeconds is the backoff time in seconds before a pod group can be scheduled again.
 	PodGroupBackoffSeconds *int64 `json:"podGroupBackoffSeconds,omitempty"`
+	// PodGroupRejectThreshold is the fraction (0.0-1.0) of unassigned pods relative to
+	// minMember below which PostFilter will not reject the PodGroup.
+	// Default: 0.1 (10%). Set to 0.0 to always reject on any failure.
+	// Set to 1.0 to never reject (disable PostFilter group rejection).
+	PodGroupRejectThreshold *float64 `json:"podGroupRejectThreshold,omitempty"`
```
Huang-Wei: I think `float64` is not recommended in k8s's type definition: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

> Avoid floating-point values as much as possible, and never use them in spec. Floating-point values cannot be reliably round-tripped (encoded and re-decoded) without changing, and have varying precision and representations across languages and architectures.

I'd recommend to change it to integer and rename to  `PodGroupRejectPercentage`

PR #957, opened 2026-04-13
coscheduling: fix data race in Permit() by upgrading RLock to Lock

binacs:
#### What type of PR is this?

/kind bug

Permit() writes to assignedPodsByPG map and inserts into the set, but incorrectly held a read lock (RLock). This races with concurrent access from binding cycle callbacks (PostBind/Unreserve) and pod event handlers that operate under write locks on the same mutex.

Fixes #

```release-note
NONE
```


Huang-Wei (commented):
/lgtm
/approve

PR #958, opened 2026-04-17
nrt: cache: fix assumedResource drift

ffromani:
#### What type of PR is this?

/kind bug
/kind cleanup

Previously, when a node was marked dirty, a successful Reserve()  would have marked it clean. The reasoning was that if a node gets selected - thus Reserve() called upon it, then obviously the overallocation was wrong and the node could have run more workload.

If Reserve() happens concurrently with Resync(), Reserve() can clear the dirty node flag and corrupt the
internal state. If that happens, unflushed nodes become effectively invisible to future resync cycles.
The assumedResources become stuck (never flushed) on the affected nodes.
    
The safest fix is remove the dirty node clearing shortcut in Reserve().

Furthermore, ReserveNodeResources() was unconditionally tracking assumedResources
regardless if a node actually had NRT data or not. This caused `assumedResources` to grow unbounded in these cases.  The reason why the data is never cleared is that all the paths that mark a node for resync (NodeMaybeOverReserved, NodeHasForeignPods, nodesWithAttrUpdate) require the NRT to already be in the store.

Fixes #N/A

Identified during the investigation of another issue: noderesourcetopologymatch doesn't handle correctly NRT objects created *after* the scheduler start. While this is not the recommended usage flow, it may happen on real clusters and must be supported. The fix for that issue depends on these fixes.

```release-note
Fix noderesourcetopology cache drift caused by TOCTOU race with its resync loop
```

### DIFF
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -149,6 +149,10 @@ func (ov *OverReserve) ReserveNodeResources(nodeName string, pod *corev1.Pod) {
 	lh := ov.lh.WithValues(logging.KeyPod, klog.KObj(pod), logging.KeyPodUID, logging.PodUID(pod), logging.KeyNode, nodeName)
 	ov.lock.Lock()
 	defer ov.lock.Unlock()
+	if !ov.nrts.Contains(nodeName) {
+		lh.V(5).Info("ignoring reserve", "nrtinfo", "missing")
+		return
+	}
 	nodeAssumedResources, ok := ov.assumedResources[nodeName]
 	if !ok {
 		nodeAssumedResources = newResourceStore(ov.lh)
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -157,9 +157,6 @@ func (ov *OverReserve) ReserveNodeResources(nodeName string, pod *corev1.Pod) {
 
 	nodeAssumedResources.AddPod(pod)
 	lh.V(2).Info("post reserve", logging.KeyNode, nodeName, "assumedResources", nodeAssumedResources.String())
-
-	ov.nodesMaybeOverreserved.Delete(nodeName)
-	lh.V(6).Info("reset discard counter", logging.KeyNode, nodeName)
 }
 
 func (ov *OverReserve) UnreserveNodeResources(nodeName string, pod *corev1.Pod) {
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -268,19 +268,26 @@ func (ov *OverReserve) Resync() {
 		return
 	}
 
+	nrtUpdates := ov.MakeNRTUpdatesForNodes(context.Background(), lh_, nodes)
+
+	ov.FlushNodes(lh_, nrtUpdates...)
+}
+
+func (ov *OverReserve) MakeNRTUpdatesForNodes(ctx context.Context, lh_ logr.Logger, nodes DesyncedNodes) []*topologyv1alpha2.NodeResourceTopology {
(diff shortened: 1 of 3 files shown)


pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -366,8 +366,18 @@ func (ov *OverReserve) FlushNodes(lh logr.Logger, nrts ...*topologyv1alpha2.Node
 }
 
 // to be used only in tests
-func (ov *OverReserve) Store() *nrtStore {
-	return ov.nrts
+func (ov *OverReserve) Inject(nrt *topologyv1alpha2.NodeResourceTopology) {
```
swatisehgal: Great that we no longer directly mutate internals but the name Inject() feels a bit vague for what it actually does. How about  `UpdateNRTForTest` or `StoreNRTForTest`?

Something slightly more explicit might improve readability for future test authors. Not a blocker in anyway, just naming feedback.

swatisehgal (commented):
`TestResyncFingerprintMismatchKeepsNodeDirty` is a great addition and covers an important failure path. However, we are not fully exercising the original race condition that motivated this fix:

1. node marked dirty
2. Resync() begins
3. concurrent Reserve() clears dirty bit
4. resync misses flush
5. assumed resources get stuck forever

The current tests validate the resulting behavior changes (which is good), but they  don’t explicitly model this sequence.

I understand writing a true concurrent test may be difficult and potentially flaky, but I was wondering if we can look into adding a deterministic regression test that simulates this ordering to ensure we’re directly covering the original failure.

pkg/noderesourcetopology/cache/overreserve.go
ffromani: good point, changing

ffromani:
> `TestResyncFingerprintMismatchKeepsNodeDirty` is a great addition and covers an important failure path. However, we are not fully exercising the original race condition that motivated this fix:
> 
>     1. node marked dirty
> 
>     2. Resync() begins
> 
>     3. concurrent Reserve() clears dirty bit
> 
>     4. resync misses flush
> 
>     5. assumed resources get stuck forever
> 
> 
> The current tests validate the resulting behavior changes (which is good), but they don’t explicitly model this sequence.
> 
> I understand writing a true concurrent test may be difficult and potentially flaky, but I was wondering if we can look into adding a deterministic regression test that simulates this ordering to ensure we’re directly covering the original failure.

the problem here is that if we want to avoid concurrency (which we should), then we need to somehow call Reserve() during Resync(). There is not a clear path to do that without test-only code (which I'd like to avoid) or adding a test-only hook (likewise)

ffromani:
> > `TestResyncFingerprintMismatchKeepsNodeDirty` is a great addition and covers an important failure path. However, we are not fully exercising the original race condition that motivated this fix:
> > ```
> > 1. node marked dirty
> > 
> > 2. Resync() begins
> > 
> > 3. concurrent Reserve() clears dirty bit
> > 
> > 4. resync misses flush
> > 
> > 5. assumed resources get stuck forever
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The current tests validate the resulting behavior changes (which is good), but they don’t explicitly model this sequence.
> > I understand writing a true concurrent test may be difficult and potentially flaky, but I was wondering if we can look into adding a deterministic regression test that simulates this ordering to ensure we’re directly covering the original failure.
> 
> the problem here is that if we want to avoid concurrency (which we should), then we need to somehow call Reserve() during Resync(). There is not a clear path to do that without test-only code (which I'd like to avoid) or adding a test-only hook (likewise)

OTOH I may have found a good enough decompoisition. Updating.

ffromani:
/hold

because WIP commits

ffromani:
/hold cancel

happy with the shape now and reviewable again

swatisehgal (commented):
/lgtm

PR #964, opened 2026-04-19
release: pin gh actions to commit SHAs and replace softprops/action-gh-release with gh CLI

Huang-Wei:
#### What type of PR is this?

/kind cleanup
/area release

#### What this PR does / why we need it:

It seems latests GitHub action requires a sha commit instead of tag:

```
Error: The actions actions/checkout@v4, azure/setup-helm@v3, and softprops/action-gh-release@v1 are not allowed in kubernetes-sigs/scheduler-plugins because all actions must be pinned to a full-length commit SHA.
```

This PR introduces changes to:

- Pin the following actions to their verified commit SHAs:
    - actions/checkout@v4 → 34e114876b0b11c390a56381ad16ebd13914f8d5
    - azure/setup-helm@v3 → 5119fcb9089d432beecbf79bb2c7915207344b78
- Eliminate dependency on softprops/action-gh-release to use `gh` CLI directly

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #966, opened 2026-04-19
fix: create GitHub release before uploading helm chart asset

Huang-Wei:
#### What type of PR is this?

/kind bug
/kind cleanup
/area release

The previous fix (fdb7d62) replaced softprops/action-gh-release with `gh release upload`, but `gh release upload` only attaches assets to an existing release - it does not create one. The old action handled release creation implicitly, so the helm chart upload now fails with "release not found" when no release exists for the tag.

```
Run gh release upload v0.34.7 *.tgz
  gh release upload v0.34.7 *.tgz
  shell: /usr/bin/bash -e {0}
  env:
    GH_TOKEN: ***
release not found
Error: Process completed with exit code 1.
```

This PR fixes this by first running `gh release create` (with `|| true` to no-op if the release already exists), then uploading the chart with `--clobber` to allow safe re-runs.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -25,4 +25,6 @@ jobs:
       if: startsWith(github.ref, 'refs/tags/')
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      run: gh release upload ${{ github.ref_name }} *.tgz
+      run: |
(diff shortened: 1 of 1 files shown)


Huang-Wei:
cc @tenzen-y 

A follow-up of #964

tenzen-y (commented):
/lgtm
/approve

PR #970, opened 2026-04-21
upgrade k8s to v1.35

fanyang-real:
#### What type of PR is this?

/kind cleanup

Upgrade k8s to v1.35.4

Fixes #

In k8s v1.35, several scheduler framework types (`Handle`, `Plugin`, `PreFilterPlugin`, `ScorePlugin`, `NodeScoreList`, PostFilterResult`, `SharedLister`, `WaitingPod`, `PermitPlugin`, etc.) were removed from `k8s.io/kubernetes/pkg/scheduler/framework` and are now only available from `k8s.io/kube-scheduler/framework`. All references updated from `framework.X` to `fwk.X` accordingly. `NodeToStatusMap` was also replaced with `NodeToStatusReader`.

Deleted `hack/net_tls_stat.go` as it was a patch for an older prometheus/procfs version that now includes TLS stat support natively.

```release-note
Upgrade k8s to v1.35.4
```


Huang-Wei (commented):
/lgtm
/approve

Thanks @fanyang-real !

PR #971, opened 2026-04-21
restrict helm release workflow to exact vX.Y.Z tags

Huang-Wei:
#### What type of PR is this?

/kind cleanup
/area release

Avoid triggering the release workflow on pre-release or suffixed tags like v1.2.3-rc1 by narrowing the tag filter pattern.

Fixes #

```release-note
release: restrict helm release workflow to exact vX.Y.Z tags
```

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -2,7 +2,7 @@ name: Releases
 on:
   push:
     tags:
-    - v*
+    - 'v[0-9]+.[0-9]+.[0-9]+'
 
 jobs:
   releases:


zwpaper:
Hi @Huang-Wei, sorry for the long silence on this project. I've recently transitioned to a new field for work, and getting up to speed there has been taking up most of my time and energy.

I just take a look at the changes, this might not work as you expected since github workflow do NOT support regexp.

you may have to use some condition like the following to achieve your need:

ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpull_requestpull_request_targetbranchesbranches-ignore

```yaml
on:
  push:
    tags:
      - 'v*.*.*'

jobs:
  check-tag:
    runs-on: ubuntu-latest
    outputs:
      is_valid: ${{ steps.regex_check.outputs.is_valid }}
    steps:
      - name: Validate tag format with Regex
        id: regex_check
        run: |
          # Use Bash regex matching
          if [[ "${{ github.ref_name }}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
            echo "Tag is valid. Proceeding."
            echo "is_valid=true" >> "$GITHUB_OUTPUT"
          else
            echo "Tag does not match strict vX.Y.Z format. Skipping downstream jobs."
            echo "is_valid=false" >> "$GITHUB_OUTPUT"
          fi
```

Huang-Wei:
Thanks @zwpaper, and welcome back. No worries about the silence! 🙏

I _think_ the current syntax may work according to https://github.com/kubernetes-sigs/scheduler-plugins/pull/976:

> You can use special characters in path, branch, and tag filters.
> - `+` Matches one or more of the preceding character.
> - `[]` Matches one alphanumeric character listed in the brackets or contained in ranges. Ranges can only include `a-z`, `A-Z`, and `0-9`.

image

Let me know if you spot a gap - thanks again for the review!

ffromani (commented):
/lgtm

the current syntax ` 'v[0-9]+.[0-9]+.[0-9]+'` AFAIK works - I'm using something similar myself in other projects.

ffromani:
mh, no longer sure it's a flake, at glance the coscheduling failure looks consistent now

Huang-Wei:
> mh, no longer sure it's a flake, at glance the coscheduling failure looks consistent now

Rebasing now to see if it helps.

Huang-Wei:
> Rebasing now to see if it helps.

No luck. I think I need to revert the coscheduling queuinghint change: #983

PR #972, opened 2026-04-22
add QueueingHintFn to coscheduling EventsToRegister

Huang-Wei:
#### What type of PR is this?

/kind feature
/kind flake

Add precise hint functions so the scheduler only retries a pod when:
- A sibling pod (same PodGroup label) is added/deleted
- Pods get deleted
- The pod's own PodGroup is created or updated

~~This should be able to also reduce long duration (sometime it even timed out) of integration tests.~~ No, it doesn't seem so.

Fixes #

```release-note
feat: add QueueingHintFn to coscheduling EventsToRegister
```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -28,6 +28,7 @@ import (
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
 	fwk "k8s.io/kube-scheduler/framework"
+	schedutil "k8s.io/kubernetes/pkg/scheduler/util"
 	"sigs.k8s.io/scheduler-plugins/apis/config"
 	"sigs.k8s.io/scheduler-plugins/apis/config/validation"
 	"sigs.k8s.io/scheduler-plugins/apis/scheduling"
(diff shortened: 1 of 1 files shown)


Huang-Wei:
cc @tenzen-y in a previous PR, this was proposed.

ffromani (commented):
IIUC/IIRC the podGroup doesn't necessarily require all pods to be scheduled on the same node - I guess this can actually backfire if all the pods in a podGroup require access to an expensive device e.g. GPU.

I wonder if we have a way to measure the gains.

all that said, the change per se LGTM, I can add a formal LGTM once the question above is answered.

Huang-Wei:
> IIUC/IIRC the podGroup doesn't necessarily require all pods to be scheduled on the same node

This is true.

> - I guess this can actually backfire if all the pods in a podGroup require access to an expensive device e.g. GPU.

I'm not sure I fully understand the correlation of this statement with this PR. Could you elaborate?

> I wonder if we have a way to measure the gains.

The gain doesn't seem to be measurable by integration test. However, a straightforward comparison vs. w/o this pr is:

- in the before, all Pod & PodGroup's add will trigger pods moving insider scheduler's internal queue, which sometimes is a waste.
- now, only the pods matches the Queuing function will be legit to be re-queued.

pkg/coscheduling/coscheduling.go
```diff
@@ -111,16 +112,71 @@ func New(ctx context.Context, obj runtime.Object, handle fwk.Handle) (fwk.Plugin
...
-	// Please follow: eventhandlers.go#L403-L410
 	pgGVK := fmt.Sprintf("podgroups.v1alpha1.%v", scheduling.GroupName)
 	return []fwk.ClusterEventWithHint{
-		{Event: fwk.ClusterEvent{Resource: fwk.Pod, ActionType: fwk.Add}},
-		{Event: fwk.ClusterEvent{Resource: fwk.EventResource(pgGVK), ActionType: fwk.Add | fwk.Update}},
+		// A new pod being added might complete the sibling count needed for a PodGroup's minMember.
+		{Event: fwk.ClusterEvent{Resource: fwk.Pod, ActionType: fwk.Add}, QueueingHintFn: cs.isSchedulableAfterPodAdded},
+		// An assigned pod being deleted (or un-assumed after Permit timeout) frees cluster resources,
+		// which may allow a previously resource-blocked PodGroup to schedule.
+		{Event: fwk.ClusterEvent{Resource: fwk.Pod, ActionType: fwk.Delete}, QueueingHintFn: cs.isSchedulableAfterPodDeleted},
```
Huang-Wei, line 121: Add fwk.Delete is b/c we support MinResources in PodGroup

ffromani:
> > IIUC/IIRC the podGroup doesn't necessarily require all pods to be scheduled on the same node
> 
> This is true.
> 
> > * I guess this can actually backfire if all the pods in a podGroup require access to an expensive device e.g. GPU.
> 
> I'm not sure I fully understand the correlation of this statement with this PR. Could you elaborate?

Because is not! I was thinking aloud about the reason why we don't _want_ all the podgroup pods to land on the same node. No consequences for this PR. Sorry for the confusion.

Huang-Wei:
This PR may help de-flake recent integration timeouts observed: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_scheduler-plugins/976/pull-scheduler-plugins-integration-test-master/2060500882667081728

ffromani:
sorry for the delay. Thanks for the clarifications. I'll have another review pass ASAP to unblock this PR

ffromani (commented):
/lgtm

PR #973, opened 2026-04-22
nrt: overreserve: support ingestion of nrt objects created after startup

### LINKED ISSUE #967: [NodeTopologyMatch] NodeResourceTopology objects *created* after scheduler startup are ignored
### Area

- [x] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

The NodeResourceTopology objects created (not updated!) post scheduler startup are ignored by the NodeTopologyMatch plugin, causing incorrect scheduling.

With V>=4 we have evidence the data is being ignored ("modified non-existent NRT"") but otherwise the issue is kinda hard to pinpoint.

The recommended deployment and usage flow is to make sure all NRT objects are *created* (API create) before to deploy the scheduler w/ NodeTopologyMatch enabled, but in practices nodes can be added, reprovisioned, updated and long story short there's quite few legit cases on which this issue can emerge.

The NodeTopologyMatch plugin should just handle NRT objects created after it started.

The only current workaround is to restart the scheduler, which kinda works but is not great.

NodeTopologyMatch should handle newly-created NRT objects created (kube API create) after its startup.

Add a node to your cluster, or alternative deploy a NRT updating agent like nfd-topology-updater, after the scheduler with NodeTopologyMatch configured started.

N/A

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

ffromani:
#### What type of PR is this?

/kind bug
/kind cleanup

Adds support to ingest NRT objects created after the scheduler started

Fixes #967 

Depends on #958 
AI usage disclosure: AI was used for self-review, partially for initial prototyping, for iterating over tests and for boilerplate code.
I run `update-codegen.sh` and intentionally discarded the changes the tool produces for `pkg/generated` because didn't seem relevant for this PR.

```release-note
The NodeTopologyMatch plugin gained an option to optionally reject nodes without topology data.
The NodeTopologyMatch plugin now handles NodeResourceTopology objects created after its startup
```

### DIFF
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -276,26 +276,25 @@ func (ov *OverReserve) GetDesyncedNodes(lh logr.Logger) DesyncedNodes {
 	nodes := ov.nodesWithForeignPods.Clone()
 	foreignCount := nodes.Len()
 
-	overreservedCount := ov.nodesMaybeOverreserved.Len()
 	for _, node := range ov.nodesMaybeOverreserved.Keys() {
 		nodes.Incr(node)
 	}
 
 	// always use local copies
 	configChangeNodes := ov.nodesWithAttrUpdate.Clone()
-	configChangeCount := configChangeNodes.Len()
-
 	newlyAddedNodes := ov.nodesNewlyAdded.Clone()
 
-	if nodes.Len() > 0 {
-		lh.V(4).Info("found dirty nodes", "foreign", foreignCount, "discarded", overreservedCount, "configChange", configChangeCount, "total", nodes.Len())
-	}
-	return DesyncedNodes{
+	dn := DesyncedNodes{
 		Generation:        ov.generation,
 		MaybeOverReserved: nodes.Keys(),
 		ConfigChanged:     configChangeNodes.Keys(),
 		NewlyAdded:        newlyAddedNodes.Keys(),
 	}
+	if dn.Len() > 0 {
+		lh.V(4).Info("desynced nodes", "foreignCount", foreignCount, "stats", dn.String())
+	}
+	return dn
+
 }
 
 // Resync implements the cache resync loop step. This function checks if the latest available NRT information received matches the
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -36,9 +36,17 @@ import (
 
 	apiconfig "sigs.k8s.io/scheduler-plugins/apis/config"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/logging"
+	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/nodeconfig"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/podprovider"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/resourcerequests"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/stringify"
+)
+
+const (
+	// defaultMaxNRTUpdates is the watcher update channel size initial heuristic, no hard data.
+	// we used a bounded channel to prevent excessive event accumulation. The watcher code
+	// is now in charge to detect and handle overflow using retries.
+	defaultMaxNRTUpdates = 128
 )
 
 type OverReserve struct {
@@ -58,6 +66,7 @@ type OverReserve struct {
 	resyncMethod           apiconfig.CacheResyncMethod
 	resyncScope            apiconfig.CacheResyncScope
 	isPodRelevant          podprovider.PodFilterFunc
+	nrtUpdateCh            chan string
 }
 
 func NewOverReserve(ctx context.Context, lh logr.Logger, cfg *apiconfig.NodeResourceTopologyCache, client ctrlclient.WithWatch, podLister podlisterv1.PodLister, isPodRelevant podprovider.PodFilterFunc) (*OverReserve, error) {
@@ -83,6 +92,7 @@ func NewOverReserve(ctx context.Context, lh logr.Logger, cfg *apiconfig.NodeReso
 		nodesMaybeOverreserved: newCounter(),
 		nodesWithForeignPods:   newCounter(),
 		nodesWithAttrUpdate:    newCounter(),
+		nrtUpdateCh:            make(chan string, defaultMaxNRTUpdates),
 		podLister:              podLister,
 		resyncMethod:           resyncMethod,
 		isPodRelevant:          isPodRelevant,
@@ -90,9 +100,9 @@ func NewOverReserve(ctx context.Context, lh logr.Logger, cfg *apiconfig.NodeReso
 
 	if resyncScope == apiconfig.CacheResyncScopeAll {
 		wt := Watcher{
-			lh:    obj.lh,
-			nrts:  obj.nrts,
-			nodes: obj.nodesWithAttrUpdate,
+			lh:       obj.lh,
+			eventCh:  obj.nrtUpdateCh,
+			lastConf: make(map[string]nodeconfig.TopologyManager),
 		}
 		go wt.NodeResourceTopologies(ctx, client)
 	}
@@ -181,6 +191,8 @@ func (ov *OverReserve) UnreserveNodeResources(nodeName string, pod *corev1.Pod) 
 	lh.V(2).Info("post unreserve", logging.KeyNode, nodeName, "assumedResources", nodeAssumedResources.String())
 }
 
+func (ov *OverReserve) PostBind(nodeName string, pod *corev1.Pod) {}
+
 type DesyncedNodes struct {
 	Generation        uint64
 	MaybeOverReserved []string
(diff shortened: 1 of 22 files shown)


ffromani:
/cc @swatisehgal @Tal-or 

please be aware of the dependant PR

ffromani:
/hold

actually let me think about integration tests before review

ffromani:

unrelated failure it seems? investigating separately

ffromani:
/hold cancel

ready for review at last

pkg/noderesourcetopology/cache/watch.go
```diff
@@ -65,35 +72,33 @@ func (wt Watcher) NodeResourceTopologies(ctx context.Context, client ctrlclient.
 	}
 }
 
-func (wt Watcher) ProcessEvent(ev watch.Event) bool {
+func (wt Watcher) processEvent(ev watch.Event) {
+	// TODO: handle node deletion. How common do we expect it to be?
+	// Modified is trivially verified; Added can happen if we
```
Tal-or: Isn't that the case we're aiming to cover? i.e new NRT created (`ev.Type == watch.Added`)

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -36,11 +36,19 @@ import (
...
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/podprovider"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/resourcerequests"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/stringify"
 )
 
+const (
+	// defaultMaxNRTUpdates is the watcher update channel size initial heuristic, no hard data.
+	// we used a bounded channel to prevent excessive event accumulation. The watcher code
+	// is now in charge to detect and handle overflow using retries.
+	defaultMaxNRTUpdates = 128
```
Tal-or, line 52: Did we take into account how this number will take affect on big clusters? 
We're committing to handle 500 nodes cluster size

pkg/noderesourcetopology/cache/watch.go
Tal-or: Changed in next commit NVM

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -313,55 +315,52 @@ func (ov *OverReserve) MakeNRTUpdatesForNodes(ctx context.Context, lh_ logr.Logg
...
+	if part := ov.makeNRTUpdatesForNodes(ctx, lh_, ov.client, nodes.MaybeOverReserved, "resynced", isNRTFresher); len(part) > 0 {
+		nrtUpdates = append(nrtUpdates, part...)
 	}
 
-	for _, nodeName := range nodes.ConfigChanged {
+	alwaysPass := func(_ logr.Logger, _ *topologyv1alpha2.NodeResourceTopology) error { return nil }
+	if part := ov.makeNRTUpdatesForNodes(ctx, lh_, ov.client, nodes.ConfigChanged, "configChanged", alwaysPass); len(part) > 0 {
+		nrtUpdates = append(nrtUpdates, part...)
+	}
+
```
Tal-or, line 394: nit: extra new line (we removed the exact same on line 376)

Tal-or (commented):
Overall looking good, minor comments inside.

Also a question - with this new implementation the sole owner of the `nrtStore` is the `OverReserve` so why does the store still need a `Mutex` for an access?

ffromani:
/cc @PiotrProkop 

in this PR there's a change outside cache/overreserve (backward compatible by default) that you may found useful

PiotrProkop:
that's an interesting one, looks very interesting, will review later today.

thc1006:
I was reading the watcher to understand how the resync loop gets fed, and ended up putting its lifecycle through a few tests. Three things came out of it, all in `watch.go`.

The one I would worry about most is the closed result channel. `case ev := <-wa.ResultChan()` drops the receive `ok`, so once the apiserver closes the watch that case is permanently ready, `processEvent` discards the zero event, and the loop keeps going round. With a fake watch whose channel is already closed and the context still live, the goroutine burned 506ms of CPU over 500ms of wall clock, so a whole core, and it never tries to reconnect. The outer `for !done` reads like it is there to re-establish the watch, but the inner loop only leaves through `ctx.Done()`, which sets `done` on the way out, so nothing ever reaches it. Breaking out on `!ok` would give that outer loop something to do, with some backoff so a persistently failing watch does not turn into a hot retry loop.

Second, when the first `client.Watch` fails, the `cannot watch NRT objects` branch returns while the WaitGroup counter is still 1, so `Wait` never comes back. A fake client that returns an error from `Watch` leaves `Wait` blocked until the test gives up. `defer wt.wg.Done()` right after the `Add` would cover that and any early return added later.

Third, `wg.Add(1)` and `inited = true` both run inside the goroutine while `Wait` reads `inited` from the caller. Under `-race -count=50` the detector flagged the write at watch.go:99 against the read at watch.go:90, and one run in fifty had `Wait` return while the watcher was still blocked inside `client.Watch`. In practice `closeCache` cancels the context first so the window is small, but it is real. Doing the `Add` in the caller before the `go` would close it, and `inited` could then go away: `Wait` on a zero WaitGroup returns immediately, which is what the non-`CacheResyncScopeAll` branch is relying on the flag for today.

I can send these as a small PR against your branch if that is easier than folding them in, along with the fakes I used.

ffromani:
> I was reading the watcher to understand how the resync loop gets fed, and ended up putting its lifecycle through a few tests. Three things came out of it, all in `watch.go`.
> 
> The one I would worry about most is the closed result channel. `case ev := <-wa.ResultChan()` drops the receive `ok`, so once the apiserver closes the watch that case is permanently ready, `processEvent` discards the zero event, and the loop keeps going round. With a fake watch whose channel is already closed and the context still live, the goroutine burned 506ms of CPU over 500ms of wall clock, so a whole core, and it never tries to reconnect. The outer `for !done` reads like it is there to re-establish the watch, but the inner loop only leaves through `ctx.Done()`, which sets `done` on the way out, so nothing ever reaches it. Breaking out on `!ok` would give that outer loop something to do, with some backoff so a persistently failing watch does not turn into a hot retry loop.
> 
> Second, when the first `client.Watch` fails, the `cannot watch NRT objects` branch returns while the WaitGroup counter is still 1, so `Wait` never comes back. A fake client that returns an error from `Watch` leaves `Wait` blocked until the test gives up. `defer wt.wg.Done()` right after the `Add` would cover that and any early return added later.
> 
> Third, `wg.Add(1)` and `inited = true` both run inside the goroutine while `Wait` reads `inited` from the caller. Under `-race -count=50` the detector flagged the write at watch.go:99 against the read at watch.go:90, and one run in fifty had `Wait` return while the watcher was still blocked inside `client.Watch`. In practice `closeCache` cancels the context first so the window is small, but it is real. Doing the `Add` in the caller before the `go` would close it, and `inited` could then go away: `Wait` on a zero WaitGroup returns immediately, which is what the non-`CacheResyncScopeAll` branch is relying on the flag for today.
> 
> I can send these as a small PR against your branch if that is easier than folding them in, along with the fakes I used.

thanks for your time and your findings, these help. I think I had enough context to fix the issues you kindly identified and I integrated the changes in the originating commit (https://github.com/kubernetes-sigs/scheduler-plugins/pull/973/changes/7e4ccb0078e8d4064f446aabf1d886319e0f27c5), crediting you for the findings. If you want the attribution to be done differently in the commit message, please let me know how you wish it to be worded. Thanks!

thc1006:
No worries , I just happened to have some free time during summer vacation here in Taiwan, and I’m quite interested in this kind of development, so I took a look and analyzed it a bit for fun.

Glad it helped, and thank you too for your work and contribution to the open source community!

really very thank you!!!

thc1006:
The credit as it stands is fine, thanks for asking. No change needed on my side.

Swapping the WaitGroup for the `done` channel is nicer than what I suggested: it removes the Add/Wait ordering question entirely rather than tightening it, and `close(wt.done)` sitting past the outer loop means every exit reaches it.

Two small things I noticed while reading it, neither of which needs to hold this up.

The `!ok` path reconnects immediately. If the apiserver closes the watch as fast as it is established, that turns into a tight reconnect loop rather than the tight receive loop it replaced. A short backoff before the retry would bound it.

The other is that a failed `client.Watch` is now fatal for the watcher: it breaks out, closes `done`, and nothing restarts it. That is a real improvement over hanging in `Wait` forever, but it does mean a transient apiserver blip during startup leaves that scheduler instance without NRT watching until it is restarted, and quietly. Retrying that one alongside the `!ok` case, rather than treating it differently, would cover both.

I saw the integration test go red on the new commit. I have not looked at whether it is anywhere near this code, so I am not claiming a connection, but happy to dig if it turns out to be in this area.

ffromani:
> The other is that a failed `client.Watch` is now fatal for the watcher: it breaks out, closes `done`, and nothing restarts it. That is a real improvement over hanging in `Wait` forever, but it does mean a transient apiserver blip during startup leaves that scheduler instance without NRT watching until it is restarted, and quietly. Retrying that one alongside the `!ok` case, rather than treating it differently, would cover both.

Mh, this is a fair point and honestly it's more relevant than ever especially in the messy CI environments. I'm having a look again. I think I can learn a bit more about the rich loop/retry primitives kubernetes offers.

PR #974, opened 2026-05-14
Add KEP for node metadata scoring plugin

alikhil:
#### What type of PR is this?

/kind documentation

This PR adds KEP-942 for the NodeMetadata scoring plugin proposed in #942.

It introduces:

- `kep/942-node-metadata-scoring/README.md` with the proposal, design details, test plan, and PRR sections
- `kep/942-node-metadata-scoring/kep.yaml` with the KEP metadata

The KEP documents how the plugin scores nodes from labels or annotations using
numeric values or timestamps, along with the supported configuration, scoring
strategies, validation behavior, and expected test coverage.

None

This KEP tracks the existing NodeMetadata plugin work from #942, so the KEP
number is intentionally `942` while this documentation PR is #974.

```release-note
NONE
```

### DIFF
--- a/kep/942-node-metadata-scoring/README.md
+++ b/kep/942-node-metadata-scoring/README.md
@@ -1,0 +1,408 @@
+# KEP-942: Node Metadata Scoring
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Use cases](#use-cases)
+  - [Cost-aware placement](#cost-aware-placement)
+  - [Freshness-aware placement](#freshness-aware-placement)
+  - [Hardware lifecycle placement](#hardware-lifecycle-placement)
+- [Proposal](#proposal)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API and configuration](#api-and-configuration)
+  - [Example](#example)
+  - [Plugin registration and extension point](#plugin-registration-and-extension-point)
+  - [Score calculation](#score-calculation)
+    - [Numeric metadata](#numeric-metadata)
+    - [Timestamp metadata](#timestamp-metadata)
+    - [Normalization](#normalization)
+  - [Validation and defaults](#validation-and-defaults)
+- [Known limitations](#known-limitations)
+- [Test plan](#test-plan)
+- [Graduation criteria](#graduation-criteria)
+- [Production Readiness Review Questionnaire](#production-readiness-review-questionnaire)
+  - [Feature enablement and rollback](#feature-enablement-and-rollback)
+  - [Scalability](#scalability)
+  - [Troubleshooting](#troubleshooting)
+- [Implementation History](#implementation-history)
+- [Alternatives](#alternatives)
+- [References](#references)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a new scheduler Score plugin, `NodeMetadata`, that
+ranks feasible nodes by reading a single node label or annotation and
+interpreting its value as either a number or a timestamp. It lets cluster
+operators express preferences such as lower cost, newer hardware, older stable
+nodes, or recent maintenance without adding new APIs to Kubernetes core
+objects.
+
+The plugin is optional and composable. It participates only in scoring, uses
+the standard scheduler framework score range `[0, 100]`, and can be combined
+with other scheduler plugins through normal score weights.
+
+## Motivation
+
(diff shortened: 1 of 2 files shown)


kep/942-node-metadata-scoring/README.md
```diff
@@ -0,0 +1,408 @@
...
+
+If the key is missing or the value cannot be parsed, the plugin returns a raw
+score of `0` for that node.
+
+#### Numeric metadata
+
+Numeric metadata is parsed from the string value and converted into a raw score.
+
+- `Highest`: larger numeric values produce larger raw scores.
+- `Lowest`: smaller numeric values produce larger normalized scores by inverting
```
fanyang-real, line 254: if the key is missing and score will be 0, will that be expected for lowest?

alikhil: Great catch, you’re right!
With the current #942 logic, for `Lowest` a missing key can become better than valid nodes because `Lowest` uses inverted raw values while missing is `0`.

I'll discuss this bug with @josqu4red and will update the KEP once we address it. 
Thank you for review!

alikhil:
@fanyang-real hello!
@josqu4red [adressed](https://github.com/kubernetes-sigs/scheduler-plugins/pull/942/commits/6fe090c56965d7c779bc6a65581c4f4b5fdf64be) your comments
And I've reflected that change and limitations in the KEP [here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/974/commits/5fdafe5ee002b7c2e8932d7be253bf40a882b72c)

Could you please re-review? Thanks!

fanyang-real:
> @fanyang-real hello! @josqu4red [adressed](https://github.com/kubernetes-sigs/scheduler-plugins/pull/942/commits/6fe090c56965d7c779bc6a65581c4f4b5fdf64be) your comments And I've reflected that change and limitations in the KEP [here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/974/commits/5fdafe5ee002b7c2e8932d7be253bf40a882b72c)
> 
> Could you please re-review? Thanks!

Thanks @alikhil Left a minor comment in [the PR](https://github.com/kubernetes-sigs/scheduler-plugins/pull/942#discussion_r3494747698) (not blocking). The KEP change looks great.

kep/942-node-metadata-scoring/README.md
```diff
@@ -0,0 +1,426 @@
...
+If the key is missing or the value cannot be parsed, the plugin returns an
+internal invalid score sentinel for that node. During normalization, invalid
+scores are excluded from the valid score range and are normalized to
+`framework.MinNodeScore`. This makes nodes with valid metadata rank ahead of
+nodes where metadata is absent or malformed.
+
+#### Numeric metadata
+
+Numeric metadata is parsed from the string value as a `float64`, rounded to the
+nearest `int64`, and converted into a raw score.
```
Huang-Wei, line 251: nit: the scheduler framework only requires the final (post-`NormalizeScore()`) score to be int64. So you can keep the score as float64, do the min-max mapping into [0, 100], and round only at the end.

kep/942-node-metadata-scoring/README.md
```diff
@@ -0,0 +1,426 @@
...
+For each candidate node, the plugin:
+
+1. Reads the configured key from either `node.Labels` or `node.Annotations`.
+2. Parses the string value according to the configured metadata type.
+3. Converts the parsed value into a raw score whose ordering matches the chosen
+   strategy.
+4. Returns raw scores for normalization across all feasible nodes.
+
+If the key is missing or the value cannot be parsed, the plugin returns an
+internal invalid score sentinel for that node. During normalization, invalid
```
Huang-Wei, line 242: Do you want to distinguish the scoring of "invalid node" vs. "valid node but in the lower end if min-max normalization" - in current design, I guess they're scored identically/similarly?

Huang-Wei (commented):
LGTM overall. Some nits

kep/942-node-metadata-scoring/README.md
alikhil: Good point, but the issue is that `NormalizeScore()` receives scores after they have already been computed as a `fwk.NodeScoreList`. Since `NodeScore.Score` is an [`int64`](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-scheduler/framework/interface.go#L279-L282), we lose the original `float64` value before normalization, so there is no way to preserve it until that stage.

kep/942-node-metadata-scoring/README.md
alikhil: Oh, my bad. This is already addressed [here](https://github.com/kubernetes-sigs/scheduler-plugins/pull/942/commits/2bccd653a0c608ab7a4c85e55d060324785b47d5). I forgot to reflect this change in the KEP.

Now it's up to date

PR #976, opened 2026-05-29
fix: use configured rest.Config for controller manager

yshngg:
#### What type of PR is this?

/kind bug

The controller manager was calling `ctrl.GetConfigOrDie()` directly when creating the manager, ignoring the QPS and Burst settings that were already configured on the `rest.Config` variable. This fix uses the existing `config` variable which already has `ApiServerQPS` and `ApiServerBurst` applied.

Fixes #

The `config` variable is created at the top of `Run()` via `ctrl.GetConfigOrDie()` and then has QPS/Burst set on it. The old code was calling `ctrl.GetConfigOrDie()` a second time in `ctrl.NewManager()`, effectively discarding those settings.

```release-note
Fix controller manager to respect ApiServerQPS and ApiServerBurst flag settings.
```

### DIFF
--- a/cmd/controller/app/server.go
+++ b/cmd/controller/app/server.go
@@ -47,7 +47,7 @@ func Run(s *ServerRunOptions) error {
 
 	// Controller Runtime Controllers
 	ctrl.SetLogger(klogr.New())
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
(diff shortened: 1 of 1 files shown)


Huang-Wei (commented):
/lgtm
/approve

Thanks for the catch!

Huang-Wei:
#972 may de-flake the integration test timeout here.

/retest

PR #978, opened 2026-06-09
nrt: overreserve: consider only sidecars containers

### LINKED ISSUE #979: [NodeTopologyMatch] scheduler computed PFP includes non-guaranteed pod with non-restartable init-containers
### Area

- [x] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

 NRT is using PodResourcesAPI to determine the PFP sign which is compared with the one computed by scheduler and is expected to both be synced, hence computing both from each side needs to use the same filter on pods.

 Non-guaranteed pods with regular (non-restartable) init containers that request exclusive non-native resources (i.e. devices) are not included in the NRT-PFP sign because they are not reported in by the PodResourcesAPI (see https://github.com/kubernetes/kubernetes/issues/126765). The reason for that is that regular init containers are supposed to run fast. 

 The mismatch in PFPs causes the scheduler to stall on invalid topology data for the node on which this pod is running. This behavior starts from the moment the pod starts running (thus included in the list of running pods as per SharedInformer), and remains as long as the pod is running.

NRT scheduler should use the same standards to compute the PFP as in NRT in order to eventually have them in sync and unblock the scheduling on the affected node. 

1. Create a non-guaranteed pod, its app containers request only native resources (memory,cpu), and it has an init-container requesting non-native resources: 
```
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  initContainers:
    - name: non-restartable-init
      image: alpine:latest
      command: ["/bin/sh", "-c"]
      args: ["sleep 1"]
      resources:
        limits:
          example.com/deviceA: 3         # Correctly indented under limits
  containers:
    - name: app
      image: alpine:latest
      command: ["/bin/sh", "-c"]
      args: ["sleep 36000"]
  restartPolicy: Always
```
2. Once the pod is scheduled it will be pending util the init container finishes. Wait for it to go running. 
3. Create any simple guaranteed pod using the NRT scheduler name:
```
apiVersion: v1
kind: Pod
metadata:
  name: test-gu
spec:
  schedulerName: <NRT scheduler name goes here>
  containers:
    - name: app
      image: alpine:latest
      command: ["/bin/sh", "-c"]
      args: ["sleep 36000"]
      resources:
        limits:
          cpu: "1"
          memory: 4Mi
  restartPolicy: Always
```
4. Notice the pod is kept pending with an event like: 
'Warning  FailedScheduling  38s (x30 over 2m58s)  <NRT scheduler name>  0/5 nodes are available: 1 invalid node topology data,.."

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>
master
</details>

shajmakh:
#### What type of PR is this?
/kind bug

NRT is using PodResourcesAPI to determine the PFP sign which is compared with the one computed by scheduler and is expected to both be synced, hence computing both from each side need to use the same filter on pods.

Non-guaranteed pods with regular (non-restartable) init containers that request exclusive non-native resources (i.e. devices) are not included in the NRT-PFP sign because they are not reported in by the PodResourcesAPI (see https://github.com/kubernetes/kubernetes/issues/126765). The reason for that is that regular init containers are supposed to run fast. The mismatch in PFPs causes the scheduler to stall on `invalid topology data` for the node on which this opd is running.

This commit matches the NRT-computed-PFP filter on pods and drops pods with non-restartable init-containers from being included in the scheduler computed PFP.

Fixes #979 

```release-note
NONE
```

### DIFF
--- a/pkg/noderesourcetopology/resourcerequests/exclusive.go
+++ b/pkg/noderesourcetopology/resourcerequests/exclusive.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	v1helper "k8s.io/kubernetes/pkg/apis/core/v1/helper"
 	v1qos "k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
+	"sigs.k8s.io/scheduler-plugins/pkg/util"
 )
 
 func IncludeNonNative(pod *corev1.Pod) bool {
@@ -40,9 +41,22 @@ func IncludeNonNative(pod *corev1.Pod) bool {
 	}
 	return false
 }
+
+// AreExclusiveForPod checks if the given pod's containers are consuming exclusive resources
+// in the steady state of the pod, i.e. after the irrestartable init containers have finished running.
 func AreExclusiveForPod(pod *corev1.Pod) bool {
 	qos := v1qos.GetPodQOS(pod)
-	return areExclusiveForAnyContainer(qos, append(pod.Spec.InitContainers, pod.Spec.Containers...))
+
+	// filter out init containers with restart policy other than Always because these are *supposed* to
+	// run fast and finish, hence not consuming exclusive resources in a steady state while the pod is Running.
+	restartableInitCnts := []corev1.Container{}
+	for _, ctr := range pod.Spec.InitContainers {
+		if !util.IsSidecarInitContainer(&ctr) {
+			continue
+		}
+		restartableInitCnts = append(restartableInitCnts, ctr)
+	}
+	return areExclusiveForAnyContainer(qos, append(restartableInitCnts, pod.Spec.Containers...))
 }
 
 func areExclusiveForAnyContainer(qos corev1.PodQOSClass, containers []corev1.Container) bool {
--- a/pkg/noderesourcetopology/resourcerequests/exclusive_test.go
+++ b/pkg/noderesourcetopology/resourcerequests/exclusive_test.go
@@ -22,6 +22,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 )
 
 type testCase struct {
@@ -122,6 +123,35 @@ func coreTestCases() []testCase {
 				},
 			},
 			expectedNonNative: false,
+			expectedExclusive: false,
+		},
+		{
+			name: "single-sidecar-initcontainer-gu-no-devs",
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pod",
+					Namespace: "default",
+				},
+				Spec: corev1.PodSpec{
+					InitContainers: []corev1.Container{
+						{
+							Name: "cnt",
+							Resources: corev1.ResourceRequirements{
+								Limits: corev1.ResourceList{
+									corev1.ResourceCPU:    resource.MustParse("4"),
+									corev1.ResourceMemory: resource.MustParse("2Gi"),
+								},
+								Requests: corev1.ResourceList{
+									corev1.ResourceCPU:    resource.MustParse("4"),
+									corev1.ResourceMemory: resource.MustParse("2Gi"),
+								},
+							},
+							RestartPolicy: ptr.To(corev1.ContainerRestartPolicyAlways),
+						},
+					},
+				},
+			},
+			expectedNonNative: false,
 			expectedExclusive: true,
 		},
 		{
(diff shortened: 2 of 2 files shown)


ffromani:
/ok-to-test

please follow more closely the PR template in the cover letter (https://github.com/kubernetes-sigs/scheduler-plugins/pull/978#issue-4622284263)

shajmakh:
@ffromani Thanks for the comment, I updated the PR's description. 

> /ok-to-test
> 
> please follow more closely the PR template in the cover letter ([#978 (comment)](https://github.com/kubernetes-sigs/scheduler-plugins/pull/978#issue-4622284263))

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -40,11 +41,26 @@ func IncludeNonNative(pod *corev1.Pod) bool {
 	}
 	return false
 }
+
 func AreExclusiveForPod(pod *corev1.Pod) bool {
 	qos := v1qos.GetPodQOS(pod)
 	return areExclusiveForAnyContainer(qos, append(pod.Spec.InitContainers, pod.Spec.Containers...))
 }
```
shajmakh: This is no longer used, should be removed

pkg/noderesourcetopology/cache/foreign_pods.go
```diff
@@ -90,7 +90,7 @@ func IsForeignPod(pod *corev1.Pod) bool {
 	if !onlyExclusiveResources {
 		return true
 	}
-	return resourcerequests.AreExclusiveForPod(pod)
+	return resourcerequests.AreExclusiveForSteadyState(pod)
```
ffromani: no need to rename. You caught a bug which was in how we filterd pods. The `steady state` requirement can be documented in a code comment in the implementation of `AreExclusiveForPod`.

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -391,9 +391,9 @@ func makeNodeToPodDataMap(lh logr.Logger, podLister podlisterv1.PodLister, isPod
 		}
 		nodeObjs := nodeToObjsMap[pod.Spec.NodeName]
 		nodeObjs = append(nodeObjs, podData{
-			Namespace:             pod.Namespace,
-			Name:                  pod.Name,
-			HasExclusiveResources: resourcerequests.AreExclusiveForPod(pod),
+			Namespace:                   pod.Namespace,
+			Name:                        pod.Name,
+			HasSteadyExclusiveResources: resourcerequests.AreExclusiveForSteadyState(pod),
```
ffromani: likewise, no need to rename

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -40,9 +41,19 @@ func IncludeNonNative(pod *corev1.Pod) bool {
 	}
 	return false
 }
-func AreExclusiveForPod(pod *corev1.Pod) bool {
+
+// AreExclusiveForSteadyState checks if the given pod's containers are consuming exclusive resources
+// in the steady state of the pod, i.e. after the irrestartable init containers have finished running.
+func AreExclusiveForSteadyState(pod *corev1.Pod) bool {
```
ffromani: no need to rename, document the (correct) requirement as code comment and do the filtering as you do now.

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
```diff
@@ -22,13 +22,15 @@ import (
...
 type testCase struct {
-	name              string
-	pod               *corev1.Pod
-	expectedNonNative bool
-	expectedExclusive bool
+	name                    string
+	pod                     *corev1.Pod
+	expectedNonNative       bool
+	expectedExclusive       bool
+	expectedSteadyExclusive bool
```
ffromani: no need for `expectedSteadyExclusive`, we can just use `expectedExclusive`

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
```diff
@@ -65,8 +67,9 @@ func coreTestCases() []testCase {
 					Namespace: "default",
 				},
 			},
-			expectedNonNative: false,
-			expectedExclusive: false,
+			expectedNonNative:       false,
+			expectedExclusive:       false,
```
ffromani: furthermore as it is now, `expectedExclusive` seems like dead code - another reason to repurpose it vs adding `expectedSteadyExclusive`

ffromani (commented):
Thanks for the PR. I agree we need this fix - and we would need a better kube fix. Please avoid abbreviations like 'PFP sign' - should be 'PFP signature'. Each commit message should also define acronyms (PFP, NRT) first time it uses them. Other than that, please just avoid the rename - I think it's unnecessary, and we can merge.

pkg/noderesourcetopology/cache/foreign_pods.go
shajmakh: I see, addressed. Thanks.

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
shajmakh: totally right, I missed removing that field but with the new updates this is no longer needed

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -40,9 +41,22 @@ func IncludeNonNative(pod *corev1.Pod) bool {
 	}
 	return false
 }
+
+// AreExclusiveForPod checks if the given pod's containers are consuming exclusive resources
+// in the steady state of the pod, i.e. after the irrestartable init containers have finished running.
```
ffromani, line 46: I think we just use `init containers` for standard init containers and `restartable` for sidecars. Not worth a resubmit I believe

ffromani (commented):
/approve
/lgtm

thanks!

pkg/noderesourcetopology/resourcerequests/exclusive.go
shajmakh: I added that "irrestartable" just to be on the safe side because I read in other places (not in this project) these containers are referred to as "regular init containers", so I thought this would make it clearer

shajmakh:
Thanks @ffromani for the quick review!

PR #980, opened 2026-06-12
Fix CRD of SySched symlink name incorrect that causes it to no longer be updated

wzshiming:
#### What type of PR is this?

/kind bug

Fixes #

`seccompprofiles.security-profiles-operator.x-k8s.io_sysched.yaml` and `security-profiles-operator.x-k8s.io_seccompprofiles.yaml` should be the same file,
but one is a regular file and the other is a symlink.
The regular file name should be the correct CRD file name, and the symlink file has the incorrect file name.
This PR deletes the incorrect symlink and recreates the symlink to the correct file name.

```release-note
NONE
```


Huang-Wei (commented):
/lgtm
/approve

PR #983, opened 2026-06-22
Revert "add QueueingHintFn to coscheduling EventsToRegister"

Huang-Wei:
#### What type of PR is this?

/kind failing-test
/kind flake
/kind regression

This reverts the QueueingHintFns added in #972, which regressed the `pull-scheduler-plugins-integration-test-master` CI job: TestCoschedulingPlugin consistently times out (120s) in the "different priority" cases (t5/t6) where a higher-priority workload must win a contended node against a lower-priority PodGroup.

With SchedulerQueueingHints GA in k8s 1.35, an unschedulable pod is only re-queued when a registered event's hint returns Queue. The added hints are too restrictive for coscheduling's cross-PodGroup, priority-driven behavior:

  - Pod/Add was limited to same-PodGroup siblings, so a higher-priority group that becomes schedulable is no longer re-queued when another group changes.
  - A new Pod/Delete registration lets a lower-priority gang re-grab memory freed by its own Permit timeouts, livelocking ahead of higher-priority pods.

Under load a lower-priority gang reserves a Permit slot first, and the precise hints suppress the retries that previously let higher-priority pods recover or preempt, so scheduling stalls until the test times out. Reproduced locally: with the hints, 4/4 parallel runs fail; reverted, 4/4 pass.

Restoring the original EventsToRegister (Pod/Add and PodGroup Add|Update with no hints) brings back the behavior that has been stable for years. Precise hints can be revisited separately with integration coverage that exercises these priority/anti-affinity scenarios under load.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/coscheduling/coscheduling.go
+++ b/pkg/coscheduling/coscheduling.go
@@ -28,7 +28,6 @@ import (
 	corev1helpers "k8s.io/component-helpers/scheduling/corev1"
 	"k8s.io/klog/v2"
 	fwk "k8s.io/kube-scheduler/framework"
-	schedutil "k8s.io/kubernetes/pkg/scheduler/util"
 	"sigs.k8s.io/scheduler-plugins/apis/config"
 	"sigs.k8s.io/scheduler-plugins/apis/config/validation"
 	"sigs.k8s.io/scheduler-plugins/apis/scheduling"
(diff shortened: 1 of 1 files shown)


Huang-Wei:
cc @ffromani

ffromani:
/lgtm

too bad we need to revert, but thanks for the quick fix.

PR #984, opened 2026-07-01
[NodeTopologyMatch]: improve exclusive resource detection

Tal-or:
#### What type of PR is this?

/kind bug

[Extended resources](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/) are only treated as exclusive when they appear in the node's NRT, aligning exclusive-resource detection with topology exporter accounting and pod fingerprint (PFP) matching.

* Gate `IsExclusive` non-native resources on `nrtResources.Has(resource)` instead of treating every extended resource as exclusive.
* Add `ResourceNamesFromNRT` and pass a per-node resource set from PFP resync and foreign-pod detection
* Native CPU/memory/hugepage exclusivity rules are unchanged.

**Test plan**

- [x] go test ./pkg/noderesourcetopology/resourcerequests/...

- [x] go test ./pkg/noderesourcetopology/cache/...

- [x] Deploy locally built scheduler image and verify foreign-pod / cache resync behavior with pods requesting extended resources not listed in NRT

Fixes #
https://github.com/kubernetes-sigs/scheduler-plugins/issues/985

```release-note
NONE
```

### DIFF
--- a/pkg/noderesourcetopology/cache/foreign_pods.go
+++ b/pkg/noderesourcetopology/cache/foreign_pods.go
@@ -39,14 +39,15 @@ var (
 	onlyExclusiveResources = false
 )
 
-func SetupForeignPodsDetector(lh logr.Logger, schedProfileName string, podInformer k8scache.SharedInformer, cc Interface) {
+func SetupForeignPodsDetector(lh logr.Logger, schedProfileName string, podInformer k8scache.SharedInformer, cc *OverReserve) {
 	foreignCache := func(obj interface{}) {
 		pod, ok := obj.(*corev1.Pod)
 		if !ok {
 			lh.V(3).Info("unsupported object", "kind", fmt.Sprintf("%T", obj))
 			return
 		}
-		if !IsForeignPod(pod) {
+		nrtResources := cc.nrtResNames.Get(pod.Spec.NodeName)
+		if !IsForeignPod(pod, nrtResources) {
 			return
 		}
 
@@ -78,7 +79,7 @@ func RegisterSchedulerProfileName(lh logr.Logger, schedProfileName string) {
 	lh.V(5).Info("registered scheduler profiles", "names", schedProfileNames.UnsortedList())
 }
 
-func IsForeignPod(pod *corev1.Pod) bool {
+func IsForeignPod(pod *corev1.Pod, nrtResources sets.Set[corev1.ResourceName]) bool {
 	if pod.Spec.NodeName == "" {
 		// nothing to do yet
 		return false
@@ -90,7 +91,7 @@ func IsForeignPod(pod *corev1.Pod) bool {
 	if !onlyExclusiveResources {
 		return true
 	}
-	return resourcerequests.AreExclusiveForPod(pod)
+	return resourcerequests.AreExclusiveForPod(pod, nrtResources)
 }
 
 // for testing only; NOT thread safe
--- a/pkg/noderesourcetopology/cache/foreign_pods_test.go
+++ b/pkg/noderesourcetopology/cache/foreign_pods_test.go
@@ -22,15 +22,18 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 )
 
 func TestIsForeignPod(t *testing.T) {
 	tests := []struct {
-		name         string
-		profileNames []string
-		pod          *corev1.Pod
-		expected     bool
+		name          string
+		profileNames  []string
+		pod           *corev1.Pod
+		nrtResources  sets.Set[corev1.ResourceName]
+		exclusiveOnly bool
+		expected      bool
 	}{
 		{
 			name: "empty",
@@ -191,6 +194,57 @@ func TestIsForeignPod(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:          "exclusive-only-dev-in-nrt",
+			profileNames:  []string{"secondary-scheduler"},
+			exclusiveOnly: true,
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pod",
+					Namespace: "default",
+				},
+				Spec: corev1.PodSpec{
+					NodeName: "random-node",
+					Containers: []corev1.Container{
+						{
+							Name: "cnt",
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									corev1.ResourceName("veryfast.io/fpga"): resource.MustParse("1"),
+								},
+							},
+						},
+					},
+				},
+			},
+			nrtResources: sets.New(corev1.ResourceName("veryfast.io/fpga")),
+			expected:     true,
+		},
+		{
+			name:          "exclusive-only-dev-not-in-nrt",
+			profileNames:  []string{"secondary-scheduler"},
+			exclusiveOnly: true,
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pod",
+					Namespace: "default",
+				},
+				Spec: corev1.PodSpec{
+					NodeName: "random-node",
+					Containers: []corev1.Container{
+						{
+							Name: "cnt",
+							Resources: corev1.ResourceRequirements{
+								Requests: corev1.ResourceList{
+									corev1.ResourceName("veryfast.io/fpga"): resource.MustParse("1"),
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: false,
+		},
 	}
 
 	for _, tt := range tests {
@@ -199,8 +253,13 @@ func TestIsForeignPod(t *testing.T) {
 				RegisterSchedulerProfileName(klog.Background(), profileName)
 			}
 			defer CleanRegisteredSchedulerProfileNames()
+			defer TrackAllForeignPods()
 
-			got := IsForeignPod(tt.pod)
+			if tt.exclusiveOnly {
+				TrackOnlyForeignPodsWithExclusiveResources()
+			}
+
+			got := IsForeignPod(tt.pod, tt.nrtResources)
 			if got != tt.expected {
 				t.Errorf("%s: pod %q foreign status got %v expected %v", tt.name, tt.pod.Name, got, tt.expected)
 			}
--- a/pkg/noderesourcetopology/cache/overreserve.go
+++ b/pkg/noderesourcetopology/cache/overreserve.go
@@ -47,6 +47,7 @@ type OverReserve struct {
 	lock             sync.Mutex
 	generation       uint64
 	nrts             *nrtStore
+	nrtResNames      *nrtResourcesStore
 	assumedResources map[string]*resourceStore // nodeName -> resourceStore
 	// nodesMaybeOverreserved counts how many times a node is filtered out. This is used as trigger condition to try
 	// to resync nodes. See The documentation of Resync() below for more details.
@@ -77,6 +78,7 @@ func NewOverReserve(ctx context.Context, lh logr.Logger, cfg *apiconfig.NodeReso
 		lh:                     lh,
 		client:                 client,
 		nrts:                   newNrtStore(lh, nrtObjs.Items),
+		nrtResNames:            newNrtResourcesStore(nrtObjs.Items),
 		assumedResources:       make(map[string]*resourceStore),
 		nodesMaybeOverreserved: newCounter(),
 		nodesWithForeignPods:   newCounter(),
@@ -277,7 +279,7 @@ func (ov *OverReserve) MakeNRTUpdatesForNodes(ctx context.Context, lh_ logr.Logg
 	var nrtUpdates []*topologyv1alpha2.NodeResourceTopology
 
 	// node -> pod identifier (namespace, name)
-	nodeToObjsMap, err := makeNodeToPodDataMap(lh_, ov.podLister, ov.isPodRelevant)
+	nodeToObjsMap, err := makeNodeToPodDataMap(lh_, ov.podLister, ov.isPodRelevant, ov.nrtResNames.Get)
 	if err != nil {
 		lh_.Error(err, "cannot find the mapping between running pods and nodes")
 		return nrtUpdates
@@ -355,6 +357,7 @@ func (ov *OverReserve) FlushNodes(lh logr.Logger, nrts ...*topologyv1alpha2.Node
 	for _, nrt := range nrts {
 		lh.V(2).Info("flushing", logging.KeyNode, nrt.Name)
 		ov.nrts.Update(nrt)
+		ov.nrtResNames.Update(nrt)
 		delete(ov.assumedResources, nrt.Name)
 		ov.nodesMaybeOverreserved.Delete(nrt.Name)
 		ov.nodesWithForeignPods.Delete(nrt.Name)
@@ -379,7 +382,7 @@ func (ov *OverReserve) TestOnlyUpdateNRT(nrt *topologyv1alpha2.NodeResourceTopol
 	ov.nrts.Update(nrt)
 }
 
-func makeNodeToPodDataMap(lh logr.Logger, podLister podlisterv1.PodLister, isPodRelevant podprovider.PodFilterFunc) (map[string][]podData, error) {
+func makeNodeToPodDataMap(lh logr.Logger, podLister podlisterv1.PodLister, isPodRelevant podprovider.PodFilterFunc, nrtResourcesLookup NRTResourcesLookupFunc) (map[string][]podData, error) {
 	nodeToObjsMap := make(map[string][]podData)
 	pods, err := podLister.List(labels.Everything())
 	if err != nil {
@@ -389,11 +392,12 @@ func makeNodeToPodDataMap(lh logr.Logger, podLister podlisterv1.PodLister, isPod
 		if !isPodRelevant(lh, pod) {
 			continue
 		}
+		nrtResources := nrtResourcesLookup(pod.Spec.NodeName)
 		nodeObjs := nodeToObjsMap[pod.Spec.NodeName]
 		nodeObjs = append(nodeObjs, podData{
 			Namespace:             pod.Namespace,
 			Name:                  pod.Name,
-			HasExclusiveResources: resourcerequests.AreExclusiveForPod(pod),
+			HasExclusiveResources: resourcerequests.AreExclusiveForPod(pod, nrtResources),
 		})
 		nodeToObjsMap[pod.Spec.NodeName] = nodeObjs
 	}
--- a/pkg/noderesourcetopology/cache/store.go
+++ b/pkg/noderesourcetopology/cache/store.go
@@ -21,6 +21,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/go-logr/logr"
 	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
@@ -67,6 +68,20 @@ func (nrs *nrtStore) GetNRTCopyByNodeName(nodeName string) *topologyv1alpha2.Nod
 		return nil
 	}
 	return obj.DeepCopy()
+}
+
+// ResourceNamesFromNRT returns the set of resource names listed in the given NRT zones.
+func ResourceNamesFromNRT(nrt *topologyv1alpha2.NodeResourceTopology) sets.Set[corev1.ResourceName] {
+	if nrt == nil {
+		return nil
+	}
+	res := sets.New[corev1.ResourceName]()
+	for _, zone := range nrt.Zones {
+		for _, resInfo := range zone.Resources {
+			res.Insert(corev1.ResourceName(resInfo.Name))
+		}
+	}
+	return res
 }
 
 // Update adds or replace the Node Resource Topology associated to a node. Always do a copy.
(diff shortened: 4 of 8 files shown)


pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -44,7 +45,7 @@ func IncludeNonNative(pod *corev1.Pod) bool {
 
 // AreExclusiveForPod checks if the given pod's containers are consuming exclusive resources
 // in the steady state of the pod, i.e. after the irrestartable init containers have finished running.
-func AreExclusiveForPod(pod *corev1.Pod) bool {
+func AreExclusiveForPod(pod *corev1.Pod, nrtResources sets.Set[corev1.ResourceName]) bool {
```
shajmakh, line 48: other places that call this is not being updated in the same commit, which would lead to compiling errors with this commit only

shajmakh: I believe this should be squashed to later commit 004070fd3efd83f516b3ae23b797b9d413ca71c6

pkg/noderesourcetopology/cache/nrt_resources.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+)
+
+// ResourceNamesFromNRT returns the set of resource names listed in the given NRT zones.
+func ResourceNamesFromNRT(nrt *topologyv1alpha2.NodeResourceTopology) sets.Set[corev1.ResourceName] {
```
shajmakh: missing unit testing

pkg/noderesourcetopology/cache/foreign_pods_test.go
```diff
@@ -199,8 +253,13 @@ func TestIsForeignPod(t *testing.T) {
...
 			}
 			defer CleanRegisteredSchedulerProfileNames()
+			defer TrackAllForeignPods()
+
+			if tt.exclusiveOnly {
+				TrackOnlyForeignPodsWithExclusiveResources()
+			}
 
-			got := IsForeignPod(tt.pod)
+			got := IsForeignPod(tt.pod, tt.nrtResources)
```
shajmakh, line 262: similarly here WRT updating the function calls in the same commit that updates teh function signature

shajmakh (commented):
Thanks for addressing, the direction of handling this LGTM given that we trust NRT to report the actual exclusive resources only.

pkg/noderesourcetopology/cache/nrt_resources.go
Tal-or: Added. thanks

pkg/noderesourcetopology/cache/nrt_resources_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+package cache
+
+import (
+	"testing"
+
+	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
```
shajmakh: nit: k8s deps first then the 3rd party

pkg/noderesourcetopology/cache/nrt_resources.go
ffromani: our junkyard is `store.go` and lacking better place I'd add this code there, deferring a proper cleanup which is a bigger concern anyway

pkg/noderesourcetopology/cache/foreign_pods.go
```diff
@@ -46,7 +47,9 @@ func SetupForeignPodsDetector(lh logr.Logger, schedProfileName string, podInform
 			lh.V(3).Info("unsupported object", "kind", fmt.Sprintf("%T", obj))
 			return
 		}
-		if !IsForeignPod(pod) {
+		nrt, _ := cc.GetCachedNRTCopy(context.Background(), pod.Spec.NodeName, pod)
```
ffromani: calling a function which requires a context and hardcoding `context.Background()` makes me pause.
Can we compute `nrtResources` once at setup time and pass the set to the callback?

ffromani (commented):
directionwise +1. Implementation questions/suggestions inline.

pkg/noderesourcetopology/cache/foreign_pods.go
Tal-or: `GetCachedNRTCopy` implemented by the `nrtcache.OverReserve` which accepts `ctx` but does nothing with it.
I could pass `context.TODO()` which would work the same but here both have the same meaning.

Tal-or: > Can we compute nrtResources once at setup time and pass the set to the callback?

We need to work with updated data (which we're getting from the cache) for every call. 
think of a case when new  resource which is bound to the HW is being added to the node.
In the case of using the same set without updating it, we'll lose track of this new resource.

pkg/noderesourcetopology/cache/nrt_resources.go
Tal-or: done.

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
```diff
@@ -291,6 +297,53 @@ func coreTestCases() []testCase {
...
+									corev1.ResourceName("slow.io/accelerator"): resource.MustParse("1"),
+								},
+							},
+						},
+					},
+				},
+			},
+			expectedNonNative: true,
+			expectedExclusive: false,
+		},
```
swatisehgal: From scheduler perspective, both these cases are funtionally identical, devices are represented as extended resources (both of which in this case are not in NRT). The different resource names don't exercise different behavior so I don't see the need for us to have two test cases unless I am missin something here?

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -56,25 +57,26 @@ func AreExclusiveForPod(pod *corev1.Pod) bool {
...
 	return false
 }
 
-func IsExclusive(qos corev1.PodQOSClass, resource corev1.ResourceName, quantity resource.Quantity) bool {
-	// devices accessed via device plugins are non-shareable
-	// note until we reach better clarity we treat extended resources as devices
+func IsExclusive(qos corev1.PodQOSClass, resource corev1.ResourceName, quantity resource.Quantity, nrtResources sets.Set[corev1.ResourceName]) bool {
+	// devices accessed via device plugins are non-shareable.
+	// extended resources are not necessarily bound to the node topology which means they may not be exclusive.
+	// the fact that a resource is being tracked in the NRT object provides certantiy about its topological criteria.
```
swatisehgal: NIT: typo  s/certantiy/certainty

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -56,25 +57,26 @@ func AreExclusiveForPod(pod *corev1.Pod) bool {
...
 	}
 	return false
 }
 
-func IsExclusive(qos corev1.PodQOSClass, resource corev1.ResourceName, quantity resource.Quantity) bool {
-	// devices accessed via device plugins are non-shareable
-	// note until we reach better clarity we treat extended resources as devices
+func IsExclusive(qos corev1.PodQOSClass, resource corev1.ResourceName, quantity resource.Quantity, nrtResources sets.Set[corev1.ResourceName]) bool {
+	// devices accessed via device plugins are non-shareable.
+	// extended resources are not necessarily bound to the node topology which means they may not be exclusive.
```
swatisehgal: Extended resource not being tied to topology shouldn't have anything to do with exclusivity. Can you please clarify?

pkg/noderesourcetopology/resourcerequests/exclusive.go
```diff
@@ -56,25 +57,26 @@ func AreExclusiveForPod(pod *corev1.Pod) bool {
...
+	// the fact that a resource is being tracked in the NRT object provides certantiy about its topological criteria.
 	if !v1helper.IsNativeResource(resource) {
-		return true
+		return nrtResources.Has(resource)
```
swatisehgal, line 80: In case where NRT is nil due to unavailability (e.g. initial setup/ node reboot) this returns false which implies extended resource is treated as non-exclusive. This means that multiple pods could request the same resource if NRT is unavailable. It doesn't seem right for extended resource (irrespective of whether it backed by device plugin or not) to be considered shared?

swatisehgal (commented):
The approach of gating extended resource exclusivity on NRT presence is reasonable. However, there's an important edge case: when a pod requests an extended resource but the node's NRT is missing/stale, this PR treats it as "non-exclusive". This behavior needs careful evaluation.

pkg/noderesourcetopology/cache/foreign_pods.go
ffromani: I don't think we support this case all across the stack and I'm pretty sure we already have the assumption creeped in that the set of resources is stable.

ffromani: Another, and more serious, thing that worries me is the fact `GetCachedNRTCopy` will (and must) fail once we have foreign pods so there's a built-in risk of misuse if we combine them

ffromani:
we can probably plug in the `attr_watch` flow on which we observe updates anyway and maintain asynchronously a set of `nrtResources`. This will allow us to still have as-soon-as-possible update but will break the GetCachedNRTCopy/foreign pods detection cycle with a minimal cost.

Tal-or:
> The approach of gating extended resource exclusivity on NRT presence is reasonable. However, there's an important edge case: when a pod requests an extended resource but the node's NRT is missing/stale, this PR treats it as "non-exclusive". This behavior needs careful evaluation.

extended-resources will never be seen under the NRT object.

[extended-resources](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/) are representation of node-level resources in a form of devices under the node status.

Those resources are NOT backed-up by any device plugin under the hood, so they are not providing any topology-hints which in turn means they won't be taken into account by the PodResourceAPI.
So even if NRT is not stale (i.e. fresh), we won't see the extended resources.

However there's the opposite direction - a new resources which IS backed by a device plugin (non-shareable) is not seen under the NRT (stale NRT) so `nrtResources.Has(resource) ` return `false` and the container is not consider to be exclusive. 
But even before that, if the resource is no listed in the NRT, the scheduler will reject the pod anyway, since from its perspective the node is not the right fit, because the container requests for a device the node doesn't have.

pkg/noderesourcetopology/resourcerequests/exclusive.go
Tal-or: extended-resources are node level resources which are patched directly under the node status.
They are not being advertise by device-plugin which has a way to expose topology parameters:
https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#device-plugin-integration-with-the-topology-manager

pkg/noderesourcetopology/resourcerequests/exclusive.go
Tal-or: In case NRT is nil, we're not getting that deep into the flow:
https://github.com/kubernetes-sigs/scheduler-plugins/blob/025090fe41228899930d1ffe8ec04ca3e2021375/pkg/noderesourcetopology/filter.go#L198

pkg/noderesourcetopology/resourcerequests/exclusive_test.go
Tal-or: good catch. I'll remove

pkg/noderesourcetopology/cache/foreign_pods.go
Tal-or: This assumption is simplifying the flow, although it exposes a risk for the edge case I just mentioned.
We can start with single initialization and reevaluate again if needed.

Tal-or: I changed the logic to accept an existing NRT resources set which is being initialized once when we're setting up the cache.

pkg/noderesourcetopology/cache/foreign_pods.go
```diff
@@ -90,7 +91,15 @@ func IsForeignPod(pod *corev1.Pod) bool {
 	if !onlyExclusiveResources {
 		return true
 	}
-	return resourcerequests.AreExclusiveForPod(pod)
+	return resourcerequests.AreExclusiveForPod(pod, nrtResources)
+}
+
+func resourceNamesForNode(cc Interface, nodeName string) sets.Set[corev1.ResourceName] {
+	ov, ok := cc.(*OverReserve)
```
ffromani: type casts always make me pause. Plus, we add the resourceNames in the existing store. So I still think we can and should leverage the `attr_watch` flow. What about this: introduce a dedicated nrtResourcesStore — a lightweight, thread-safe (with its own sync.RWMutex) per-node cache of resource names extracted from NRT zones.

It can be populated at startup from the initial NRT list and kept up to date through the existing Watcher/attr_watch flow and FlushNodes, the same paths we use to feed the NRT cache.
The foreign-pods detector can get a callback (NRTResourcesLookupFunc?) rather than reaching into the cache through a type assertion.

This decouples resource-name lookups from the main OverReserve lock, avoiding contention with the scheduling hot path (GetCachedNRTCopy) and removing any circular dependency between
foreign-pods detection and NRT cache lookups.

Tal-or: > What about this: introduce a dedicated nrtResourcesStore — a lightweight, thread-safe (with its own sync.RWMutex) per-node cache of resource names extracted from NRT zones.

Fair enough, make sense.

> It can be populated at startup from the initial NRT list and kept up to date through the existing Watcher/attr_watch flow and FlushNodes, the same paths we use to feed the NRT cache

This is the only part I'm struggling to understand:
Are you suggesting to update the resourceNames every time the Watcher detects a `MODIFICATION` change, or only at the time that `FlushNodes` is being called?

> The foreign-pods detector can get a callback (NRTResourcesLookupFunc?) rather than reaching into the cache through a type assertion.

Fair enough, make sense

ffromani: managed to express myself poorly. Updating only inside processEvent should be fine. Note we _already_ call `GetCachedNRTCopy` in there, but that's ok, we're not adding new calls, and clean up further it's a different endeavour

ffromani: I was thinking more about your question and I think we can simplify even more:
1. keep the separate store with its RWLock - still needed, the foreign pods logic is fully async
2. update the logic to update the `nrtResourcesStore` only in Resync, aligned with what we do for `nrtStore` - same lifecycle
3. SetupForeignPodsDetector can now take a OverReserve concrete type - it is getting a OverReserve always anyway, and its logic is closely tied to the OverReserve type anyway.
4. everything else stays the same

the key realization is we can't depend on the `attr_watch` flow anyway, because the watcher is started only if resync scope == all, so we need the update in FlushNodes anyway. Too bad, but refactoring further is out of the picture

There's also an argument to avoid the RWLock and depend on `ov.lock`. But then we would need to move `makeNodeToPodDataMap` in the critical region, which goes in a wrong direction. Adding a separate RWLock goes in the direction of more granular locking, which is something good in general

pkg/noderesourcetopology/resourcerequests/exclusive.go
swatisehgal: I understand that extended resources can be node-level without topology info, but I think there's a conceptual separation we need to address. Topology-awareness is not equal to exclusivity. These are orthogonal concepts:

Topology-aware: Resource has NUMA affinity, in case of devices is done throught TopologyInfo (or some equivaent field) in case of device plugins in case of CPU and memory determined by querying sysfs
Exclusive implies resource cannot be shared between pods (e.g., GPUs, specific devices)
A resource can be:

- Exclusive + topology-aware (GPU with device plugin reporting NUMA affinity in NRT)
- Exclusive + NOT topology-aware (GPU with device plugin that doesn't report topology, NOT in NRT, but still exclusive)
- Non-exclusive + NOT topology-aware (shared pool, NOT in NRT, shareable)

Bottomline, your comment needs to be updated.

Tal-or: Just did, thanks

pkg/noderesourcetopology/cache/foreign_pods.go
Tal-or: Align the PR according to the above.
I also think it's more accurate to change `makeNodeToPodDataMap` to be a receiver function (method) of  `OverReserve` since it's solely used by it and most of the arguments it gets are members of latter.

We can suggest this change in future PR since this one is quite big already.

pkg/noderesourcetopology/resourcerequests/exclusive.go
swatisehgal: That's correct! But `nodeTopology == nil `means the node is schedulable (Filter passes through). We explicitly return `fwk.NewStatus(fwk.Unschedulable, ...) `only when we want to filter out a node.

My concern is about the foreign pod detection path when NRT is missing. When nrt = nil, `nrtResources.Has(resource)` returns false, which causes extended resources to be treated as non-exclusive.

Consider this scenario, we are installing topology-aware scheduler
- Cluster already has workloads (scheduled by default scheduler)
- RTE pods are rolling out node-by-node
- Some nodes: NRT exists
- Some nodes: NRT not published yet

Node worker-5:
- Has existing Pod-A with vendor.io/gpuX (scheduled by default scheduler)
- RTE not started yet (nrt = nil)
- Topology-aware scheduler sees Pod-A

Without this PR: Pod-A  is treated as foreign, node is blocked.
With this PR (nrt = nil): Pod-A NOT treated as foreign (IsExclusive() returns false => AreExclusiveForPod() returns false => IsForeignPod() returns false => SetupForeignPodsDetector() hits "if !IsForeignPod()" and returns early and NodeHasForeignPods() is never called => Node stays Fresh = true, schedulable)

Topology-aware scheduler schedules Pod-B also requesting GPU which is deviation from the current behavior and can potentially lead to kubelet admission rejections in case of GPU resource contention.

 A simple fix could be:
```go
func IsExclusive(..., nrtResources sets.Set[corev1.ResourceName]) bool {
    if !v1helper.IsNativeResource(resource) {
        if nrtResources == nil {
            return true  //  assume exclusive when NRT unavailable
        }
        return nrtResources.Has(resource)
    }
    // ...
}

Comment (swatisehgal) [reply pkg/noderesourcetopology/resourcerequests/exclusive.go:80]: @ffromani Would be good to have your input on this as well.

Comment (Tal-or) [reply pkg/noderesourcetopology/resourcerequests/exclusive.go:80]: hmm so `nil` nrtResources might change the behavior for a very specific case where foreign pods are no longer foreign. 

You have a point here but there's already a protection in  the code for that:
> Without this PR: Pod-A is treated as foreign, node is blocked.

for that to happen the node should be marked as dirty in `NodeHasForeignPods`:
```go
func (ov *OverReserve) NodeHasForeignPods(nodeName string, pod *corev1.Pod) {
	lh := ov.lh.WithValues(logging.KeyPod, klog.KObj(pod), logging.KeyPodUID, logging.PodUID(pod), logging.KeyNode, nodeName)
	ov.lock.Lock()
	defer ov.lock.Unlock()
	if !ov.nrts.Contains(nodeName) { // <---- [1]  return early NTR for the node was not found
		lh.V(5).Info("ignoring foreign pods", "nrtinfo", "missing")
		return
	}
	val := ov.nodesWithForeignPods.Incr(nodeName) //  <------- [2] here we're marking the node as dirty
	lh.V(2).Info("marked with foreign pods", logging.KeyNode, nodeName, "count", val)
}
```

see my comments inside the code snippet - in [2] we're marking the node as dirty so it will be blocked.
but even before that in [1] we're returning early since nrt was not found for the node, which is exactly the case we're discussing here.

swatisehgal: The PR maintains the existing behavior (nodes without NRT remain schedulable, foreign pods ignored) while correctly gating exclusivity on NRT presence once NRT exists. The concern I raised doesn't apply since the protection was already there.

Thanks for the clarification!

Tal-or: Appreciate the thorough review thanks!

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -277,7 +279,7 @@ func (ov *OverReserve) MakeNRTUpdatesForNodes(ctx context.Context, lh_ logr.Logg
 	var nrtUpdates []*topologyv1alpha2.NodeResourceTopology
 
 	// node -> pod identifier (namespace, name)
-	nodeToObjsMap, err := makeNodeToPodDataMap(lh_, ov.podLister, ov.isPodRelevant)
+	nodeToObjsMap, err := makeNodeToPodDataMap(lh_, ov.podLister, ov.isPodRelevant, ov.nrtResNames.Get)
```
ffromani, line 282: honestly I'd just pass a `nrtResourceStore` reference and save us a `type`, but it's minor and not worthy a resubmit

pkg/noderesourcetopology/cache/store.go
```diff
@@ -69,6 +71,58 @@ func (nrs *nrtStore) GetNRTCopyByNodeName(nodeName string) *topologyv1alpha2.Nod
...
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	rs.data[nrt.Name] = ResourceNamesFromNRT(nrt)
+}
+
+func (rs *nrtResourcesStore) Delete(nodeName string) {
+	rs.mu.Lock()
+	defer rs.mu.Unlock()
+	delete(rs.data, nodeName)
+}
```
ffromani, line 124: not sure this is ever used

ffromani (commented):
/approve
/lgtm

thanks!

pkg/noderesourcetopology/cache/overreserve.go
Tal-or: yea, see https://github.com/kubernetes-sigs/scheduler-plugins/pull/984#discussion_r3542270563

PR #992, opened 2026-07-20
KEP: add standalone HighLoadFilter scheduler plugin

Feelings0220:
/kind feature

## What

This PR proposes `HighLoadFilter`, an optional, standalone scheduler framework plugin that rejects nodes whose observed CPU or memory load is already too high for an incoming Pod.

## Why

Request-based filtering cannot account for the gap between requested resources and real node utilization. Existing load-aware scoring can prefer a less-loaded node, but a score alone does not establish a hard admission boundary for overloaded nodes.

This proposal is deliberately separate from Trimaran and KEP-61. The plugin has its own configuration API, lifecycle, cache, failure policy, tests, and documentation. It consumes `load-watcher` directly and does not import or configure Trimaran plugins.

## Proposed design

- `PreFilter` captures one immutable metrics snapshot and the incoming Pod's effective CPU and memory requests.
- `Filter` predicts post-placement utilization as observed utilization plus `pod request / node allocatable`.
- CPU and memory thresholds are independently configurable.
- Metrics may come from an independent load-watcher service, or from an embedded collector backed by Kubernetes Metrics Server or Prometheus.
- Missing, stale, or partial metrics follow an explicit fail-open or fail-closed policy.
- The plugin remains optional and is enabled only when configured in the scheduler profile.

## Scope of this PR

This PR contains the enhancement proposal only. The implementation is available as [Draft PR #993](https://github.com/kubernetes-sigs/scheduler-plugins/pull/993) so API and behavioral feedback can be incorporated before merge.

## Validation

- The KEP table of contents was generated and verified with the repository-pinned `mdtoc` version.
- The proposal includes API defaults and validation, failure semantics, security considerations, scalability risks, test plans, graduation criteria, and production-readiness review material.

```release-note
NONE
```

### DIFF
--- a/kep/tbd-high-load-filter/README.md
+++ b/kep/tbd-high-load-filter/README.md
@@ -1,0 +1,542 @@
+# KEP-TBD: High Load Filter
+
+## Table of Contents
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User stories](#user-stories)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [Plugin lifecycle and extension points](#plugin-lifecycle-and-extension-points)
+  - [Metrics pipeline](#metrics-pipeline)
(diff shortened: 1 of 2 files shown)


Feelings0220:
Hi @ffromani , thanks for /ok-to-test. When convenient, could you please take a look at this KEP? The corresponding implementation is in #993. Thanks!

PR #993, opened 2026-07-20
add standalone HighLoadFilter scheduler plugin

Feelings0220:
/kind feature
/kind api-change

## Summary

This PR adds `HighLoadFilter`, an optional scheduler framework plugin that filters nodes using observed CPU and memory utilization before placement.

The plugin is intentionally standalone. It does not import `pkg/trimaran`, use `TrimaranSpec`, or depend on the LoadVariationRiskBalancing plugin. It consumes `github.com/paypal/load-watcher` directly and owns its configuration, metrics collector, cache, failure policy, lifecycle, tests, and documentation.

## Motivation

Resource-request predicates protect allocatable capacity, but cannot detect nodes whose real utilization is already high. Load-aware scoring may rank such a node lower, but does not create a hard boundary. `HighLoadFilter` adds an explicit, configurable filter for that case.

## Design

- `PreFilter` records one immutable load snapshot and calculates the Pod's effective CPU and memory requests once per scheduling cycle.
- `Filter` predicts post-placement utilization:

  `observed utilization + pod request / node allocatable * 100`

- A node is rejected only when predicted CPU or memory utilization is greater than its configured threshold; equality is allowed.
- Metrics can be obtained from either:
  - an external, independently deployed load-watcher service; or
  - an embedded collector backed by Kubernetes Metrics Server or Prometheus.
- Stale, missing, or partial metrics use an explicit fail-open/fail-closed policy.
- The collector retains the last successful snapshot only until its configured expiration and shuts down with the scheduler context.
- The plugin participates in scheduler batching through `SignPlugin` using CPU and memory requests.

## Configuration and documentation

This change adds the versioned and internal configuration API, generated conversions/defaults/deep-copy code, validation, scheduler registration, plugin documentation, root README registration, and an integration test.

The default thresholds are 100% CPU and memory, so enabling the plugin without lowering a threshold does not introduce an implicit utilization ceiling. The default policy is fail-open.

## Enhancement proposal

This implementation follows the standalone design proposed in [KEP PR #992](https://github.com/kubernetes-sigs/scheduler-plugins/pull/992). This PR should remain Draft until the KEP direction and API are accepted.

## Testing

- `go test -timeout=40m ./cmd/... ./pkg/... ./apis/...`
- `go test -race ./pkg/highloadfilter`
- `go test ./test/integration -run TestHighLoadFilterPlugin -count=1 -timeout=10m`
- `go vet ./pkg/highloadfilter ./apis/config/validation ./cmd/scheduler`
- `go build ./cmd/scheduler`
- Code generation repeated with a clean working tree afterward
- `verify-gomod.sh`, `verify-gofmt.sh`, and `verify-crdgen.sh`
- `git diff --check`

The repository's structured-logging verification tool could not be completed locally because its existing pinned `logcheck` dependency resolves an older `golang.org/x/tools` release that does not compile with Go 1.25. This change introduces no new occurrence reported before that toolchain failure; upstream CI remains the canonical check.

The original prototype has also been exercised in a real 115-node cluster. That is functional deployment evidence, not a scalability benchmark; the KEP calls for formal scale and failure-mode testing before graduation.

```release-note
Adds an optional standalone HighLoadFilter scheduler plugin that can reject nodes based on observed and predicted CPU or memory utilization.
```


Feelings0220:
Hi @zwpaper , this implementation PR has been ready for review since July 25, and the related KEP #992 has passed its test checks. Could you please /ok-to-test when convenient so that Prow can run? Thanks!

PR #994, opened 2026-07-21
nrt:filter: return early for non-gu pod

shajmakh:
#### What type of PR is this?
/kind cleanup

Native resources of the burstable pods are consumed from a shared pool hence not reflected in NRT. Return early in Filter phase on non-guaranteed pods that do not request exclusive resources.

Fixes #

```release-note

```

### DIFF
--- a/pkg/noderesourcetopology/filter.go
+++ b/pkg/noderesourcetopology/filter.go
@@ -178,7 +178,7 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState fwk.CycleState, 
 		return fwk.NewStatus(fwk.Error, "node not found")
 	}
 	qos := v1qos.GetPodQOS(pod)
-	if qos == v1.PodQOSBestEffort && !resourcerequests.IncludeNonNative(pod) {
+	if qos != v1.PodQOSGuaranteed && !resourcerequests.IncludeNonNative(pod) {
 		return nil
 	}
 
--- a/pkg/noderesourcetopology/filter_test.go
+++ b/pkg/noderesourcetopology/filter_test.go
@@ -418,8 +418,8 @@ func TestNodeResourceTopology(t *testing.T) {
 		{
 			name: "Burstable QoS, pod fit",
 			pod: makePodByResourceList(&v1.ResourceList{
-				v1.ResourceCPU:  *resource.NewQuantity(4, resource.DecimalSI),
-				nicResourceName: *resource.NewQuantity(3, resource.DecimalSI)}),
+				v1.ResourceCPU: *resource.NewQuantity(4, resource.DecimalSI),
+			}),
 			node:       nodes[1],
 			wantStatus: nil,
 		},


Tal-or (commented):
/ok-to-test

PiotrProkop:
Some teams(like mine) use `topo-aware` scheduler for burstable pods. We also don't use `cpumanager` and or `memorymanager` and still want to align resources based on topology. This patch would prevent us from using this plugin, at least can we make this `enhancement` as `opt-in` and don't change current behaviour?

shajmakh:
> Some teams(like mine) use `topo-aware` scheduler for burstable pods. We also don't use `cpumanager` and or `memorymanager` and still want to align resources based on topology. This patch would prevent us from using this plugin, at least can we make this `enhancement` as `opt-in` and don't change current behaviour?

If we track the flow of handling the burstable pod in Filter, it eventually always fit in the node according to https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/numaresources.go#L138
what this patch does is return the success status early for burstable QoS pods and pass it the NodeResourceFit to decide. NodeResourcesMatch plugin by design does not own the decision about non-guaranteed pods that don't request exclusive resources, thus I don't see how this can affect the scenario in the case you are describing.

ffromani:
> Some teams(like mine) use `topo-aware` scheduler for burstable pods. We also don't use `cpumanager` and or `memorymanager` and still want to align resources based on topology. This patch would prevent us from using this plugin, at least can we make this `enhancement` as `opt-in` and don't change current behaviour?

By all means we must be careful to not break existing users of the plugin with different configurations - `discardReserved` cache implementation comes to mind.
We should make sure these usecases are recorded so we have more safety nets than review - thanks @PiotrProkop for chiming in. Is there a chance to add/extend testcases to make sure your usecases are recorded?

ffromani:
that said, @shajmakh 's points have merit, at glance is not immediately obvious to me how this change can break users because non-guaranteed pods or pods without exclusive resources don't contribute to per-numa accounting. I'll need to take into account past convos in my review.

/hold

PiotrProkop:
> that said, @shajmakh 's points have merit, at glance is not immediately obvious to me how this change can break users because non-guaranteed pods or pods without exclusive resources don't contribute to per-numa accounting. I'll need to take into account past convos in my review.
> 

I agree that https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/numaresources.go#L138 makes it so only Guaranteed pods can benefit from upstream noderesourcetopology plugin. If I remember this constraint was not there from the beginning and it was only introduced with some later refactor. That's why we have to maintain such patch in our downstream release:
```go
-	// if it's a non-guaranteed pod, every node is considered to be a good fit
-	if v1qos.GetPodQOS(pod) != v1.PodQOSGuaranteed {
+	if v1qos.GetPodQOS(pod) == v1.PodQOSBestEffort {
		return framework.MaxNodeScore, nil
	}
```
This PR made me think if we can somehow make it configurable in the plugin itself, instead of us having to maintain yet another downstream patch.

PR #982, opened 2026-06-12
[NodeResourceTopologyMatch] Enable default preemption 

shajmakh:
#### What type of PR is this?
/kind bug

#### What this PR does / why we need it:
Default preemption on NRT scheduler only frees node-level resources in 
NodeInfo during dry-run. The NUMA filter reads per-zone NRT data that still 
reflects victim consumption on their NUMA zones, causing valid preemption
candidates to be rejected.

This PR implements eviction simulation for the NUMA-aware filter for 
**Overreserve cache only** so dry-run correctly evaluates whether a preemptor 
can be NUMA-aligned after evicting victim pods. 

For the post-eviction simulation, add PreFilter and PreFilterExtensions to track
victim pods via the framework's RemovePod/AddPod callbacks into a preemption 
stack stored in CycleState. During Filter, when victims are present,
reconstruct a post-eviction NRT view using numaplacement-encoded
container localities before running NUMA alignment checks. Skip
NodeMaybeOverReserved marking during preemption dry-run since the
node is being evaluated for eviction, not rejected for over-reservation.

**ON/OFF option**: The plugin provides a configuration option to control whether eviction dry-run and 
all its required setup will run or not. By default, the functionality is disabled and
can be enabled by adding `preemptionMode: "Enabled"` as an args field under the 
NodeResourceTopologyMatch plugin config.

**Cache type note**: The preemption workflow is tackled only for Overreserve cache in this PR. 
Other caches are not handled here/now, and their workflows are not affected by this code as 
they return `nil` NUMA-placement data, which reserves the old flow for Filter. 
Handling the preemption for other cached boils down to implementing `GetCachedNUMAPlacementInfo`
to return a valid parsed containers NUMA placement data instead of `nil`, and will be proposed in a different PR. 

#### Which issue(s) this PR fixes:

Fixes #981 

#### Special notes for your reviewer:

1. Requires https://github.com/kubernetes-sigs/scheduler-plugins/pull/978

2. AI-assisted work notes:
* This support depends on a 3rd party project that was heavily AI-assisted while also carefully reviewed by humans. 
* Part of the unit tests are AI-assisted, reviewed and approved by the reporter (@shajmakh me :) )
* The feature documentation was initiated by the reporter, primarily written by AI, and reviewed by the reporter. 

3. Follow-up work needed:
https://github.com/kubernetes-sigs/scheduler-plugins/pull/982#issuecomment-5292966985

#### Does this PR introduce a user-facing change?

```release-note
[...]

PR #1000, opened 2026-08-18
[NodeResourceTopology] watcher code redesign and simplification

ffromani:
#### What type of PR is this?

/kind cleanup

Redesign and simplify the nodresourcetopology watcher code to remove shared data access. This PR includes the prefactoring needed by #973 and the fixes to watcher code we identified as pre-existing gap in #982

Fixes #N/A

Extracted from #973 ; #973 will be trivially rebased on top of this later on

```release-note
NONE
```


pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -39,12 +39,20 @@ import (
...
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/resourcerequests"
 	"sigs.k8s.io/scheduler-plugins/pkg/noderesourcetopology/stringify"
 	"sigs.k8s.io/scheduler-plugins/pkg/util"
 )
 
+const (
+	// defaultMaxNRTUpdates is the watcher update channel size initial heuristic, no hard data.
+	// we used a bounded channel to prevent excessive event accumulation. The watcher code
+	// is now in charge to detect and handle overflow using retries.
+	defaultMaxNRTUpdates = 128
```
Tal-or, line 52: Copy my comment from #973 

Did we take into account how this number will take affect on big clusters?
We're committing to handle 500 nodes cluster size

ffromani: Thanks for the comment. This means we will process at most 128 events in one go - it's a throttling knob essentially, and no, there's no hard data backing the value.

Should the channel overflow, this means the events will be lost temporarily but safely detected on the next change, so the system eventually converges

...provided the node change again, which is actually a bold assumption.
So in the next commit I added a self-healing mechanism which handles the overflow and keep the resource consumption bound, which is the key reason to use fixed-size channels.

ffromani:

this is a race somewhere in the coscheduling plugin. When I catch my breath I'll try to have a look and join the hunt

pkg/noderesourcetopology/cache/attr_watch.go
```diff
@@ -66,35 +72,33 @@ func (wt Watcher) NodeResourceTopologies(ctx context.Context, client ctrlclient.
...
-	if nrtCur == nil {
-		wt.lh.Info("modified non-existent NRT", logging.KeyNode, nrtObj.Name)
-		return false
-	}
+	newConf := nodeconfig.TopologyManagerFromNodeResourceTopology(wt.lh, nrtObj)
 
-	if !areAttrsChanged(nrtCur, nrtObj) {
-		return false
+	oldConf := wt.lastConf[nrtObj.Name]
+	if oldConf.Equal(newConf) {
```
shajmakh: I understand there is visible change by this modification, but do we want to keep it specific to TM attributes only or include all other attributes?

ffromani:

bad streak?

pkg/noderesourcetopology/cache/attr_watch.go
```diff
@@ -66,35 +72,33 @@ func (wt Watcher) NodeResourceTopologies(ctx context.Context, client ctrlclient.
 	}
 }
 
-func (wt Watcher) ProcessEvent(ev watch.Event) bool {
+func (wt Watcher) processEvent(ev watch.Event) {
+	// TODO: handle node deletion. How common do we expect it to be?
+	// Modified is trivially verified; Added can happen if we
+	// happen to run the scheduler before node update agents,
+	// so turns out not so uncommon.
```
shajmakh: is Added part of the TODO?

pkg/noderesourcetopology/cache/overreserve_test.go
```diff
@@ -1045,6 +1045,7 @@ func mustOverReserve(t *testing.T, client ctrlclient.WithWatch, podLister podlis
 	if err != nil {
```
shajmakh, line 1046: is it possible to have a test that checks if Close actually stops the watcher?

pkg/noderesourcetopology/cache/cache.go
```diff
@@ -138,4 +138,9 @@ type Interface interface {
 	// up. If a plugin needs to clean up its state after a pod is scheduled and
 	// bound, PostBind is the extension point that it should register.
 	PostBind(nodeName string, pod *corev1.Pod)
+
+	// Close shuts down the cache, releasing all held resources.
```
shajmakh, line 142: held resources is meant to be the attribute watcher right?

pkg/noderesourcetopology/cache/overreserve.go
```diff
@@ -381,13 +365,45 @@ func (ov *OverReserve) MakeNRTUpdatesForNodes(ctx context.Context, lh_ logr.Logg
...
+	}
+
+	if part := ov.makeNRTUpdatesForNodes(ctx, lh_, ov.client, nodeUpdatePool{
+		NodeToObjsMap: nodeToObjsMap,
+		Names:         nodes.ConfigChanged,
+		Reason:        "configChanged",
+		GateCheck:     nullGate,
+	}); len(part) > 0 {
+		nrtUpdates = append(nrtUpdates, part...)
+	}
```
shajmakh, line 379: I see more abstraction here which is good in a way, but readability wise is more complex. how is this better than the old version?

pkg/noderesourcetopology/cache/attr_watch.go
ffromani: we focus only on Topology Manager (TM) attributes, there are no use cases I'm aware of for other stuff

pkg/noderesourcetopology/cache/attr_watch.go
ffromani: no, this is only about `Deleted` items - and it's more a note than else

pkg/noderesourcetopology/cache/overreserve_test.go
ffromani: worth adding it, thanks

pkg/noderesourcetopology/cache/cache.go
ffromani: no, this is for the cache proper. The fact the `overreserve` cache owns a Watcher is a detail which should be transparent for the callers of the API.

pkg/noderesourcetopology/cache/overreserve.go
ffromani: I see this as positive because the update generation logic is isolated and consistent. What changes per-node-set is the gating condition which makes us decide if we should generate an update for a node in the set.

For the nodes in the "configChanged" set, we always do (and we will for the nodes in the "newlyAdded" set which we introduce in #973). For the nodes in the `maybeOverreserved` we actually have a gate which includes logic, which is now self-contained.

I don't claim a function pointer is the best isolation mechanism, yet it's the simplest I could find to enforce this isolation, which would make implementing #973 straightforwarded.

ffromani: aaaand there it was a real bug in this area I just noticed, fixing

pkg/noderesourcetopology/cache/attr_watch.go
shajmakh: I read the comment: "so turns out not so uncommon."  such that Added event post scheduler startup need to be tackled, or do I miss anything? or do we rely on a later Modified event to handle the first addition?

pkg/noderesourcetopology/cache/attr_watch.go
shajmakh: understood. I was thinking about pfp but the foreign pod detection already handles that.

pkg/noderesourcetopology/cache/overreserve.go
shajmakh: ack, fair enough

pkg/noderesourcetopology/cache/attr_watch.go
ffromani: let me rephrase the sentence. It is meant to convey that "Added" events were initially expected to be rare, but turned out relatively common, especially in CI envs.

Tal-or (approved):
/lgtm

Thanks!

PR #1001, opened 2026-08-18
coscheduling: address wakeup race

ffromani:
#### What type of PR is this?

/kind bug
/kind flake

Conditionally waking up pods on the first occurrence opens the door to races and lost events, which we observed in CI.

The fix is unconditionally try to wake up each time a pod belonging to a group comes. This is admittedly less efficient but should be more robust.

Fixes no issues, but another attempt to deflake CI after #983 

Being not much familiar with the `coscheduling` codebase, I used extensive AI help to investigate the failure and draft a solution. I reviewed the code and from my very limited `coscheduling` experience the fix makes sense.
I can't estimate how much worse the performance would be.

```release-note
NONE
```

### DIFF
--- a/pkg/coscheduling/core/core.go
+++ b/pkg/coscheduling/core/core.go
@@ -222,7 +222,9 @@ func (pgMgr *PodGroupManager) ActivateSiblings(ctx context.Context, pod *corev1.
 		}
 	}
 
-	if len(pods) != 0 {
+	if len(pods) == 0 {
+		lh.V(4).Info("ActivateSiblings found no other pods to activate", "pod", klog.KObj(pod), "podGroup", pgName)
+	} else {
 		if c, err := state.Read(framework.PodsToActivateKey); err == nil {
 			if s, ok := c.(*framework.PodsToActivate); ok {
 				s.Lock()
@@ -231,6 +233,7 @@ func (pgMgr *PodGroupManager) ActivateSiblings(ctx context.Context, pod *corev1.
 					s.Map[namespacedName] = pod
 				}
 				s.Unlock()
+				lh.V(3).Info("ActivateSiblings requested activation for sibling pods", "pod", klog.KObj(pod), "podGroup", pgName, "activated", len(pods))
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


ffromani:
/hold
/cc @Huang-Wei 

would this direction make sense? should we accept the performance loss?

ffromani:

good luck or good fix? :)

pkg/coscheduling/core/core.go
```diff
@@ -329,17 +332,9 @@ func (pgMgr *PodGroupManager) Permit(ctx context.Context, state fwk.CycleState,
...
-		// - new pods added
-		// In either case, we should and only should use this 0-th pod to trigger activating
-		// its siblings.
-		// It'd be in-efficient if we trigger activating siblings unconditionally.
-		// See https://github.com/kubernetes-sigs/scheduler-plugins/issues/682
-		state.Write(permitStateKey, &PermitState{Activate: true})
-	}
+	// Request sibling activation on every pod that reaches this point, not just the
+	// first one to join `assigned`. This is meant to avoid wakeup races.
+	state.Write(permitStateKey, &PermitState{Activate: true})
```
Huang-Wei, line 337: this would revert the logic to pre-#700, which could cause CPU spike as #682 reported.

Let me think of a solution to keep re-queue efficient as well as keeping cpu usage non-spiky.

Huang-Wei: @ffromani it seems unconditional wakeup doesn't really fix the integration test flake - `t6: different priority, not sequentially pg1 meet min and 3 regular pods, 120s timeout` still timed out.

ffromani:
@Huang-Wei yes, I noticed this fix doesn't even improve the flakiness side. Considering the perf regression I'm closing it. Tried throwing AI at the problem but (very much unsurprisingly) didn't help

Huang-Wei:
> @Huang-Wei yes, I noticed this fix doesn't even improve the flakiness side. Considering the perf regression I'm closing it. Tried throwing AI at the problem but (very much unsurprisingly) didn't help

yup, let's use https://github.com/kubernetes/kubernetes/issues/141694 to track the upstream fix.

PR #1004, opened 2026-09-03
nrt: cleanup: use ContainerIter to iterate in place

ffromani:
#### What type of PR is this?

/kind cleanup

The pre-iterator standard pattern of using `append(pod.Spec.InitContainers, pod.Spec.Containers...)` to iterate over containers was causeing allocation of backing slices and copies, creating GC pressure.

We now have a kube utility to iterate in place, let's use it. No intended change of behavior.

Fixes N/A

Note: this means the previous code MUST not mutate containers - that was never the intent anyway, but this change may now expose latent hidden bugs.

```release-note
NONE
```

### DIFF
--- a/pkg/noderesourcetopology/resourcerequests/exclusive.go
+++ b/pkg/noderesourcetopology/resourcerequests/exclusive.go
@@ -20,21 +20,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/sets"
+	podutil "k8s.io/kubernetes/pkg/api/v1/pod"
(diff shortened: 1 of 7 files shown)


shajmakh:
Thanks for posting this. Good to learn it. LGTM.

pkg/noderesourcetopology/resourcerequests/exclusive.go
shajmakh: if handling both types of containers is exactly the same (as seems here) , then can these reduce the duplication of the inner for using `podutil.ContainerIter(&pod.Spec, podutil.InitContainers|podutil.Containers)  (saw something similar in your other similar PRs)

ffromani: nice idea, let me play with it

Issue #1007, opened 2026-09-07
Flaky test: TestCoschedulingPlugin "different priority, not sequentially pg1 meet min and 3 regular pods" times out in ~38% of CI runs


Huang-Wei:
### Area

- [x] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

`test/integration/coscheduling_test.go`, job `pull-scheduler-plugins-integration-test-master`

### What happened?

`TestCoschedulingPlugin/different_priority,_not_sequentially_pg1_meet_min_and_3_regular_pods`
(t6) hangs for its full 120 s poll budget and fails with
`Waiting expectedPods error: context deadline exceeded`.

It is not rare. Enumerating every retained build of
`pull-scheduler-plugins-integration-test-master` from GCS (126 builds with both
`build-log.txt` and `finished.json`, 2026-06-09 → 2026-09-04):

| | |
|---|---|
| builds where the integration suite ran to completion | 105 |
| of those, t6 failed | **40 (38.1%)** |
| distinct PRs affected | **17** |
| other `TestCoschedulingPlugin` subtests that ever failed | only t5, 4 times, all inside the #972 window |

The PR under test is irrelevant — the top offenders touch nothing in coscheduling:

| PR | t6 failures | what it changes |
|---|---|---|
| #982 | 9 | NodeResourceTopologyMatch preemption |
| #971 | 4 | one file: `.github/workflows/helm.yaml` |
| #1000 | 4 | NRT watcher redesign |
| #1004 | 3 | NRT cleanup |
| #992 | 1 | a KEP markdown file and its `kep.yaml` |
| #1005 | 1 | docs/manifests for the v0.35.7 release |

It is also not a slow-machine problem, and not caused by #972:

- The t6 subtest duration is **always** 120.03–120.10 s across all 40 failures — never an
  intermediate value. It either completes in ~2 s or it never completes.
- Subtracting the burned 120 s, failing runs finish the rest of the suite ~2 s *faster* than
  passing runs of the same PR in the same week.
- The three earliest failures (builds `2064598502557618176` / #978, `2065280715448127488` / #980,
  `2066254212131459072` / #971, all 2026-06-10 → 06-14) ran on base SHAs that are **not**
  descendants of `ec94a648` (the #972 merge), i.e. the flake predates the queueing hints that
  #983 reverted. t5 really was hint-induced — 0 t5 failures in the runs since the revert — but t6
[...]

PR #1005, opened 2026-09-06
doc update for v0.35.7

Huang-Wei:
#### What type of PR is this?

/kind documentation

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

> [!NOTE]
> Merging this PR doesn't meant registry.k8s.io/scheduler-plugins/kube-scheduler:v0.35.7 is live.
> In terms of release process, merging this and then cut a v0.35.7 tag can ensure tag-based Helm chart is in-sync.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #1008, opened 2026-09-08
Deflake TestCoschedulingPlugin by disabling async preemption

Huang-Wei:
#### What type of PR is this?

/kind flake
/kind failing-test

`TestCoschedulingPlugin/different priority, not sequentially pg1 meet min and 3 regular pods` (t6)
times out in **~38% of CI runs** — 40 of the 105 completed runs of
`pull-scheduler-plugins-integration-test-master` retained in GCS (2026-06-09 → 2026-09-04), across
**17 unrelated PRs** (e.g. #971, which changes one line of `.github/workflows/helm.yaml`; #992, a
KEP markdown file; #1005, release docs). The subtest always burns exactly its 120 s poll budget,
never an intermediate value — it deadlocks rather than runs slowly.

The stall is an upstream kube-scheduler wake-up gap, not a coscheduling bug:

1. The `pg6-1` Pods park on Permit. They are *assumed*, so they hold node memory in the scheduler
   cache, but their API objects carry no `.spec.nodeName`.
2. `t6-p4` (high priority) fails Filter. `DefaultPreemption` selects those assumed-and-waiting Pods
   as victims, starts async preemption, and gates `t6-p4` in its `PreEnqueue`.
3. Coscheduling's `PostFilter` rejects the waiting siblings, so `ForgetPod` releases their memory
   and `NominatedNodeName` is cleared. The resulting `AssignedPodDelete` reaches `t6-p4` while the
   gate is still shut, and it bounces back to `unschedulablePods`.
4. The preemption goroutine then deletes the victims. They are by now neither assumed nor
   nominated, so `deletePodFromSchedulingQueue` returns without emitting a cluster event.
5. `preempting` is cleared — the gate is open, but nothing re-evaluates it.

`t6-p4` is then stuck indefinitely: since 1.34, `movePodsToActiveOrBackoffQueue` skips gated Pods
unless the event matches the gating plugin's registered events, and `DefaultPreemption` registers
only `Pod/Delete`, so `flushUnschedulablePodsLeftover` cannot rescue it either.

This PR turns `SchedulerAsyncPreemption` off for that one test, which removes the `PreEnqueue` gate
entirely. Measured locally over 16 parallel runs per variant on a 12-core machine:

| variant | t6 failures |
|---|---|
| master | 5/16 |
| `--pod-max-in-unschedulable-pods-duration=10s` | 5/16 |
| **this PR** | **0/16** (and 0/32 on a confirmation run) |

`TestPodCompleted` and `TestPodgroupBackoff` use a single priority, so preemption never fires there
and they are left as-is.

It also registers Coscheduling at `Reserve` in all three tests. The plugin implements
`Reserve`/`Unreserve` and the shipped `multiPoint` config enables it there, so without this the
tests have never exercised `Unreserve` — `pgMgr.Unreserve`, `DeletePermittedPodGroup` and
`MarkPodGroupScheduleFailure` were all dead code under test. This is a fidelity fix and does not
affect the flake rate on its own (4/16 measured).

Related: #1007 (kept open — this is a workaround, not the fix) and
kubernetes/kubernetes#141694 upstream.

This is deliberately a workaround with a `TODO`, not a fix. The real fix belongs in
`prepareCandidateAsync`, which should activate the preemptor on completion rather than relying on a
victim-deletion event that a no-op or already-forgotten victim never produces. Upstream #141694
tracks two other triggers for the same gap; the coscheduling one is a third.

The cost of this workaround is that `TestCoschedulingPlugin` no longer exercises async preemption.
The alternative that also measured 0/16 — dropping coscheduling's optimistic rejection of waiting
siblings in `PostFilter` — removes a real feature, so I did not take it.

```release-note
NONE
```

### DIFF
--- a/test/integration/coscheduling_test.go
+++ b/test/integration/coscheduling_test.go
@@ -28,7 +28,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/apimachinery/pkg/util/wait"
+	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/client-go/kubernetes"
+	featuregatetesting "k8s.io/component-base/featuregate/testing"
+	"k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/scheduler"
 	schedapi "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	fwkruntime "k8s.io/kubernetes/pkg/scheduler/framework/runtime"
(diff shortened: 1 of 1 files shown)


Huang-Wei:
cc @ffromani this can be served as a temporary workaround to unblock annoying flaky CI.

googs1025 (commented):
LGTM. The workaround is justified by the upstream async-preemption lost-wakeup bug.

ffromani:
late to the party, but LGTM as well. Thanks for the nice writeup, I learned quite some things.

PR #1011, opened 2026-09-08
upgrade k8s to v1.36

Huang-Wei:
#### What type of PR is this?

/kind cleanup

Upgrade k8s to v1.36.

- Go 1.26.0
- `k8s.io/*` to v0.36.4 / `k8s.io/kubernetes` to v1.36.4
- `sigs.k8s.io/controller-runtime` to v0.24.1
- re-ran `hack/update-codegen.sh`

Fixes #

**New staging repos.** k8s 1.36 splits out `k8s.io/cri-streaming` and `k8s.io/streaming`. They need `replace` directives, otherwise `go mod tidy` fails outright because k/k pins them at `v0.0.0`.

**Preemption `Evaluator` / `Executor` split.** `preemption.NewEvaluator()` no longer takes an `enableAsyncPreemption bool`; the actuation logic moved into a new `preemption.Executor`. `capacityscheduling` and `preemptiontoleration` now pass `preemption.NewExecutor(fh, feature.Features{})` — the zero-valued `Features` leaves `EnableAsyncPreemption` off, exactly like the `false` it replaces.

**Real bug that 1.36 uncovered.** `NodeInfo.RemovePod()` now clears the vacated slot of the pods slice (`// clear the reference to prevent potential memory leak` in `removeFromSlice()`). Both `capacityscheduling` and `preemptiontoleration` called `removePod()` from inside `for _, pi := range nodeInfo.GetPods()`, so the still-in-flight range observed a `nil` `PodInfo` and panicked on `pi.GetPod()`. Inside `DryRunPreemption` the panic surfaced only as "0 preemption candidates" / `TestPostFilter` failures. Both plugins now collect the victims first and remove them in a second loop, which is the same restructuring upstream applied to `defaultpreemption`.

**Interface / default changes reflected in tests:**
- `fwk.SharedLister` gained `PodGroupStates()`; implemented in the fake listers.
- `DynamicResourcesArgs` gained `BindingTimeout` (defaults to 10m) — updated `apis/config/scheme` expectations.
- New `PlacementScore` extension point, defaulted to `NodeResourcesFit` with weight 1 — updated `cmd/scheduler` expectations (most of that file's churn is gofmt realignment).
- `TestPostFilter` now passes `frameworkruntime.WithPodsInPreBind()`, because `Executor` calls `Handle.GetPodInPreBind()` and `frameworkImpl.podsInPreBind` is nil unless the option is set. Production is unaffected — `scheduler.New()` wires it — and upstream's own preemption tests do the same.

`hack/install-envtest.sh` now installs `setup-envtest@release-0.24` to match controller-runtime.

Unit tests and integration tests both pass locally, as do `verify-gofmt`, `verify-gomod`, `verify-crdgen`, `verify-structured-logging` and `verify-toc`.

```release-note
Upgrade k8s to v1.36
```


Huang-Wei:
cc @ffromani @googs1025

fanyang-real (approved):
/lgtm

PR #1012, opened 2026-09-08
Speed up the integration test suite by ~30%

Huang-Wei:
#### What type of PR is this?

/kind cleanup

`pull-scheduler-plugins-integration-test-master` currently takes ~19 minutes, of which ~16 minutes is the test binary itself (`ok sigs.k8s.io/scheduler-plugins/test/integration 976.241s`). This PR cuts the test binary from **976s to 686s (-30%)** without changing a single assertion, test case, or timeout.

Two independent observations drive it.

**1. The suite is wall-clock-bound, not CPU-bound.**

Running the same suite on a 12-core laptop takes the same time as on the 2-CPU Prow pod:

```
Prow (2 CPU):  ok ... 976.241s
laptop (12 core): ok ... 976.264s
```

So essentially all of the 976s is a test *sleeping*, not working.

**2. Most of that sleeping is the poll interval, not the thing being waited for.**

Most wait helpers use a 1s interval (5s in two places, 10s in one) with `immediate=false`, so every wait sleeps a full tick before it checks the condition even once — and the conditions are typically satisfied in milliseconds.

`allocatable_test.go` is the clearest case: it waits for five pods sequentially, so a subtest doing ~50ms of real work reliably takes 5.2s.

The first commit lowers those intervals to 200ms (250ms for the NRT preemption loop, which was at 10s) and sets `immediate=true`. **Every timeout is left exactly as it was**, so no test can fail sooner than it does today; only the time spent asleep between checks changes. `capacity_scheduling`, `elasticquota_controller` and `podgroup_controller` already used this pattern — this makes the rest consistent with them.

The second commit lowers `defaultCacheResyncPeriodSeconds` from 5 to 2. That is a test-local constant — the plugin has no default of its own (`pluginhelpers.go` treats `<= 0` as passthrough) — and the NRT tests derive their waits from it (fixed sleeps at 3x and 5x, `flushNRTViaConfigChange` at 2x), so it acts as a plain multiplier on runtime while exercising the identical resync path. The multipliers still leave 2-5 resync periods of slack on every wait.

Measured end to end, full suite, `-count=1`:

| | package time |
|---|---|
| master | 976s |
| + poll intervals | 767s |
| + resync period | **686s** |

Biggest movers: `TestTopologyMatchPlugin` 123s -> 14s, `TestTopologyMatchPreemption` 275s -> 203s, `TestAllocatablePlugin` 10.5s -> 0.4s, `TestCoschedulingPlugin` 12.0s -> 3.0s.

None.

- **Nothing is skipped, deleted, weakened or given a shorter deadline.** The mechanical check on the diff is that on every changed line the timeout argument is identical before and after; only the interval and the `immediate` boolean move.
- `podIsScheduled()` keeps its existing overall budget (`times * interval`) and just polls more often inside it.
- Negative assertions ("this pod must stay pending") are untouched — those are genuine waits and still run for their full window.
- Verified with three full local runs (976s baseline, 767s after commit 1, 686s after commit 2); the final state is green with 25/25 top-level tests passing and `go vet ./test/integration/` clean.

There is more available beyond this PR, which I'd rather keep separate:

- `TestMain` starts one envtest control plane per *process*, so sharding the compiled test binary across N processes gives each shard a private apiserver with no source changes. Measured at **220s wall clock across 5 shards** (3.1x), bounded by the longest single test.
- ~50s of the ~165s build phase looks recoverable (`go install setup-envtest@release-0.22` compiles a tool from source every run; `go test` re-vets the 1120-package dependency tree; the envtest asset download is serialized ahead of the compile rather than overlapped).
- `podIsPending()` never returns early, so it always burns `times * interval` = a hard 50s, three times over. Instrumenting it shows the second scheduling attempt actually lands at t≈30.1-31.2s (the hard-coded 30s `flushUnschedulablePodsLeftover` interval), so ~20s of each window is padding.

Also worth noting: CI runs `go test` without `-v`, so per-test timings have never been visible in the job log. That is largely why this went unnoticed while the suite grew from ~697s (Jun) to ~976s (Aug).

```release-note
NONE
```

### DIFF
--- a/test/integration/allocatable_test.go
+++ b/test/integration/allocatable_test.go
@@ -175,7 +175,7 @@ func TestAllocatablePlugin(t *testing.T) {
 			defer cleanupPods(t, testCtx, tc.pods)
 
 			for _, pod := range tc.pods {
-				err := wait.PollUntilContextTimeout(testCtx.Ctx, 1*time.Second, 60*time.Second, false, func(ctx context.Context) (bool, error) {
+				err := wait.PollUntilContextTimeout(testCtx.Ctx, 200*time.Millisecond, 60*time.Second, true, func(ctx context.Context) (bool, error) {
 					return podScheduled(t, cs, pod.Namespace, pod.Name), nil
 				})
 				if err != nil {
--- a/test/integration/coscheduling_test.go
+++ b/test/integration/coscheduling_test.go
@@ -378,7 +378,7 @@ func TestCoschedulingPlugin(t *testing.T) {
 					t.Fatalf("Failed to create Pod %q: %v", tt.pods[i].Name, err)
 				}
 			}
-			err = wait.PollUntilContextTimeout(testCtx.Ctx, 1*time.Second, 120*time.Second, false, func(ctx context.Context) (bool, error) {
+			err = wait.PollUntilContextTimeout(testCtx.Ctx, 200*time.Millisecond, 120*time.Second, true, func(ctx context.Context) (bool, error) {
 				for _, v := range tt.expectedPods {
 					if !podScheduled(t, cs, ns, v) {
 						return false, nil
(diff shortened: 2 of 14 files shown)


Huang-Wei:
cc @ffromani 

The running time for integration test is shorten to ~14m. ([log](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_scheduler-plugins/1012/pull-scheduler-plugins-integration-test-master/2097445476218441728))

ffromani (commented):
confidently LGTM all the NRT tests. I authored most of them and I can say the existing timeouts were intentionally conservative, e.g. there is indeed _not_ a clear reason to have timeouts so long.

Also LGTM all the other parts.

@Huang-Wei would you like more review or can I just tag it? I'm fine doing so, just checking for due diligence

PR #1014, opened 2026-09-14
nrt: preemption: account for init containers

### LINKED ISSUE #1013: [NodeResourcesTopologyMatch] DefaultPreemption does not account for restartable init-containers
### Area

- [x] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

High GU priority pod, managed by the NRT scheduler while the preemption is enabled, fails to evict lower priority pod that has restartable init-container with the same exclusive resource requests as the preemptor.  

should evict the pod with init container to free room for more important workloads (preemptor).

 1. fill node in pods that has restartable init containers with most of the CPU requests (or any exclusive resources type combined with relevant QoS). 
2. create a pod with TAS scheduler that requests the same resources as one init container and is higher in priority that all running pods.

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

shajmakh:
#### What type of PR is this?
/kind bug

A victim pod may have restartable init containers that use execlusive
resources, these too should be accounted for during the accumulation of
the victims resources as pre-step for the dry-run eviction.
This commit includes the init containers in the scan and adds a unit
test that was missing before.

Fixes #1013

```release-note
Fixed NUMA-aware preemption in the NodeResourceTopologyMatch scheduler plugin to account for exclusive resources requested by restartable init containers when simulating victim pod eviction.
```


ffromani:
/approve
/lgtm

thanks!

PR #1016, opened 2026-09-15
nrt: preemption: request a resync on missing NUMA mapping

### LINKED ISSUE #1015: [NodeTopologyMatch] Stale NRT cache with no NUMA mapping for one victim container will continuesly fail preemption
### Area

- [x] Scheduler
- [ ] Controller
- [ ] Helm Chart
- [ ] Documents

### Other components

_No response_

### What happened?

When preemption is enabled and a node is full, scheduling a high-priority pod requesting exclusive resources fails when:
-a previous low-priority pod with exclusive resources was scheduled with NRT scheduler last on that node
-non-NRT filter plugin failed which evaluating the high-priority pod because node-level resources are not enough, hence triggering a preemption. 
-preemption first victim includes all running pods with lower priority than the preemptor
-eviction simulation fails because the cached NRT is stale in terms of NUMA-placement attributes

A workaround would be to schedule a foreign pod; change the TM config; or initiate another pod scheduling while it is guaranteed to pass the non-NRT filter stage and fail NRT filter plugin. These all share the same idea for fixing the issue: sending a signal to the plugin to refresh the NRT cache. 

Evict lower priority pod successfully because it is running and should have valid NUMA mapping as per the APIServer reflects. 

1. A low-priority pod-k passed the NRT plugins.
2. The NRT cache got updated with overreserved resources of this pod
   because soon it will be scheduled on the node.
3. pod-k is scheduled and cluster node resources are updated.

Note here that the NRT cache node is still "clean" because no other pod
failed the NRT filter.

4.new higher-priority pod-A is created and managed by NRT scheduler,
which first runs the other Filter plugins.
5. NodeResourceFit fails because it sees that the new available
   resources quantities are not enough to fit pod-A
6. preemption is triggered, and a victim set is chosen from the running
   pods on the cluster. The first victim set will include all running pods
   on this node that are with lower priority than pod-A; this means the set
   includes pod-k.
6.1 eviction simulation will start according to the container-to-NUMA
mapping in the current NRT cache. The cache will provide the latest
update it had which is before scheduling pod-k, meaning that pod-k
will not be part of the PFP in the cached NRT, hence the NUMA placement
attributes will not include this pod either. This leads to failing the
eviction simulation -> failing preemption/postFilter.

_No response_

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Scheduler Plugins version

<details>

</details>

shajmakh:
#### What type of PR is this?
/kind bug
/kind cleanup

TLDR: When pod fails filter plugin that is not NRT, it does not request a
resync of the cache and moves directly to PostFilter where it runs the
preemption. And when it runs, it may pick up victims that are newly
turned to running and were not yet part of the captured NRT cache.
This leads to failing the preemption and scheduling cycle due to missing
mapping. The solution to that is to refresh the NRT in hope to have a
fresher NUMA-placement data that includes NUMA mapping of all the
relevant victims.
This commit requests a resync by marking the node as
`MaybeOverreserved`.

Note about current limitation by design: when a pod is pending after a
failed preemption attempt it will be retried again on events like nodes
getting more space (allocatable or available resources) or after unschedulable
timeout period ends. While successful resync may update the availability
of the resources, it current has no way to trigger a scheduling retry on
the pending pods, leading to the need to wait max
~unscheduable-timout-period fr the next scheduling cycle.

Fixes #1015

```release-note
NONE
```


shajmakh:
/cc @ffromani 

This will fix the attached issue, yet there is still some room to avoid the resync when victims are not relevant for NRT, meaning they are not requesting exclusive resources. It can be in a new commit under this PR or we can do that as a cleanup PR. Let me know what you think is better please. Thanks. 

EDIT: Avoiding resync on shared-resources only victims appeared to be a small enough change (nil out the error) thus I added it under this PR.

shajmakh:
/hold 
for testing results

shajmakh:
/unhold
passed e2e manual test

pkg/noderesourcetopology/filter.go
```diff
@@ -213,7 +213,9 @@ func (tm *TopologyMatch) Filter(ctx context.Context, cycleState fwk.CycleState,
 		if numaPlacementInfo != nil && numaPlacementInfo.Containers() != 0 {
 			nodeTopology, err = preemption.GetNRTPostPodsEviction(lh, nodeTopology.DeepCopy(), victims, numaPlacementInfo)
 			if err != nil {
-				return fwk.NewStatus(fwk.Unschedulable, "eviction simulation in NRT is not possible:"+err.Error())
+				// valid NUMA placement info aligns with having a fresh PFP hence use OverReserved signal to resync with the PFP gate
+				tm.nrtCache.NodeMaybeOverReserved(nodeName, pod)
```
ffromani, line 217: I don't agree with the premise just yet.
1. the fact we miss placement data is not a synonym of the fact the node is overreserved. These are 2 different reasons which may possibly trigger a resync, we should not conflate them
2. contrary to the PR head title, marking a node overreserved does _not_ request a resync - which implied a request/response: it _enables_ the node to be resynced as soon as possible, which depending on the NRT influx from the node updating agent, can happen significantly later in the future

ffromani (commented):
I think I get the problem here, but unfortunately I don't get yet why this is a correct fix

pkg/noderesourcetopology/filter.go
shajmakh: I think I get your 2 points, let me try to explain why I went with this. 
The preemption bug is that when at least one victim pod (with exclusive resources) does not have container-NUMA mapping. The old Filter-preemption version would keep rejecting the preemptor  always starting the first preemption cycle. 
The fix for this issue is to ultimately have NUMA mapping to all victims with exclusive resources. 
If we agree so far then we can proceed to how we can get to have an updated NUMA placement that has all the mapping that the preemption needs.

Before that we need to highlight that the postFilter has the same or more recent(>=) view of the running pods on the node as the NRT cache; this is because of the overreserve and how resync infinite loop works. 

Based on the above, when a preemption cycle first starts it will have ALL running lower-priority pods (doesn't care if they are with exclusive resources or not). It knows these victim pods are all consuming some resources on the node. 
In NRT-Filter, we care about the NUMA placements because we need them to do a proper add-back of the victims resources (dry-run the victim eviction and update NRT cache with additional free resources).
When preemption is triggered, given the above, all victims pods are certainly running (a terminating pod will abort preemption). so for sure these pods are expected to already have NUMA mapping (relevant ones = with exclusive resources). and here hits the problem, what if some pods don't have NUMA mapping? 
It means we cannot do dry-eviction of this pod, and possible handling is either:

**A.** ignore this pod and move on in the preemption cycle: the consequence of this is that it Filter-preemption may succeed and and eventually evict some pods, however, it will not guarantee that it evicted the correct set of pods: the victim with no-NUMA mapping may be the correct candidate with lowest priority. other inaccurate scenarios may happen here too, and the ultimate solution for the results remains to have an updated NUMA mapping. 

**B.** we abort preemption and wait for a valid NUMA mapping: we don't make assumptions; that's why we have the NUMA placement. The NUMA-placement at that moment in time may still lack some running containers, but that is only a problem if the victims are among the missing ones. 


That said we proceed with option B. But what problem is there with the old code before this PR? 
it is that we may wait infinitely without having NUMA-placement ever updated. 
This is rare but it may happen. 
so to address that we look at what refreshes the NUMA-placement; the answer is a fresh PFP. 
and how can we get a fresher cached PFP? it's by either have the node marked as MaybeOverreserved (will keep the node dirty until the view matches between APIServer NRT and scheduler overreserved cache); or by side effect of updating the cache due to forieng pod or topology-manager config.  
Hence the `tm.nrtCache.NodeMaybeOverReserved(nodeName, pod)` after failing NUMA mapping. 

If we agree so far, we know that a resync is ultimately needed. 
As for the signal we send, and why I think NodeMaybeOverReserved is fine to be used here, the reason is that an invalid NUMA mapping means missing NUMA mapping (probably need to change this error message), and that happens when PostFilter knows about running pods that the scheduler do not know about them. 
The foreign pod case is solved anyway by having the resync already requested. OTOH what about NRT-scheduler managed pods? these are exactly the pods that have their resources overreserved in the NRT cache (as assumed resources) because this is how Overreserve works. To my knowledge, this is the only cause to have the misalignment between PostFilter view and NRT-scheduler view **concerning additional pods that PostFilter is aware of**. 
There is another mismatching reason but is in the reverse form: NRT scheduler accounts for pods as existing but they are actually removed so PostFilter do not consider them. <- for this we need another fix in a completely different PR.

ffromani (commented):
/approve
/lgtm

thanks for clarifying the commit messages and for the detailed answers to my comments. It took me some time to navigate the data, but your later clarifications highlighted better the context and now I agree about these fixes.

We probably can use more invasive changes to enable better numa-aware preemption, but this seems a good improvement anyway.

